PR #2, opened 2019-06-21
Use Kubernetes environment variables to find IP addresses

ebaron:
Kubernetes/OpenShift automatically populates certain environment variables in containers to help detect local services those containers may want to communicate with [1]. This PR uses those environment variables to find services that are listening on TCP port 9091. If no such environment variables are found, it will fall back to using the port scanning approach.

Example from robotcontroller pod [2]:
```bash
$ env | grep ROBOTMAKER | sort
ROBOTMAKER_PORT=tcp://172.30.95.111:9091
ROBOTMAKER_PORT_8080_TCP=tcp://172.30.95.111:8080
ROBOTMAKER_PORT_8080_TCP_ADDR=172.30.95.111
ROBOTMAKER_PORT_8080_TCP_PORT=8080
ROBOTMAKER_PORT_8080_TCP_PROTO=tcp
ROBOTMAKER_PORT_8778_TCP=tcp://172.30.95.111:8778
ROBOTMAKER_PORT_8778_TCP_ADDR=172.30.95.111
ROBOTMAKER_PORT_8778_TCP_PORT=8778
ROBOTMAKER_PORT_8778_TCP_PROTO=tcp
ROBOTMAKER_PORT_9091_TCP=tcp://172.30.95.111:9091
ROBOTMAKER_PORT_9091_TCP_ADDR=172.30.95.111
ROBOTMAKER_PORT_9091_TCP_PORT=9091
ROBOTMAKER_PORT_9091_TCP_PROTO=tcp
ROBOTMAKER_PORT_9779_TCP=tcp://172.30.95.111:9779
ROBOTMAKER_PORT_9779_TCP_ADDR=172.30.95.111
ROBOTMAKER_PORT_9779_TCP_PORT=9779
ROBOTMAKER_PORT_9779_TCP_PROTO=tcp
ROBOTMAKER_SERVICE_HOST=172.30.95.111
ROBOTMAKER_SERVICE_PORT=9091
ROBOTMAKER_SERVICE_PORT_8080_TCP=8080
ROBOTMAKER_SERVICE_PORT_8778_TCP=8778
ROBOTMAKER_SERVICE_PORT_9091_TCP=9091
ROBOTMAKER_SERVICE_PORT_9779_TCP=9779
```

[1] https://kubernetes.io/docs/concepts/containers/container-environment-variables/
[2] https://github.com/rh-jmc-team/jmc-robots-demo

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -19,6 +20,8 @@ import com.redhat.rhjmc.containerjfr.tui.ClientWriter;
 
 @Singleton
 class PortScanCommand implements SerializableCommand {
+
+    private static final String KUBERNETES_ENV_SUFFIX = "_PORT_9091_TCP_ADDR";
 
     private final ClientWriter cw;
 
@@ -48,15 +51,41 @@ class PortScanCommand implements SerializableCommand {
 
     @Override
     public void execute(String[] args) throws Exception {
-        scan().forEach(m -> cw.println(String.format("%s -> %s", m.hostname, m.ip)));
+        findCompatibleJvms().forEach(m -> cw.println(String.format("%s -> %s", m.hostname, m.ip)));
     }
 
     @Override
     public Output<?> serializableExecute(String[] args) {
         try {
-            return new ListOutput<>(scan());
+            return new ListOutput<>(findCompatibleJvms());
         } catch (Exception e) {
             return new ExceptionOutput(e);
+        }
+    }
+
+    private List<IpHostMapping> findCompatibleJvms() throws UnknownHostException, InterruptedException {
+        // Check for environment variables added by Kubernetes that indicate
+        // IP addresses listening on port 9091.
+        List<IpHostMapping> mapping = System.getenv().entrySet().parallelStream()
+                .filter(e -> e.getKey().endsWith(KUBERNETES_ENV_SUFFIX))
+                .map(e -> {
+                    try {
+                        return testHost(InetAddress.getByName(e.getValue()));
+                    } catch (IOException ignored) {
+                        return null;
+                    }
+                }).filter(m -> m != null).collect(Collectors.toList());
+        if (mapping.isEmpty()) {
+            // No matches from environment variables, use port scan
+            mapping = scan();
+        }
+        return mapping;
+    }
+
+    private IpHostMapping testHost(InetAddress addr) throws IOException {
+        try (Socket s = new Socket()) {
+            s.connect(new InetSocketAddress(addr, 9091), 100);
+            return new IpHostMapping(addr.getHostAddress(), addr.getCanonicalHostName());
         }
     }
 
@@ -76,11 +105,9 @@ class PortScanCommand implements SerializableCommand {
             };
             executor.submit(() -> {
                 try {
-                    Socket s = new Socket();
                     InetAddress addr = InetAddress.getByAddress(remote);
-                    s.connect(new InetSocketAddress(addr, 9091), 100);
-                    s.close();
-                    result.add(new IpHostMapping(addr.getHostAddress(), addr.getCanonicalHostName()));
+                    IpHostMapping mapping = testHost(addr);
+                    result.add(mapping);
                 } catch (IOException ignored) {
                 } finally {
                     latch.countDown();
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
```diff
@@ -48,18 +51,44 @@ public boolean validate(String[] args) {
...
+        // Check for environment variables added by Kubernetes that indicate
+        // IP addresses listening on port 9091.
+        List<IpHostMapping> mapping = System.getenv().entrySet().parallelStream()
+                .filter(e -> e.getKey().endsWith(KUBERNETES_ENV_SUFFIX))
+                .map(e -> {
+                    try {
+                        return testHost(InetAddress.getByName(e.getValue()));
+                    } catch (IOException ignored) {
+                        return null;
+                    }
```
andrewazores: Could you move the exception handling and `null` return into `testHost` so that the `.map` here can be a one-liner?

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
```diff
@@ -48,18 +51,44 @@ public boolean validate(String[] args) {
...
+                }).filter(m -> m != null).collect(Collectors.toList());
```
andrewazores: Just a matter of taste but I'd prefer to break the `.filter` and `.collect` onto new lines (ie one stream operation per line).

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
```diff
@@ -48,18 +51,44 @@ public boolean validate(String[] args) {
...
+            return new ListOutput<>(findCompatibleJvms());
         } catch (Exception e) {
             return new ExceptionOutput(e);
         }
     }
 
+    private List<IpHostMapping> findCompatibleJvms() throws UnknownHostException, InterruptedException {
+        // Check for environment variables added by Kubernetes that indicate
+        // IP addresses listening on port 9091.
+        List<IpHostMapping> mapping = System.getenv().entrySet().parallelStream()
```
andrewazores: Please abstract the `System.getenv()` call into `com.redhat.rhjmc.containerjfr.sys.Environment`. This will help eventual testability for this class.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
```diff
@@ -48,18 +51,44 @@ public boolean validate(String[] args) {
...
+            // No matches from environment variables, use port scan
+            mapping = scan();
+        }
+        return mapping;
+    }
+
+    private IpHostMapping testHost(InetAddress addr) throws IOException {
+        try (Socket s = new Socket()) {
+            s.connect(new InetSocketAddress(addr, 9091), 100);
+            return new IpHostMapping(addr.getHostAddress(), addr.getCanonicalHostName());
```
andrewazores: Does `getCanonicalHostName` here return a human-readable hostname (as it does in a plain Docker network), or do we still get back the cluster-internal IP?

If it's just the IP then I think a really nice enhancement would be to scan over the environment variables and extract the service names from them and use that to populate this structure.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
```diff
@@ -20,6 +21,8 @@
 @Singleton
 class PortScanCommand implements SerializableCommand {
```
andrewazores: Maybe this command should be renamed now, since it has more ways to detect targets than port scanning. `scan-targets` maybe? We can then do some refactoring to have some pluggable and configurable interface for various detection methods.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
ebaron: Sure. These also catch the UnknownHostException thrown by `InetAddress.getBy{Name,Address}`, so I also added `testHostBy{Name,Address}` helpers to catch these as well. I think it looks cleaner now.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
ebaron: Done

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
ebaron: Sure, I think I've done so correctly now.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
ebaron: It is showing the cluser-internal hostnames for me now.
Screenshot from 2019-06-21 17-56-07

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/PortScanCommand.java
ebaron: Sure, I've done this. I'll see if I can do this on the container-jfr-web side of things too.

PR #25, opened 2019-08-09
[WIP] Serve container-jfr-web from container-jfr embedded webserver

andrewazores:


andrewazores:
This is now generally working, but there's work to be done to clean things up.

1) container-jfr-web overrides the port specified by container-jfr's `list` and `list-saved` commands with port 80, which was a hack to get it working quickly with a specific Kubernetes demo deployment. This should be somehow configurable so that container-jfr can always report the actual correct port number, even if hidden behind something like a Kubernetes route (which may change the port number a client must use).

2) Similarly, container-jfr currently reports the port number it listens for websocket connections on along with the hostname it is configured to know it is accessible via, and container-jfr-web expects to connect with that host/port pair. However, behind a Kubernetes route, that port number may be incorrect. This also needs to be configurable so that container-jfr can report a different port for access than what it knows itself to be listening on.

3) Client assets, recording streams, and rules reports are all currently served by an ugly hacked-together NanoHTTPD instance within container-jfr WebServer class, and websocket connections are handled by a Jetty instance in the MessagingServer class. I had a branch where I tried to refactor this so that everything was running with NanoHTTPD "Nanolets", but ran into issues with the nanolet instantiation when they required dependencies to be injected (ie NanoHTTPD was only attempting to invoke the default no-args constructor for the Nanolets). This may have just been an error on my part, but I scrapped the approach and am of the opinion that the refactor should instead consolidate WebServer and MessagingServer into a single Jetty server with appropriate Jetty servlets for each provided service (client, recording streams, rules reports, and websockets).

Issue #6, opened 2019-06-26
Network-accessing classes need to be testable


andrewazores:
Currently there are several classes which create TCP sockets or HTTP client connections (some examples: RecordingExporter, UploadSavedRecordingsCommand, SocketClientReaderWriter, SocketInteractiveShellExecutor, JMCConnection), and which are partly or entirely missing unit tests since the networking codepaths can not be exercised in unit testing without actually producing network connections. We need some enhanced mocking utilities, or a test harness to intercept connections, in order to properly test and verify these classes.

andrewazores:
`JMCConnection` has been moved to rh-jmc-team/container-jfr-core

andrewazores:
This is improved since #23 since fewer classes directly establish connections, but NetworkResolver, RecordingExporter, and SocketClientReaderWriter need testing. Particularly the latter two as they are larger and relatively complex.

PR #33, opened 2019-09-23
Improve tests for SocketClientReaderWriter and NetworkResolver

tabjy:
This PR improves #6.

For `NetworkResolver::testConnection`, I'm unsure how to mock the internal Socket and InetSocketAddress. The function is therefore untested.

As for the WebServer, it's better to refactor it, making it easier to adapts arbitrary HTTP server implementations. I'm currently working on this.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriter.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriter.java
@@ -5,6 +5,7 @@ import java.io.OutputStreamWriter;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketException;
+import java.nio.charset.StandardCharsets;
 import java.util.Scanner;
 import java.util.concurrent.Semaphore;
 
@@ -17,7 +18,7 @@ class SocketClientReaderWriter implements ClientReader, ClientWriter {
     private final Logger logger;
     private final Thread listenerThread;
     private final ServerSocket ss;
-    private final Semaphore semaphore = new Semaphore(0, true);
+    private final Semaphore semaphore;
     private volatile Socket s;
     private volatile Scanner scanner;
     private volatile OutputStreamWriter writer;
@@ -25,6 +26,7 @@ class SocketClientReaderWriter implements ClientReader, ClientWriter {
     SocketClientReaderWriter(Logger logger, int port) throws IOException {
         this.logger = logger;
         ss = new ServerSocket(port);
+        semaphore = new Semaphore(0, true);
         listenerThread = new Thread(() -> {
             System.out.println(String.format("Listening on port %d", port));
             while (true) {
@@ -38,8 +40,8 @@ class SocketClientReaderWriter implements ClientReader, ClientWriter {
                     System.out.println(String.format("Connected: %s", sock.getRemoteSocketAddress().toString()));
                     try {
                         s = sock;
-                        scanner = new Scanner(sock.getInputStream(), "utf-8");
-                        writer = new OutputStreamWriter(sock.getOutputStream(), "utf-8");
+                        scanner = new Scanner(sock.getInputStream(), StandardCharsets.UTF_8);
+                        writer = new OutputStreamWriter(sock.getOutputStream(), StandardCharsets.UTF_8);
                     } finally {
                         semaphore.release();
                     }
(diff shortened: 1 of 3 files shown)


src/test/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriterTest.java
```diff
@@ -0,0 +1,120 @@
...
+        verify(writer).close();
+        verify(s).close();
+    }
+
+    @Test
+    void shouldReadLine() throws Exception {
+        String result = "read line result";
+        when(scanner.nextLine()).thenReturn(result);
+
+        reset(semaphore);
```
andrewazores: What's the purpose of this reset?

src/test/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriterTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        verify(logger).warn(e);
+    }
+
+    @Test
+    void shouldReadLineReleaseOnException() throws Exception {
+        RuntimeException e = new RuntimeException();
+        when(scanner.nextLine()).thenThrow(e);
+
+        reset(semaphore);
```
andrewazores: Same for this `reset`.

src/main/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriter.java
```diff
@@ -53,6 +55,20 @@
...
 
+    // Testing-only constructor
+    SocketClientReaderWriter(Logger logger, Semaphore semaphore, Socket s, Scanner scanner, OutputStreamWriter writer) {
+        this.logger = logger;
+        this.semaphore = semaphore;
+        this.scanner = scanner;
+        this.writer = writer;
+        this.listenerThread = null;
+        this.ss = null;
+        this.s = s;
```
andrewazores: Since you're already here refactoring, could you please rename `s` and `ss` more appropriately?

src/test/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriterTest.java
tabjy: On `L63` and `L64`, there are verifications for `semaphore` to be properly released for exactly once. Since there is already a `.release()` call in the `SocketClientReaderWriter::new`, which is called in `setup()`, `semaphore` needs to be reset or there would be two `.release()` calls detected at the time of verification.

src/test/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriterTest.java
tabjy: Same as above.

src/main/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriter.java
tabjy: Yes. This will be fixed in the next commit.

src/test/java/com/redhat/rhjmc/containerjfr/tui/tcp/SocketClientReaderWriterTest.java
andrewazores: That's what I figured. I'm not sure if I would rather have the `reset` or just `verify(semaphore, times(2)).whatever()`, but in either case there should be a comment explaining that there's an implicit/hidden `release` at construction time. Or perhaps there should just be a single `reset` done in the initial test setup?

andrewazores: Or perhaps I should ask why the testing-only constructor needs to perform this initial `release` on the semaphore to begin with - since it's testing-only we can/probably should assume that the passed in instance will be a mock, so what does the first release achieve?

tabjy: I'd prefer a `reset()` as the last statement in `setup()` and comment there. Will fix in the next commit.

tabjy: In the non-testing constructor, `semaphore` is created with `0` initial permits. The `.release()` call in that constructor is critical or else there would be a deadlock. I thought the test constructor should follow the same pattern. But you are right, I now come to a realization it is indeed unnecessary as the passed in `semaphore` is mocked anyway.

I'll delete this initial `.release()` call in the testing constructor, as well as all `.reset()`s.

tabjy:
@andrewazores 

Follow-ups will be submitted as separate PRs. Thank you very much!

PR #41, opened 2019-10-09
Relocate generated static web assets to resources directory

tabjy:
Having `/web-client` placed under `/` seems awkward that the application resource is not in a place relative to the application itself. This pull request makes it more sensible by relocating generated static web assets to `resources` directory.

closes tabjy/container-jfr#1


build.gradle
```diff
@@ -52,19 +60,16 @@ task buildWebClient(type: Exec) {
     commandLine 'npm', 'run', 'build'
 }
 
-jib.extraDirectory = file('build/extra-directory')
-
 task copyWebClient(type: Copy) {
     from file('web-client/dist/container-jfr-web')
-    into file('build/extra-directory/web-client')
+    into file("${buildDir}/generated/sources/web-client/resources/main/com/redhat/rhjmc/containerjfr/net")
```
andrewazores: Hmm. I get that this puts the assets somewhere where the bootstrap classloader will find them, but it seems strange to put non-sources like PNGs, TTFs, and the sort-of-source-built-artefacts like actual HTML/CSS/JS of the client into "generated sources". Is there anywhere else that the assets can be placed which will end up with them searchable on the classpath? Or, can a new location be specified for them and added to the search classpath?

src/main/java/com/redhat/rhjmc/containerjfr/net/WebServer.java
```diff
@@ -234,14 +233,14 @@ private Response serveClientIndex() {
...
-            if (!assetPath.toFile().isFile()) {
+            URL url = WebServer.class.getResource(assetName);
+            if (url == null) {
                 return newFixedLengthResponse(Status.NOT_FOUND, NanoHTTPD.MIME_PLAINTEXT,
                         String.format("%s not found", assetName));
             }
             try {
-                String mime = NanoHTTPD.getMimeTypeForFile(assetPath.toUri().toString());
-                InputStream assetStream = Files.newInputStream(assetPath);
+                String mime = NanoHTTPD.getMimeTypeForFile(url.toString());
```
andrewazores: In retrospect, passing the full file path to `getMimeTypeForFile` was probably not any better than simply passing in `assetName`. Please try that out, and if it still works as expected, let's go with that simple way instead.

src/main/java/com/redhat/rhjmc/containerjfr/net/WebServer.java
```diff
@@ -234,14 +233,14 @@ private Response serveClientIndex() {
...
+                InputStream assetStream = url.openStream();
```
andrewazores: Then if the above works, we can forego the `url` altogether and use `WebServer.class.getResourceAsStream` and skip this `openStream` step.

build.gradle
tabjy: In theory, is location can be anywhere, as long as the paths in `copyWebClient` and `sourceSets` are consistent. I put them under `generated` as I thought they are, indeed, resources generated by another program (`npm`) from external sources (`container-jfr-web`).

Anyway, the location can be arbitrary, but they will be eventually copied to `build/resources`, by `processResources` task, for the bootstrap class loader to discover. Any idea where you want it to be?

src/main/java/com/redhat/rhjmc/containerjfr/net/WebServer.java
```diff
@@ -234,14 +233,14 @@ private Response serveClientIndex() {
 
         @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
```
tabjy: Also, I just realized this spotbugs suppression can be now removed. It will be removed in the next commit.

src/main/java/com/redhat/rhjmc/containerjfr/net/WebServer.java
tabjy: Examing the source of `getMimeTypeForFile`, it determines to extension name by looking for the last occurrence of `.` in the input. Therefore, regardless of the input being a full URL, a pathname, a base name, or even just an extension name begins with `.`, it returns the same mine-type.

I agree with skipping obtaining `url` and using `assetName` directly.

src/main/java/com/redhat/rhjmc/containerjfr/net/WebServer.java
tabjy: Will switch to using `getResourceAsStream` on next commit.

build.gradle
andrewazores: I suppose. `generated/sources` sounds to me more like Java sources, ex. stuff produced by Dagger for direct injection and the likes. These things themselves aren't source files though, especially not after they've been processed through the npm build.

Something like `${buildDir}/assets` might do?

src/main/java/com/redhat/rhjmc/containerjfr/net/WebServer.java
andrewazores: Nice, that's always a welcome bonus.

PR #42, opened 2019-10-09
Remove CORS headers from WebServer responses

andrewazores:
CORS headers are no longer necessary since container-jfr is also the
webserver for container-jfr-web

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,7 +69,7 @@ copyWebClient.dependsOn buildWebClient
 processResources.dependsOn copyWebClient
 
 def baseImage = 'gcr.io/distroless/java:11';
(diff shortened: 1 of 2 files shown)


ebaron:
Would this affect command channel connections at all? I'm about to start work on sending commands to container-jfr from the container-jfr-operator.

tabjy:
@ebaron As far as I know, the command channel is managed by a different web service running on a different port. Also, the command channel uses Websocket, which I believe is not bounded by CORS policies.

andrewazores:
The headers are also typically enforced by the end user's browser, not the server, so connections from a "headless" client like the operator would be unaffected anyway. But yea, good question.

PR #43, opened 2019-10-11
JDP Local autodiscovery

andrewazores:
This PR replaces the 9091 portscanning target discovery mechanism with a JDP-based multicast discovery. This should be more robust, makes fewer assumptions (target port number may vary), and has significantly less latency on invocation (though it takes a few seconds for new multicasting targets to be noticed).

For testing purposes, container-jfr can be run as a local machine JVM process with `./gradlew run` and accessed via `localhost:8181`. This can be repeated with https://github.com/andrewazores/container-jmc-simple-demo to run another JVM process on the local machine which has the integrated JDP autodiscovery enabled.


andrewazores:
> Looking forward though, is this something we can make work for detecting JVMs in the cloud?

Possibly, it's just based on multicasting (see JEP: https://openjdk.java.net/jeps/168), so I think it's potentially usable in any environment given a correctly configured network. I haven't tested it in a Kubernetes environment to see what happens with an out-of-the-box configuration there but I'll take a look at that next week and see if it's useful.

If running in plain Docker containers, it does detect and successfully connect to targets in other containers if both containers are added to the same network, but the JDPClient pruning seems to hang somehow when the target goes away so that'll need some investigation (probably in upstream JMC). I've never tested it but this means it's probably a useful starting point for detection in setups like Docker Swarm once that pruning bug is ironed out.

andrewazores:
Just rebased on top of a spotbugs violation fix.

PR #57, opened 2019-11-05
Refactor WebServer and MessagingServer to use Vert.x

tabjy:
Notable changes from original implementation:
- Both `WebServer` and (Ws)`MessagingServer` use the same `HttpServer` and, therefore, listen on the same host.
- `HttpServer` listens on a concrete hostname obtained from `netConf.getWebServerHost()` (instead of `0.0.0.0`)
- JSON responses now have `Content-Type` set as `application/json` instead of `text/plain`
- `WebServer` no longer restart upon `connectionChanged()`. It simply updates `service` and `recordings` fields.
- Tests added for HTTP request handlers.

Closes tabjy#2. Relates to rh-jmc-team#6.


andrewazores:
Before I read through the changeset too much more thoroughly, I've tried testing this with `./gradlew jibDockerBuild && sh run.sh` and I get a "connection reset by peer" when loading "http://localhost:8181", either with my browser or curl. It works when running with `./gradlew run`, though. Any idea why that is?

andrewazores:
Well, it works if I back out the "listen on concrete host" commit, which I guess makes sense. The container is defined to have hostname `container-jfr`, but the webserver is being told to listen on `localhost`. I'll think about how to solve this while still 1) giving the Docker container a meaningful name, 2) keeping the service accessible on the host machine without some additional Docker DNS tricks

andrewazores:
Maybe for the purposes of the run.sh script and testing, it's sufficient to just set `    CONTAINER_JFR_WEB_HOST="0.0.0.0"`?

tabjy:
@andrewazores This sounds rather like a misconfiguration. The `run.sh` explicitly set `-e CONTAINER_JFR_WEB_HOST=localhost`. I would recommend leaving this env var unset. And yes, setting `-e CONTAINER_JFR_WEB_HOST="0.0.0.0"` is sufficient enough to make it listens on all interfaces and hostname.

andrewazores:
Yea, makes sense. Could you add a commit to fix that?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -0,0 +1,327 @@
...
+        // blocking function, must be called from a blocking handler
+        byte[] buff = new byte[WRITE_BUFFER_SIZE]; // 64 KB
+        int n;
+        while (true) {
+            n = inputStream.read(buff);
+            if (n == -1) {
+                break;
+            }
+
+            response.write(Buffer.buffer().appendBytes(buff, 0, n));
```
andrewazores: I'm not familiar with vertx's stream writing pieces, but this looks suspicious. Why is a new buffer (`Buffer.buffer()`) being allocated on each iteration of the loop?

And, is there no way to make this non-blocking since the source is a stream?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -0,0 +1,327 @@
...
+        try {
+            Optional<InputStream> recording = getRecordingInputStream(recordingName);
+            if (recording.isEmpty()) {
+                throw new HttpStatusException(404, String.format("%s not found", recordingName));
+            }
+
+            ctx.response().setChunked(true);
+            ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, MIME_TYPE_OCTET_STREAM);
+            writeInputStream(recording.get(), ctx.response())
+                    .endHandler((e) -> downloadCounts.merge(recordingName, 1, Integer::sum))
```
andrewazores: Fancy. :-)

src/test/java/com/redhat/rhjmc/containerjfr/net/web/WebServerTest.java
```diff
@@ -1,173 +1,133 @@
...
-import static org.mockito.Mockito.inOrder;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.verifyZeroInteractions;
-import static org.mockito.Mockito.when;
+package com.redhat.rhjmc.containerjfr.net.web;
 
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
```
andrewazores: Please undo the wildcard imports here

src/test/java/com/redhat/rhjmc/containerjfr/tui/ws/MessagingServerTest.java
```diff
@@ -1,19 +1,17 @@
 package com.redhat.rhjmc.containerjfr.tui.ws;
 
 import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.verifyZeroInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
andrewazores: Wildcard import

src/test/java/com/redhat/rhjmc/containerjfr/tui/ws/WsClientReaderWriterTest.java
```diff
@@ -2,33 +2,30 @@
...
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyZeroInteractions;
-import static org.mockito.Mockito.when;
-
-import java.io.IOException;
-import java.net.InetSocketAddress;
+import static org.mockito.Mockito.*;
```
andrewazores: Wildcard import

tabjy:
Fixed and pushed

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public boolean isAlive() {
+        return this.server.actualPort() != 0;
+    } 
+
+    public Vertx getVertx() {
+        return vertx;
+    }
+
+    private class HandlerDelegate<T> implements Handler<T> {
```
andrewazores: I get a SpotBugs warning due to this class:

```
Should com.redhat.rhjmc.containerjfr.net.HttpServer$HandlerDelegate be a _static_ inner class?

This class is an inner class, but does not use its embedded reference to the object which created it.  This reference makes the instances of the class larger, and may keep the reference to the creator object alive longer than necessary.  If possible, the class should be made static. 
```

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
andrewazores: (This isn't a complaint, I just like the `merge`)

tabjy: My IDE promoted this suggested change. Very fancy indeed. XD

src/test/java/com/redhat/rhjmc/containerjfr/net/web/WebServerTest.java
tabjy: Will do.

src/test/java/com/redhat/rhjmc/containerjfr/tui/ws/MessagingServerTest.java
tabjy: Will do.

src/test/java/com/redhat/rhjmc/containerjfr/tui/ws/WsClientReaderWriterTest.java
tabjy: Will do.

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
tabjy: I agree. HandlerDelegate will be changed to a static class.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
tabjy: Buffer allocation will be moved outside the loop in the next commit.

The source is a stream, but a `java.io` blocking stream. `read()` calls are guaranteed to return immediately. Although there is a way to convert a blocking `InputStream` into a `java.nio` async `ReadableByteChannel`, but such a method conversion cannot utilize epoll/select/kqueue and signals and achieve true asynchronous, but instead creating a worker thread that still calls blocking functions. (I'm not 100% sure about this.) Additionally, since the report generator is a synchronous API anyway, I chose the straight forward way of using blocking handlers.

andrewazores: This is used when the client attempts to download a recording file though, isn't it? If it's made asynchronous, even if backed by a blocking stream, it will allow for data to flow out over the network and be in-flight as the rest of the recording is loaded. It would also reduce the memory footprint of sending out the recording, since it wouldn't need to be copied entirely into a buffer before being handed off to the network stack.

tabjy: If that's the concern, the data is already being streamed to clients with HTTP chunked transfer encoding. Every call to `.write` will queue a 64 KB chunk (in FIFO order of course) to send on the next appropriate event loop tick.

And this reminds me why I was keeping allocating `Buffer.buffer()` in the loop: if the content of a buffer is rewritten with the next chunk before it is sent out, the file client receives will be corrupted. So there must be a new buffer allocated **inside the loop** for every new chunk.

tabjy: Or I can code it so it only reads a new chunk after the last chunk is sent out. In this case only one 64 KB buffer is needed.

andrewazores: I see - it seems odd that the API uses the `.write` calls to also delineate the chunks within the write queue, but that is what the docs say so I suppose it's how it's done.

Using a single buffer and only refilling it after the previous contents have been flushed will give us a constant memory size footprint, which is very nice, but also has the potential to reduce the network throughput (although also lowering initial response latency). Maybe this is worth trying out. Or, could you implement both and make the selection configurable by a property or env var? I think we'll want to test both solutions in the real-world for a while and see what works best. It may even be something we leave for users to decide.

andrewazores: Also, while looking through the vertx API, I noticed the `.sendFile` method. Since the requested recording can be either in-memory or on-disk, this is something that we can/should take advantage of. That information about whether the recording is in-memory or on-disk is currently lost by the `getRecordingInputStream` implementation, but I think it's worth refactoring things so that this distinction is preserved and the appropriate API used. This would make the download of on-disk recordings *way* more efficient, so the cost vs. benefit analysis of how we treat the buffers in the discussion above would only apply to in-memory recordings.

In the future it might even make sense to drop the ability to directly download files out of memory without first flushing the recording to disk storage.

tabjy: No problem. This will be implemented in the next commit.

tabjy: I have implemented a synchronous method writing an `InputStream`. It can be enabled by defining `USE_SYNC_WRITE_INPUT_STREAM` environment variable, regardless of the value.

`.sendFile()` is very efficient because the copying of the content between two fds is directly handled by the kernel. The actual content doesn't route around to userspace. However, it's worth noting this advantage doesn't apply if we enable end-to-end SSL encryption in the future, as the encryption process still needs to be done in userspace.

tabjy: I have implemented a synchronous method writing an InputStream. It can be enabled by defining `USE_SYNC_WRITE_INPUT_STREAM` environment variable, regardless of the value.

`.sendFile()` is very efficient because the copying of the content between two fds is directly handled by the kernel. The actual content doesn't route around to userspace. However, it's worth noting this advantage doesn't apply if we enable end-to-end SSL encryption in the future, as the encryption process still needs to be done in userspace.

andrewazores: That's a good point about SSL. It will be something to take into consideration since flight recordings can easily contain sensitive information.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -223,6 +225,53 @@ private void endWithJsonKeyValue(String key, String value, HttpServerResponse re
...
+                        future.completeExceptionally(res.cause());
+                        return;
+                    }
+                    run(); // recursive call on this runnable itself
+                });
+            }
+        }).run();
+
+        try {
+            future.join();
```
andrewazores: This looks like it's going to do what we talked about, but I can imagine this easily becoming confusing in the future when someone goes back to look at this implementation. In particular, I think it should be pointed out that the first execution of the runnable is done by whatever thread is handling the request, while the subsequent invocations are actually going to be run by some other thread specified by the HttpServerResponse and however it handles its internal write queue. This actually seems a bit off to begin with - would it make sense to submit the initial Runnable to an ExecutorService, and then in the `write` completion handler, submit the subsequent re-run to the same executor? ie the only things done by the vertx write queue handler are `future.completeExceptionally` and `executor.submit(this)`, with all of the buffer/chunk management and handing off the chunk data to the write queue done by the same ExecutorService thread. You could also use `shutdown/awaitTermination` on the ExecutorService to replace your explicitly managed `future`.

andrewazores: Actually maybe you still would need the `future`, since I don't know how else you would nicely signal exceptional statuses.

tabjy: You're right. Sorry, I was lacking some consideration there, and thanks for pointing out using `ExecutorService`. I didn't eliminate the usage of `future` as `.awaitTermination()` requires a timeout parameter, and such timeout can not be easily determined since the recording size and network condition is unknown. (And having a timeout makes little senses if we are streaming a file.)

andrewazores: Fair enough. The timeout could have simply been arbitrarily large, like Long.MAX_VALUE * TimeUnit.DAYS (or maybe be a little more sensible), but the implementation you have here now looks good.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -0,0 +1,384 @@
...
+                throw new HttpStatusException(404, String.format("%s not found", recordingName));
+            }
+
+            ctx.response().setChunked(true);
+            ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, MIME_TYPE_OCTET_STREAM);
+            ctx.response().endHandler((e) -> downloadCounts.merge(recordingName, 1, Integer::sum));
+
+            if (System.getenv("USE_SYNC_WRITE_INPUT_STREAM") != null) {
+                writeInputStreamSync(recording.get(), ctx.response());
+                logger.info("useWriteInputStreamSync");
```
andrewazores: Almost done with the review.

Since this check is based on an environment variable, which is constant after JVM startup, could you log the strategy used just once at WebServer creation or startup?

And, since both methods are blocking calls and really synchronous to the same degree (ie async only happens within the Response's write queue handling, but we block until all the data is copied there), I think this env var and the "sync" version of the method could use a rename.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
andrewazores: I think there's still some optimization to be done to make the response stream the data better using a vertx "Pump", but this is an optimization we can look at later. I'll mark this as resolved for now and we can file an issue for exploring performance optimization.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
tabjy: I agree. Will move to logging to the static initializer. Also, I'm wondering if you have any suggestions for renaming...

andrewazores: Maybe something relating to "fixed size copy buffer"?

andrewazores: If we're going to leave this as a configurable option then the name should be more indicative of the result than the implementation detail, so maybe the environment variable name should relate to "low memory pressure streaming", or some other keywords like that.

As for the names of the methods themselves, perhaps it could reflect that same environment variable naming. I'll try to think of some other options. "Sequential" came to mind, but I don't think that's really any more accurate than "synchronous" for this situation.

tabjy: Please see commit 23820bc.

andrewazores:
Looking great. Please document the low pressure env var in the README, and also add a note in there that the LISTEN_PORT variables do not apply to WebSocket mode (raw TCP socket mode only).

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -73,6 +74,12 @@
         this.savedRecordingsPath = savedRecordingsPath;
         this.logger = logger;
         this.reportGenerator = reportGenerator;
+
+        if (env.getEnv(USE_LOW_MEM_PRESSURE_STREAMING_ENV) != null) {
```
andrewazores: I'd rather use `hasEnv` here than a null-check. Setting the variable to the empty string should still count as leaving it disabled, IMO. `hasEnv` checks for null-ness and blank-ness.

You could then also add a `-e` in run.sh for this variable.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -0,0 +1,389 @@
...
+
+        router.get("/grafana_datasource_url").handler(this::handleGrafanaDatasourceUrlRequest)
+                .failureHandler(failureHandler);
+
+        router.get("/grafana_dashboard_url").handler(this::handleGrafanaDashboardUrlRequest)
+                .failureHandler(failureHandler);
+
+        router.get("/recordings/:name").blockingHandler(ctx -> {
+            String recordingName = ctx.pathParam("name");
+            if (recordingName != null && recordingName.indexOf(".jfr") == recordingName.length() - 4) {
```
andrewazores: This looks unsafe. If I have a recording with a name of length 3, then `recordingName.indexOf(".jfr")` must be `-1` (because that substring is longer than the whole recording name), and `recordingName.length() - 4` is also `-1`. So we fall into this if-statement and attempt to take the substring of the name from 0 to -1.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
tabjy: I'm aware of the existant of `.hasEnv()`, but I was thinking any non-null value, including an empty string, should be considered to enabling such feature.

But yes, I'll change it to use `.hasEnv()` and keep things consistent.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
tabjy: Ah, nice catch. I knew, `.indexOf()` returns `-1` if not found, but didn't realize this flaw. I'll address this problem in the next commit.

andrewazores: You can probably just use `.endsWith(".jfr")`

run.sh
```diff
@@ -57,4 +57,5 @@ docker run \
     -e CONTAINER_JFR_EXT_LISTEN_PORT=$CONTAINER_JFR_EXT_LISTEN_PORT \
     -e GRAFANA_DATASOURCE_URL=$GRAFANA_DATASOURCE_URL \
     -e GRAFANA_DASHBOARD_URL=$GRAFANA_DASHBOARD_URL \
+    -e USE_LOW_MEM_PRESSURE_STREAMING=$USE_LOW_MEM_PRESSURE_STREAMING
```
andrewazores: This is missing the backslash at the end of the line

tabjy: oops...

PR #67, opened 2019-11-12
Secure communication channels with SSL/TLS

### LINKED ISSUE #9: Communication channels must be secured
HTTPS and WSS should be used, not plain HTTP and WS.

tabjy:
This pull request implements securing HTTP and WS with SSL/TLS. The usage is documented in README.md.

Closes #9.


run.sh
```diff
@@ -58,4 +58,8 @@ docker run \
     -e GRAFANA_DATASOURCE_URL=$GRAFANA_DATASOURCE_URL \
     -e GRAFANA_DASHBOARD_URL=$GRAFANA_DASHBOARD_URL \
     -e USE_LOW_MEM_PRESSURE_STREAMING=$USE_LOW_MEM_PRESSURE_STREAMING \
+    -e KEYSTORE_PATH=KEYSTORE_PATH \
+    -e KEYSTORE_PASS=KEYSTORE_PASS \
+    -e KEY_PATH=KEY_PATH \
+    -e CERT_PATH=CERT_PATH \
```
andrewazores: These new lines are setting the values to the literal names of the variables, ex. the variable `CERT_PATH` will have the value `"CERT_PATH"`. Is this intentional?

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
```diff
@@ -19,18 +24,92 @@
...
+        this.sslEnabled = setSslOptions(env, options);
+
+        this.server = vertx.createHttpServer(options);
+    }
+
+    private boolean setSslOptions(Environment env, HttpServerOptions options) {
+        String keystorePath = null;
+        if (env.hasEnv(KEYSTORE_PATH_ENV)) {
+            keystorePath = env.getEnv(KEYSTORE_PATH_ENV);
+        } else if (new File(System.getProperty("user.home"), "container-jfr-keystore.jks").exists()) {
```
andrewazores: Rather than `new File(foo).exists()`, it would be preferable IMO to use NIO APIs, ex. `Files.exists(Path.get(System.getProperty("user.home"), somePath))`. Although, this is another feature that should probably be added to the `container-jfr-core` `Filesystem` class to allow testing.

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
```diff
@@ -19,18 +24,92 @@
...
+        } else if (keyPath != null ^ certPath != null) {
+            IllegalArgumentException e = new IllegalArgumentException("both a key and a certificate are required");
+            logger.error(e);
+            throw e;
+        }
+
+        logger.warn("SSL parameters not set. Fallback to plain HTTP.");
+        options.setSsl(false);
+        return false;
+    }
```
andrewazores: Overall, this is quite a long method with a lot of branching logic. Can this be split into smaller methods, and tests added?

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
```diff
@@ -19,18 +24,92 @@
...
         this.logger = logger;
         this.vertx = Vertx.vertx();
-        this.server = vertx.createHttpServer(new HttpServerOptions()
+
+        HttpServerOptions options = new HttpServerOptions()
                 .setCompressionSupported(true)
-                .setLogActivity(true)
-        );
+                .setLogActivity(true);
+        this.sslEnabled = setSslOptions(env, options);
```
andrewazores: Is the return value here useful? It seems like `sslEnabled` should always have the same value as `options.isSsl()`.

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
```diff
@@ -19,18 +24,92 @@
...
+    private boolean setSslOptions(Environment env, HttpServerOptions options) {
+        String keystorePath = null;
+        if (env.hasEnv(KEYSTORE_PATH_ENV)) {
+            keystorePath = env.getEnv(KEYSTORE_PATH_ENV);
+        } else if (new File(System.getProperty("user.home"), "container-jfr-keystore.jks").exists()) {
+            keystorePath = System.getProperty("user.home") + File.separator + "container-jfr-keystore.jks";
+        } else if (new File(System.getProperty("user.home"), "container-jfr-keystore.pfx").exists()) {
+            keystorePath = System.getProperty("user.home") + File.separator + "container-jfr-keystore.pfx";
+        } else if (new File(System.getProperty("user.home"), "container-jfr-keystore.p12").exists()) {
+            keystorePath = System.getProperty("user.home") + File.separator + "container-jfr-keystore.p12";
```
andrewazores: The keystorePath should probably be an actual `Path` type rather than a String, and methods like `Path.resolve()` used to construct the path rather than appending separators and filenames.

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
andrewazores: Actually, this seems like it probably even warrants its own "SSLConfiguration" class or something.

run.sh
tabjy: No. This is a copy-paste error. I'll fix them in the next commit.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -0,0 +1,207 @@
+package com.redhat.rhjmc.containerjfr.net;
+
+import com.redhat.rhjmc.containerjfr.core.log.Logger;
+import com.redhat.rhjmc.containerjfr.core.sys.Environment;
+import com.redhat.rhjmc.containerjfr.core.sys.FileSystem;
+import io.vertx.core.http.HttpServerOptions;
+import io.vertx.core.net.JksOptions;
+import io.vertx.core.net.PemKeyCertOptions;
+import io.vertx.core.net.PfxOptions;
+import org.openjdk.jmc.common.util.Pair;
```
andrewazores: I'm trying to decouple container-jfr from direct dependencies on JMC classes[0]. Could you replace this Pair type with 1) a homegrown implementation (should be simple enough), 2) an Apache commons-lang Pair, 3) an abused AbstractMap.SimpleEntry, 4) a Map or some other structure

[0] some of this currently just means moving that dependency into container-jfr-core for potential independent reimplementation later

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+        logger.warn("No available SSL certificates. Fallback to plain HTTP.");
+    }
+
+    Path obtainKeyStorePathIfSpecified() throws IllegalArgumentException {
+        if (!env.hasEnv(KEYSTORE_PATH_ENV)) {
+            return null;
+        }
+
+        Path path = Path.of(env.getEnv(KEYSTORE_PATH_ENV)).normalize();
+        if (!fs.exists(keystorePath)) {
```
andrewazores: I think we're missing a corresponding PR to -core to add this `exists` method.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+        {
+            Path path = obtainKeyStorePathIfSpecified();
+            if (path != null) {
+                keystorePath = path;
+                keystorePass = env.getEnv(KEYSTORE_PASS_ENV, "");
+                keyPath = null;
+                certPath = null;
+                return;
+            }    
+        }
```
andrewazores: This is much easier to read than it was before, but these anonymous blocks with early returns are still a little bit funky to read. Maybe this is a sign that the class could use some further refactoring to use something like a strategy pattern?

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+
+        if (keyPath == null || certPath == null) {
+            return null;
+        }
+
+        return ret;
+    }
+
+    HttpServerOptions setToHttpServerOptions(HttpServerOptions options) {
```
andrewazores: `setTo` sounds kinda funny. `applyTo` maybe? Not too important.

If it mutates the HttpServerOptions that is passed in to it, does it need to also return back that same reference?

This is another place that looks like it could use some refactoring - there's a good chunk of nested branching logic in here.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+                return options
+                        .setKeyStoreOptions(new JksOptions().setPath(keystorePath.toString()).setPassword(keystorePass));
+            } else if (keystorePath.endsWith(".pfx") || keystorePath.endsWith(".p12")) {
+                return options
+                        .setPfxKeyCertOptions(new PfxOptions().setPath(keystorePath.toString()).setPassword(keystorePass));
+            }
+
+            IllegalArgumentException e = new IllegalArgumentException("unrecognized keystore type");
+            logger.error(e);
+            throw e;
```
andrewazores: Creating a runtime exception, logging it, and then throwing it seems weird. IMO this should be a checked exception and whoever is responsible for catching it should then also deal with logging it. And `IllegalArgumentException` generally seems like not quite the right fit either, since it's being thrown not based on the arguments passed to the method throwing the exception but based on some previously set internal state of the object. For existing runtime exceptions I can think of, IllegalStateException sounds more applicable.

Maybe this warrants a new, checked, `SslConfigurationException` for this purpose.

tabjy:
Sorry, I should have marked my commits WIP. I'm currently working on unit tests for `SslConfiguration`.
 Also, I'm planning to open a PR to [rh-jmc-team/container-jfr-core](https://github.com/rh-jmc-team/container-jfr-core) adding some methods to `FileSystem` class.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: Yes, it mutates the `HttpServerOptions` passed it. `HttpServerOptions` implements fluent API so what's returned and what's passed in are actually the same reference.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: I agree. A new checked `SslConfigurationException` sounds desirable. I'll create and switch to that instead.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: Thanks for the reminder. I have a PR ready and will submit once ready.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: Replaced with `org.apache.commons.lang3.tuple.Pair`

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: I adopted the strategy pattern in the latest version. Please confirm I'm using it correctly. Thanks!

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: Refactored and adopted the strategy pattern.

tabjy:
TODO: Test SSL Strategies.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -145,74 +118,117 @@ Path discoverKeyStorePathInDefaultLocations() {
...
+
+    HttpServerOptions applyToHttpServerOptions(HttpServerOptions options) {
+        return strategy.applyKeyStoreOptions(options);
+    }
+
+    boolean enabled() {
+        return strategy.enabled();
+    }
+
+    SslConfigurationStrategy keyStoreStrategy(Path path, String password) throws SslConfigurationException {
```
andrewazores: Do you have some specific future plans for this method and the others like it? They look like they can just be inlined as it is now.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
```diff
@@ -145,74 +118,117 @@ Path discoverKeyStorePathInDefaultLocations() {
...
+    
+    SslConfigurationStrategy noSslStrategy() {
+        return new NoSslStrategy();
+    }
+
+    interface SslConfigurationStrategy {
+        HttpServerOptions applyKeyStoreOptions(HttpServerOptions options);
+        
+        default boolean enabled() {
+            return !(this instanceof NoSslStrategy);
```
andrewazores: I think this is a little bit *too* clever :-). Rather than defaulting to an `instanceof` check, how about simply defaulting to return `true`, and then the `NoSslStrategy` can override that to return `false`?

tabjy: Yes, that sounds better. I'll update as suggested.

src/main/java/com/redhat/rhjmc/containerjfr/net/SslConfiguration.java
tabjy: No. As far as I know, these are all what vertx supports. I'll inline these two calls.

src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
tabjy: This is outdated. Such a field is no longer used.

andrewazores:
Looks good. I'll do a few rounds of manual testing in various scenarios and if everything checks out, I'll merge it. Thanks for the good work here.

andrewazores:
Cool. Running without SSL as a local JVM process and in Docker work normally, as expected. Then I generated a self-signed .pem, named it accordingly within my $HOME, and ran as a JVM process, and it worked perfectly. Very nice.

PR #78, opened 2019-11-29
File autorename

### LINKED ISSUE #56: When saving files into archive, filenames should be automatically updated to avoid clobbering old data
Example scenario:

User has ongoing in-memory recording named foo from target app.svc.local. Saving this recording to archive produces a file on disk named "app-svc-local_foo.jfr". Currently, saving the recording again will overwrite "app-svc-local_foo.jfr". This should not be the case.

1) A timestamp should be attached to saved filenames, ex. "app-svc-local_foo_20191105T090738.jfr"

2) If two recordings happen to have the same filename, whether from rapid repeated saving of the same active recording or re-upload or any other reason, then some differentiating suffix should be applied to the filename such as an incrementing count of the number of copies ("local_foo_20191105T090738_1.jfr")

andrewazores:
Fixes #56

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/SaveRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/SaveRecordingCommand.java
@@ -4,19 +4,20 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.StandardCopyOption;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Optional;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
-import org.openjdk.jmc.rjmx.services.jfr.FlightRecorderException;
-import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
-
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/SaveRecordingCommand.java
```diff
@@ -88,16 +92,29 @@ public boolean isAvailable() {
...
+        String destination = String.format("%s_%s_%s", targetName, recordingName, timestamp);
+        // TODO byte-sized rename limit is arbitrary. Probably plenty since recordings are also differentiated by second-resolution timestamp
+        byte count = 1;
+        while (fs.exists(recordingsPath.resolve(destination + ".jfr"))) {
+            destination = String.format("%s_%s.%d", targetName, recordingName, count++);
+            if (count == Byte.MAX_VALUE) {
+                throw new IOException("Recording could not be saved. File already exists and rename attempts were exhausted.");
+            }
+        }
+        if (!destination.endsWith(".jfr")) {
```
tabjy: This check seems redundant? Looking at L103 and L100, it's impossible for `destination` to end with `.jfr`.

andrewazores: Very true. Added and then forgot to remove that in an earlier version of this changeset. I'll pull it out in a minute.

PR #90, opened 2020-01-02
Allow re-uploading recordings from clients

tabjy:
This patch allows JFR recordings to be (re-)uploaded to container-jfr from clients. The implement expects a recording with its name in form of *`<target-name>_<recording_name>_<timestamp>[.\d+].jfr`* and rejects otherwise.

**RFC**: Looking at the JMC API, I was unable to find a way which just simply validates a recording. Therefore, I called `JfrLoaderToolkit.loadEvents()` to verify the file uploaded is a valid JFR recording. This could lead to degraded performance.  

Closes rh-jmc-team#55. Blocked by rh-jmc-team/container-jfr-web#24 (needs sub-module commit hash).


tabjy:
@andrewazores I've split it up into smaller functions. Please see 3e72aca.

PR #97, opened 2020-01-03
Fix code comment in StartRecordingCommand

jiekang:
Adjusts the comment to be accurate. Another possibility is to remove these kinds of comments and make the variable names more verbose, but they can be useful, e.g. the events list sample. I don't have a strong opinion either way.


andrewazores:
Descriptive variable names are definitely better than relying on such comments, but at least with the current interface design, all commands accept a String[], so it's difficult to name this meaningfully. Some commands don't expect any args so that parameter could be named "unused" for example, which would be effective, but for ex. the "dump" command it would be tough to come up with a useful name IMO.

There is also the COMMANDS.md document which came about later and also serves as documentation for this purpose, but that suffers the same problem of requiring manual checks and updates that these comments do.

PR #101, opened 2020-01-13
#87: Auth dialog messages should have tailored messages

### LINKED ISSUE #87: Platform auth managers should have tailored messages for users
See rh-jmc-team/container-jfr-web#23

tabjy:
This is a work-in-progress PR: tests needed for `DocumentationMessageManager`.

An injectable `DocumentationMessageManager` provides a centralized registry for documentations messages, with localization in mind, for components like `AuthManager` and `WebServer`.

The `/documentation_messages` endpoint returns a dictionary in JSON format for the client to use. It looks through the `Accepted-Languages` header value and finds the best match.

Closes #87


andrewazores:
Are more tests planned? I see that it's WIP awaiting tests for the documentation message manager, but that you've also added some tests in the latest commit.

tabjy:
> Are more tests planned? I see that it's WIP awaiting tests for the documentation message manager, but that you've also added some tests in the latest commit.

@andrewazores Yes, it is ready. Sorry, I forgot to update the pr status. My bad. :(

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -416,7 +420,7 @@ void handleAuthRequest(RoutingContext ctx) {
     }
 
     void handleClientUrlRequest(RoutingContext ctx) {
-        ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, MIME_TYPE_JSON);
+
```
andrewazores: Why was this removal needed?

src/main/java/com/redhat/rhjmc/containerjfr/localization/LocalizationManager.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Add or overwrite a message of the given key
+     *
+     * @param locale the locale of the message being added
+     * @param key the unique message key
+     * @param message the message content
+     */
+    public void putMessage(Locale locale, String key, String message) {
+        messages.putIfAbsent(locale, new HashMap<>());
```
andrewazores: It would probably make sense to return a boolean indicating whether the `put` succeeded (ie value was absent).

src/main/java/com/redhat/rhjmc/containerjfr/documentation_messages/DocumentationMessageModule.java
```diff
@@ -0,0 +1,17 @@
+package com.redhat.rhjmc.containerjfr.documentation_messages;
```
andrewazores: Using an underscore in a package name is a bit unconventional. I would prefer `documentationmessages`, or ideally something a bit more concise. IMO it would make sense to lump this in to a `localization` package or something like that.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
tabjy: I'm afraid you're looking at an out-dated commit. I had this line removed by mistake. It was re-added later.

src/main/java/com/redhat/rhjmc/containerjfr/localization/LocalizationManager.java
tabjy: Sorry, I don't understand here. The put operation is always gonna success. What we have here is essentially a 2-D map, and this line it for creating a new inner map if one hasn't been created yet.

src/main/java/com/redhat/rhjmc/containerjfr/documentation_messages/DocumentationMessageModule.java
tabjy: Should `DocumentationManager` and `DocumentationManagerModule` change to `LocalizationManager` and `LocalizationModule` respectively?

tabjy: And should I update the name in the responding container-jfr-web pr?

andrewazores: Yea, I think that all makes sense.

andrewazores:
Upon a re-read, I think there are really just two changes I'd like to see before merging.

First and very minor one is just the package and class rename we discussed above.

Second is a little more involved. Currently, the localization strings are added into the localization manager by anyone who consumes that API - ex. the auth managers which have the manager injected into themselves so that they can insert a message back into the manager. I think we should avoid this as much as possible. At the very least, the message added to the manager should be written somewhere like a `messages.properties` resource file, not in source code. But IMO ideally those property files should be read and the contents merged into the localization manager either by the localization manager itself or some other class written for this express purpose only. It probably makes sense to split up the property files by Dagger module and/or Java package, so perhaps there's some useful way to do it so that a shared common message loader class can be written, and then one instance of each added to each module to load that module's messages. The localization manager could then have the set of all such loaders injected into itself and merge all of their messages.

PR #105, opened 2020-01-22
Configurable authmanager

andrewazores:


andrewazores:
Re-assigned review to @jiekang . You probably already looked over this part when reviewing #107 , which has been based on top of this one since the start.

PR #107, opened 2020-01-23
Basic authmanager

### LINKED ISSUE #103: Implement a "basic" auth manager
Currently, the only auth manager implementations are "noop" (permissive) and "OpenShift" (delegates to the OpenShift API). There should be an auth manager implementation that instead allows some configuration of basic key-value pairs for usernames and password hashes. Ideally this should allow for more than one such user account to be configured. This could be via some environment variables following some predefined naming convention, or by some config file. Combined with #102 , this would allow a scenario like basic auth in OpenShift, and enabling auth for the desktop app or simple Docker deployment scenarios.

andrewazores:
Fixes #103

Based on top of currently open PR #105

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
@@ -71,7 +71,7 @@ class BasicAuthManager extends AbstractAuthManager {
     public Future<Boolean> validateWebSocketSubProtocol(Supplier<String> subProtocolProvider) {
         String subprotocol = subProtocolProvider.get();
         if (subprotocol == null) {
-            return null;
+            return CompletableFuture.completedFuture(false);
         }
         Pattern pattern =
                 Pattern.compile(
@@ -79,7 +79,7 @@ class BasicAuthManager extends AbstractAuthManager {
                         Pattern.CASE_INSENSITIVE);
         Matcher matcher = pattern.matcher(subprotocol);
         if (!matcher.matches()) {
-            return null;
+            return CompletableFuture.completedFuture(false);
         }
         String b64 = matcher.group(1);
         try {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
@@ -23,15 +23,15 @@ class BasicAuthManager extends AbstractAuthManager {
 
     static final String USER_PROPERTIES_FILENAME = "container-jfr-users.properties";
 
-    private final Properties users;
     private final FileSystem fs;
+    private final Properties users;
 
     // TODO inject FileSystem, but this also means changing the assumed constructor signature
     // TODO salted hashes
-    BasicAuthManager(Logger logger) {
+    BasicAuthManager(Logger logger, FileSystem fs) {
         super(logger);
+        this.fs = fs;
         this.users = new Properties();
-        this.fs = new FileSystem();
         loadConfig();
     }
 
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
@@ -27,7 +27,6 @@ class BasicAuthManager extends AbstractAuthManager {
     private final Properties users;
     private volatile boolean configLoaded = false;
 
-    // TODO inject FileSystem, but this also means changing the assumed constructor signature
     // TODO salted hashes
     BasicAuthManager(Logger logger, FileSystem fs) {
         super(logger);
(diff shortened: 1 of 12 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/net/BasicAuthManager.java
```diff
@@ -0,0 +1,121 @@
...
+
+    // TODO salted hashes
+    BasicAuthManager(Logger logger, FileSystem fs) {
+        super(logger);
+        this.fs = fs;
+        this.users = new Properties();
+    }
+
+    @Override
+    public Future<Boolean> validateToken(Supplier<String> tokenProvider) {
```
jiekang: Would it be overkill to operate on character arrays for anything dealing with the incoming credentials and nulling out the password characters immediately after use?

jiekang: Er, overwriting with zeroes rather than nulling in this case.

jiekang: Actually it's coming in from a library anyways, it's probably overkill.

andrewazores: It would be doable, but I wouldn't consider that to be too much of a security improvement, if that's the concern. It narrows the timing window where the password is in memory, but not considerably so, and doesn't do anything about the other places in memory where the password had to pass through first - for example, the webserver's HTTP header we parsed it out of. 

And if an attacker has the ability to read your ContainerJFR credentials out of JVM memory, then they probably already have access to whatever targets your ContainerJFR would have connected to and more...

jiekang:
When I run
```
CONTAINER_JFR_AUTH_MANAGER='com.redhat.rhjmc.containerjfr.net.BasicAuthManager' ./gradlew run
```

and visit the web ui via browser, an Auth Token pop-up appears asking for a valid access token. What should I enter to get through here?

andrewazores:
https://github.com/rh-jmc-team/container-jfr-web/issues/38

At the moment, there's nothing you can do about it from -web, you'll just be locked out when using basic auth. This change would need to be tested with CLI tools, ex. `curl`/`wget` and `websocat`.

ex. `websocat -kE --protocol basic.authorization.containerjfr.$(echo foo:bar | base64) wss://localhost:8181/command`

jiekang:
> [rh-jmc-team/container-jfr-web#38](https://github.com/rh-jmc-team/container-jfr-web/issues/38)
> 
> At the moment, there's nothing you can do about it from -web, you'll just be locked out when using basic auth. This change would need to be tested with CLI tools, ex. `curl`/`wget` and `websocat`.
> 
> ex. `websocat -kE --protocol basic.authorization.containerjfr.$(echo foo:bar | base64) wss://localhost:8181/command`

If I don't deploy container-jfr with SSL, then I should connect to ws:// instead correct? The basic auth should still be present there, right? What's the expected output of this command if the credentials are correct? Incorrect? And if the properties file does not exist?

andrewazores:
Correct, SSL being enabled and which authmanager are selected are independent, so you would use `ws` or `http` to connect if you don't have SSL configured.

If the provided credentials are correct then curl/wget should give you a 401. container-jfr would drop the websocket connection on the first message, since the WebSocket protocol doesn't actually deliver the auth header to us on the handshake but with each message instead, so if you use websocat and use the -E flag then you would see the connection simply drop if you had bad auth and tried to send any message. If the properties file does not exist then you'll see a log message about that from container-jfr, and the auth manager will have an empty mapping of users, so any credentials you provide will be handled as invalid ones.

andrewazores:
I should also explain the justification for doing "basic" auth via the subprotocol header here, rather than the more typical proper HTTP-style basic authorization header, which would also support URL credentials (`ws://user:pass@localhost:8181/command`). The reason being that the primary expected authentication method is auth tokens, and currently in browser implementations of WebSockets there is no way to add custom headers or Authorization headers to requests to include a Bearer token. So, the existing implementation of token-based auth in container-jfr (and -web and -operator) relies on the subprotocol technique for passing the auth token, and this "basic" auth manager simply follows that pattern for some semblance of API consistency between auth managers.

However, a good argument could be made that doing basic auth in this way would be surprising for anyone else trying to target this API. Although a browser client can't add custom headers or specify the Authorization header directly, they should be able to add credentials in the URL, which the browser should recognize and convert into the appropriate `Authorization: Basic` header. The -operator would also be able to do this. This would be more in line with typical expectations of basic auth, but also make things a little more muddy on the client side when determining how the credentials/token should be sent.

It would be easy enough to fix that though - the provider passed in to `validateWebSocketSubProtocol` by the MessagingServer would just need to grab the headers, rather than the subprotocol, and `validateWebSocketSubProtocol` would be renamed process the headers instead. The credentials validation steps would all remain the same once the credentials are parsed out of the subprotocol/headers.

andrewazores:
Rebased, no changes.

PR #113, opened 2020-01-28
upload-recording command enhancements and fixes

andrewazores:
These commits add tests for the upload-recording command, fix a bug where in-memory recordings were closed after upload, fix a validation failure output message format bug, and add buffering to disk reads for better upload speeds when uploading archived recordings.


andrewazores:
Rebased with no changes. `gradlew check` still passing

PR #117, opened 2020-02-03
Add "id" field to command and response messages

### LINKED ISSUE #116: Messages/responses should be uniquely identified
See discussion in #114 . Currently, messages do not explicitly contain any identifying information about the sending client (though the server can of course tell which socket connection received a message), and all responses to such messages are multicast out to all clients, with no indication for the clients whether they are receiving a broadcast or a direct response to one of their messages. Due to the overall design and architecture it does not make sense at this time to move to only using unicast responses, so including some kind of client/message identifier on client messages which is included in the corresponding broadcast response will allow clients to selectively react to responses.

andrewazores:
id is copied from command request messages into the corresponding
response (assuming that the command message was well-formed and
understood) so that clients can provide some unique ID on request, and
identify the correct associated response in the interleaved broadcasts
from the server

Fixes #116


andrewazores:
Yea, clients will only see a null ID on a message if they or some other client sent a message with a null/missing ID. For clients that care about that, like the -operator when it's trying to automate some workflow, it should always include a message ID and ignore any message it sees without that matching ID.

PR #127, opened 2020-02-06
Update build requirements in README.md

jiekang:
Addresses https://github.com/rh-jmc-team/container-jfr/issues/108

It's a tiny bit weird since node bundles a specific version of npm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,9 @@ Build:
 - Git
 - JDK11+
 - Gradle
+- Podman
+- npm 6+
+- Node 12+
 
 Run:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,6 +19,9 @@ Build:
 - Git
 - JDK11+
 - Gradle
+- Docker
```
andrewazores: Docker is no longer used for the image builds actually, I just recently changed it to use podman.

jiekang: Ah shoot; I did a git pull to check for changes but forgot the default was my fork, not upstream...

My bad.

jiekang: I've force-pushed an updated commit that specifies Podman instead.

PR #129, opened 2020-02-27
Replace gradle-based build setup with Maven

### LINKED ISSUE #128: Maven build
We should consider using Maven to build rather than Gradle.

Things to consider:
- `-web` needs to be built by invoking `npm`
- there needs to be some configuration flags or similar to allow building standard and minimal (no web)
- it would be nice to have tasks for building container images, but not strictly required (can be done by an external build script, Makefile, etc.)

andrewazores:
This patch set adds a Maven `pom.xml` which replaces the existing `build.gradle`
while retaining all major functionality and tooling. All gradle scripts and
configurations are removed and the README is updated to reflect the new build
tool.

Fixes #128


pom.xml
```diff
@@ -0,0 +1,378 @@
...
+                <workingDirectory>web-client</workingDirectory>
+                <arguments>
+                  <argument>-rf</argument>
+                  <argument>node_modules</argument>
+                  <argument>dist</argument>
+                </arguments>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
```
ebaron: Might I suggest we handle clean using maven-clean-plugin? Seems a bit more portable.

```suggestion
          </executions>
        </plugin>
        <plugin>
          <artifactId>maven-clean-plugin</artifactId>
          <version>3.1.0</version>
          <configuration>
            <filesets>
              <fileset>
                <directory>web-client</directory>
                <includes>
                  <include>node_modules/**</include>
                  <include>dist/**</include>
                </includes>
              </fileset>
            </filesets>
          </configuration>
        </plugin>
```

pom.xml
```diff
@@ -0,0 +1,378 @@
...
+
+<properties>
+  <java.version>11</java.version>
+  <maven.compiler.target>${java.version}</maven.compiler.target>
+  <maven.compiler.source>${java.version}</maven.compiler.source>
+  <imageBuilder>/usr/bin/podman</imageBuilder>
+  <containerjfr.minimal>false</containerjfr.minimal>
+  <containerjfr.imageStream>quay.io/rh-jmc-team/container-jfr</containerjfr.imageStream>
+  <containerjfr.webPort>8181</containerjfr.webPort>
+  <containerjfr.listenPort>9090</containerjfr.listenPort>
```
ebaron: Adding `<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>` to properties prevents a few warnings about guessing the encoding.

andrewazores:
Yea, I see that web-client build failure too. I'll take a look at it and open up a PR separately. It probably is just as you said, I need to cut a new -web release and update the submodule to point to that. For whatever reason, the currently pointed-to version has un-buildable deps.

pom.xml
```diff
@@ -0,0 +1,416 @@
...
+        <plugin>
+          <artifactId>maven-clean-plugin</artifactId>
+          <version>${org.apache.maven.plugins.clean.version}</version>
+          <configuration>
+            <filesets>
+              <fileset>
+                <directory>web-client</directory>
+                <includes>node_modules/**</includes>
+                <includes>dist/**</includes>
+              </fileset>
```
ebaron: This only cleans `dist` for me and not also `node_modules`. I think the directories need to be listed as `<include>` tags within `<includes>`, and not just `<includes>`.

PR #131, opened 2020-03-05
Use frontend plugin to build web-client

andrewazores:
The "frontend" plugin is a nice improvement over directly using the exec plugin
to invoke "npm" for building the web-client. For one thing, the intent is more
clear. But the real advantage is that the frontend plugin will automatically
download and install Node and npm binaries into the bin directory, removing
the need for developers (or CI servers) to install Node and npm in order to
build container-jfr. This also keeps versions consistent and makes builds more
reproducible. These additional binaries will be removed on `mvn clean`, and
are also cached in `.m2`, so the build time impact is very minimal after the first
run.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,8 @@
   <maven.compiler.source>${java.version}</maven.compiler.source>
 
   <imageBuilder>/usr/bin/podman</imageBuilder>
+  <node.version>v12.5.0</node.version>
+  <npm.version>6.13.4</npm.version>
 
   <containerjfr.minimal>false</containerjfr.minimal>
   <containerjfr.imageStream>quay.io/rh-jmc-team/container-jfr</containerjfr.imageStream>
@@ -31,6 +33,7 @@
   <org.apache.maven.plugins.info.reports.version>3.0.0</org.apache.maven.plugins.info.reports.version>
   <org.apache.maven.plugins.clean.version>3.1.0</org.apache.maven.plugins.clean.version>
   <org.apache.maven.plugins.resources.version>3.1.0</org.apache.maven.plugins.resources.version>
+  <com.github.eirslett.frontend.plugin.version>1.9.1</com.github.eirslett.frontend.plugin.version>
   <org.codehaus.mojo.exec.plugin.version>1.6.0</org.codehaus.mojo.exec.plugin.version>
   <com.google.cloud.tools.jib.maven.plugin.version>2.1.0</com.google.cloud.tools.jib.maven.plugin.version>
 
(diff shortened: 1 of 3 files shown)


andrewazores:
This is also based on top of #130

andrewazores:
Rebased

pom.xml
```diff
@@ -163,6 +166,65 @@
...
+              <argument>CONTAINER_JFR_WEB_HOST=0.0.0.0</argument>
+              <argument>--env</argument>
+              <argument>CONTAINER_JFR_LISTEN_HOST=0.0.0.0</argument>
+              <argument>--env</argument>
+              <argument>CONTAINER_JFR_WEB_PORT=${containerjfr.itest.webPort}</argument>
+              <argument>--env</argument>
+              <argument>CONTAINER_JFR_EXT_WEB_PORT=${containerjfr.itest.webPort}</argument>
+              <argument>--detach</argument>
+              <argument>--rm</argument>
+              <argument>${containerjfr.imageStream}:latest</argument>
```
ebaron: Should we use `${containerjfr.imageVersion}` instead of `latest` here? Also I had this task hang for me, but couldn't reproduce it. Not sure why.

```
INFO] --- jib-maven-plugin:2.1.0:dockerBuild (build-oci-image) @ container-jfr ---
[WARNING] Setting image creation time to current time; your image may not be reproducible.
[INFO] 
[INFO] Containerizing application to Docker daemon as quay.io/rh-jmc-team/container-jfr, quay.io/rh-jmc-team/container-jfr:0.16.0...
[WARNING] Base image 'gcr.io/distroless/java:11' does not use a specific image digest - build may not be reproducible
[INFO] Using base image with digest: sha256:c94feda039172152495b5cd60a350a03162fce4f8986b560ea555de4d276ce19
[INFO] 
[INFO] Container entrypoint set to [java, -XX:+CrashOnOutOfMemoryError, -Dcom.sun.management.jmxremote.port=9091, -Dcom.sun.management.jmxremote.ssl=false, -Dcom.sun.management.jmxremote.authenticate=false, -cp, /app/resources:/app/classes:/app/libs/*, com.redhat.rhjmc.containerjfr.ContainerJfr]
[INFO] 
[INFO] Built image to Docker daemon as quay.io/rh-jmc-team/container-jfr, quay.io/rh-jmc-team/container-jfr:0.16.0
[INFO] Executing tasks:
[INFO] [==============================] 100.0% complete
[INFO] 
[INFO] 
[INFO] --- exec-maven-plugin:1.6.0:exec (start-container) @ container-jfr ---
^C
```

andrewazores: Sure, I'll push a commit momentarily to use `imageVersion` rather than `latest`.

I've seen the same hang on start-container. I just filed an issue for it (#134). I think it's a podman thing, but I'm going to test that hypothesis by running the suite a few times using docker instead and see what happens.

PR #137, opened 2020-03-12
Include WWW-Authenticate Header on 401 responses

andrewazores:


andrewazores:
Actually, I'm marking this WIP - please don't bother reviewing it for now. I just tested -web against this and the behaviour isn't quite what I was looking for, since this causes the browser to display an auth prompt to the user. I need to evaluate if this is going to be useful...

andrewazores:
Tested and now working as expected with -web, too. The header used is now X-WWW-Authenticate (note the X- prefix), so that it doesn't trigger web browsers' built-in HTTP Basic client auth dialogs, while still clearly relaying the intent of the header information. This is used in rh-jmc-team/container-jfr-web#43 to handle Basic auth in addition to the already supported Bearer tokens (and no-op auth).

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -165,6 +165,11 @@ public void start() throws FlightRecorderException, SocketException, UnknownHost
                         logger.error(exception);
                     }
 
+                    if (exception.getStatusCode() == 401) {
+                        // Use X- prefix so as to not trigger web-browser auth dialogs
+                        ctx.response().putHeader("X-WWW-Authenticate", auth.getScheme().toString());
```
tabjy: Not sure if it's better put header name into a constant

tabjy:
LGTM. (Other than header name)

andrewazores:
Other than extracting the header name to a constant, did you have some other concern with its name?

PR #140, opened 2020-03-18
Implement remote target VM event template settings

andrewazores:


pom.xml
```diff
@@ -66,6 +65,18 @@
     <artifactId>containerjfr-core</artifactId>
     <version>${com.redhat.rhjmc.containerjfr.core.version}</version>
   </dependency>
+  <!-- TODO move this into -core -->
```
jiekang: This is from jmc -> jmc-core right? I think there are some comments in code below for container-jfr -> container-jfr-core.

andrewazores: Well, both. `container-jfr` shouldn't have any JMC dependencies, they should all be abstracted away into `container-jfr-core`. But also, a lot of those JMC dependencies are still "internal" JMC artifacts, and should be moved into jmc-core so that they can be properly published and consumed by other projects.

src/main/java/org/openjdk/jmc/ui/UIPlugin.java
```diff
@@ -0,0 +1,18 @@
+package org.openjdk.jmc.ui;
+
+import java.util.logging.Logger;
+
+/**
+ * TODO move this stub into -core This is only used for stubbing out a Logger instance used by JMC's
+ * JFCXMLValidator !!! This is a major classloading hackaround !!!
+ */
```
andrewazores: @jiekang here's a good RFE for JMC - the JFCXMLValidator currently imports this UIPlugin class from another artifact, which is a subclass of an Eclipse AbstractUIPlugin class. So those UIPlugin classes are actually somehow related to the rendered application components and widgets, but clearly the XML Validator and the rest of the XML processing stuff I'm using here are not. And yet, the JMC code pathway for processing XML calls into UI components simply for the purpose of reusing their static logger instance...

A fix for this is probably going to take a while to make it into a snapshot release version so I'll have to rely on this hack for a while, but if any portion of this XML stuff can be cleaned up and moved into jmc-core, then it would be really nice and useful for us to use here in container-jfr(-core).

pom.xml
jiekang: Ah right you are. Thanks!

src/main/java/org/openjdk/jmc/ui/UIPlugin.java
jiekang: Uh wow okay, that's quite a curiosity. This sounds like something warranting an issue opened in JMC Jira. If you agree I can write one up.

andrewazores: Yea, I would agree with that. The XML stuff does currently live within a `ui` maven module, although within a `model` package, so that part makes sense. But IMO it's definitely wrong for a model to be using a view or controller's logger.

andrewazores: And it's even worse than that really, because it isn't even the specific parent controller instance's logger or something - it's just a general static one on a class named UIPlugin...

andrewazores: I think I can work around this in a slightly less ugly way by using reflection to make some constructors visible and then not use the JFCXMLValidator at all. This would let us use JMC's XML parsing, which is nice, but JFC validation would be great to retain - especially when I get to #143 and users can upload their own .jfc files for defining new templates.

Maybe I can do those reflection tricks plus implement a JFC validator based on the JMC one that doesn't call out to a UIPlugin logger and pass that in rather than allowing the JMC classes to default to their own validator.

andrewazores: Nevermind - the `IXMLValidator` interface I'd need to implement is package-private, so making my own implementation doesn't really work. And the `JFCXMLValidator` class implementation is `final`. Well then.

andrewazores:
WIP label removed. There is a new command, `list-event-templates`, which retrieves the list of templates from the target JVM using the container-jfr-core TemplateService. The `AbstractRecordingCommand`, which both the `start` and `dump` commands subclass, is modified to also accept `template=Foo` syntax for the event specifier string. Finally, the `ALL` special event specifier string has been replaced with a pseudo-template with the name `Foo`, which is included in the results from the `list-event-templates` command and will also be accepted by the `template=All` template specifier syntax.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/AbstractRecordingCommand.java
```diff
@@ -7,11 +7,18 @@
...
+import com.redhat.rhjmc.containerjfr.core.templates.Template;
 import com.redhat.rhjmc.containerjfr.core.tui.ClientWriter;
 
 abstract class AbstractRecordingCommand extends AbstractConnectedCommand {
 
-    private static final Pattern ALL_EVENTS_PATTERN = Pattern.compile("^ALL$", Pattern.MULTILINE);
+    static final Template ALL_EVENTS_TEMPLATE =
+            new Template(
+                    "ALL",
+                    "Enable all available events in the target JVM, with default option values. This may be expensive and is intended primarily for testing ContainerJFR's own capabilities.",
```
jiekang: Heh, I'd replace may be expensive with "will be very expensive" :P

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/ListEventTemplatesCommandTest.java
```diff
@@ -0,0 +1,97 @@
...
+        Mockito.verifyZeroInteractions(cw);
+        cmd.execute(new String[0]);
+        InOrder inOrder = Mockito.inOrder(cw);
+        inOrder.verify(cw).println("Available recording templates:");
+        inOrder.verify(cw).println("\t[Foo Inc.]\tFoo:\ta foo-ing template");
+        inOrder.verify(cw).println("\t[Bar Inc.]\tBar:\ta bar-ing template");
+        inOrder.verify(cw).println("\t[Baz Inc.]\tBaz:\ta baz-ing template");
+        inOrder.verify(cw)
+                .println(
+                        "\t[ContainerJFR]\tALL:\tEnable all available events in the target JVM, with default option values. This may be expensive and is intended primarily for testing ContainerJFR's own capabilities.");
```
jiekang: Does this string need changing also?

jiekang:
Atm seeing test failures when running `mvn clean verify`:

```
[ERROR] Failures: 
[ERROR]   MessagingServerTest.shouldHandleRemovedConnections:145 
crw2.readLine();
Wanted 1 time:
-> at com.redhat.rhjmc.containerjfr.tui.ws.WsClientReaderWriter.readLine(WsClientReaderWriter.java:65)
But was 2 times:
-> at com.redhat.rhjmc.containerjfr.tui.ws.MessagingServer.lambda$addConnection$6(MessagingServer.java:123)
-> at com.redhat.rhjmc.containerjfr.tui.ws.MessagingServer.lambda$addConnection$6(MessagingServer.java:123)


[INFO] 
[ERROR] Tests run: 560, Failures: 1, Errors: 0, Skipped: 0

```

andrewazores:
Oh whoops. Kept that change locally apparently.

PR #148, opened 2020-04-09
Support for React-based web-client

andrewazores:
This allows an env var to be set to enable CORS headers, allowing the React dev server for the new web-client to target a separate ContainerJFR instance, and modifies the Maven plugin for preparing the client assets with the updated build location.

See rh-jmc-team/container-jfr-web#50


andrewazores:
@jiekang the last two commits here support the routing bug you discovered. The first one moves requests endpoints like `/clienturl`, `/auth`, `/recordings/:name`, and `/reports/:name` into an `/api/v1/` prefix. The second simply adds a fallback handler for requests to `/*` - if the request hasn't matched one of the `/api` paths, then it falls through to here. If a static asset is found with the requested name (`app.css` for example, or `images/pfbg_768.jpg`) then that is returned. If that can't be found then the web-client `index.html` is returned. This allows the user to visit a path like `/events`, which ContainerJFR's WebServer doesn't know anything about, and receive the React application and preserve the `/events` path, which the React router then sees and sends the user to the in-app Events route.

andrewazores:
Yea, that's just a consequence of how the in-app routing works. The only route available initially is the login page, then once that's passed it's only the Dashboard. Once that's loaded then the application checks if ContainerJFR is connected to a target, and if so, the other target-specific routes like `/events` are enabled. That's entirely due to how things are working within the React client, so it will behave the same way no matter how the ContainerJFR instance is configured or running, locally or not.

jiekang:
Hmm; I'd definitely revisit this in another issue and PR. Looks like something we'll need to think about.

Maybe the change you described is okay? I need to analyze the code more in depth. The path you describe of loading routes (login -> dashboard -> events) sounds strange to me.

andrewazores:
It's all in this file:

https://github.com/rh-jmc-team/container-jfr-web/blob/react/src/app/routes.tsx

There are just two checks - first that the user is logged in, and then that there is a ContainerJFR connection to a target. Without the user being logged in they should of course not be able to view any route, just the login page. So really, the only flow under question is basically this:

https://github.com/rh-jmc-team/container-jfr-web/blob/react/src/app/routes.tsx#L120

Here, the "dynamic" routes (`/events`, /`recordings`/`/recordings/create` currently) are only enabled if a target connection is seen. It's really not much of a path and not as sequential as it sounds. There is simply a base set of always-available routes, and an expanded set that is added when there is a target connection. If the user tries to visit one of these "dynamic" routes directly by URL, rather than navigating within the app, then of course it needs to do all of its internal initialization first and check whether there is a target connection. If the user visits the Dashboard first and proceeds through the usual flow then there is no 404 flicker or other "strange" behaviour since this is all already initialized.

jiekang:
Ah... I wonder if dynamic routing is the right tool here. I guess it also stems from cjfr itself having the concept of an active connection to a single jvm. 

It feels somewhat arbitrary at the moment for the recordings/events pages to be only active when a jvm is targeted. On the other side, it seems reasonable to allow users to visit those pages with an empty state and target a jvm from there.


Anyways, I think we'll revisit this when looking at
https://github.com/rh-jmc-team/container-jfr/issues/153

I think this PR is fine.

PR #152, opened 2020-04-22
Replace Service discovery with Endpoint discovery

### LINKED ISSUE #151: OpenShift target discovery sees load-balanced Pods
When ContainerJFR uses the OpenShift platform, the target discovery mechanism discovers Services. This means that pod replicas backing this service are all treated as one logical unit, and connections to this service will be load-balanced across the replicas. This allows for arbitrary sampling of data within a replica set, but has problems:

1) *only* arbitrarily sampling within the set is possible. Users cannot retrieve recordings for *all* members of the set, which may be useful when diagnosing various classes of bug or performance degradation (ex. a hard-to-trigger bug which is known to be occurring live in production, or even known to have been triggered on a specific replica)

2) if the user connects to a Service and starts a recording, then disconnects, then there is no way to return back to the same replica to retrieve that recording later without repeatedly connecting and disconnecting from the service and hoping to be load-balanced back to the correct replica

andrewazores:
Fixes #151

Screenshot_2020-04-21 ContainerJFRWeb(1)

### DIFF
--- a/src/test/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClientTest.java
+++ b/src/test/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClientTest.java
@@ -1,0 +1,159 @@
+package com.redhat.rhjmc.containerjfr.platform.openshift;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import com.redhat.rhjmc.containerjfr.core.log.Logger;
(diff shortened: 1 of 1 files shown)


src/test/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClientTest.java
```diff
@@ -0,0 +1,159 @@
...
+        EndpointPort port2 = Mockito.mock(EndpointPort.class);
+        Mockito.when(port2.getPort()).thenReturn(9999);
+        Mockito.when(port2.getName()).thenReturn("jfr-jmx");
+        EndpointPort port3 = Mockito.mock(EndpointPort.class);
+        Mockito.when(port3.getPort()).thenReturn(9091);
+        Mockito.when(port3.getName()).thenReturn("tcp-9091");
+
+        EndpointSubset subset1 = Mockito.mock(EndpointSubset.class);
+        // Mockito.when(subset1.getAddresses()).thenReturn(Collections.singletonList(address1));
+        Mockito.when(subset1.getPorts()).thenReturn(Collections.singletonList(port1));
```
ebaron, line 121: Are these commented-out lines meant to be kept?

andrewazores: Sort of. I set up all that structure of a mock endpoints API response, but because of the way the selection algorithm works, the IP and TargetRef of the Address are never checked if the corresponding port isn't deemed compatible. This caused Mockito to complain about unnecessary stubbing, although if those lines were completely removed then it might be a little more difficult to come back to that test later and see how things were intended to be structured when certain fields from the mock response are just omitted entirely.

PR #156, opened 2020-04-23
Add UPL license, manage with license-maven-plugin

ebaron:
This PR adds a LICENSE file to the root of the project, along with license headers to all Java source files. UPL was chosen to match the JMC project.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,35 @@
+Copyright (c) 2020 Red Hat, Inc.
+
+The Universal Permissive License (UPL), Version 1.0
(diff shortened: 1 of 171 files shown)


andrewazores:
Can the Maven license plugin (https://www.mojohaus.org/license-maven-plugin/) be used to help automate this process for next time something needs to be changed/updated?

ebaron:
Sure, here's a version with the headers managed by license-maven-plugin. There are some unsightly delimiters, but I believe those are mandatory.

PR #164, opened 2020-05-22
Add more integration tests

andrewazores:
This adds more integration tests, primarily targeting the HTTP API to make up for the current lack of unit test coverage in the area, but also exercising some command channel workflow and downloading real recordings/reports.


andrewazores:
@vic-ma or @Alexjsenn would either (or even better, both) of you care to review these integration test additions?

vic-ma:
I'd like to finish my `upload-recording` PR first, but I can take a look at this after.

PR #167, opened 2020-05-27
Refactor WebServer and split behaviour into handlers

andrewazores:
This introduces a RequestHandler interface, where implementations are injected into a set of handlers and processed into a sorted list by the WebServer, and then applied to its router as before. Each of these new handlers extracts existing logic and behaviour from the WebServer, as well as moving any relevant tests. No functional changes are made, which has been verified by running the integration tests in #164 against these changes as well as with manual testing. This makes the WebServer significantly smaller and simpler, so that the individual RequestHandlers can be unit tested much more easily and the WebServer itself does not have as many direct responsibilities.

~~Still in progress: handle CORS using these new handlers.~~

There are some TODOs left behind or newly added, which are intended to be addressed in a later cleanup. Some of them also require supporting changes in -core. There is also opportunity for more tests to be added, before and particularly after these cleanups are done.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/HttpMimeType.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/HttpMimeType.java
@@ -39,24 +39,21 @@
  * SOFTWARE.
  * #L%
  */
-package com.redhat.rhjmc.containerjfr.net.web.handlers;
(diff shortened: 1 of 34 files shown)


andrewazores:
@ebaron @jiekang any volunteers for a review?

@vic-ma I think you'll get some value from reading over this, too.

PR #168, opened 2020-05-29
Add download URL and report URL to `list` command in interactive mode

### LINKED ISSUE #160: `list` command is missing some information
The `list` command is missing some useful information. When executed in "serialized" form, ie when ContainerJFR is in WebSocket execution mode, the output includes the download URL of the recording but not the report URL. When running in non-serialized, interactive mode (ie. interactive shell execution mode), neither download or report URLs are included in the printed output.

Steps to reproduce:

(interactive)
```
$ mvn clean package
$ sh run.sh -it
- connect localhost
> dump foo 30 template=Continuous
> list
```

(serialized)
```
$ mvn clean package
$ sh run.sh
$ websocat ws://localhost:8181/api/v1/command # in another terminal
{command:dump,args:[foo,30,"template=Continuous"]}
{command:list}
```

vic-ma:
Fixes #160 

The code for the test has a lot duplicated with the test for the serializable list command, `shouldReturnListOutput()`, but I'm not sure if there's a good way to put it all into a method or if this would even be worth it, since it's just two tests.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
@@ -88,7 +88,13 @@ class ListCommand extends AbstractConnectedCommand implements SerializableComman
                         cw.println("\tNone");
                     }
                     for (IRecordingDescriptor recording : recordings) {
-                        cw.println(toString(recording));
+                        String downloadURL =
+                                "\tgetDownloadURL\t\t"
+                                + exporter.getDownloadURL(connection, recording.getName());
+                        String reportURL =
+                                "\tgetReportURL\t\t"
+                                + exporter.getReportURL(connection, recording.getName());
+                        cw.println(toString(recording) + downloadURL + "\n" + reportURL);
                     }
                     return null;
                 });
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
@@ -49,6 +49,8 @@ import java.util.List;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import org.apache.commons.lang3.ArrayUtils;
+
 import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
 
 import com.redhat.rhjmc.containerjfr.commands.SerializableCommand;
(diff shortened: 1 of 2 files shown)


andrewazores:
The test looks nice, very robust. I'm not too worried about a bit of duplication in the tests as you mention, in a sense trying to refactor the tests to share their implementations between the interactive and serialized cases to mirror the sharing in the actual command implementation is also coupling the tests to that implementation detail.

Rather than manually constructing and appending the "get Url" strings to the interactive output, do you think there's a way you can make use of the `HyperlinkedSerializableRecordingDescriptor` class that the serialized execution uses?

andrewazores:
It might take a bit more work to figure out how to modify the `ListCommand.toString` method as well. Just as a hint. But I think it's worthwhile to have all the properties printed by the `list` command done programmatically, so that it's as flexible as possible in case any properties are added or removed from the recording descriptors in the future, without also needing to modify the `list` command or any others like it.

vic-ma:
Okay, I think I have an idea of what to do.

But on the topic of `ListCommand.toString()`, I noticed that since it always uses two tabs between the method name and the data, some of the output is not properly aligned, since sometimes the method name is either too short or two long for two tabs.

Would this be worth fixing? I think it's probably just a simple fix: use 1 tab if len >= X; use 3 tabs if len <= Y; and that should cover all reasonable length method names.

andrewazores:
Yes, that does happen occasionally. If you'd like to tackle that as well then please feel free! I think it should be a separate issue with a separate PR, however.

And if you do go on to fix that alignment issue, I think it's worth fixing it for all cases - not just "reasonable" length method names. Those are famous last words.

Here's a throwback to OG Thermostat and how that rendered similar tables nicely in CLI:
https://icedtea.classpath.org/hg/thermostat/file/77b119517732/common/core/src/main/java/com/redhat/thermostat/common/cli/TableRenderer.java

andrewazores:
Here's a class that indirectly consumes that TableRenderer:
https://icedtea.classpath.org/hg/thermostat/file/77b119517732/client/cli/src/main/java/com/redhat/thermostat/client/cli/internal/ListVMsCommand.java

It's fairly analogous to the ListCommand here, so it's good reading.

vic-ma:
I've made the change to use `HyperlinkedSerializableRecordingDescriptor`.

Now, some of the output that used to be in the interactive command is no longer there, but I guess this is a good thing since it gets rid of a duplicate (`getDataStartTime`) and some unwanted data (e.g. `getObjectName`), and it makes the output equivalent to the serialized output.

Question about convention: is there a difference between when we use "URL" vs. "Url" in method names? For example, in `WebServer.getDownloadURL()` vs. `HyperlinkedSerializableRecordingDescriptor.getDownloadUrl()`.

Thanks for the info on the formatting issue. Looks like it would take more work than I originally thought haha. I'll open up an issue for it.

andrewazores:
Nice, the change looks good. I'll test it out this afternoon, but I expect to approve it.

> Question about convention: is there a difference between when we use "URL" vs. "Url" in method names? For example, in WebServer.getDownloadURL() vs. HyperlinkedSerializableRecordingDescriptor.getDownloadUrl().

No, that's just inconsistency and a lack of convention, heh. We could also file an issue for that and go through fixing those naming inconsistencies.

andrewazores:
Ah, whoops. These changes don't pass the Spotless formatting check. I only did a `mvn test` rather than a full `mvn verify` when I tested this, so the formatting tool wasn't run. I'll open a PR in a minute that just contains the result of `mvn spotless:apply`.

andrewazores:
https://github.com/rh-jmc-team/container-jfr/pull/170

PR #172, opened 2020-06-03
Set expected archived recordings filepath to absolute path

### LINKED ISSUE #171: Archiving commands unavailable
Since #163, commands implementing archive-related functionality (`save`, `list-saved`, `delete-saved`) are broken because the expected flight recordings path is not found.

andrewazores:
Using previous gcr.io base image, a relative path to the archived
recordings directory happened to work. Since switching base images to
UBI8 this assumption no longer holds, so the path is explicitly made
absolute.

No new tests are included, but it's worth noting that the integration tests in #164 do actually catch this regression. However, that PR is still outstanding and was not tested against the base image change.

Fixes #171

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
@@ -59,6 +59,7 @@ import com.redhat.rhjmc.containerjfr.tui.TuiModule;
 
 import dagger.Module;
 import dagger.Provides;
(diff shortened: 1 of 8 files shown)


andrewazores:
Thanks, I'll come back around and fix those. Those came from the WebServer RequestHandler PR which was just merged, so at the time that this PR was written those didn't exist yet.

andrewazores:
Just amended the commit and force-pushed it to include that change.

PR #178, opened 2020-06-04
Report caching

andrewazores:


andrewazores:
@jiekang this adds a caching layer in front of reports requests for archived recordings. I've tested this along with the -web inline report frame PR, and though it certainly helps, it's still not the best UX. Reports re-render much more quickly since network response time falls from ex. ~1s (30 seconds of "template=ALL") on the first request to ~10ms on reloads, but there is still definitely a noticeable visual effect of the report disappearing, the loading spinner very briefly appearing, and then the report iframe rendering again. So I'll still have more work to do on the front-end to try to smooth everything out, even after we do implement caching of reports everywhere.

andrewazores:
This is nearly ready now but I'm leaving it as a draft until I complete my own self-review and some more cleanup and refactoring, and handle IO failures in the archived recording disk cache (ex. cache directory not writeable, filesystem out of space, etc.)

An in-memory cache is used for reports from active recordings, which expires entries after 30 minutes without access and considers entries "fresh" for 5 minutes after they're written, so any repeated requests for reports on the same active recording will only trigger a new report generation if it has been at least 5 minutes since the last generation. Currently there is no maximum cache size applied since the important metric is not the number of reports or even the size of the reports themselves, but the size of the source recording (and therefore the cost of computing the report), which does not really correlate with the size of the reports, which will always live in a small range of fairly small values. Rather, the cache uses soft references for its reports to ensure that caching does not interfere with the generation of new reports or other activities, even if a large number of reports are cached. The cache can later be extended to cap the number of reports in-memory and evict to disk, or load from disk, if this turns out to be useful. For now, any disk caching of active recording reports is avoided due to the extra complexity of maintaining cache consistency, entry refreshing, and eviction policies, which are all handled by the Caffeine caching library which is strictly in-memory.

Archived recording reports are cached to disk immediately on first request, and subsequent requests for reports from the same archived recording will always be served from the same cached report. Since archived recordings are static and unchanging there is no need to ever refresh their reports, which greatly simplifies the caching logic.

In either case, the `delete` or `delete-saved` commands will also invalidate and remove cache entries for the reports corresponding to the deleted recording.

andrewazores:
Ready for review. This combined with the frontend client's own in-memory caching (not yet in container-jfr master, which is still pointing to an older submodule revision) makes repeated report generation much snappier. There is still a bit of weirdness on the frontend with scroll position, but that's a separate concern. The container resource consumption for generating reports is still equally heavy, but at least now it won't happen as frequently in most scenarios.

andrewazores:
@jiekang or @ebaron , could either of you take a look at this when you have the time?

Issue #175, opened 2020-06-04
Add endpoint for health status


jiekang:
container-jfr service should have an endpoint such as root '/' that returns health status of the service. This can also include health status of it's child services such as jfr-datasource and grafana

jiekang:
Resolved in https://github.com/rh-jmc-team/container-jfr/pull/179

PR #179, opened 2020-06-05
Add health check

jiekang:
Addresses #175

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
@@ -1,0 +1,111 @@
+/*-
+ * #%L
+ * Container JFR
+ * %%
+ * Copyright (C) 2020 Red Hat, Inc.
+ * %%
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ * #L%
+ */
+package com.redhat.rhjmc.containerjfr.net.web.handlers;
+
+import java.io.IOException;
+import java.util.Map;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.impl.client.CloseableHttpClient;
+import com.google.gson.Gson;
+
+import com.redhat.rhjmc.containerjfr.core.sys.Environment;
+import com.redhat.rhjmc.containerjfr.net.web.HttpMimeType;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+
+class HealthGetHandler implements RequestHandler {
+
+    static final String GRAFANA_DATASOURCE_ENV = "GRAFANA_DATASOURCE_URL";
+
+    private final Provider<CloseableHttpClient> httpClientProvider;
+    private final Environment env;
+    private final Gson gson;
+
+    @Inject
+    HealthGetHandler(Provider<CloseableHttpClient> httpClientProvider, Environment env, Gson gson) {
+        this.httpClientProvider = httpClientProvider;
+        this.env = env;
+        this.gson = gson;
+    }
+
+    @Override
+    public String path() {
+        return "/health";
+    }
+
+    @Override
+    public HttpMethod httpMethod() {
+        return HttpMethod.GET;
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -534,6 +542,25 @@ void handleAuthRequest(RoutingContext ctx) {
...
+    void handleHealthRequest(RoutingContext ctx) {
+        ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, MIME_TYPE_JSON);
+        ctx.response().setStatusCode(200);
+
+        boolean datasourceAvailable = false;
+
+        if (this.env.hasEnv(GRAFANA_DATASOURCE_ENV)) {
+            HttpGet grafanaHealth = new HttpGet(this.env.getEnv(GRAFANA_DATASOURCE_ENV));
+            try {
+                CloseableHttpResponse response = this.client.execute(grafanaHealth);
```
andrewazores, line 554: I think the `response` needs to be cleaned up - simply `try (CloseableHttpResponse response = ...) { ... }` would suffice.

jiekang:
@andrewazores Ah... looks like the handling has been refactored a lot; I'll have to work on this some more.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -147,6 +152,7 @@
             TargetConnectionManager targetConnectionManager,
             Logger logger) {
         this.server = server;
+        this.client = client;
```
andrewazores, line 155: I think this is probably okay, but what do you think about injecting a `Provider<CloseableHttpClient>`? Then later on when you need to use a client, you can create, use, and clean up the client as needed. Right now there will be a single client sticking around for the life of the webserver. That's not likely a real problem, but I guess just a stylistic thing.

andrewazores:
> @andrewazores Ah... looks like the handling has been refactored a lot; I'll have to work on this some more.

Yes, I was going to ask if you could refactor and rebase on top of the latest "RequestHandlers" change. It really won't be a lot of work for you though, you just need to extract everything you've added into a very simple event handler, basically.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/WebServer.java
```diff
@@ -534,6 +542,25 @@ void handleAuthRequest(RoutingContext ctx) {
...
+            HttpGet grafanaHealth = new HttpGet(this.env.getEnv(GRAFANA_DATASOURCE_ENV));
+            try {
+                CloseableHttpResponse response = this.client.execute(grafanaHealth);
+                if (response.getStatusLine().getStatusCode() == 200) {
+                    datasourceAvailable = true;
+                }
+            } catch (IOException e) {
+            }
+        }
+        endWithJsonKeyValue("datasourceAvailable", datasourceAvailable, ctx.response());
```
andrewazores, line 561: When you do the refactor, this `endWithJsonKeyValue` isn't around anymore, so the preferred equivalent would look like `ctx.response().end(gson.toJson(Map.of("datasourceAvailable", datasourceAvailable)))`.

On that note, what do you think about adding an equivalent JSON property for the Grafana dashboard's health as well? It would be something like a `new HttpGet(this.env.getEnv(GRAFANA_DASHBOARD_ENV) + "/api/health"))`, and then the rest of the response processing you've done here would be the same I think.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
```diff
@@ -0,0 +1,111 @@
...
+        if (this.env.hasEnv(GRAFANA_DATASOURCE_ENV)) {
+            try (CloseableHttpResponse response =
+                    this.httpClientProvider
+                            .get()
+                            .execute(new HttpGet(this.env.getEnv(GRAFANA_DATASOURCE_ENV))); ) {
+                if (response.getStatusLine().getStatusCode() == 200) {
+                    datasourceAvailable = true;
+                }
+            } catch (IOException e) {
+                // Do nothing
```
andrewazores, line 103: I think at least putting a Logger warning here or something would make sense. What do you think?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
+    @Override
+    public void handle(RoutingContext ctx) {
+        boolean datasourceAvailable = false;
+
+        if (this.env.hasEnv(GRAFANA_DATASOURCE_ENV)) {
+            try (CloseableHttpResponse response =
+                    this.httpClientProvider
+                            .get()
+                            .execute(new HttpGet(this.env.getEnv(GRAFANA_DATASOURCE_ENV))); ) {
```
andrewazores, line 98: If the env var is not set, or is blank what happens? Is a request still fired? I think in that scenario it's worth just short-circuiting and skipping the network request.

jiekang: This is inside an if block for env.hasEnv so it should be skipped when env isn't set. I think it's fine to continue without checking the value; it will just error out in it's attempt in the client code

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
jiekang: Added in latest commit

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
```diff
@@ -0,0 +1,140 @@
...
+            } catch (IOException e) {
+                logger.warn(e);
+            }
+        }
+
+        ctx.response()
+                .putHeader(HttpHeaders.CONTENT_TYPE, HttpMimeType.JSON.mime())
+                .end(
+                        gson.toJson(
+                                new TreeMap<>(
```
andrewazores, line 133: Is the `TreeMap` here jjust to ensure consistent ordering of the keys?

jiekang: Yeah; let me know if there's a smarter way to do it D:

andrewazores: Hmm. I don't know of a particularly better way, although I wonder if it really matters - a typical client parsing the JSON response shouldn't care about the key order. Maybe it makes testing easier, but then perhaps the tests should grab a Gson instance (from the MainModule), parse the response, and make assertions on what is found in the deserialized structure?

jiekang: Hey Andrew. I think that clients will appreciate consistent responses and the cost for us to do so is minimal, though a bit... hacky (here). If you'd like, I can keep it random and adjust the test, let me know.

andrewazores: I don't really see the benefit to clients of having consistently ordered JSON keys, though. JSON parsing of an object should be independent of what order the keys appear in, so any client using a spec-compliant parser shouldn't even really notice whether the keys are sorted or even in consistent order or not. If we start enforcing a particular ordering of the keys here then a) we should be doing it everywhere and b) that ordering becomes an implicit part of our API, and I don't quite agree with either of those.

jiekang: Okay. Updated in the latest commit

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/HealthGetHandler.java
```diff
@@ -0,0 +1,138 @@
...
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
+    @Override
+    public void handle(RoutingContext ctx) {
+        boolean datasourceAvailable = false;
+        boolean dashboardAvailable = false;
+
+        if (this.env.hasEnv(GRAFANA_DATASOURCE_ENV)) {
+            try (CloseableHttpResponse response =
+                    this.httpClientProvider
+                            .get()
```
andrewazores, line 106: Last thing: this auto-closes the response, but not the client itself. I think it should be something like `try (CloseableHttpClient client = this.httpClientProvider.get(); CloseableHttpResponse response = client.execute())`. There's a similar example in the UploadRecordingCommand

jiekang: Should be fixed now. Thanks for the note.

Issue #149, opened 2020-04-15
Users should be able to download templates from target


andrewazores:
Parent: #153

A user should be able to download the XML definitions of a target's event templates.

andrewazores:
See #153 . This can be an API enhancement following parent's point 4.

andrewazores:
Closed: #199

Issue #143, opened 2020-03-18
Users should be able to create custom event templates


andrewazores:
See #141 

A user should be able to create a custom event template (ex. by uploading an XML `.jfc` file), which would subsequently be included in the set of event templates that can be applied to a recording. This should be stored in ContainerJFR's own local storage, and available as a template for any and all target JVMs. The template may not then be applicable to every target in case it specifies events/options not known to the target, but that's up to the users to determine compatibility when they're uploading custom templates.

andrewazores:
See #153 . This can be an API enhancement following parent's point 4.

PR #185, opened 2020-06-17
Add HTTP API handlers for various "lists"

andrewazores:
This adds HTTP API handlers for various paths with a GET verb, mirroring current WebSocket APIs (`scan-targets`, `list`, `list-saved`, `list-templates`, `list-event-types`). Those WebSocket APIs should be deprecated, although there is currently no defined way to mark them as such in the API responses. This helps bring the API closer to a typical HTTP REST design, moving further away from the original interactive execution mode that was initially implemented and is currently mirrored in the WebSocket command channel API.


andrewazores:
Once this makes its way through reviews and is merged my intent is to use the same approach to address #149 by adding a handler for `GET /api/v1/targets/:targetId/templates/:templateName`, and #143 by adding handlers for `POST /api/v1/targets/:targetId/templates`/`PUT/DELETE /api/v1/targets/targetId/templates/:templateName`.

PR #186, opened 2020-06-18
Archive path should be configurable

Alexjsenn:
Adds environment variable `ARCHIVE_PATH` which specifies path when saving flight recordings locally. If not set, the default path remains `/flightrecordings`.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
@@ -90,6 +90,8 @@ public abstract class MainModule {
     @Provides
     @Named(RECORDINGS_PATH)
     static Path provideSavedRecordingsPath(Logger logger, Environment env) {
-        return Paths.get(env.getEnv("CONTAINER_JFR_ARCHIVE_PATH", "/flightrecordings"));
+        String ARCHIVE_PATH = env.getEnv("CONTAINER_JFR_ARCHIVE_PATH", "/flightrecordings");
+        logger.info(String.format("Local save path for flight recordings set as %s", ARCHIVE_PATH));
+        return Paths.get(ARCHIVE_PATH);
     }
 }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
@@ -52,6 +52,7 @@ import com.google.gson.GsonBuilder;
 
 import com.redhat.rhjmc.containerjfr.commands.CommandsModule;
 import com.redhat.rhjmc.containerjfr.core.log.Logger;
+import com.redhat.rhjmc.containerjfr.core.sys.Environment;
 import com.redhat.rhjmc.containerjfr.net.web.WebModule;
 import com.redhat.rhjmc.containerjfr.platform.PlatformModule;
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
```diff
@@ -89,6 +89,17 @@ public static Gson provideGson() {
     @Provides
     @Named(RECORDINGS_PATH)
     static Path provideSavedRecordingsPath() {
-        return Paths.get("/", "flightrecordings");
+        String ARCHIVE_PATH = null;
+        final Logger logger = provideLogger();
```
andrewazores, line 93: You should be able to have Dagger inject a Logger instance into this method as a parameter, rather than having to grab the logger by manually calling this method like this. Just add a Logger to the method's argument list.

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
```diff
@@ -89,6 +89,17 @@ public static Gson provideGson() {
...
+        try {
+            ARCHIVE_PATH = System.getenv("ARCHIVE_PATH");
```
andrewazores, line 95: Can you refactor this to use the `Environment` utility class, which is also available by dependency injection? The source for that class is in `-core`, so you can take a look and see what methods might be useful for you here.

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
andrewazores: You can see examples of this in other DI modules around the codebase, for example the `NetworkModule`.

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
andrewazores: Also, what do you think of prefixing this new env var with `CONTAINER_JFR_`? That's how the other configuration variables are generally named.

Alexjsenn: After using the Environment class I realized I needed way less code!

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
Alexjsenn: Ok, fixed!

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
```diff
@@ -88,7 +89,7 @@ public static Gson provideGson() {
     @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
     @Provides
     @Named(RECORDINGS_PATH)
-    static Path provideSavedRecordingsPath() {
-        return Paths.get("/", "flightrecordings");
+    static Path provideSavedRecordingsPath(Logger logger, Environment env) {
```
andrewazores, line 92: The `Logger` isn't used anymore. It can either be removed, or maybe there should be a message logged if the environment variable is (or isn't) set so that it's clear from the logs what path was used during a particular run.

Alexjsenn: Ok, logger now reports set path.

Issue #132, opened 2020-03-06
Intermittent unit test failure


andrewazores:
```
[ERROR] Failures: 
[ERROR]   MessagingServerTest.shouldHandleRemovedConnections:160 
Expected: "another message"
     but: was "hello world"
```

This test seems to intermittently fail. Possibly related to Maven parallel execution configuration, since I don't recall ever seeing this failure before with the Gradle build setup.

andrewazores:
Here's another one:

```
[ERROR] Failures: 
[ERROR]   MessagingServerTest.shouldHandleRemovedConnections:145 
crw2.readLine();
Wanted 1 time:
-> at com.redhat.rhjmc.containerjfr.tui.ws.WsClientReaderWriter.readLine(WsClientReaderWriter.java:65)
But was 2 times:
-> at com.redhat.rhjmc.containerjfr.tui.ws.MessagingServer.lambda$addConnection$6(MessagingServer.java:123)
-> at com.redhat.rhjmc.containerjfr.tui.ws.MessagingServer.lambda$addConnection$6(MessagingServer.java:123)
```

vic-ma:
Ah, I've experienced this for some time now and was planning to ask you about it, since I wasn't sure if it was just me. Now I see it is already a known issue.

But I also started getting this same problem with a `list-command` test, right after I made the changes in #168. Do you get this as well sometimes?

```
[ERROR] Failures: 
[ERROR]   ListCommandTest.shouldPrintDownloadAndReportURL:192 
Verification in order failure
Wanted but not invoked:
cw.println(
    contains("	getDownloadUrl		http://example.com:1234/api/v1/targets/fooHost:1/recordings/foo
	getReportUrl		http://example.com:1234/api/v1/targets/fooHost:1/reports/foo")
);
-> at com.redhat.rhjmc.containerjfr.core.tui.ClientWriter.println(ClientWriter.java:54)
Wanted anywhere AFTER following interaction:
cw.println("Available recordings:");
-> at com.redhat.rhjmc.containerjfr.commands.internal.ListCommand.lambda$execute$0(ListCommand.java:86)
```

andrewazores:
[...]

PR #187, opened 2020-06-22
list-saved enhancements

Alexjsenn:
For each saved file it now lists the URLs as well as the name (using the same format as list).

### DIFF
--- a/src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/ListSavedRecordingsCommandTest.java
+++ b/src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/ListSavedRecordingsCommandTest.java
@@ -144,39 +144,41 @@ class ListSavedRecordingsCommandTest {
     @Test
     void shouldPrintURLs() throws Exception {
         when(fs.listDirectoryChildren(recordingsPath)).thenReturn(Arrays.asList("foo", "bar"));
-        when(exporter.getArchivedDownloadURL(Mockito.anyString())).thenAnswer(
-            new Answer<String>() {
-                @Override
-                public String answer(InvocationOnMock invocation) throws Throwable {
-                    return String.format(
-                            "http://example.com:1234/api/v1/recordings/%s",
-                            invocation.getArguments()[0]);
-                }
-            });
-        when (exporter.getArchivedReportURL(Mockito.anyString())).thenAnswer(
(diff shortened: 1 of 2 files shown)


Alexjsenn:
I also realized a test to verify the URLs should be added, which I am working on now, but if I use the same method as the `ListCommandTest.shouldPrintDownloadAndReportURL` test, it will likely fail intermittently (#132)

Alexjsenn:
So I added the test and then ran them a few times, and from the limited iterations I ran, either all tests passed, or both `ListCommandTest.shouldPrintDownloadAndReportURL` and my new test `ListSavedRecordingsCommandTest.shouldPrintURLs` failed together. I guess this makes sense given I wrote the new test in a similar fashion to the old one.

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/ListSavedRecordingsCommandTest.java
```diff
@@ -137,8 +137,46 @@ void shouldExecuteAndPrintSavedRecordings() throws Exception {
...
+        when(exporter.getArchivedDownloadURL(Mockito.anyString())).thenAnswer(
+            new Answer<String>() {
+                @Override
+                public String answer(InvocationOnMock invocation) throws Throwable {
+                    return String.format(
+                            "http://example.com:1234/api/v1/recordings/%s",
+                            invocation.getArguments()[0]);
+                }
+            });
+        when (exporter.getArchivedReportURL(Mockito.anyString())).thenAnswer(
```
andrewazores, line 156: There's an extra whitespace after the `when` on this line. Have you done a `mvn spotless:apply` to auto-format the code?

andrewazores:
I'm still really confused about how those tests can intermittently fail, but after all of Victor's investigations I don't feel any closer to discovering what the real root cause might be. Anyway, for the purposes of this PR I'm fine with the tests as implemented, since they mirror the other pre-existing ones for the standard `list` command.

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/ListSavedRecordingsCommandTest.java
Alexjsenn: Fixed!

PR #189, opened 2020-06-22
Replace Apache HttpClient with Vertx WebClient

andrewazores:
This PR is based on top of #164 , since that already previously introduced the Vertx WebClient dependency. Only the most recent five commits are specific to this PR.

The goal here is to remove the usage of the Apache HttpComponents HttpClient since it overlaps with the Vertx WebClient (or Vertx HttpClient). There is no need for both to be used in the project, and since we have a stronger dependency on Vertx for our webserver it makes sense to standardize on that.

There is one remaining use of an Apache URIBuilder class which comes from the HttpComponents dependency, but this has been deferred from this PR as it isn't as simple to fix as it sounds. A suitable replacement which properly escapes path parameters is needed, which may be a custom implementation or supplied by another dependency to be determined later.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,11 +100,6 @@
     <version>${org.apache.httpcomponents.version}</version>
   </dependency>
   <dependency>
-    <groupId>org.apache.httpcomponents</groupId>
-    <artifactId>httpmime</artifactId>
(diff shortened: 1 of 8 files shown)


andrewazores:
Rebased and ready.

PR #192, opened 2020-06-24
Fix intermittent print URL bug

Alexjsenn:
By strictly enforcing the order in which they print it seems to not fail the tests anymore.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
@@ -156,7 +156,7 @@ class ListCommand extends AbstractConnectedCommand implements SerializableComman
         for (Method m : methods) {
             if (m.getParameterTypes().length == 0
                     && (m.getName().startsWith("get") || m.getName().startsWith("is"))) {
-                if (m.getName().contains("Url")) {
+                if (m.getName().toLowerCase().contains("url")) {
                     urls.add(String.format("\t%s\t\t%s%n", m.getName(), m.invoke(descriptor)));
                 } else {
                     sb.append(String.format("\t%s\t\t%s%n", m.getName(), m.invoke(descriptor)));
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
@@ -44,6 +44,7 @@ package com.redhat.rhjmc.containerjfr.commands.internal;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -150,17 +151,22 @@ class ListCommand extends AbstractConnectedCommand implements SerializableComman
                 ArrayUtils.addAll(
                         descriptor.getClass().getSuperclass().getDeclaredMethods(),
                         descriptor.getClass().getDeclaredMethods());
+
+        List<String> urls = new ArrayList<String>();
         for (Method m : methods) {
             if (m.getParameterTypes().length == 0
                     && (m.getName().startsWith("get") || m.getName().startsWith("is"))) {
-                sb.append("\t");
-                sb.append(m.getName());
-
-                sb.append("\t\t");
-                sb.append(m.invoke(descriptor));
-
-                sb.append("\n");
+                if (m.getName().contains("Url")) {
+                    urls.add(String.format("\t%s\t\t%s%n", m.getName(), m.invoke(descriptor)));
+                } else {
+                    sb.append(String.format("\t%s\t\t%s%n", m.getName(), m.invoke(descriptor)));
+                }
             }
+        }
+
+        Collections.sort(urls);
+        for (String s : urls) {
+            sb.append(s);
         }
 
         return sb.toString();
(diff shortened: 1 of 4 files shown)


andrewazores:
Could you run a repeated series of tests before and after this change and post what the failure rate stats are for each series? ---- On Wed, 24 Jun 2020 06:58:45 -0700  notifications@github.com  wrote ----By strictly enforcing the order in which they print it seems to not fail the tests anymore.



You can view, comment on, or merge this pull request online at:
  https://github.com/rh-jmc-team/container-jfr/pull/192

Commit Summary

  Fix ListCommand and ListSavedRecordingsCommand to enforce order in which URLs are printed (previously causing intermittent test failures)


File Changes

  
    M
    src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
    (20)
  
  
    M
    src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListSavedRecordingsCommand.java
    (20)
  


Patch Links:

  https://github.com/rh-jmc-team/container-jfr/pull/192.patch
  https://github.com/rh-jmc-team/container-jfr/pull/192.diff

Alexjsenn:
So I ran the tests 60 times before and after this fix.
Before: failed 25 times (6 of the failures were actually other bugs from #132)
After: Only failed 4 times (none of them from print URL tests)
So after this fix it should be failing well under 10% of the times.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
```diff
@@ -150,19 +151,24 @@ private static String toString(HyperlinkedSerializableRecordingDescriptor descri
...
             if (m.getParameterTypes().length == 0
                     && (m.getName().startsWith("get") || m.getName().startsWith("is"))) {
-                sb.append("\t");
-                sb.append(m.getName());
-
-                sb.append("\t\t");
-                sb.append(m.invoke(descriptor));
-
-                sb.append("\n");
+                if (m.getName().contains("Url")) {
```
andrewazores, line 159: This is a bit fragile, in case the method is renamed from `getDownloadUrl` to `getDownloadURL` for example. Could it be made case-insensitive?

vic-ma:
Would it be worth sorting all the non-URL methods as well? There aren't currently any tests that check their order so nothing breaks because of it, but probably worth enforcing an order for them as well, I think?

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ListCommand.java
Alexjsenn: Ok, should've fixed it.

Alexjsenn:
I tried that at first but it makes things much harder to read because they're in aphabetical order. So its a bit of a tradeoff.

vic-ma:
I was thinking more like if you ordered all the non-URL methods, then appended the ordered URL methods after.

Thinking more about it, I do like how `getName` is the first field with the current behaviour, since it makes sense that the name of the recording is the first thing you see. But this is just a lucky coincidence anyway and I suppose if that's what we really wanted, it should be forced by the code.

And if we really cared about the order that `execute()` returns, I imagine we'd have no choice but to hard-code it, in order to get exactly the order we want. But the non-serial execute is used more for debugging reasons, right? Since a regular user would just use the web-client?

andrewazores:
Yea, the interactive mode is mostly a legacy holdover at this point. I don't image it would really get used by end users for the most part, certainly not in an OpenShift (or general Kubernetes) deployment. Normally we're expecting ContainerJFR to be run in WebSocket execution mode, so all commands run through the `serializableExecute` and get returned through an API response as JSON, so ordering the keys really isn't something to bother with. These two commands also now have preferred equivalents in the WebServer HTTP API, so the interactive mode output here is really not a focus area IMO. That doesn't mean it has to be ignored or that it isn't worth improving, just that it doesn't need to be a particularly high priority.

I do have another thought on fixing the tests however - why don't we split the test that checks for the URLs into two tests? They would be nearly identical to each other, but just test that the output contains just one of the URLs. Taken as a whole the tests still exercise everything, but then the order of the strings in the printed result really just doesn't matter.

vic-ma:
>I do have another thought on fixing the tests however - why don't we split the test that checks for the URLs into two tests? They would be nearly identical to each other, but just test that the output contains just one of the URLs. Taken as a whole the tests still exercise everything, but then the order of the strings in the printed result really just doesn't matter.

I think that also makes sense, especially if somehow more URLs get added in the future.

At the same time, I like the idea of having the output be sorted, since it feels strange knowing that the ordering is X *most of the times*, and also that the ordering is completely dependent on what `getMethods()` does underneath and could change at random, and could also be impacted by the running environment.

But I do think having the tests split up is a good idea, even just as a standalone idea, not necessarily to fix this issue. So maybe we could do both?

andrewazores:
@Alexjsenn are you planning to make any further changes to this PR?

Alexjsenn:
For the order of printing, given we have never encountered an issue I think it makes sens to leave it as it is for now since it provides the most intuitive layout (name at the top, urls at the bottom, etc). With regards to the tests I can split the URL checks into 2 if that makes more sense?

andrewazores:
Yea, I think splitting the test up makes sense. One test for each expected line of output seems to be reasonable - it will make tracking down any bugs or regressions easier, since seeing exactly which tests fail and pass is better than seeing that the one big test fails at some point. If the single big test fails then it's possible that the implementation is broken in more than one way, but you'll only see the first assertion that fails in the failing test.

PR #193, opened 2020-06-25
Change `upload-url` Command to Assume Datasource URL from Environment Variable

### LINKED ISSUE #166: `upload-recording` command should assume datasource URL
Currently, the `upload-recording` command expects to be provided three arguments: targetId, recordingName, and datasourceUrl. The datasourceUrl should not be provided by the client. Instead, `container-jfr` should simply assume that the URL to upload the recording to is the configured datasource URL in its environment variables (ie the same one reported by the WebServer on GET /api/v1/grafana_datasource_url).

vic-ma:
Fixes #166.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,7 @@
 
   <org.apache.commons.lang3.version>3.9</org.apache.commons.lang3.version>
   <org.apache.commons.codec.version>1.13</org.apache.commons.codec.version>
+  <org.apache.commons.validator.version>1.6</org.apache.commons.validator.version>
   <org.apache.commons.io.version>2.6</org.apache.commons.io.version>
   <org.apache.httpcomponents.version>4.5.11</org.apache.httpcomponents.version>
   <io.fabric8.openshift.client.version>4.6.4</io.fabric8.openshift.client.version>
@@ -93,6 +94,11 @@
     <groupId>commons-codec</groupId>
     <artifactId>commons-codec</artifactId>
     <version>${org.apache.commons.codec.version}</version>
+  </dependency>
+  <dependency>
+    <groupId>commons-validator</groupId>
+    <artifactId>commons-validator</artifactId>
+    <version>${org.apache.commons.validator.version}</version>
   </dependency>
   <dependency>
     <groupId>org.apache.httpcomponents</groupId>
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
@@ -167,6 +167,7 @@ class UploadRecordingCommand extends AbstractConnectedCommand implements Seriali
 
         String targetId = args[0];
         String recordingName = args[1];
+        String datasourceUrl = env.getEnv(GRAFANA_DATASOURCE_ENV);
 
         boolean isValidTargetId = validateTargetId(targetId);
         if (!isValidTargetId) {
@@ -179,10 +180,12 @@ class UploadRecordingCommand extends AbstractConnectedCommand implements Seriali
         }
 
         boolean isValidDatasourceUrl =
-                new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS)
-                        .isValid(env.getEnv(GRAFANA_DATASOURCE_ENV));
+                new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS).isValid(datasourceUrl);
         if (!isValidDatasourceUrl) {
-            cw.println(String.format("%s is an invalid datasource URL", GRAFANA_DATASOURCE_ENV));
+            cw.println(
+                    String.format(
+                            "$%s=%s is an invalid datasource URL",
+                            GRAFANA_DATASOURCE_ENV, datasourceUrl));
         }
 
         return isValidTargetId && isValidRecordingName && isValidDatasourceUrl;
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
@@ -43,7 +43,6 @@ package com.redhat.rhjmc.containerjfr.commands.internal;
 
 import java.io.BufferedInputStream;
 import java.io.InputStream;
-import java.net.MalformedURLException;
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.Optional;
@@ -53,6 +52,7 @@ import javax.inject.Named;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
+import org.apache.commons.validator.routines.UrlValidator;
 import org.apache.http.StatusLine;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
@@ -178,12 +178,11 @@ class UploadRecordingCommand extends AbstractConnectedCommand implements Seriali
             cw.println(String.format("%s is an invalid recording name", recordingName));
         }
 
-        boolean isValidDatasourceUrl = true;
-        try {
-            new java.net.URL(env.getEnv(GRAFANA_DATASOURCE_ENV));
-        } catch (MalformedURLException e) {
+        boolean isValidDatasourceUrl =
+                new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS)
+                        .isValid(env.getEnv(GRAFANA_DATASOURCE_ENV));
+        if (!isValidDatasourceUrl) {
             cw.println(String.format("%s is an invalid datasource URL", GRAFANA_DATASOURCE_ENV));
-            isValidDatasourceUrl = false;
         }
 
         return isValidTargetId && isValidRecordingName && isValidDatasourceUrl;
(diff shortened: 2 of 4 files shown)


andrewazores:
You haven't updated the tests for the command to reflect the added constructor parameter - might that be where your build failure is coming from?

vic-ma:
I'm just running `mvn compile`, so that should completely ignore tests right?

Here is the compilation error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project container-jfr: Compilation failure: Compilation failure: 
[ERROR] /home/vma/container-jfr/target/generated-sources/annotations/com/redhat/rhjmc/containerjfr/DaggerContainerJfr_Client.java:[449,94] method create in class com.redhat.rhjmc.containerjfr.commands.internal.UploadRecordingCommand_Factory cannot be applied to given types;
[ERROR]   required: javax.inject.Provider<com.redhat.rhjmc.containerjfr.core.tui.ClientWriter>,javax.inject.Provider<com.redhat.rhjmc.containerjfr.net.TargetConnectionManager>,javax.inject.Provider<com.redhat.rhjmc.containerjfr.core.sys.FileSystem>,javax.inject.Provider<java.nio.file.Path>,javax.inject.Provider<org.apache.http.impl.client.CloseableHttpClient>
[ERROR]   found: javax.inject.Provider<com.redhat.rhjmc.containerjfr.core.tui.ClientWriter>,javax.inject.Provider<com.redhat.rhjmc.containerjfr.net.TargetConnectionManager>,javax.inject.Provider<com.redhat.rhjmc.containerjfr.core.sys.FileSystem>,javax.inject.Provider<com.redhat.rhjmc.containerjfr.core.sys.Environment>,javax.inject.Provider<java.nio.file.Path>,javax.inject.Provider<org.apache.http.impl.client.CloseableHttpClient>
[ERROR]   reason: actual and formal argument lists differ in length
[ERROR] /home/vma/container-jfr/target/generated-sources/annotations/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand_Factory.java:[57,12] constructor UploadRecordingCommand in class com.redhat.rhjmc.containerjfr.commands.internal.UploadRecordingCommand cannot be applied to given types;
[ERROR]   required: com.redhat.rhjmc.containerjfr.core.tui.ClientWriter,com.redhat.rhjmc.containerjfr.net.TargetConnectionManager,com.redhat.rhjmc.containerjfr.core.sys.FileSystem,com.redhat.rhjmc.containerjfr.core.sys.Environment,java.nio.file.Path,javax.inject.Provider<org.apache.http.impl.client.CloseableHttpClient>
[ERROR]   found: com.redhat.rhjmc.containerjfr.core.tui.ClientWriter,com.redhat.rhjmc.containerjfr.net.TargetConnectionManager,com.redhat.rhjmc.containerjfr.core.sys.FileSystem,java.nio.file.Path,javax.inject.Provider<org.apache.http.impl.client.CloseableHttpClient>
```

andrewazores:
It's complaining about the argument list not matching up within Dagger's generated sources. `mvn clean`?

vic-ma:
Ah yup that fixed it, thanks. I'm never too sure when to add in clean...but I guess seeing this it may be good to just always use it ha.

andrewazores:
Yea, I always include `clean.` It hardly slows anything down and it avoids weirdness like this.

vic-ma:
>2. The validate method should do a basic parse of the URL retrieved from the Environment, just a simple new java.net.URL(urlStringFromEnv) would suffice. It doesn't need to validate that the URL is an actual reachable jfr-datasource, just that the value looks like a URL

With this, will I run into the issue of the `ValidationTestable` tests not working anymore? I originally had a `hasEnv()` check inside `validate()` but then I realized that this breaks all the validation tests. The impression I got was that `validate()` should only check the arguments the command gets, and only with `validate*()` methods, since it seems like any other checks wouldn't be compatible with how the `ValidationTestable`-based tests work since we can't modify them or add mocks, etc.?

andrewazores:
Ah, yea, that sounds like it might happen with the way that the 
validation mixin currently works, so it might require a bit of reworking 
to do since it does require a mock injected. Or, at least for now, you 
might have to actually restore those tests you just remove and expand 
upon them to cover more validation cases, and stop using the validation 
mixin(s). There is some precedent for checking the URL within the 
validation though - all commands that expect a target ID argument do 
also validate that argument as a `JMXServiceURL`, or at least they are 
meant to. Check the `Command` interface to see that (yea yea, that's 
also sort of a mixin on its own... but it isn't quite as egregious as 
the validation testing ones. It's more what default methods were really 
meant for).

vic-ma:
I've made the requested changes.

For the validation tests, I just got rid of the interfaces and rewrote each test directly into `UploadRecordingCommandTest`.

I also added two new `validate` tests for the datasource URL. They aren't too stringent, since `java.net.URL` doesn't catch a lot of bad URLs (like `http:// localhost` or `http:///localhost`). I considered using the Apache Commons `UrlValidator`, since it would catch more and get rid of the need for a try-catch block, but I couldn't get the import working, so I'm guessing we just don't have the needed dependency in the `pom`. And I didn't want to add it in unless you think we do need the extra-strength URL verification.

Another thing: in my manual testing, I found that if I set my `GRAFANA_DATASOURCE_URL` to an invalid URL, thus causing the `validate()` method to return `false`, then on websocat, I will get this response:
```
{"id":null,"commandName":"upload-recording","status":-1,"payload":"[localhost, foo] are invalid arguments to upload-recording"}
```
even though technically, the arguments are both correct, and it is only the environment variable that is wrong. (I do of course also see the "invalid datasource URL" message in the backend logs.) I assume this is just the standard response whenever `validate()` is false. Is this something to worry about at all?

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
```diff
@@ -179,17 +170,28 @@ public boolean validate(String[] args) {
...
+            isValidDatasourceUrl = false;
+        }
 
-        return isValidTargetId && isValidRecordingName;
+        return isValidTargetId && isValidRecordingName && isValidDatasourceUrl;
+    }
+
+    @Override
+    public boolean isAvailable() {
+        return env.hasEnv(GRAFANA_DATASOURCE_ENV);
```
andrewazores, line 194: Just to be safe, this should also call through to `super.isAvailable()` and `&&` the results together.

andrewazores:
> I also added two new validate tests for the datasource URL. They aren't too stringent, since java.net.URL doesn't catch a lot of bad URLs (like http:// localhost or http:///localhost). I considered using the Apache Commons UrlValidator, since it would catch more and get rid of the need for a try-catch block, but I couldn't get the import working, so I'm guessing we just don't have the needed dependency in the pom. And I didn't want to add it in unless you think we do need the extra-strength URL verification.

`commons-validator` seems like a pretty reasonable dependency to pull in. We have plenty of other places where we could shore up our validation efforts, too.

> Another thing: in my manual testing, I found that if I set my GRAFANA_DATASOURCE_URL to an invalid URL, thus causing the validate() method to return false, then on websocat, I will get this response:

Yea, that's just the standard response, since up until now that message has always been appropriate. That message should probably be reworded somehow (separately). Worth filing an issue to come back around to it.

vic-ma:
Updated to use `UrlValidator` and added the superclass' constructor call in `isAvailable()`; rebased on top of master.

>Yea, that's just the standard response, since up until now that message has always been appropriate. That message should probably be reworded somehow (separately). Worth filing an issue to come back around to it.

Okay, I'll create an issue for it after this gets merged.

pom.xml
```diff
@@ -45,6 +45,7 @@
 
   <org.apache.commons.lang3.version>3.9</org.apache.commons.lang3.version>
   <org.apache.commons.codec.version>1.13</org.apache.commons.codec.version>
+  <org.apache.commons.validator.routines.version>1.6</org.apache.commons.validator.routines.version>
```
andrewazores, line 48: I think this property should just be named `org.apache.commons.validator.version`, no? The `routines` is part of the specific package you're importing something from right now, but not part of the artifact name. Well, even the `org.apache` isn't part of the artifact name, but still.

vic-ma: Ah, yeah that makes sense...I wasn't too sure so I just copied it from the import statement haha.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
```diff
@@ -151,29 +159,38 @@ private ResponseMessage doPost(String targetId, String recordingName, String upl
...
         if (!isValidRecordingName) {
             cw.println(String.format("%s is an invalid recording name", recordingName));
         }
 
-        // TODO validate upload URL
+        boolean isValidDatasourceUrl =
+                new UrlValidator(UrlValidator.ALLOW_LOCAL_URLS)
+                        .isValid(env.getEnv(GRAFANA_DATASOURCE_ENV));
+        if (!isValidDatasourceUrl) {
+            cw.println(String.format("%s is an invalid datasource URL", GRAFANA_DATASOURCE_ENV));
```
andrewazores, line 185: Ah, I just noticed this on a final read-through. Should this printed error message include the (invalid) URL that was set as well, along with the name of the env var?

vic-ma: Yeah, I think that would be a good idea. I'll add that now.

vic-ma:
Error message now looks something like this:
```
[INFO] $GRAFANA_DATASOURCE_URL=localhost is an invalid datasource URL
```
I considered making it more explicit by having something like `var=GRAFANA_DATASOURCE_URL; value=localhost` but I feel like that is a bit too verbose, and for the purposes of a simple error message, this way gets the point across adequately without being too cluttered.

Also I noticed I forgot to add a check for the error message in the wrong datasource URLs unit test, so I added that in this commit.

PR #196, opened 2020-07-02
Change `upload-url` Command to Assume Datasource URL from Environment Variable

andrewazores:
See #193. Reopening after default branch rename.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
@@ -201,6 +201,7 @@ class UploadRecordingCommand extends AbstractConnectedCommand implements Seriali
 
         String targetId = args[0];
         String recordingName = args[1];
+        String datasourceUrl = env.getEnv(GRAFANA_DATASOURCE_ENV);
         StringJoiner combinedErrorMessage = new StringJoiner("; ");
 
         if (!validateTargetId(targetId)) {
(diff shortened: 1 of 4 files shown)


andrewazores:
@vic-ma could you check on that `pom.xml` conflict and rebase this PR against the latest?

vic-ma:
Re-ran `mvn verify` after rebasing so it seems to be working.

andrewazores:
Yea, I didn't think anything in the other PR should interfere with this. Cool.

vic-ma:
I see that the most recent commit changed some stuff relating to the `upload-recording` command. Do you think it's better to rebase and resolve conflicts on new commits as they come out, or wait until this PR is ready to be merged (when the web PR is done), and then rebase everything at once then?

andrewazores:
IMO it's best to immediately rebase as soon as anything else changes, particularly if there are merge conflicts. Otherwise the rebase just gets harder and harder over time.

There probably won't be anything else coming that touches `upload-recording` between now and the -web PR corresponding to this one, so in this case it works out the same either way really.

andrewazores:
@vic-ma this needs a rebase again since the Apache HttpClient removal, I think.

vic-ma:
Yeah, I saw and was planning to do the rebase after I finish my current PR.

After, I'll probably do the web PR that this one needs if Alex isn't working on it, so that this PR can finally be merged heh.

vic-ma:
Added commit to make `upload-recording` append `/load` to the URL itself.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
```diff
@@ -110,7 +110,7 @@ public String getName() {
     public void execute(String[] args) throws Exception {
         String targetId = args[0];
         String recordingName = args[1];
-        String datasourceUrl = env.getEnv(GRAFANA_DATASOURCE_ENV);
+        String datasourceUrl = env.getEnv(GRAFANA_DATASOURCE_ENV).concat("/load");
```
andrewazores, line 113: Does this work if the user sets the env var to ex. `http://localhost:8080/`, with a trailing slash?

vic-ma: Yes, tested just now and it works.

andrewazores: Cool. I thought a double-slash might break things, since it could be interpreted as a null or empty string as the first path segment. If the Vertx webclient and the jfr-datasource both accept this then no worries.

vic-ma: Yeah, when I was working on the web client I had a trailing slash at first and I noticed that even though the logs said it was uploading to `http://localhost:8080//load` which seemed like it'd break for sure, it was working. I guess it's cool that that works; makes stuff like this a lot easier.

andrewazores: It isn't too hard to clean it up anyway. There are cleaner ways using URIBuilder and such, but with just plain JDK classes:

`new URL(env.getEnv(GRAFANA_DATASOURCE_ENV)).toURI().resolve("/load").normalize().toString()`

should do it.

PR #199, opened 2020-07-02
Custom Templates

### LINKED ISSUE #143: Users should be able to create custom event templates

andrewazores:
Fixes #143

See also https://github.com/rh-jmc-team/container-jfr-core/pull/40 - -core dependency changes

This adds HTTP API endpoints to list available event templates, retrieve XML definitions of specific templates, and to add new templates (by XML file POST upload) to container-jfr's local filesystem storage. These templates can then be used when creating new flight recordings using the pre-existing `template=Foo` syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,13 +177,13 @@ deployments, port 9091 is expected for automatic port-scanning target discovery.
 
 ## EVENT TEMPLATES
 
-JDK Flight Recorder has a concept of "event templates", which are a sort of
-preset definition of a set of events, and for each a set of options and option
-values. A given JVM is likely to have some built-in templates ready for use
-out-of-the-box, but ContainerJFR also hosts its own small catalog of templates
-within its own local storage. This catalog is stored at the path specified by
-the environment variable `CONTAINER_JFR_TEMPLATE_PATH`. Templates can be
-uploaded to ContainerJFR and then used to create recordings.
+JDK Flight Recorder has event templates, which are preset definition of a set of
+events, and for each a set of options and option values. A given JVM is likely
+to have some built-in templates ready for use out-of-the-box, but ContainerJFR
+also hosts its own small catalog of templates within its own local storage. This
+catalog is stored at the path specified by the environment variable
+`CONTAINER_JFR_TEMPLATE_PATH`. Templates can be uploaded to ContainerJFR and
+then used to create recordings.
 
 ## ARCHIVING RECORDINGS
 
--- a/README.md
+++ b/README.md
@@ -175,6 +175,16 @@ number and the deployment network configuration allows connections on the
 configured port. As noted above, the final caveat is that in non-Kube
 deployments, port 9091 is expected for automatic port-scanning target discovery.
 
+## EVENT TEMPLATES
+
+JDK Flight Recorder has a concept of "event templates", which are a sort of
+preset definition of a set of events, and for each a set of options and option
+values. A given JVM is likely to have some built-in templates ready for use
+out-of-the-box, but ContainerJFR also hosts its own small catalog of templates
+within its own local storage. This catalog is stored at the path specified by
+the environment variable `CONTAINER_JFR_TEMPLATE_PATH`. Templates can be
+uploaded to ContainerJFR and then used to create recordings.
+
 ## ARCHIVING RECORDINGS
 
 `container-jfr` supports a concept of "archiving" recordings. This simply means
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -175,18 +175,28 @@ number and the deployment network configuration allows connections on the
 configured port. As noted above, the final caveat is that in non-Kube
 deployments, port 9091 is expected for automatic port-scanning target discovery.
 
+## EVENT TEMPLATES
+
+JDK Flight Recorder has a concept of "event templates", which are a sort of
```
jiekang, line 180: I'd prefer to be more concise, e.g.

JDK Flight Recorder event templates (.jfc files) are...

So things like "has a concept of..." or "are a sort of..." should be removed.

run.sh
```diff
@@ -41,10 +41,12 @@ if [ -z "$CONTAINER_JFR_AUTH_MANAGER" ]; then
     CONTAINER_JFR_AUTH_MANAGER="com.redhat.rhjmc.containerjfr.net.NoopAuthManager"
 fi
 
+
```
jiekang, line 44: Extra line can be removed.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/AbstractRecordingCommand.java
```diff
@@ -90,7 +94,24 @@ protected AbstractRecordingCommand(
             if (ALL_EVENTS_TEMPLATE.getName().equals(templateName)) {
                 return enableAllEvents(connection);
             }
-            return connection.getTemplateService().getEventsByTemplateName(templateName);
+            // if a template name is specified, try to find a Custom template by that name. If none,
+            // fall back on finding a target built-in template by that name. If not, throw an
+            // exception and bail out.
+            return connection
```
jiekang, line 100: I guess a side-effect here to keep in mind for the c-jfr-web UI listing of profiles is if a custom profile has the same name as a JVM one, it will make the JVM one no longer accessible with this API.

jiekang: Oh wait... this is the command while the TargetTemplateGetHandler supports specification of template type. Would it make sense to support that here too?

andrewazores: Yea, that's something I had in mind to figure out. I don't know how I want to change the `template=Foo` syntax, though. Any suggestions for how to specify custom vs remote template?

jiekang: Looking at the event syntax, comma delimited would be fine.

`template=X,type=Y.`

Or if you want to be more concise, though I guess disallowing commas in the name:
`template=X,Y`

Either seems fine to me. And for ease of use maybe the existing code can be kept for the case where users don't supply a template type.

andrewazores: I'll implement the `template=X,type=Y` format. I'll take the shortcut for now and require the keys to be in that order so I can just extend the existing validation regex, but maybe this should use a CSV parser instead... ?

The `template=Name,Type` format looks to me like a list of templates to enable, as if they would be unioned. That's also an interesting feature idea though.

jiekang: Okay, sounds fine. I think restricting order and using regex for now is okay. Re. list of templates, that's a good point. I'm not sure we'd ever want to bother with multiple template input; I don't think JFR/JMC does that with the jfc stuff, does it? But it could be a neat feature. Anyways heh

andrewazores: Yea I don't think the unioned template thing is done in JMC, but I could be wrong. Although the union `A U B U C ... U Z` does implicitly happen if you simultaneously create recordings from each template `A...Z` and then retrieve the contents of those recordings, doesn't it? But then you have a bunch of redundant recordings for no good reason.

jiekang: Ah true, I forgot about that. Recordings each have a single jfc but the event configuration in JFR is a union of all the recording event configs, which is in the end a union of the jfc files, heh.

PR #204, opened 2020-07-13
Make CORS origin configurable 

Alexjsenn:


Alexjsenn:
[log.txt](https://github.com/rh-jmc-team/container-jfr/files/4914863/log.txt)

src/test/java/com/redhat/rhjmc/containerjfr/net/web/handlers/CorsEnablingHandlerTest.java
```diff
@@ -93,6 +98,16 @@ void availabilityShouldDependOnEnvVar(boolean available) {
...
     }
 
+    @Test
+    void shouldSetCustomOrigin() {
+        Mockito.when(
+                        env.getEnv(
+                                CorsEnablingHandler.ENABLE_CORS_ENV,
+                                CorsEnablingHandler.DEV_ORIGIN))
+                .thenReturn("http://localhost:9065");
+        MatcherAssert.assertThat(handler.getOrigin(), Matchers.equalTo("http://localhost:9065"));
```
andrewazores, line 108: This just tests that the utility method `getOrigin` returns the expected value, but not that the CORS origin header is actually set. If you look at the `HandlingTest`s lower down in this same file, ex. `shouldAddHeadersToCORS`, you could copy how that test works and assert that the CORS origin header is actually added to the HTTP response with the expected origin.

Alexjsenn:
I modified the name of the env var, the README for cjfr doesn't make any mention of it, I will need to update the readme of the web-client. As for the tests, I think using a single custom origin for all of them essentially tests that it is working as expected.

Alexjsenn:
There's actually a stubbing error that I have to fix before this can be reviewed.

Alexjsenn:
Ok, should be good to go.

Alexjsenn:
[https://github.com/rh-jmc-team/container-jfr-web/pull/104](url)

andrewazores:
Please add a note to the README then. This is a configuration on the backend, so it should be documented here too.

Alexjsenn:
Ah, very true, I think that should do it.

PR #206, opened 2020-07-14
Change Response Message from `validate()`

### LINKED ISSUE #195: Response Message from a Failed `validate()` Should be Changed
After #193, it no longer makes sense for the response message from a failed `validate()` call to only talk about improper arguments, since the `validate()` in `UploadRecordingCommand` can now fail due to an improper environment variable.

vic-ma:
Fixes #195.

In commands, `validate()` is now a `void` method which `throws FailedValidationException`. If the validation fails, then the method prints the error message with its `ClientWriter`, and then throws a `FailedValidationException` which contains the same error message. If the validation is successful, the method simply returns. 

In command registries, the `validate()` method has been to match `Command.validate()`.

In `WsCommandExecutor`, the `ResponseMessage` from a failed validation has its `payload` changed to say that the command could not be validated, and includes the error message from the `validate()` call (obtained via the caught exception). Previously the response just said that the given arguments were invalid.
```
{command:list}
{"id":null,"commandName":"list","status":-1,"payload":"Could not validate \"list\" command; Expected one argument: hostname:port, ip:port, or JMX service URL"}
```

In the `AbstractCommandExecutor`, `validateCommands()` has been changed to output a more generic error message upon a failed validation; it now just says that the validation failed, whereas previously it specified that the arguments passed in were invalid. It prints a generic error message instead of the specific error message from the caught exception, because the `validate()` method in the commands will have already printed the specific error message.
```
> list
Expected one argument: hostname:port, ip:port, or JMX service URL
	Command "list" could not be validated
```

Additional, somewhat unrelated things of note: I did some light refactoring as I was working on the `*CommandTest` tests, like adding a check for the error message when there wasn't before, and changing `argc` validation test names to be more consistent. The `connect` and `disconnect` in `BatchModeExecutorTest` were changed to existing commands. Also I made a small fix to `COMMANDS.md`.

**Edit:** The commit, "Change tests to reflect the re-added cw.println() in validate()" should really just be, "Change Command unit tests to work with new validate()"; the unit tests for commands weren't actually ever modified before that commit.

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -249,7 +249,7 @@ formatted as a JSON response.
 
 * #### `list`
     ###### usage
-    `list target`
+    `list targetId`
     ###### synopsis
     Lists recordings in the specified target JVM. The name provided in this list
     is the name to pass to other commands which operate upon recordings.
@@ -275,11 +275,11 @@ formatted as a JSON response.
 
 * #### `wait-for`
     ###### usage
-    `wait-for foo`
+    `wait-for targedId foo`
     ###### synopsis
-    Waits for the given recording (`foo`) to stop running. If the recording is
-    continuous and not already stopped then the command will refuse to wait for
-    the recording to complete, since this would lock up the client and require
-    another client to be connected in order to stop the recording. Once this
-    command has begun awaiting completion of the recording it cannot be
-    interrupted.
+    Waits for the given recording (`foo`) of the specified target JVM to stop
+    running. If the recording is continuous and not already stopped then the
+    command will refuse to wait for the recording to complete, since this would
+    lock up the client and require another client to be connected in order to
+    stop the recording. Once this command has begun awaiting completion of the
+    recording it cannot be interrupted.
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -275,7 +275,7 @@ formatted as a JSON response.
 
 * #### `wait-for`
     ###### usage
-    `wait-for targedId foo`
+    `wait-for targetId foo`
     ###### synopsis
     Waits for the given recording (`foo`) of the specified target JVM to stop
     running. If the recording is continuous and not already stopped then the
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/StartRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/StartRecordingCommand.java
@@ -135,33 +135,29 @@ class StartRecordingCommand extends AbstractRecordingCommand implements Serializ
     }
 
     @Override
-    public boolean validate(String[] args) {
+    public void validate(String[] args) throws FailedValidationException {
         if (args.length != 3) {
-            cw.println(
+            throw new FailedValidationException(
                     "Expected three arguments: target (host:port, ip:port, or JMX service URL), recording name, and event types");
-            return false;
         }
 
         String targetId = args[0];
         String name = args[1];
         String events = args[2];
 
-        boolean isValidTargetId = validateTargetId(targetId);
-        boolean isValidName = validateRecordingName(name);
-        boolean isValidEvents = validateEvents(events);
-
-        if (!isValidTargetId) {
-            cw.println(String.format("%s is an invalid connection specifier", args[0]));
+        if (!validateTargetId(targetId)) {
+            throw new FailedValidationException(
+                    String.format("%s is an invalid connection specifier", args[0]));
         }
 
-        if (!isValidName) {
-            cw.println(String.format("%s is an invalid recording name", name));
+        if (!validateRecordingName(name)) {
+            throw new FailedValidationException(
+                    String.format("%s is an invalid recording name", name));
         }
 
-        if (!isValidEvents) {
-            cw.println(String.format("%s is an invalid events specifier", events));
+        if (!validateEvents(events)) {
+            throw new FailedValidationException(
+                    String.format("%s is an invalid events specifier", events));
         }
-
-        return isValidTargetId && isValidName && isValidEvents;
     }
 }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/StartRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/StartRecordingCommand.java
@@ -41,6 +41,8 @@
  */
 package com.redhat.rhjmc.containerjfr.commands.internal;
 
+import java.util.StringJoiner;
+
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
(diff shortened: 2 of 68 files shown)


vic-ma:
Force pushed to rebase onto main.

COMMANDS.md
```diff
@@ -275,11 +275,11 @@ formatted as a JSON response.
 
 * #### `wait-for`
     ###### usage
-    `wait-for foo`
+    `wait-for targedId foo`
```
andrewazores, line 278: Typo: `s/targedId/targetId/`

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/StartRecordingCommand.java
```diff
@@ -135,33 +135,34 @@ public void execute(String[] args) throws Exception {
...
         String events = args[2];
 
-        boolean isValidTargetId = validateTargetId(targetId);
-        boolean isValidName = validateRecordingName(name);
-        boolean isValidEvents = validateEvents(events);
-
-        if (!isValidTargetId) {
-            cw.println(String.format("%s is an invalid connection specifier", args[0]));
+        if (!validateTargetId(targetId)) {
+            String errorMessage = "%s is an invalid connection specifier";
```
andrewazores, line 151: In this kind of scenario where there are multiple validation checks that can fail, beyond the most basic "wrong number of arguments" failure, how about performing all of the checks and constructing a message informing the client about all of the failures, rather than just the "first" in the arbitrary implementation order?

andrewazores: This would match more closely with the previous behaviour, which would print out all of the failures before returning the value of the validation result.

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/AbstractConnectedCommandTest.java
```diff
@@ -141,8 +141,8 @@ public String getName() {
         public void execute(String[] args) {}
 
         @Override
-        public boolean validate(String[] args) {
-            return true;
+        public void validate(String[] args) throws FailedValidationException {
+            return;
```
andrewazores, line 145: The `return` is implicit for a `void` method, so you can just have an empty method body between the curly braces.

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/AbstractRecordingCommandTest.java
```diff
@@ -217,8 +217,8 @@ public String getName() {
         }
 
         @Override
-        public boolean validate(String[] args) {
-            return true;
+        public void validate(String[] args) throws FailedValidationException {
+            return;
```
andrewazores, line 221: Same here re: empty method body/implicit return

andrewazores:
Change looks great overall, very nice work @vic-ma . The only significant thing I would like to see done in addition is to fix that "short-circuiting" behaviour of the validation that I pointed out above, where clients will be given a message only informing them about one validation error in cases where there are really multiple.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/StartRecordingCommand.java
vic-ma: Ah, I didn't realize that.

For combining them, what do you think about just concatenating them together with some sort of separator? For example:
```
Multiple validation checks failed: %s is an invalid connection specifier; %s is an invalid recording name; %s is an invalid events specifier
```

And do you have any thoughts on adding tests to support this? I feel like being completely exhaustive and testing all the possible multiple-failure combinations is probably a bit too much? So maybe we could just do a 2-fail and all-fail test for `validate()` methods with more than two checks, and obviously just the 2-fail for any that have two checks?

andrewazores: Yea, semicolons to separate them seems good to me.

I think it's probably worth exhaustively testing all the validation cases, even if it does seem repetitive/monotonous to do. It's easy for a refactor or some other change in the future to accidentally change the semantics of the method, and that's a regression case that unit tests are great for catching. The command validation step should be considered as part of our command channel API, so we want to be very sure that it's stable between versions.

vic-ma:
I've fixed the short-circuiting. Now, a failed validation with multiple failed checks looks like this:
```
{command:dump,args:[localhost, ":", 1, ":"]}
{"id":null,"commandName":"dump","status":-1,"payload":"Could not validate \"dump\" command: : is an invalid recording name; : is an invalid events specifier"}
```

Also, as I was working on this commit, I realized that in quite a few `validate()` methods, instead of having something like this:
```
String errorMessage = String.format("% is invalid", args[0]);
throw new FailedValidationException(errorMessage);
```
I had something like this:
```
String errorMessage = "% is invalid";
throw new FailedValidationException(String.format(errorMessage, args[0]));
```
While it worked, it was definitely not intended, and must've been a mess-up with my macro. I've fixed it in this commit.

Also, force pushed to rebase onto main.

andrewazores:
Just noticed something in testing:

```
> dump . . . .  
. is an invalid events pattern
. is an invalid recording name
. is an invalid recording length
. is an invalid events specifier
	"[., ., ., .]" are invalid arguments to dump
```

The command's own validation method prints a message that `. is an invalid events specifier`, but the `AbstractRecordingCommand#validateEvents` method that it calls also prints `. is an invalid events pattern`.

andrewazores:
This is a minor complaint, but the `stringBuilder.toString().substring(2)` pattern that appears at the end of the validation methods now seems a little unfortunate (because of the substring, which relies on the separator being prepended and being two characters long).

```
List<String> validationErrors = new ArrayList<>();
if (!firstCheck()) {
  validationErrors.add("check 1 failed");
}
if (!secondCheck()) {
  validationErrors.add("check 2 failed');
}
throw new ValidationFailureException(String.join("; ", validationErrors));
```

might be a little cleaner?

vic-ma:
>The command's own validation method prints a message that . is an invalid events specifier, but the AbstractRecordingCommand#validateEvents method that it calls also prints . is an invalid events pattern.

Ah, I think I noticed there were two messages about events when I was testing `dump` but thought it was just something about recording commands I didn't understand. But I see what you mean now. It looks like the double printing is in upstream as well.

I think it's probably better to remove the print from `validateEvents()` than `validate()`? Since the other validate methods like `validateTargetId` don't print, and it's probably better for the check itself not to have a hidden print statement?

>This is a minor complaint, but the stringBuilder.toString().substring(2) pattern that appears at the end of the validation methods now seems a little unfortunate (because of the substring, which relies on the separator being prepended and being two characters long).

Yes I think your way makes more sense. I'll change it to that now.

andrewazores:
Yea, both of those comments sound good to me.

vic-ma:
I've finished making the changes, but now that I'm reading about it, would it be better to use `StringJoiner` over `ArrayList` and `join()`? That way we get rid of the `ArrayList` and do the joining sort of more directly.

andrewazores:
Sure. I've never seen `StringJoiner` before but it looks like a good tool for the job.

vic-ma:
I've pushed the changes. (Figuring out the right logic for the `validateEvents()` return statement took me a second heh.)

andrewazores:
This seems like it's probably a pre-existing problem, but:

```
{command:dump,args:[.,,10,.]}
<... hanging>
```

```
[INFO] (0:0:0:0:0:0:0:1:35150): CMD {command:dump,args:[.,,10,.]}

[WARNING] 
    at com.redhat.rhjmc.containerjfr.commands.Command.validateRecordingName (Command.java:74)
    at com.redhat.rhjmc.containerjfr.commands.internal.DumpCommand.validate (DumpCommand.java:165)
    at com.redhat.rhjmc.containerjfr.commands.internal.SerializableCommandRegistryImpl.validate (SerializableCommandRegistryImpl.java:116)
    at com.redhat.rhjmc.containerjfr.tui.ws.WsCommandExecutor.run (WsCommandExecutor.java:113)
    at com.redhat.rhjmc.containerjfr.ContainerJfr.main (ContainerJfr.java:115)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:566)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:282)
    at java.lang.Thread.run (Thread.java:834)
```

vic-ma:
It looks like this is just because our `validate()` methods don't check for null arguments, so any null strings in our args lists can get passed to the `validateXYZ()` methods as usual, where a NPE is eventually raised. There are also a few cases, like the `seconds` arg in `dump`, where a `String` instance method is called on the null reference.

I think the way to fix this would be to just add a null check for all the arguments, in between the argc check and the actual arg validation checks.

But since this change is sort of big, and isn't directly related to this PR (which is big enough already as it is hah) I think it's probably better to make a separate issue and PR for it?

andrewazores:
Sure, please file the new issue then and I'll approve this PR.

Issue #191, opened 2020-06-23
WebServer "low memory pressure" log message is emitted twice


andrewazores:
`[INFO] low memory pressure streaming disabled for web server` is seen twice in the startup logs.

See also: #190

PR #208, opened 2020-07-16
Remove duplicate "low memory pressure" log message

Alexjsenn:
#191


Alexjsenn:
You're right, they're not necessary, I removed them.

Issue #205, opened 2020-07-14
[API] Serialized command arguments should be map, not array


andrewazores:
Currently, the format for interactively executing a command is `commandName arg1 arg2`, and this is mirrored in the JSON serialization of command messages: `{"command":"commandName","args":["arg1","arg2"]}`. This was easy to implement at the time, but misses an opportunity to make better use of the JSON formatted requests.

For example, the `dump` command is currently issued like:
`{ "command": "dump", "args":["localhost", "foo", "10", "template=Profiling,type=TARGET"]}`

This would be much better serialized similar to:
```
{
  "command": "dump",
  "args": {
    "targetId": "localhost",
    "recordingName": "foo",
    "duration": 10,
    "events": {
      "template": "Profiling",
      "type": "TARGET"
    }
  }
}
```

There is still some question as to how best to serialize arguments that can be represented in more than one format - the event specifiers for example can currently be of the template form `template=Profiling,type=TARGET` (or simply `template=Profiling`), but can also be the individual event option specifier lists (`jdk.SocketRead:enabled=true,jdk.SocketWrite:enabled=true`).

This is a nice API enhancement since it makes it easier for clients to implement the API, since the ordering of the arguments becomes irrelevant and they are named instead, and it also makes it much cleaner and simpler to modify command implementations to add or remove arguments, or even to add optional arguments. This will be useful in #5 for example, to allow optional credentials.

andrewazores:
On a side note - perhaps this is a good opportunity to re-evaluate the WebSocket API vs the HTTP (REST-ish) API. Maybe it should be that the WebSocket API should not implement JMX auth at all, ie that WebSocket commands should only be commands for the ContainerJFR instance itself, and nothing to do with targets. This would very substantially reduce the number of commands implemented, requiring just about everything to be re-implemented as an HTTP request handler instead. The WebSocket connection would still be useful for sending notifications from the server to clients so that they can keep up-to-date state without polling.

andrewazores:
Removing myself, no longer planning to work on this in the near term.

This is still a worthwhile enhancement to the Command Channel API, however we're moving in a direction where the Command Channel will have many fewer uses than it currently does with most of that functionality moving to the HTTP REST API instead. With a smaller API surface and commands which require many fewer arguments to begin with there is less pressure to implement a better argument handling system.

andrewazores:
See #222

andrewazores:
[...]

PR #209, opened 2020-07-16
Initial JMX authentication foundation

andrewazores:
Related to #5

See also rh-jmc-team/container-jfr-core#42

This PR lays down initial support for connections to targets using JMX access control/authentication. No mechanism for the user to supply JMX credentials is yet implemented, but connections to targets configured with JMX security will now fail with an identifiable exception, including an appropriate HTTP header for HTTP API requests or an appropriate status code for WebSocket API commands. HTTP API support for supplying credentials will be relatively simple - the `AbstractAuthenticatedRequestHandler#getConnectionDescriptorFromContext` method will simply be updated to check the `Proxy-Authorization` request header for Basic auth credentials to pass on to the target JMX connection. WebSocket API support for optional credentials will require more work (see: #205 for one proposed solution), or perhaps the WebSocket API will simply never gain JMX auth support and all such requests will need to go through the HTTP REST API instead.

Additionally, the ContainerJFR container image is now built with a custom entrypoint script, and the entrypoint by default enables JMX access control on the container's own remote JMX port, with an autogenerated password for the default "containerjfr" user. This password can be seen in the ContainerJFR instance's logs at startup time. This authentication (and the password generation) can be disabled for testing/development purposes. The SelfDisoveryPlatformClient is also removed and JDP autodiscovery is enabled instead so that the ContainerJFR instance can use JDP to discover itself within its container in a more uniform manner, providing itself with its own JMXServiceURL and Main-Class name. In environments where JDP autodiscovery is not used, ex. Kubernetes/OpenShift, this also has the net effect of preventing the self-connection list item from appearing in the discovered targets list.


src/main/java/com/redhat/rhjmc/containerjfr/net/TargetConnectionManager.java
```diff
@@ -93,56 +97,55 @@
...
+    public JFRConnection connect(ConnectionDescriptor connectionDescriptor) throws Exception {
         try {
-            return attemptConnectAsJMXServiceURL(targetId);
+            return attemptConnectAsJMXServiceURL(connectionDescriptor);
+        } catch (MalformedURLException mue) {
+            return attemptConnectAsHostPortPair(connectionDescriptor);
         } catch (Exception e) {
-            return attemptConnectAsHostPortPair(targetId);
+            throw e;
         }
```
ebaron, line 107: I think we can remove this catch, or does it still serve a purpose?

src/main/java/com/redhat/rhjmc/containerjfr/net/TargetConnectionManager.java
```diff
@@ -93,56 +97,55 @@
...
         String host = m.group(1);
         String port = m.group(2);
         if (port == null) {
             port = "9091";
         }
-        return connect(host, Integer.parseInt(port));
+        return connect(
+                new JMXServiceURL(
+                        "rmi", "", 0, String.format("/jndi/rmi://%s:%s/jmxrmi", host, port)),
```
ebaron, line 131: Real minor thing, but this would be more readable IMO if we used JMC's ConnectionToolkit.createServiceURL. Maybe worth adding a wrapper for it in core next time you make changes there.

src/main/extras/app/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+set -x
+set -e
+
+PWFILE="/tmp/jmxremote.password"
+function createJmxPassword() {
+    PASS="$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c${1:-32})"
```
ebaron, line 8: I don't see this being called anywhere with an argument. We could probably simplify this to `head -c32`, unless you want to keep this length parameter for the future.

src/main/java/com/redhat/rhjmc/containerjfr/net/TargetConnectionManager.java
andrewazores: That's a good call. I'll file a -core issue so it doesn't get forgotten.

andrewazores: https://github.com/rh-jmc-team/container-jfr-core/issues/44

src/main/extras/app/entrypoint.sh
andrewazores: Yea it is currently unused. It can always be put back in the future, so I'll take it out for now and leave it at fixed-length of 32.

andrewazores:
Yea, I wasn't sure what made the most sense from the Operator's POV, but that makes sense. If a password is provided via env var then use that, otherwise generate it as implemented here. Do you have any preference for how that env var is named?

ebaron:
> Yea, I wasn't sure what made the most sense from the Operator's POV, but that makes sense. If a password is provided via env var then use that, otherwise generate it as implemented here. Do you have any preference for how that env var is named?

Maybe just `CONTAINER_JFR_RJMX_{USER,PASS}` or something? That can be in a future PR though, this one looks good!

Issue #207, opened 2020-07-14
[packaging] Base image digest should be specified


andrewazores:
The base image is currently specified without a version tag or digest, so implicitly the image build step will use the `latest` tag of the base image. This is not good for reproducibility of our downstream images, and is an additional testing variable that we are not currently controlling.

PR #210, opened 2020-07-16
specify version for base image

Alexjsenn:
#207 
Not sure if we need to specify the exact version (1.3-3.1593114401) or if just 1.3 is enough. Both would work.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
   <maven.compiler.source>${java.version}</maven.compiler.source>
 
   <imageBuilder>/usr/bin/podman</imageBuilder>
-  <baseImage>registry.access.redhat.com/ubi8/openjdk-11</baseImage>
+  <baseImage>registry.access.redhat.com/ubi8/openjdk-11:1.3-3.1593114401</baseImage>
   <node.version>v12.5.0</node.version>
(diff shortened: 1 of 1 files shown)


andrewazores:
I think the full `1.3-3.1593114401` is better, as it points to exactly one image version. `1.3` happens to point to that same version right now, but when the next release is made it may also be tagged as a `1.3` release and that tag updated to point to that next version.

pom.xml
```diff
@@ -17,7 +17,7 @@
   <maven.compiler.source>${java.version}</maven.compiler.source>
 
   <imageBuilder>/usr/bin/podman</imageBuilder>
-  <baseImage>registry.access.redhat.com/ubi8/openjdk-11</baseImage>
+  <baseImage>registry.access.redhat.com/ubi8/openjdk-11:1.3-3.1593114401</baseImage>
```
andrewazores, line 20: Rather than appending it directly to the `baseImage` property, what do you think about creating a new property like `baseImageTag` and specifying the version there? Further down in the file, when the `baseImage` is referenced by the Jib plugin, you would then specify the `from` image like `<from>${baseImage}:${baseImageTag}</from>`.

Alexjsenn: ah, looks better, done!

PR #212, opened 2020-07-20
Remove `demos/`

### LINKED ISSUE #211: `demos/print_help` Doesn't Work After the Connectionless API Changes
The file is:
```
help;
connect container-jfr:9091;
help;
```
But we don't have the `connect` command anymore, so running it with `sh run.sh "$(more demos/print_help)"` doesn't work.

We could fix it by having the file just be:
```
help;
```

But at that point maybe it would be better to just tell the reader to run `help` themselves? Since now there isn't two sets of commands (connected/unconnected), so all they need to do is type `help` into the interactive execution.

Or was the point of `print_help` to demonstrate the batch mode execution?

vic-ma:
Fixes #211.

In the README, I also removed a mention of `connect localhost`.

Also I removed the, "all three execution modes described above" part, since it looks like [the description](https://github.com/rh-jmc-team/container-jfr/commit/9719664f51ce49d9d080dee943b46c7e9d0c1529#diff-04c6e90faac2675aa89e2176d2eec7d8R39) it's referring to doesn't exist in our current README anymore.

Also I added a backslash to a `1.` at the end of a sentence, so it isn't interpreted as the start of a numbered list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,7 @@ WebSocket mode with a web frontend.
 
 The `run.sh` script can be used to spin up a `podman` container of the Container
 JFR Client, running alone but set up so that it is able to introspect itself
-with JFR. This can be achieved by doing `sh run.sh -it` and then typing
-`connect localhost` into the client shell that appears. When running in
-this container, all three execution modes described above are still available
-and accessible using the same mthods.
+with JFR. This can be achieved by doing `sh run.sh -it`.
 
 There are six network-related environment variables that the client checks
 during its runtime:
@@ -88,7 +85,7 @@ will be allowed. If this is not set then the default value is 2. Once the
 maximum number of concurrent connections is reached, the server will reject
 handshakes for any new incoming connections until a previous connection is
 closed. The maximum acceptable value is 64 and the minimum acceptable value is
-1. Values outside of this range will be ignored and the default value set
+1\. Values outside of this range will be ignored and the default value set
 instead.
 
 The environment variable `CONTAINER_JFR_LOG_LEVEL` is used to control the level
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -63,12 +57,7 @@ WebSocket mode with a web frontend.
 
 The `run.sh` script can be used to spin up a `podman` container of the Container
 JFR Client, running alone but set up so that it is able to introspect itself
-with JFR. This can be achieved by doing `sh run.sh -it` and then typing
-`connect localhost` into the client shell that appears. When running in
-this container, all three execution modes described above are still available
-and accessible using the same mthods. Some client shell demo scripts are also
-available in the `demos` directory. These can be used with batch mode, ex.
-`sh run.sh "$(more demos/print_help)"`.
+with JFR. This can be achieved by doing `sh run.sh -it`.
```
andrewazores, line 60: While we're here, let's remove that `-it`. I'd rather point users to a tool like `websocat` for use with the WebSocket mode than encourage usage of the interactive mode. Maybe we should link to `websocat`?

vic-ma:
Yes, how about something like this:

This can be achieved by running `sh run.sh` and connecting to Container JFR in a separate terminal using [websocat](https://github.com/vi/websocat). The WebSocket URL to connect to can be found by running `curl localhost:8181/api/v1/clienturl`. Once you are connected, you can issue commands by entering them into the websocat client in JSON form. For example, `{command:ping}` or `{command:dump,args:[localhost,foo,10,"template=Continuous"]}`.

andrewazores:
Sounds good to me.

Issue #183, opened 2020-06-16
PlatformClients should report JMX Service URLs


andrewazores:
Currently, some PlatformClients' discovery mechanisms return lists containing JMX Service URLs, and others return `host:port` pairs. Both of these are acceptable in the various APIs requiring a targetID, however the `host:port` pairs should only really be expected to be used in users' hand-written requests, ie in development/debugging/testing as a shorthand. API responses should uniformly contain JMX Service URLs to eliminate ambiguity.

This also means that the ServiceRef type would no longer need its `port` field, only the connection URL and an optional display alias.

Alexjsenn:
I looked into removing the port field, but there are a few places where Im not sure if the behaviour is appropriate. For example, in scanTargets it would only give the ULR and not the port, is this acceptable?

andrewazores:
A JMX Service URL should already include the port number, so yes.

andrewazores:
Fixed in #215

PR #215, opened 2020-07-22
Unify API responses to only contain JMX service URLs

Alexjsenn:
#183

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
@@ -86,11 +86,11 @@ public abstract class MainModule {
     // public since this is useful to use directly in tests
     @Provides
     @Singleton
-    public static Gson provideGson() {
+    public static Gson provideGson(Logger logger) {
         return new GsonBuilder()
                 .serializeNulls()
                 .disableHtmlEscaping()
-                .registerTypeAdapter(JMXServiceURL.class, new GsonJmxServiceUrlAdapter())
+                .registerTypeAdapter(JMXServiceURL.class, new GsonJmxServiceUrlAdapter(logger))
                 .create();
     }
 
--- a/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
@@ -46,6 +46,7 @@ import java.nio.file.Paths;
 
 import javax.inject.Named;
 import javax.inject.Singleton;
+import javax.management.remote.JMXServiceURL;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -58,6 +59,7 @@ import com.redhat.rhjmc.containerjfr.platform.PlatformModule;
 import com.redhat.rhjmc.containerjfr.sys.SystemModule;
 import com.redhat.rhjmc.containerjfr.templates.TemplatesModule;
 import com.redhat.rhjmc.containerjfr.tui.TuiModule;
+import com.redhat.rhjmc.containerjfr.util.GsonJmxServiceUrlAdapter;
 
 import dagger.Module;
 import dagger.Provides;
@@ -85,7 +87,11 @@ public abstract class MainModule {
     @Provides
     @Singleton
     public static Gson provideGson() {
-        return new GsonBuilder().serializeNulls().disableHtmlEscaping().create();
+        return new GsonBuilder()
+                .serializeNulls()
+                .disableHtmlEscaping()
+                .registerTypeAdapter(JMXServiceURL.class, new GsonJmxServiceUrlAdapter())
+                .create();
     }
 
     @SuppressFBWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/PlatformClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/PlatformClient.java
@@ -43,8 +43,6 @@ package com.redhat.rhjmc.containerjfr.platform;
 
 import java.util.List;
 
-import com.redhat.rhjmc.containerjfr.core.log.Logger;
-
 public interface PlatformClient {
-    List<ServiceRef> listDiscoverableServices(Logger logger);
+    List<ServiceRef> listDiscoverableServices();
 }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/PlatformClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/PlatformClient.java
@@ -43,6 +43,8 @@ package com.redhat.rhjmc.containerjfr.platform;
 
 import java.util.List;
 
+import com.redhat.rhjmc.containerjfr.core.log.Logger;
+
 public interface PlatformClient {
-    List<ServiceRef> listDiscoverableServices();
+    List<ServiceRef> listDiscoverableServices(Logger logger);
 }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -53,11 +53,11 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 public class ServiceRef {
 
     private final JMXServiceURL JMXServiceURL;
-    private Optional<String> alias = Optional.empty();
+    private final String alias; // nullable
 
     public ServiceRef(JMXServiceURL jmxServiceUrl, String alias) throws MalformedURLException {
         this.JMXServiceURL = jmxServiceUrl;
-        this.alias = Optional.ofNullable(alias);
+        this.alias = alias;
     }
 
     public ServiceRef(JMXServiceURL jmxServiceUrl) throws MalformedURLException {
@@ -65,7 +65,7 @@ public class ServiceRef {
     }
 
     public ServiceRef(String host, int port, String alias) throws MalformedURLException {
-        this(new JMXServiceURL(null, host, port), null);
+        this(new JMXServiceURL(null, host, port), alias);
     }
 
     public JMXServiceURL getJMXServiceUrl() {
@@ -73,7 +73,7 @@ public class ServiceRef {
     }
 
     public Optional<String> getAlias() {
-        return alias;
+        return Optional.ofNullable(alias);
     }
 
     @Override
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -50,8 +50,11 @@ import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 
+import com.google.gson.annotations.SerializedName;
+
 public class ServiceRef {
 
+    @SerializedName("connectUrl")
     private final JMXServiceURL JMXServiceURL;
     private final String alias; // nullable
 
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -68,7 +68,10 @@ public class ServiceRef {
     }
 
     public ServiceRef(String host, int port, String alias) throws MalformedURLException {
-        this(new JMXServiceURL(null, host, port), alias);
+        this(
+                new JMXServiceURL(
+                        "rmi", "", 0, String.format("/jndi/rmi://%s:%s/jmxrmi", host, port)),
+                alias);
     }
 
     public JMXServiceURL getJMXServiceUrl() {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -41,6 +41,8 @@
  */
 package com.redhat.rhjmc.containerjfr.platform;
 
+import java.util.Optional;
+
 import javax.management.remote.JMXServiceURL;
 
 import org.apache.commons.lang.builder.EqualsBuilder;
@@ -50,27 +52,27 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 public class ServiceRef {
 
     private final JMXServiceURL JMXServiceURL;
-    private final String alias;
+    private Optional<String> alias = Optional.empty();
 
     public ServiceRef(JMXServiceURL JMXServiceURL) {
-        this(JMXServiceURL, JMXServiceURL.toString());
+        this.JMXServiceURL = JMXServiceURL;
     }
 
     public ServiceRef(JMXServiceURL JMXServiceURL, String alias) {
         this.JMXServiceURL = JMXServiceURL;
-        this.alias = alias;
+        this.alias = Optional.of(alias);
     }
 
-    public ServiceRef(String host, String alias, int port) throws Exception {
+    public ServiceRef(String host, int port, String alias) throws Exception {
         this.JMXServiceURL = new JMXServiceURL(null, host, port);
-        this.alias = alias;
+        this.alias = Optional.of(alias);
     }
 
     public JMXServiceURL getJMXServiceUrl() {
         return JMXServiceURL;
     }
 
-    public String getAlias() {
+    public Optional<String> getAlias() {
         return alias;
     }
 
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -70,7 +70,7 @@ public class ServiceRef {
     public ServiceRef(String host, int port, String alias) throws MalformedURLException {
         this(
                 new JMXServiceURL(
-                        "rmi", "", 0, String.format("/jndi/rmi://%s:%s/jmxrmi", host, port)),
+                        "rmi", "", 0, String.format("/jndi/rmi://%s:%d/jmxrmi", host, port)),
                 alias);
     }
 
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -41,26 +41,33 @@
  */
 package com.redhat.rhjmc.containerjfr.platform;
 
+import javax.management.remote.JMXServiceURL;
+
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 
 public class ServiceRef {
 
-    private final String connectUrl;
+    private final JMXServiceURL JMXServiceURL;
     private final String alias;
 
-    public ServiceRef(String connectUrl) {
-        this(connectUrl, connectUrl);
+    public ServiceRef(JMXServiceURL JMXServiceURL) {
+        this(JMXServiceURL, JMXServiceURL.toString());
     }
 
-    public ServiceRef(String connectUrl, String alias) {
-        this.connectUrl = connectUrl;
+    public ServiceRef(JMXServiceURL JMXServiceURL, String alias) {
+        this.JMXServiceURL = JMXServiceURL;
         this.alias = alias;
     }
 
-    public String getConnectUrl() {
-        return connectUrl;
+    public ServiceRef(String host, String alias, int port) throws Exception {
+        this.JMXServiceURL = new JMXServiceURL(null, host, port);
+        this.alias = alias;
+    }
+
+    public JMXServiceURL getJMXServiceUrl() {
+        return JMXServiceURL;
     }
 
     public String getAlias() {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -49,13 +49,13 @@ import javax.management.remote.JMXServiceURL;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 import org.apache.commons.lang3.builder.ToStringBuilder;
-
 import com.google.gson.annotations.SerializedName;
 
 public class ServiceRef {
 
     @SerializedName("connectUrl")
     private final JMXServiceURL JMXServiceURL;
+
     private final String alias; // nullable
 
     public ServiceRef(JMXServiceURL jmxServiceUrl, String alias) throws MalformedURLException {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -49,16 +49,14 @@ public class ServiceRef {
 
     private final String connectUrl;
     private final String alias;
-    private final int port;
 
-    public ServiceRef(String connectUrl, int port) {
-        this(connectUrl, connectUrl, port);
+    public ServiceRef(String connectUrl) {
+        this(connectUrl, connectUrl);
     }
 
-    public ServiceRef(String connectUrl, String alias, int port) {
+    public ServiceRef(String connectUrl, String alias) {
         this.connectUrl = connectUrl;
         this.alias = alias;
-        this.port = port;
     }
 
     public String getConnectUrl() {
@@ -67,10 +65,6 @@ public class ServiceRef {
 
     public String getAlias() {
         return alias;
-    }
-
-    public int getPort() {
-        return port;
     }
 
     @Override
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
@@ -41,6 +41,7 @@
  */
 package com.redhat.rhjmc.containerjfr.platform;
 
+import java.net.MalformedURLException;
 import java.util.Optional;
 
 import javax.management.remote.JMXServiceURL;
@@ -54,18 +55,17 @@ public class ServiceRef {
     private final JMXServiceURL JMXServiceURL;
     private Optional<String> alias = Optional.empty();
 
-    public ServiceRef(JMXServiceURL JMXServiceURL) {
-        this.JMXServiceURL = JMXServiceURL;
+    public ServiceRef(JMXServiceURL jmxServiceUrl, String alias) throws MalformedURLException {
+        this.JMXServiceURL = jmxServiceUrl;
+        this.alias = Optional.ofNullable(alias);
     }
 
-    public ServiceRef(JMXServiceURL JMXServiceURL, String alias) {
-        this.JMXServiceURL = JMXServiceURL;
-        this.alias = Optional.of(alias);
+    public ServiceRef(JMXServiceURL jmxServiceUrl) throws MalformedURLException {
+        this(jmxServiceUrl, null);
     }
 
-    public ServiceRef(String host, int port, String alias) throws Exception {
-        this.JMXServiceURL = new JMXServiceURL(null, host, port);
-        this.alias = Optional.of(alias);
+    public ServiceRef(String host, int port, String alias) throws MalformedURLException {
+        this(new JMXServiceURL(null, host, port), null);
     }
 
     public JMXServiceURL getJMXServiceUrl() {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
@@ -67,7 +67,7 @@ class DefaultPlatformClient implements PlatformClient {
                         u -> {
                             try {
                                 return new ServiceRef(
-                                        u.getJmxServiceUrl().toString(), u.getMainClass(), 0);
+                                        u.getJmxServiceUrl().toString(), u.getMainClass());
                             } catch (MalformedURLException e) {
                                 log.info(e);
                                 return null;
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
@@ -52,22 +52,19 @@ import com.redhat.rhjmc.containerjfr.platform.ServiceRef;
 
 class DefaultPlatformClient implements PlatformClient {
 
-    private final Logger log;
     private final JvmDiscoveryClient discoveryClient;
 
-    DefaultPlatformClient(Logger log, JvmDiscoveryClient discoveryClient) {
-        this.log = log;
+    DefaultPlatformClient(JvmDiscoveryClient discoveryClient) {
         this.discoveryClient = discoveryClient;
     }
 
     @Override
-    public List<ServiceRef> listDiscoverableServices() {
+    public List<ServiceRef> listDiscoverableServices(Logger log) {
         return discoveryClient.getDiscoveredJvmDescriptors().stream()
                 .map(
                         u -> {
                             try {
-                                return new ServiceRef(
-                                        u.getJmxServiceUrl().toString(), u.getMainClass());
+                                return new ServiceRef(u.getJmxServiceUrl(), u.getMainClass());
                             } catch (MalformedURLException e) {
                                 log.info(e);
                                 return null;
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
@@ -52,21 +52,23 @@ import com.redhat.rhjmc.containerjfr.platform.ServiceRef;
 
 class DefaultPlatformClient implements PlatformClient {
 
+    private final Logger log;
     private final JvmDiscoveryClient discoveryClient;
 
-    DefaultPlatformClient(JvmDiscoveryClient discoveryClient) {
+    DefaultPlatformClient(Logger log, JvmDiscoveryClient discoveryClient) {
+        this.log = log;
         this.discoveryClient = discoveryClient;
     }
 
     @Override
-    public List<ServiceRef> listDiscoverableServices(Logger log) {
+    public List<ServiceRef> listDiscoverableServices() {
         return discoveryClient.getDiscoveredJvmDescriptors().stream()
                 .map(
                         u -> {
                             try {
                                 return new ServiceRef(u.getJmxServiceUrl(), u.getMainClass());
                             } catch (MalformedURLException e) {
-                                log.info(e);
+                                log.warn(e);
                                 return null;
                             }
                         })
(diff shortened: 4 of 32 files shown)


Alexjsenn:
In the following files:
* src/test/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClientTest.java
* src/test/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClientTest.java

there is some strange behaviour regarding some of the Mockito expressions that have .getPort(), which essentially doesn't allow me to keep or remove them all at the same time. If I leave them in, I get stubbing errors pointing to some of those lines, and if I remove them all then the tests fail. I don't understand why for example in `OpenShiftPlatformClientTest.java` if I remove the line
`Mockito.when(port3.getPort()).thenReturn(9091);` 
then the test fails because the endpoint that uses port 3 does not show up in the output (but is expected).

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
```diff
@@ -49,16 +49,14 @@
...
-    private final int port;
 
-    public ServiceRef(String connectUrl, int port) {
-        this(connectUrl, connectUrl, port);
+    public ServiceRef(String connectUrl) {
+        this(connectUrl, connectUrl);
     }
 
-    public ServiceRef(String connectUrl, String alias, int port) {
+    public ServiceRef(String connectUrl, String alias) {
```
andrewazores, line 57: The point of the change is to ensure that all `PlatformClients` report JMXServiceURLs, right? So, should the `String connectUrl` here perhaps be replaced by a `JMXServiceURL`?

src/main/java/com/redhat/rhjmc/containerjfr/platform/SelfDiscoveryPlatformClient.java
```diff
@@ -47,8 +47,7 @@
 
 class SelfDiscoveryPlatformClient implements PlatformClient {
 
-    private static final ServiceRef VM_SELF_REF =
-            new ServiceRef("localhost", "This ContainerJFR", 0);
+    private static final ServiceRef VM_SELF_REF = new ServiceRef("localhost", "This ContainerJFR");
```
andrewazores, line 50: `localhost` probably isn't a valid JMXServiceURL

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/DefaultPlatformClient.java
```diff
@@ -67,7 +67,7 @@
                         u -> {
                             try {
                                 return new ServiceRef(
-                                        u.getJmxServiceUrl().toString(), u.getMainClass(), 0);
+                                        u.getJmxServiceUrl().toString(), u.getMainClass());
```
andrewazores, line 70: If the `ServiceRef` contains a service URL then the `toString` here can be removed

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
```diff
@@ -80,13 +80,7 @@
...
                     .peek(spec -> logger.trace("Service spec: " + spec.toString()))
                     .filter(s -> s.getPorts() != null)
-                    .flatMap(
-                            s ->
-                                    s.getPorts().stream()
-                                            .map(
-                                                    p ->
-                                                            new ServiceRef(
-                                                                    s.getClusterIP(), p.getPort())))
+                    .flatMap(s -> s.getPorts().stream().map(p -> new ServiceRef(s.getClusterIP())))
```
andrewazores, line 83: The cluster IP is definitely not a service URL

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
```diff
@@ -105,7 +99,7 @@ private ServiceRef resolveServiceRefHostname(ServiceRef in) {
         try {
             String hostname = resolver.resolveCanonicalHostName(in.getConnectUrl());
             logger.debug(String.format("Resolved %s to %s", in.getConnectUrl(), hostname));
-            return new ServiceRef(hostname, in.getPort());
+            return new ServiceRef(hostname);
```
andrewazores, line 102: And the hostname is also not a fully-formed service URL

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeEnvPlatformClient.java
```diff
@@ -76,7 +76,6 @@ private static ServiceRef envToServiceRef(Map.Entry<String, String> entry) {
             return null;
         }
         String alias = matcher.group(1).toLowerCase();
-        int port = Integer.parseInt(matcher.group(2));
-        return new ServiceRef(entry.getValue(), alias, port);
+        return new ServiceRef(entry.getValue(), alias);
```
andrewazores, line 79: Same here - the form of these environment variables looks like `REDIS_MASTER_PORT_6379_TCP_ADDR=10.0.0.11` ( https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables ) , so this will be a raw IP address.

src/main/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClient.java
```diff
@@ -101,12 +101,7 @@ private boolean isCompatiblePort(EndpointPort port) {
 
     private List<ServiceRef> createServiceRefs(EndpointSubset subset, EndpointPort port) {
         return subset.getAddresses().stream()
-                .map(
-                        addr ->
-                                new ServiceRef(
-                                        addr.getIp(),
-                                        addr.getTargetRef().getName(),
-                                        port.getPort()))
+                .map(addr -> new ServiceRef(addr.getIp(), addr.getTargetRef().getName()))
```
andrewazores, line 104: Also a raw IP address

Alexjsenn:
I haven't fixed the tests yet, I first wanted to run by you the other changes I had made in case some things still need modifying. The one file I am still not sure about, I can keep looking into it, is: src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
It has the `resolveCanonicalHostName` which I do not fully understand.

src/main/java/com/redhat/rhjmc/containerjfr/platform/SelfDiscoveryPlatformClient.java
```diff
@@ -45,21 +45,28 @@
...
 
     @Override
-    public List<ServiceRef> listDiscoverableServices() {
+    public List<ServiceRef> listDiscoverableServices(Logger logger) {
         List<ServiceRef> list = new ArrayList<>();
+        try {
+            VM_SELF_REF = new ServiceRef("localhost", "This ContainerJFR", 0);
+        } catch (Exception e) {
+            logger.info(e);
+            VM_SELF_REF = null;
```
andrewazores, line 66: This doesn't seem right. Why would a `null` entry to added to the list in this case?

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
```diff
@@ -41,36 +41,37 @@
...
 
-    private final String connectUrl;
+    private final JMXServiceURL JMXServiceURL;
     private final String alias;
-    private final int port;
 
-    public ServiceRef(String connectUrl, int port) {
-        this(connectUrl, connectUrl, port);
+    public ServiceRef(JMXServiceURL JMXServiceURL) {
+        this(JMXServiceURL, JMXServiceURL.toString());
```
andrewazores, line 56: It doesn't seem very useful to use the URL itself as the alias (which is intended to be a human-friendly identifier) for the URL.

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
```diff
@@ -41,36 +41,37 @@
...
-        this.connectUrl = connectUrl;
+    public ServiceRef(JMXServiceURL JMXServiceURL, String alias) {
+        this.JMXServiceURL = JMXServiceURL;
         this.alias = alias;
-        this.port = port;
     }
 
-    public String getConnectUrl() {
-        return connectUrl;
+    public ServiceRef(String host, String alias, int port) throws Exception {
```
andrewazores, line 64: `host`/`alias`/`port` feels like an unnatural ordering of these params IMO. Maybe host/port/alias? Or alias/host/port? I think the host/port should be adjacent to each other.

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeEnvPlatformClient.java
```diff
@@ -57,26 +58,33 @@
...
         if (!matcher.matches()) {
             return null;
         }
         String alias = matcher.group(1).toLowerCase();
         int port = Integer.parseInt(matcher.group(2));
-        return new ServiceRef(entry.getValue(), alias, port);
+        try {
+            return new ServiceRef(entry.getValue(), alias, port);
+        } catch (Exception e) {
+            logger.info(e);
```
andrewazores, line 86: This goes for all of these new logger statements - maybe a Warning level message would be more appropriate?

src/main/java/com/redhat/rhjmc/containerjfr/platform/PlatformClient.java
```diff
@@ -43,6 +43,8 @@
 
 import java.util.List;
 
+import com.redhat.rhjmc.containerjfr.core.log.Logger;
+
 public interface PlatformClient {
-    List<ServiceRef> listDiscoverableServices();
+    List<ServiceRef> listDiscoverableServices(Logger logger);
```
andrewazores, line 49: This seems wrong. Why does the `Logger` need to be provided as an argument to this method? I would expect it to be dependency injected into the `PlatformClient` implementations' constructors.

andrewazores:
> It has the `resolveCanonicalHostName` which I do not fully understand.

I'm happy to explain, just let me know which piece or aspect you're having trouble with.

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
Alexjsenn: Should be fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/SelfDiscoveryPlatformClient.java
Alexjsenn: should be fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
Alexjsenn: Should be fixed, haven't tested yet if the Optional gets serialized properly, I will do that next.

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
Alexjsenn: fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeEnvPlatformClient.java
Alexjsenn: fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/PlatformClient.java
Alexjsenn: I removed it!

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
```diff
@@ -41,36 +41,39 @@
...
+        this.JMXServiceURL = JMXServiceURL;
     }
 
-    public ServiceRef(String connectUrl, String alias, int port) {
-        this.connectUrl = connectUrl;
-        this.alias = alias;
-        this.port = port;
+    public ServiceRef(JMXServiceURL JMXServiceURL, String alias) {
+        this.JMXServiceURL = JMXServiceURL;
+        this.alias = Optional.of(alias);
```
andrewazores, line 63: For robustness, maybe this should be `Optional.ofNullable()`? `Optional.of(null)` throws an exception IIRC, which we could avoid by just using `ofNullable` which produces an empty Optional if passed null.

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
```diff
@@ -41,36 +41,39 @@
...
-        this(connectUrl, connectUrl, port);
+    public ServiceRef(JMXServiceURL JMXServiceURL) {
+        this.JMXServiceURL = JMXServiceURL;
     }
 
-    public ServiceRef(String connectUrl, String alias, int port) {
-        this.connectUrl = connectUrl;
-        this.alias = alias;
-        this.port = port;
+    public ServiceRef(JMXServiceURL JMXServiceURL, String alias) {
```
andrewazores, line 61: Just a style thing but the constructor param should be named like `jmxServiceUrl` or `jmxServiceURL`, not `JMXServiceURL`.

Also if the other constructor has a `throws` declaration, they probably all should. In fact, this constructor should be the "primary" one, and the others should defer to it, ex:

```
public ServiceRef(JMXServiceURL jmxServiceUrl) {
  this(jmxServiceUrl, null);
}
```

That way as much logic is shared as possible.

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
```diff
@@ -100,15 +107,16 @@
...
-            return new ServiceRef(hostname, in.getPort());
+            String hostname = resolver.resolveCanonicalHostName(in.getJMXServiceUrl());
+            logger.debug(String.format("Resolved %s to %s", in.getJMXServiceUrl(), hostname));
+            return new ServiceRef(hostname);
         } catch (Exception e) {
             logger.debug(e);
             return null;
         }
     }
+    */
```
andrewazores, line 121: Is this intended to be removed? The point of this method was to map IP addresses to more readable hostnames for the entry aliases, since the IP address itself isn't likely to be meaningful to a human reader without them having to go look at their cluster configuration and cross-reference the addresses. A hostname is more likely to give some telling information about what the user is looking at.

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
Alexjsenn: Makes sense!

src/main/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClient.java
```diff
@@ -99,14 +100,21 @@ private boolean isCompatiblePort(EndpointPort port) {
...
+                            try {
+                                return new ServiceRef(
                                         addr.getIp(),
-                                        addr.getTargetRef().getName(),
-                                        port.getPort()))
+                                        port.getPort(),
+                                        addr.getTargetRef().getName());
+                            } catch (Exception e) {
+                                logger.warn(e);
+                                return null;
```
andrewazores, line 115: Since there is a possible `return null` in this chain, there should also be a corresponding filter to remove null entries, like the `KubeEnv` client has for example.

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
Alexjsenn: Yeah no, I couldn't figure it out so I had temprarily removed it. Is the resolveCanonicalHostName only important for the alias? Because I could just incorporate it when Im creating the serviceRef above, and insert the resolved hostname as alias?

andrewazores: Yes, that's all it does is use a reverse DNS lookup to turn an InetAddress into a hostname string (under the hood it's just https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#getCanonicalHostName()), ex. `127.0.0.1 -> localhost.local` or `123.45.6.7 -> someProductionService.cluster.local`.

Using the actual IP to construct the JMXServiceURL makes sense, since if that URL used the hostname then it would need to do another DNS lookup back in the other direction to resolve the IP anyway (whoops - that's how it used to work, re-reading this). But that reverse-lookup-resolved hostname should be kept around for use as the alias as I said above.

What part did you have trouble figuring out? Was it the stream mapping that was also commented out above?

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
Alexjsenn: Fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
Alexjsenn: fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClient.java
Alexjsenn: Fixed.

src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
Alexjsenn: I tried reducing it given that there was no point in creating the serviceRef and then making a new one again, im not sure if I should add back in the log message?

andrewazores: I think it's fine to remove it.

andrewazores:
Do you have any more changes to make, other than testing and the Optional serialization? I think the changes look good now from a read-over, so I'll start testing them out (soon) if there is nothing else planned.

Alexjsenn:
I have to fix the tests, which I am looking at right now.

Alexjsenn:
The serialization of the Optional apparently is either Optional.empty or Optional[alias]. Should I try returning the string instead? or can the way it is serialized be changed?

andrewazores:
You mean the JSON comes out like `{"alias":"Optional.empty"}`?

It is possible to add custom (de)serializations to Gson using TypeAdapters. Here's an example, but there are plenty more out there:

https://stackoverflow.com/questions/12161366/how-to-serialize-optionalt-classes-with-gson

This customization should be applied to the "global" Gson instance provided by the MainModule.

andrewazores:
It's worth trying to work around it first though. What if the data structure looks like this? Does this get serialized to `{"alias":null}` and `{"alias":"foo"}`?

```
class Something {
  private final String alias; // nullable
  Something(String alias) {
    this.alias = alias;
  }
  Something() {
    this(null);
  }

  Optional<String> getAlias() {
    return Optional.ofNullable(alias);
  }
}
```

```
gson.toJson(new Something());
gson.toJson(new Something("foo"));
```

Alexjsenn:
When running scan-targets with websocat, it returns the Optional alias as "alias":{"value":"This ContainerJFR"}. I will look into the custom deserialization.

andrewazores:
Please, look into restructuring the data structure as I said in my previous comment first. It's a better fix and probably less work than the custom serialization with the TypeAdapter, too.

Alexjsenn:
Ok, so with that change the ws response now formats alias as: "alias":"This ContainerJFR"

andrewazores:
Sounds correct. That's the same format as before this PR, right?

Alexjsenn:
Yes!

Alexjsenn:
So, the jmx URL is serialized like this:
"JMXServiceURL":{"protocol":"jmxmp","host":"localhost","port":0,"urlPath":""}
And Im not sure how I can modify the webclient to properly reconstruct the url.

andrewazores:
I don't think we want to modify the web-client for this change, or more generally for the URLs to be serialized like that. Although they are URLs, I think we should serialize them simply as strings.

src/main/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClient.java
```diff
@@ -99,14 +101,22 @@ private boolean isCompatiblePort(EndpointPort port) {
         return "jfr-jmx".equals(port.getName()) || 9091 == port.getPort();
     }
 
-    private List<ServiceRef> createServiceRefs(EndpointSubset subset, EndpointPort port) {
+    private List<ServiceRef> createServiceRefs(
```
andrewazores, line 104: This method isn't static, so I don't think it needs the `logger` instance passed as a parameter, does it?

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
```diff
@@ -85,7 +87,11 @@ static Logger provideLogger() {
     @Provides
     @Singleton
     public static Gson provideGson() {
```
andrewazores, line 89: Could you inject a `Logger` instance here and pass it to `new GsonJmxServiceUrlAdapter(logger)`?

src/main/java/com/redhat/rhjmc/containerjfr/util/GsonJmxServiceUrlAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+public class GsonJmxServiceUrlAdapter extends TypeAdapter<JMXServiceURL> {
+
+    @Override
+    public JMXServiceURL read(JsonReader reader) throws IOException {
+        String url = reader.nextString();
+        JMXServiceURL jmxUrl;
+        try {
+            jmxUrl = new JMXServiceURL(url);
+        } catch (Exception e) {
+            jmxUrl = null;
```
andrewazores, line 61: With the injected Logger instance, this exception can be logged (at warning level).

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/ScanTargetsCommandTest.java
```diff
@@ -139,9 +139,19 @@ void shouldExecuteAndReturnExceptionOutputIfPlatformClientThrows() throws Except
...
-        ServiceRef mockB = new ServiceRef("bHost", "Host B", 2);
-        ServiceRef mockC = new ServiceRef("cHost", "Host C", 3);
+        ServiceRef mockA;
+        ServiceRef mockB;
+        ServiceRef mockC;
+        try {
+            mockA = new ServiceRef("aHost", 0, "Host A");
+            mockB = new ServiceRef("bHost", 0, "Host B");
+            mockC = new ServiceRef("cHost", 0, "Host C");
+        } catch (Exception e) {
```
andrewazores, line 149: This exception should never actually be thrown though, right? I think we can skip trying to handle it here and just add a throws-declaration to this test utility method.

src/test/java/com/redhat/rhjmc/containerjfr/net/web/handlers/TargetsGetHandlerTest.java
```diff
@@ -90,7 +90,12 @@ void shouldHandleCorrectPath() {
 
     @Test
     void shouldReturnListOfTargets() {
-        ServiceRef target = new ServiceRef("foo", 1);
+        ServiceRef target;
+        try {
+            target = new ServiceRef("foo", 1, "foo");
+        } catch (Exception e) {
+            target = null;
```
andrewazores, line 97: Same here, just add a throws-declaration to the method and let it fail right away if somehow the test ServiceRef construction fails.

src/test/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClientTest.java
```diff
@@ -153,12 +153,21 @@ void discoversAndResolvesServices() throws ApiException, UnknownHostException {
...
+            ServiceRef serv2;
+            ServiceRef serv3;
+            try {
+                serv1 = new ServiceRef("127.0.0.1", 123, "ServiceA.local");
+                serv2 = new ServiceRef("127.0.0.1", 456, "ServiceA.local");
+                serv3 = new ServiceRef("10.0.0.1", 7899, "b-service.example.com");
+            } catch (Exception e) {
+                serv1 = null;
+                serv2 = null;
+                serv3 = null;
```
andrewazores, line 166: Same as above

src/test/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClientTest.java
```diff
@@ -211,11 +220,16 @@ void ignoresUnresolveableServices() throws ApiException, UnknownHostException {
...
-                            new ServiceRef("ServiceA.local", 123),
-                            new ServiceRef("ServiceA.local", 456)));
+            ServiceRef serv1;
+            ServiceRef serv2;
+            try {
+                serv1 = new ServiceRef("127.0.0.1", 123, "ServiceA.local");
+                serv2 = new ServiceRef("127.0.0.1", 456, "ServiceA.local");
+            } catch (Exception e) {
+                serv1 = null;
+                serv2 = null;
```
andrewazores, line 230: Here too

src/test/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeEnvPlatformClientTest.java
```diff
@@ -97,11 +99,18 @@ void shouldDiscoverServicesByEnv() {
...
-                            new ServiceRef("1.2.3.4", "bar", 9999)));
+
+            ServiceRef serv1;
+            ServiceRef serv2;
+            try {
+                serv1 = new ServiceRef("127.0.0.1", 1234, "foo");
+                serv2 = new ServiceRef("1.2.3.4", 9999, "bar");
+            } catch (Exception e) {
+                serv1 = null;
+                serv2 = null;
```
andrewazores, line 110: And here

src/test/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClientTest.java
```diff
@@ -173,23 +173,26 @@ void shouldReturnListOfMatchingEndpointRefs() throws Exception {
...
+            serv2 =
+                    new ServiceRef(
+                            address3.getIp(), port2.getPort(), address3.getTargetRef().getName());
+            serv3 =
+                    new ServiceRef(
+                            address4.getIp(), port3.getPort(), address4.getTargetRef().getName());
+        } catch (Exception e) {
+            serv1 = null;
+            serv2 = null;
+            serv3 = null;
```
andrewazores, line 192: And finally, here :-)

src/main/java/com/redhat/rhjmc/containerjfr/platform/openshift/OpenShiftPlatformClient.java
Alexjsenn: No, you're right!

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
Alexjsenn: could I just inject the logger into the `GsonJmxServiceUrlAdapter`? I had trouble figuring out how to inject into the static method provideGson()

andrewazores: All you do is add the Logger to the parameter list of this method. Since it's part of a Dagger module and has the `@Provides` annotation, Dagger will see that and inject the logger instance into this method.

src/main/java/com/redhat/rhjmc/containerjfr/util/GsonJmxServiceUrlAdapter.java
andrewazores: I don't think this was resolved. The `@Inject` annotation on the Logger field you added doesn't seem right, or it's at least not usual DI practice. Nothing in the Dagger dependency graph knows how to construct and inject an instance of this class, because no module includes it and there is no `@Provides` or `@Binds` method that directs Dagger on how to make one of these adapters available.

src/main/java/com/redhat/rhjmc/containerjfr/MainModule.java
andrewazores: Just scroll down a few lines within this file for an example.

```
    @Provides
    static Path provideSavedRecordingsPath(Logger logger, Environment env) {
      ...
    }
```

This is a cleaned-up version of the next method down, which has two injected dependencies.

andrewazores: If you want to inject the Logger directly into the TypeAdapter then that's also an option, but then rather than directly `new`-ing it here, you will have to add an `@Provides` or `@Binds` method to some module to allow the adapter to have dependencies injected, and to allow the adapter itself to be injected into other places.

Alexjsenn:
[errorLog.txt](https://github.com/rh-jmc-team/container-jfr/files/4990585/errorLog.txt)

src/main/java/com/redhat/rhjmc/containerjfr/util/GsonJmxServiceUrlAdapter.java
Alexjsenn: ok, this should work now.

andrewazores:
Looks better. Please rebase onto current main and I'll do a quick last read-over and manual test, but I think it's good now.

andrewazores:
FWIW that's also what our `TargetConnectionManager` does:

```
    private JFRConnection attemptConnectAsHostPortPair(ConnectionDescriptor connectionDescriptor)
            throws Exception {
    ...
        return connect(
                new JMXServiceURL(
                        "rmi", "", 0, String.format("/jndi/rmi://%s:%s/jmxrmi", host, port)),
                connectionDescriptor.getCredentials());
  }
```

When connecting using a Service URL:

```
    private JFRConnection attemptConnectAsJMXServiceURL(ConnectionDescriptor connectionDescriptor)
            throws Exception {
        return connect(
                new JMXServiceURL(connectionDescriptor.getTargetId()),
                connectionDescriptor.getCredentials());
    }
```

It just passes the `targetId` through to the URL constructor. So I think correcting that `ServiceRef` constructor to match the TCM host:port/JMC ConnectionToolkit URL parameters is probably the correct fix.

Alexjsenn:
I pushed the changes to use rmi as the protocol, but right now when I deploy it in operator it gives the following strange error when opening the web client: 
Danger alert:WebSocket error
{"stack":"Error: Failed to construct 'WebSocket': The subprotocol 'base64url.bearer.authorization.containerjfr.N2pKY0FfU0k3LTFUZ3VLb1Q5UFFab21zQ3hTZmRoRzE3aTl2VDlNOGJyQQ==' is invalid.\n at WebSocketSubject._connectSocket (webpack-internal:///194:3985:17)\n at WebSocketSubject._subscribe (webpack-internal:///194:4082:18)\n at WebSocketSubject.Observable._trySubscribe (webpack-internal:///194:1425:25)\n at WebSocketSubject.Subject._trySubscribe (webpack-internal:///194:1638:51)\n at WebSocketSubject.Observable.subscribe (webpack-internal:///194:1411:22)\n at SafeSubscriber.eval [as _next] (webpack-internal:///194:4214:22)\n at SafeSubscriber.__tryOrUnsub (webpack-internal:///194:1263:16)\n at SafeSubscriber.next (webpack-internal:///194:1201:22)\n at Subscriber._next (webpack-internal:///194:1147:26)\n at Subscriber.next (webpack-internal:///194:1124:18)"}

andrewazores:
Very strange. I'm confused why anything to do with Webpack is showing up in that error, but I'll build my own set of images and try to reproduce this. Where do you see that error, as a graphical notification within the client? Maybe there are some other clues in the browser console or the browser network monitor?

Also, just to be sure, `echo -n N2pKY0FfU0k3LTFUZ3VLb1Q5UFFab21zQ3hTZmRoRzE3aTl2VDlNOGJyQQ== | base64 -d` gives me `7jJcA_SI7-1TguKoT9PQZomsCxSfdhG17i9vT9M8brA` - is that your CRC account token (`oc whoami -t`)?

andrewazores:
Hmm. I just built a ContainerJFR + -Operator image pair and deployed it to my CRC and I my web-client is able to connect fine.

image

The images are `quay.io/andrewazores/container-jfr:0.20.2` and `quay.io/andrewazores/container-jfr-operator:0.7.4`. Could you retry an `IMAGE_TAG=quay.io/andrewazores/container-jfr-operator:0.7.4 make undeploy deploy sample_app2` and see if it works for you with my image?

Alexjsenn:
Yes, it works!!!

andrewazores:
Okay, so we need to figure out what went wrong with your build so that you can test your own changes properly and finalize this PR. How did you build your ContainerJFR and Operator images?

andrewazores:
Also, my ContainerJFR can connect to the `sample_app2 jmx-listener`, and it can connect to the `sample_app vertx-fib-demo` if I specify `0.1.0` as the image version for that sample (need to check why `0.2.0` is broken - can't remember what I did to it), but it cannot connect to itself. I'm not sure if that's a new regression from this PR or already exists in current main, so I'll do some troubleshooting.

Alexjsenn:
I run sh build-images.sh
podman push quay.io/alexjsenn/container-jfr:0.19.93
then I run make image in container-jfr-operator, push the same way then 
IMAGE_TAG=quay.io/alexjsenn/container-jfr-operator:0.5.1 make undeploy deploy sample_app2

It actually worked this time, I think the problem was I hadn't update the web-client submodule

andrewazores:
Okay cool, those steps sound like they should generally work.

Alexjsenn:
Does the jmx autodiscovery feature work the same way when it is in a container? I see what you're saying though, it can see itself but fails when it tries to connect.

andrewazores:
JDP discovery is used as a fallback mechanism when no other more specific platform can be detected (`PlatformModule.java` and related classes), whether or not it's running in a container. When running in CRC it should automatically detect the OpenShift environment and make OpenShift API calls for discovery instead, never falling through to the case where it checks for JDP results.

andrewazores:
Oh, for the ContainerJFR inability to self-connect in this PR - I just remembered you have already rebased this so it's on top of the initial RJMX auth changes I made. So, ContainerJFR has secured its own incoming JMX connection port with autogenerated credentials, and it doesn't know how to identify itself to automatically allow self-connection, nor is there a graphical way for the user to provide the credentials when attempting to connect. So I think we can ignore that "regression" - that is actually an expected behaviour at this point. False alarm.

Alexjsenn:
Ok, that makes sense! Is the reason it works when using mvn because we specify authenticate=false? And is there anything else that needs done for this PR?

andrewazores:
Yea, if you're running using `mvn exec:java` then you probably do have `authenticate=false` set, and so it's able to connect to itself without credentials. `run.sh` also sets `CONTAINER_JFR_RJMX_AUTH=false` by default, which also causes the `authenticate=false` JVM flag to be set later on, so the same effect. But in an Operator deployment `CONTAINER_JFR_RJMX_AUTH` will default to `true`, so auth is enabled using an autogenerated password, and then the self-connection fails.

andrewazores:
As for this PR, I'll give it another look over and do a bit more manual testing, but I think it's good now. Finally :-)

src/main/java/com/redhat/rhjmc/containerjfr/platform/ServiceRef.java
```diff
@@ -41,36 +41,45 @@
...
-    public String getConnectUrl() {
-        return connectUrl;
+    public ServiceRef(JMXServiceURL jmxServiceUrl) throws MalformedURLException {
+        this(jmxServiceUrl, null);
+    }
+
+    public ServiceRef(String host, int port, String alias) throws MalformedURLException {
+        this(
+                new JMXServiceURL(
+                        "rmi", "", 0, String.format("/jndi/rmi://%s:%s/jmxrmi", host, port)),
```
andrewazores, line 73: Ah, tiny nitpick here - your format specifier has two `%s` tokens, but you pass in a string (`%s`) and an int (`%d`). This will still work because casting the port to a string still comes out the same in the formatted string, but I think it's still worth correcting.

Alexjsenn: Ah true, Ill fix it!

Issue #158, opened 2020-05-08
Platform discovery should not require polling


andrewazores:
Currently, platform discovery methods only update the list of targets on-demand via polling, which generally requires a client to send a `scan-targets` command. This is sensible in some runtime scenarios, such as `KubernetesEnv` and `OpenShiftApi`, but misses an opportunity for passing along real-time target discovery as in the `DefaultPlatformClient` (which, behind the scenes, uses JDP).

There should be a command channel message that ContainerJFR can broadcast to connected clients unprompted which reports the discovery (or loss) of target(s) from the known list. Initially this might only be sent when ContainerJFR is using a "passive" discovery mechanism like JDP, but it would also be useful following on from the ideas of #153 , where even when backed by a polled mechanism like the OpenShift API, ContainerJFR should expose a push-based API for target discovery that clients can consume.

PR #222, opened 2020-07-29
HTTP REST API feature parity

andrewazores:
This series of commits brings the HTTP REST API up to feature parity with the WebSocket Command Channel - that is, anything (recording-related, anyway - there is no equivalent for `help`, `hostname`, etc.) that can be accomplished using a Command can now also be accomplished using an HTTP request. The immediate benefit here is in the easier extensibility of HTTP requests with various headers, which will be used in a follow-up PR (#225) to enable JMX authentication credentials. All commands that operate upon a specific target have been marked deprecated as they are not planned to gain JMX authentication support. When any such command is run the registry will log a warning to indicate this.

TODO as followups:
- create a document outlining the HTTP REST API, similar to the current COMMANDS.md
- update -web and -operator to send these requests whenever applicable, rather than using command channel

Requires:
https://github.com/rh-jmc-team/container-jfr-core/pull/45

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
@@ -43,6 +43,7 @@ package com.redhat.rhjmc.containerjfr.net.web.handlers;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
@@ -150,11 +151,15 @@ class RecordingUploadPostHandler extends AbstractAuthenticatedRequestHandler {
     }
 
     Optional<Path> getRecordingPath(String recordingName) throws Exception {
-        Path archivedRecording = savedRecordingsPath.resolve(recordingName);
-        if (fs.isRegularFile(archivedRecording) && fs.isReadable(archivedRecording)) {
-            return Optional.of(archivedRecording);
+        try {
+            Path archivedRecording = savedRecordingsPath.resolve(recordingName);
+            if (fs.isRegularFile(archivedRecording) && fs.isReadable(archivedRecording)) {
+                return Optional.of(archivedRecording);
+            }
+            return Optional.empty();
+        } catch (InvalidPathException e) {
+            throw new HttpStatusException(400, e.getMessage(), e);
         }
-        return Optional.empty();
     }
 
     private static class ResponseMessage {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
@@ -41,23 +41,19 @@
  */
 package com.redhat.rhjmc.containerjfr.net.web.handlers;
 
-import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.StandardCopyOption;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
 import javax.inject.Inject;
+import javax.inject.Named;
 
-import com.redhat.rhjmc.containerjfr.core.net.JFRConnection;
+import com.redhat.rhjmc.containerjfr.MainModule;
 import com.redhat.rhjmc.containerjfr.core.sys.Environment;
 import com.redhat.rhjmc.containerjfr.core.sys.FileSystem;
 import com.redhat.rhjmc.containerjfr.net.AuthManager;
-import com.redhat.rhjmc.containerjfr.net.TargetConnectionManager;
-import com.redhat.rhjmc.containerjfr.net.internal.reports.ReportService.RecordingNotFoundException;
 import com.redhat.rhjmc.containerjfr.net.web.HttpMimeType;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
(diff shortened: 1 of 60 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/RecordingOptionsBuilderFactory.java
```diff
@@ -45,7 +45,9 @@
 import org.openjdk.jmc.flightrecorder.configuration.recording.RecordingOptionsBuilder;
 import org.openjdk.jmc.rjmx.services.jfr.IFlightRecorderService;
 
-interface RecordingOptionsBuilderFactory {
+// FIXME this should be in a more general package, not commands/internal. This is also used in
```
jiekang, line 48: Does this comment also apply to EventOptionsBuilder?

andrewazores: Yea, it does. Since essentially all of the commands are now deprecated, I figured it would make sense until they can be fully removed and then the RecordingOptionsBuilderFactory and EventOptionsBuilder can be moved somewhere close to the HTTP Handlers.

jiekang: Ah gotcha. Makes sense. Thanks

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
```diff
@@ -0,0 +1,174 @@
...
+            ResponseMessage response = doPost(recordingName, uploadUrl);
+            ctx.response().setStatusCode(response.statusCode);
+            ctx.response().setStatusMessage(response.statusMessage);
+            ctx.response().end(response.body);
+        } catch (MalformedURLException e) {
+            throw new HttpStatusException(501, e);
+        }
+    }
+
+    // FindBugs thinks the recordingPath or its properties is null somehow
```
jiekang, line 119: Would the check for this be running `mvn spotbugs:check` ?

jiekang: Er, with the suppression removed :P

andrewazores: Yea that should do it

jiekang: Dang, guess our projects really do differ in some critical way. I've removed the suppression and see green for my runs of `spotbugs:check`

jiekang: s/project/systems I suppose

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                                    new ResponseMessage(
+                                            response.statusCode(),
+                                            response.statusMessage(),
+                                            response.bodyAsString()));
+                        });
+        return future.get();
+    }
+
+    Optional<Path> getRecordingPath(String recordingName) throws Exception {
+        Path archivedRecording = savedRecordingsPath.resolve(recordingName);
```
jiekang, line 156: `resolve` can throw an `InvalidPathException`. If this happens, it will throw through `handleAuthenticated` as it only catches `MalformedURLException`. Would it make sense to handle it within instead to respond that the input was invalid? Or does it do that somewhere higher up?

I see test class for `TargetRecordingUploadPostHandler` but not for this one. Is that intended?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
andrewazores: I don't get the Spotbugs warning anymore either - maybe I had come back around at some point and refactored this in a way that avoids the previous problem. I'll remove the annotation.

jiekang: I can see why it would have this warning since there are calls that would NPE if their input was null and the values submitted aren't checked before being used as input. However it looks like the values are currently never not null and I guess it would have trouble analyzing that? Anyways; weird that I don't see the warning on my runs...

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
andrewazores: `handleAuthenticated` is called in `AbstractAuthenticatedRequestHandler.handle()`, which will catch any exception thrown by `handleAuthenticated` and handle it appropriately. For an `InvalidPathException` or any other generally unexpected exception, it will be wrapped in an `HttpStatusException` with status code 500, which will later on be caught by the `WebServer` and turned into a failure response on the wire.

andrewazores: re: missing test - nope, not intended, looks like I forgot to write one. Good catch :-) I'll add one now.

jiekang: Okay. I agree, I think it can be caught in this class and handled as a `4xx`.

jiekang:
Id be interested to see if all of the spotbug ignores that exist in the files touched here could be removed or not. The one's I've tried to remove so far result in green, but my setup seems to differ critically so I'm not sure if that's a false positive or negative or anything heh. Bleh...

andrewazores:
I think there are almost certainly more enhancements and refinements to be made in the same class as that `500 -> 4xx` response we just talked about above, but finding all of those up front is not really realistic. They'll just be dealt with as they come up.

vic-ma:
Is the WebSocket command channel planned to be removed in the future, or do we at least expect users to not use it?

andrewazores:
> Is the WebSocket command channel planned to be removed in the future, or do we at least expect users to not use it?

The WebSocket connection will not be fully removed, but most or all of the current commands will be. It will still be useful for pushing state changes to clients who might be interested to update asynchronously (ex. -web or any other graphical client) so that they can display up-to-date data without having to poll for it. For sure, any commands which operate upon targets will be removed, since there is no current way and no plan to support JMX credential passthrough over the Command Channel protocol.

As one example, clients should use `GET /api/v1/targets` to discover what targets are known to the ContainerJFR instance. The `scan-targets` command could still exist and prompt ContainerJFR to rescan for targets. And as in #158, a new type of message could also be pushed from ContainerJFR to clients unprompted on the WebSocket connection to inform clients when new clients are discovered asynchronously (ex. JDP).

PR #224, opened 2020-07-29
Add Checks for `upload-recording` ResponseMessage

### LINKED ISSUE #218: `upload-recording` Throws NPE When Datasource URL is Invalid
When the URL points to an existent but invalid server, the command will throw a NPE (or it will hang; see #217).

Here, the URL points a jfr-datasource, but it is missing the needed `/load` ending:
```
{command:upload-recording,args:[localhost,foo,"http://localhost:8080"]}
{"id":null,"commandName":"upload-recording","status":-2,"payload":"NullPointerException: "}
```

Here, the URL points to the running CJFR instance:
```
{command:upload-recording,args:[localhost,foo,"http://localhost:8181"]}
{"id":null,"commandName":"upload-recording","status":-2,"payload":"NullPointerException: "}
```

In 139649d, there isn't a NPE:
```
{command:upload-recording,args:[localhost,foo,"http://localhost:8080"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":{"protoVersion":{"protocol":"HTTP","major":1,"minor":1},"statusCode":200,"reasonPhrase":"OK"},"body":""}}
```

```
{command:upload-recording,args:[localhost,foo,"http://localhost:8181"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":{"protoVersion":{"protocol":"HTTP","major":1,"minor":1},"statusCode":405,"reasonPhrase":"Method Not Allowed"},"body":""}}
```

Though really this behaviour isn't that much better, since here the responses make it seem like the upload worked, even though it didn't; but it does show that the NPE is a problem that was introduced after that commit.

vic-ma:
Fixes #218.

This PR adds a check to make sure that the HTTP response's status code is `200` and that its status message and body are not `null`. If the `200` requirement is too strict, we could also change it to `200 <= x < 300`.

In addition to fixing the NPE issue, because of the status code check, `upload-recording` now also correctly identifies many invalid uploads to existing but non-jfr-datasource servers, such as `http://google.com`, which previously would have just returned a standard `MapResponse`.

With the datasource URL pointing to the running CJFR instance, the response from an `upload-recording` call now looks like this:

`{"id":null,"commandName":"upload-recording","status":-2,"payload":"Invalid response from server; datasource URL may be incorrect, or server may not be functioning properly: status=\"405 Method Not Allowed\"; body=\"null\""}`

The test code is just copied from the `shouldDoPost()` tests. In the serialized `shoudDoPost()`, I moved the `urlCaptor` part to before the response checking part so it's a bit more chronological. The diff looks a little weird because of it.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/UploadRecordingCommand.java
@@ -41,6 +41,8 @@
  */
 package com.redhat.rhjmc.containerjfr.commands.internal;
 
+import static com.redhat.rhjmc.containerjfr.util.HttpStatusCodeIdentifier.isSuccessCode;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
@@ -114,7 +116,9 @@ class UploadRecordingCommand extends AbstractConnectedCommand implements Seriali
         String datasourceUrl = env.getEnv(GRAFANA_DATASOURCE_ENV).concat("/load");
         ResponseMessage response = doPost(targetId, recordingName, datasourceUrl);
 
-        if (response.statusCode != 200 || response.statusMessage == null || response.body == null) {
+        if (!isSuccessCode(response.statusCode)
+                || response.statusMessage == null
+                || response.body == null) {
             cw.println(
                     String.format(
                             "Invalid response from server; datasource URL may be incorrect, or server may not be functioning properly: status=\"%d %s\"; body=\"%s\"",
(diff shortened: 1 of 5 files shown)


andrewazores:
I think the `200 <= x < 300` status check does make sense here. There may be some utility method around that we can use, otherwise I think it makes sense to create one since this is likely a type of check we'll be performing often. Otherwise this change looks good.

vic-ma:
I wasn't able to find any such method in Java or Vertx or anything else so I guess we should create our own. Should I just make a class in `util/`? And I think the methods can probably just be static since we shouldn't need to stub them in tests?

andrewazores:
Yea, static methods should be fine. `util` might be the right place, or maybe `net/web`. Or even a new `net/web/util`.

PR #225, opened 2020-07-30
RJMX authorization via HTTP header

andrewazores:
This adds a bit of logic to the `AbstractAuthenticatedRequestHandler` to fill in the implementation of retrieving target RJMX credentials from HTTP headers. This was previously added as a stub which always used null credentials in an earlier PR. Now, if credentials are provided using the `Proxy-Authorization` header (maybe this isn't the most appropriate header to use), they are checked for validity and passed along down the line when establishing the connection to the target JVM.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/HttpServer.java
@@ -78,7 +78,7 @@ public class HttpServer {
                         sslConf.applyToHttpServerOptions(
                                 new HttpServerOptions()
                                         .setPort(netConf.getInternalWebServerPort())
-                                        .setWebsocketSubProtocols("*")
+                                        .addWebSocketSubProtocol("*")
                                         .setCompressionSupported(true)
                                         .setLogActivity(true)));
 
@@ -95,7 +95,7 @@ public class HttpServer {
         CompletableFuture<Void> future = new CompletableFuture<>();
         this.server
                 .requestHandler(requestHandlerDelegate)
-                .websocketHandler(websocketHandlerDelegate)
+                .webSocketHandler(websocketHandlerDelegate)
                 .listen(
                         res -> {
                             if (res.failed()) {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/AbstractAuthenticatedRequestHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/AbstractAuthenticatedRequestHandler.java
@@ -41,7 +41,11 @@
  */
 package com.redhat.rhjmc.containerjfr.net.web.handlers;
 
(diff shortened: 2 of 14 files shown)


andrewazores:
To test, do `mvn clean package` or equivalent, then `CONTAINER_JFR_RJMX_AUTH=true sh run.sh`. Make note of the generated RJMX password which can be seen near the beginning of the startup logs. Set that to an env var like `RJMX_PASS=whatever`. Then, try:

```
curl -vvv -X POST --data 'recordingName=foo&events=template=Continuous&duration=10' localhost:8181/api/v1/targets/localhost/recordings
```

You should see a response like:

```
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#0)
> POST /api/v1/targets/localhost/recordings HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> Content-Length: 56
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 56 out of 56 bytes
< HTTP/1.1 407 Proxy Authentication Required
< proxy-authenticate: Basic
< content-type: text/plain
< content-length: 29
< 
* Connection #0 to host localhost left intact
Proxy Authentication Required
```

Then, include the `Proxy-Authorization` header in the request:

```
curl -vvv -X POST --data 'recordingName=foo&events=template=Continuous&duration=10' -H "Proxy-Authorization: Basic $(echo -n containerjfr:$RJMX_PASS | base64)" localhost:8181/api/v1/targets/localhost/recordings
```

This should succeed with a response like:

```
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#0)
> POST /api/v1/targets/localhost/recordings HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> Proxy-Authorization: Basic Y29udGFpbmVyamZyOl9lR3pQVVl6TklBMDViektOLWptR2pITzFZN0ppUGNp
> Content-Length: 56
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 56 out of 56 bytes
< HTTP/1.1 201 Created
< location: /foo
< content-length: 397
< 
* Connection #0 to host localhost left intact
{"downloadUrl":"http://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/recordings/foo","reportUrl":"http://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/reports/foo","id":1,"name":"foo","state":"RUNNING","startTime":1596752049768,"duration":10000,"continuous":false,"toDisk":true,"maxSize":0,"maxAge":0}
```

PR #228, opened 2020-08-04
Increase test timeouts to reduce likelihood of false negatives

andrewazores:


andrewazores:
@jiekang does this help with the reliability of your builds at all? I don't seem to hit these failures as often as everyone else, so it's hard for me to know for sure.

andrewazores:
Of course, a real fix for the unit tests by not relying on an actual wall clocker timer and instead using a faked executor or similar would be an even better solution, but that's a good bit more involved.

Issue #219, opened 2020-07-27
`upload-recording` Sometimes Causes CJFR to Hang


vic-ma:
If you run `upload-recording` on the same recording multiple times, it will eventually cause CJFR to hang. I'm still not totally sure how the speed at which you run the commands affects the bug. Once I ran the commands 1-2 seconds apart and was still able to cause the hang. For replicating the bug, spamming the commands is fastest.

I discovered this bug in the web client, but since I was also able to replicate the bug with a slow speed in the backend, I think this bug is unrelated to the web client.

```
{command:dump,args:[localhost,foo,3,"template=Continuous"]}
{"id":null,"commandName":"dump","status":0,"payload":null}
{command:upload-recording,args:[localhost,foo,"http://localhost:8080/load"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":"200 OK","body":"Uploaded: 15334564489196067905.tmp\nSet: 15334564489196067905.tmp\n"}}
{command:upload-recording,args:[localhost,foo,"http://localhost:8080/load"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":"200 OK","body":"Uploaded: 2781532471906468424.tmp\nSet: 2781532471906468424.tmp\n"}}
{command:upload-recording,args:[localhost,foo,"http://localhost:8080/load"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":"200 OK","body":"Uploaded: 8537316693928713982.tmp\nSet: 8537316693928713982.tmp\n"}}
{command:upload-recording,args:[localhost,foo,"http://localhost:8080/load"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":"200 OK","body":"Uploaded: 10557930756227151177.tmp\nSet: 10557930756227151177.tmp\n"}}
{command:upload-recording,args:[localhost,foo,"http://localhost:8080/load"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":"200 OK","body":"Uploaded: 1204474703124060677.tmp\nSet: 1204474703124060677.tmp\n"}}
{command:upload-recording,args:[localhost,foo,"http://localhost:8080/load"]}
{"id":null,"commandName":"upload-recording","status":0,"payload":{"status":"200 OK","body":"Uploaded: 18216638194210754985.tmp\nSet: 18216638194210754985.tmp\n"}}
[...]

PR #229, opened 2020-08-05
Update Vert.x Version to Fix `upload-recording` Hangs

### LINKED ISSUE #217: `upload-recording` Hangs When Datasource URL is Invalid
If the URL points to a non-existent server, it will hang:
```
{command:upload-recording,args:[localhost,foo,"http://abc"]}
```

Also if the URL points to an existent but invalid server, it will hang (or it will throw a NPE; see #218):
```
{command:upload-recording,args:[localhost,foo,"http://localhost"]}
```

The command is working in 139649d, which is the first commit after the Connectionless API where `upload-recording` was working:

```
{command:upload-recording,args:[localhost,foo,"http://abc"]}    
{"id":null,"commandName":"upload-recording","status":-2,"payload":"UnknownHostException: abc: Name or service not known"}

{command:upload-recording,args:[localhost,foo,"http://localhost"]}
{"id":null,"commandName":"upload-recording","status":-2,"payload":"HttpHostConnectException: Connect to localhost:80 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1] failed: Connection refused (Connection refused)"}
```

vic-ma:
Fixes #217, #219.

### DIFF
--- a/src/test/java/itest/GrafanaDatasourceIT.java
+++ b/src/test/java/itest/GrafanaDatasourceIT.java
@@ -59,7 +59,8 @@ public class GrafanaDatasourceIT extends ITestBase {
         req = webClient.get("/api/v1/grafana_datasource_url");
     }
 
-    // Disabled for now due to conflict with UploadRecordingIT; see #229
+    // Disabled for now due to conflict with UploadRecordingIT;
+    // See https://github.com/rh-jmc-team/container-jfr/pull/229
     // @Test
     public void shouldFail() throws Exception {
         CompletableFuture<Integer> future = new CompletableFuture<>();
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
   <org.apache.httpcomponents.version>4.5.11</org.apache.httpcomponents.version>
   <io.fabric8.openshift.client.version>4.6.4</io.fabric8.openshift.client.version>
   <io.kubernetes.client.java.version>5.0.0</io.kubernetes.client.java.version>
-  <io.vertx.web.version>3.8.5</io.vertx.web.version>
+  <io.vertx.web.version>3.9.2</io.vertx.web.version>
   <org.slf4j.version>1.7.30</org.slf4j.version>
   <com.google.code.gson.version>2.8.6</com.google.code.gson.version>
   <com.github.ben-manes.caffeine.version>2.8.4</com.github.ben-manes.caffeine.version>
--- a/pom.xml
+++ b/pom.xml
@@ -242,6 +242,8 @@
               <argument>CONTAINER_JFR_ARCHIVE_PATH=/flightrecordings</argument>
               <argument>--env</argument>
               <argument>CONTAINER_JFR_TEMPLATE_PATH=/templates</argument>
+              <argument>--env</argument>
+              <argument>GRAFANA_DATASOURCE_URL=http://abc</argument>
               <argument>--detach</argument>
               <argument>--rm</argument>
               <argument>${containerjfr.imageStream}:${containerjfr.imageVersion}</argument>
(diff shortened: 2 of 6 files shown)


andrewazores:
I think this is a good opportunity to write an integration test to guard against any regressions, in particular against #217 type of behaviour since it's consistently reproducible. At least for now there are no other tests that exercise `upload-recording`, so I think simply setting an invalid datasource URL in the integration test configuration in `pom.xml` is acceptable. When we get to the point of a multi-container integration test suite and can thoroughly exercise the uploading workflow then we can think about how to set up the test suite to allow multiple container-jfr runs with different datasource URL env var settings.

vic-ma:
I've got the test working, but now `GrafanaDatasourceIT.shouldFail()` is failing, because that test is checking to see that the datasource URL web handler returns a `500` when `GRAFANA_DATASOURCE_URL` is not set, but for my test I have to set the env var. Is there a workaround to this?

andrewazores:
Ah, yea that would be a case where we need that ability to set different values for it. I think that `shouldFail()` test can just be disabled for now - it isn't as important as this one.

vic-ma:
Marking as draft while I investigate a bug I just found.

vic-ma:
So the bug is that if I set `GRAFANA_DATASOURCE_URL` to `http://localhost:8080/load`, meaning the actual URL used for the POST is `http://localhost:8080/load/load`, then the first time I run `upload-recording` it will handle it correctly, saying that it did not get a success code; but when I run it again, it will hang.

And really, this bug happens with any invalid path after `http://localhost:8080`. It does not seem to happen with any other servers though, just jfr-datasource. Though if I run curl on the invalid URL multiple times, jfr-datasource responds correctly with an error message each time. And I tested the same invalid URL in the Apache version of `upload-recording`, and the bug doesn't exist there. So I don't think there's an issue with jfr-datasource. But then I can't think of what the cause of this bug might be. The bug is present in upstream as well, so it can really be seen as a third type of `upload-recording` hanging bug, in addition to the two already identified.

But in any case, the solution I found was to just add a timeout to the request. I guess in cases like `http://abc`, a timeout wasn't necessary because it never connects in the first place, but here a timeout solves the issue. Now that I think of it, I think we also could've used the `CompleteableFuture.get()` that has a timeout as a workaround to these hanging problems, though it's good that we were able to fix the root cause of course.

The timeout is also good for any other servers that connect and just hang. I can't think of any and I never really tested this, but it's definitely a good protection that I should have added in the first place.

Interestingly, with the change, after the timeout happens, the next time I run `upload-recording`, it will properly output the error message from jfr-datasource immediately, but the one after hangs again (until the timeout).

src/test/java/itest/GrafanaDatasourceIT.java
```diff
@@ -60,7 +59,8 @@ void createRequest() {
         req = webClient.get("/api/v1/grafana_datasource_url");
     }
 
-    @Test
+    // Disabled for now due to conflict with UploadRecordingIT; see #229
```
andrewazores, line 62: Just for ease of later back-reference, in case we ever move repos or something, let's expand out that issue number to a full link.

andrewazores:
Looks good from a reading. I'll run it and test it manually to be sure I see the same fixed behaviour you do, and then if everything checks out this looks good to merge.

andrewazores:
I'm pretty sure it also does apply, but could you check that this same bug applies to `POST /api/v1/targets/:targetId/recordings/:recordingName/upload` and `POST /api/v1/recordings/:recordingName/upload`? These are the HTTP API equivalents to `upload-recording` and contain basically copy-paste implementations. If this also needs to be fixed separately by adding the `.timeout` or whatever then it's up to you if this is added to this PR or (maybe more correctly) filed and fixed as a separate issue/PR.

andrewazores:
Tested it out and it seems to work as expected. I think this is ready to merge now but I'll wait until you let me know if you'll add any further commits to patch up the HTTP API as well or if that will be separate (or is unnecessary).

vic-ma:
The hanging from a nonexistent server URL is fixed in the HTTP handler version once the Vertx version is updated, as well.

For existent but non-jfr-datasource servers, the handler returns the response from the invalid server. So I think this just needs the status code check from #224.

For jfr-datasource URLs with an invalid path at the end, the handler upload works as is, because of the resolve + normalize (which isn't in the websocket command code), and there is also a timeout there already, so that would have solved it anyway.

So the only change should be the status code check. I'll do it in a seperate PR.

andrewazores:
Great, thanks for checking on all that.

Issue #220, opened 2020-07-27
Can't delete duplicate/renamed saved recording


ebaron:
While testing a feature in the operator that wasn't working correctly, I stumbled across this bug in the `delete-saved` command.
```
server failed to execute \"delete-saved\": [10-128-0-66_my-recording_20200716T214229Z.1.jfr] are invalid arguments to delete-saved (code -1) 
```
I think the regex needs to be expanded to include the _.n_ suffix.

PR #236, opened 2020-08-10
Fix delete recording bug for renamed recordings

Alexjsenn:


Alexjsenn:
Fixes #220

src/main/java/com/redhat/rhjmc/containerjfr/commands/Command.java
```diff
@@ -71,6 +71,6 @@ default boolean validateTargetId(String targetId) {
     }
 
     default boolean validateRecordingName(String name) {
-        return name.matches("[\\w-_]+(\\.jfr)?");
+        return name.matches("[\\w-_]+(\\.\\d)?(\\.jfr)?");
```
andrewazores, line 74: What happens when this gets to double-digit rename suffixes?

src/test/java/com/redhat/rhjmc/containerjfr/commands/internal/DeleteSavedRecordingCommandTest.java
```diff
@@ -107,6 +107,12 @@ void shouldNotValidateIncorrectArgc(int argc) {
         MatcherAssert.assertThat(e.getMessage(), Matchers.equalTo(errorMessage));
     }
 
+    @Test
+    void shouldValidateRenamedRecording() {
```
andrewazores, line 111: Please make this a `@ParameterizedTest` and test various different numeric suffixes to be sure they all work.

src/main/java/com/redhat/rhjmc/containerjfr/commands/Command.java
Alexjsenn: My bad, for some reason I had remembered wrong and thought it only renamed up to 9, it's fixed now.

PR #238, opened 2020-08-11
Fix username problem for RJMX credentials

### LINKED ISSUE #221: entrypoint RJMX credentials should be configurable
See discussion starting at https://github.com/rh-jmc-team/container-jfr/pull/209#pullrequestreview-456108081 .

The built OCI image contains an `entrypoint.sh` script which randomly generates a password. This password and the hard-coded username `containerjfr` (from `jmxremote.access`) are used to secure the RJMX port which ContainerJFR itself opens and listens on for incoming connections.

The end user should be able to specify the username and password to use for this authentication system by setting an environment variable for each. If those variables are set then those are the credentials that should be used. If the username is not set then `containerjfr` may be used, and if the password is not set then it should be autogenerated.

Alexjsenn:
Fixes #221 
Remove jmxremote.access, and generate it with the correct username in entrypoint.sh

### DIFF
--- a/src/main/extras/app/entrypoint.sh
+++ b/src/main/extras/app/entrypoint.sh
@@ -3,6 +3,7 @@
 set -x
 set -e
 
+USRFILE="/tmp/jmxremote.access"
 PWFILE="/tmp/jmxremote.password"
 function createJmxPassword() {
     if [ -z "$CONTAINER_JFR_RJMX_USER" ]; then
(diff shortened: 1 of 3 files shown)


src/main/extras/app/entrypoint.sh
```diff
@@ -3,6 +3,7 @@
 set -x
 set -e
 
+USRFILE="/tmp/jmxremote.access"
 PWFILE="/tmp/jmxremote.password"
 function createJmxPassword() {
```
andrewazores, line 8: I think this function should be renamed now that its responsibilities have been expanded, what do you think?

Alexjsenn: yes, definitely makes sense!

PR #239, opened 2020-08-13
Add Null Check to Command Validation

### LINKED ISSUE #213: Null Arguments Cause a NPE in `validate()`
Currently, `validate()` does not check for null, so when a null argument is passed in, it will be treated as a regular `String` and get passed to methods like `validateTargetId()`, or be used with `String` instance methods, and this will cause a `NullPointerException`.

```
{command:dump,args:[,foo,1,"template=Continuous"]}
```
```
[INFO] (192.168.0.107:33136): CMD {command:dump,args:[,foo,1,"template=Continuous"]}

[WARNING] 
    at javax.management.remote.JMXServiceURL.<init> (JMXServiceURL.java:148)
    at com.redhat.rhjmc.containerjfr.commands.Command.validateTargetId (Command.java:63)
    at com.redhat.rhjmc.containerjfr.commands.internal.DumpCommand.validate (DumpCommand.java:154)
    at com.redhat.rhjmc.containerjfr.commands.internal.SerializableCommandRegistryImpl.validate (SerializableCommandRegistryImpl.java:112)
    at com.redhat.rhjmc.containerjfr.tui.ws.WsCommandExecutor.run (WsCommandExecutor.java:111)
    at com.redhat.rhjmc.containerjfr.ContainerJfr.main (ContainerJfr.java:115)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:566)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:282)
    at java.lang.Thread.run (Thread.java:834)
```

vic-ma:
Fixes #213.

I added the null check to commands with zero arguments as well.

I had made a `ValidationTestable` sub-interface that automated the null tests, but by the end of it I realized that the logic for generating the arglist permutations was complex enough that it basically warranted being tested itself, so I ended up scrapping it.

For a command with three parameters, the order of the tests is like this (`x` = null), and for four, it's the same pattern:
```
x x x
x x o
x o x
x o o
o x x
o x o
o o x
```

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/DeleteCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/DeleteCommand.java
@@ -135,6 +135,14 @@ class DeleteCommand extends AbstractConnectedCommand implements SerializableComm
             throw new FailedValidationException(errorMessage);
         }
 
+        for (String arg : args) {
+            if (arg == null) {
+                String errorMessage = "One or more arguments were null";
+                cw.println(errorMessage);
+                throw new FailedValidationException(errorMessage);
+            }
+        }
+
         StringJoiner combinedErrorMessage = new StringJoiner("; ");
 
         if (!validateTargetId(args[0])) {
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/DeleteCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/DeleteCommand.java
@@ -135,12 +135,10 @@ class DeleteCommand extends AbstractConnectedCommand implements SerializableComm
             throw new FailedValidationException(errorMessage);
         }
 
-        for (String arg : args) {
-            if (arg == null) {
-                String errorMessage = "One or more arguments were null";
-                cw.println(errorMessage);
-                throw new FailedValidationException(errorMessage);
-            }
+        if (!validateNoNullArgs(args)) {
+            String errorMessage = "One or more arguments were null";
+            cw.println(errorMessage);
+            throw new FailedValidationException(errorMessage);
         }
 
         StringJoiner combinedErrorMessage = new StringJoiner("; ");
--- a/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ExitCommand.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ExitCommand.java
@@ -69,6 +69,13 @@ public class ExitCommand implements Command {
             cw.println(errorMessage);
             throw new FailedValidationException(errorMessage);
         }
+        for (String arg : args) {
+            if (arg == null) {
+                String errorMessage = "One or more arguments were null";
+                cw.println(errorMessage);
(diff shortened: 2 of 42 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/DeleteCommand.java
```diff
@@ -135,6 +135,14 @@ public void validate(String[] args) throws FailedValidationException {
...
         }
 
+        for (String arg : args) {
+            if (arg == null) {
+                String errorMessage = "One or more arguments were null";
+                cw.println(errorMessage);
+                throw new FailedValidationException(errorMessage);
+            }
+        }
+
```
andrewazores, line 145: This snippet looks like it's worth extracting to a utility method somewhere, rather than copy-pasting the implementation around.

src/main/java/com/redhat/rhjmc/containerjfr/commands/internal/ExitCommand.java
```diff
@@ -69,6 +69,13 @@ public void validate(String[] args) throws FailedValidationException {
             cw.println(errorMessage);
             throw new FailedValidationException(errorMessage);
         }
+        for (String arg : args) {
+            if (arg == null) {
+                String errorMessage = "One or more arguments were null";
+                cw.println(errorMessage);
+                throw new FailedValidationException(errorMessage);
+            }
+        }
```
andrewazores, line 78: For commands which expect 0 arguments, does it still make sense to make assertions about the contents of an empty array? They (should have) already validated that the array has length 0 at this point.

vic-ma: My thinking was that since the code would be dead anyway in those cases, we could just put the check in every `validate()` method, to make it uniform and to guard against possible additions of parameters. I guess the tests would need to be added anyway in that case though, and unnecessary code is probably bad, so I'll get rid of it.

andrewazores: That's a fair enough idea, although in that case it would make sense to do something like put the null validation into a superclass and then ensure that all validations call through it. ex. (rough pseudojava):

```
abstract class SuperClass {
  void doValidation(String[] args) {
    doNullValidation(args);
    validate(args);
  }
  void validate(String[] args);
}

class SubClass {
  void validate(String[] args) {
    doNumArgsValidation();
    doOtherValidations();
  }
}
```

vic-ma: I ended up just getting rid of it since I think that's a little cleaner and since there won't be many commands eventually, so it's probably fine to just keep the validation simple.

PR #242, opened 2020-08-18
Change call/response JMX auth header

andrewazores:
Related to #5

Server's "call" is changed to X-JMX-Authenticate, while corresponding
client response is left as "X-JMX-Authorization". This matches the HTTP
WWW-Authenticate/Authorization headers.

This header will be used by clients to discover the authentication scheme expected by the ContainerJFR instance. The only supported scheme at this time is `Basic`.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/AbstractAuthenticatedRequestHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/AbstractAuthenticatedRequestHandler.java
@@ -62,7 +62,8 @@ abstract class AbstractAuthenticatedRequestHandler implements RequestHandler {
 
     static final Pattern AUTH_HEADER_PATTERN =
             Pattern.compile("(?<type>[\\w]+)[\\s]+(?<credentials>[\\S]+)");
-    static final String JMX_AUTH_HEADER = "X-JMX-Authorization";
+    static final String JMX_AUTHENTICATE_HEADER = "X-JMX-Authenticate";
+    static final String JMX_AUTHORIZATION_HEADER = "X-JMX-Authorization";
 
     private final AuthManager auth;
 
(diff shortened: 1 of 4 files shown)


vic-ma:
I'm trying to figure out how to test this and how the auth stuff works in general.

Is JMX authentication always enabled, or is that what the flag `Dcom.sun.management.jmxremote.authenticate` does? Though I think that flag existed before all the JMX auth PRs, so maybe not?

How does JMX auth relate to user authentication? When you say:

>The only supported scheme at this time is `Basic`

Is this referring to the `BasicAuthManager`, and so I should enable that when testing this, or do you mean something else?

andrewazores:
JMX authentication will be enabled whenever `CONTAINER_JFR_RJMX_AUTH=true`, `CONTAINER_JFR_RJMX_USER=something`, or `CONTAINER_JFR_RJMX_PASS=something`. When any of those are true/set, then the flag `-Dcom.sun.management.jmxremote.authenticate=true` is set and so are `-Dcom.sun.management.jmxremote.password.file=$PWFILE` and `-Dcom.sun.management.jmxremote.access.file=$USRFILE`, which are what tell the JVM what users have which roles and what their passwords are for securing its own JMX listening port. These are all standard JVM flags, not ContainerJFR specific.

JMX auth is for users to authenticate themselves to targets which are configured using the JVM flags described above. User auth, using the `BasicAuthManager` for example, is for users to authenticate themselves to ContainerJFR itself. "The only supported scheme at this time is `Basic`" refers to the way that credentials are provided in the authorization header - Basic is a scheme meaning that the credentials are provided like `base64(username:password)`. This is also a standardized auth scheme, not specific to ContainerJFR or either the connection between ContainerJFR/target or client/ContainerJFR. Another common scheme is Bearer, using Bearer tokens, which we have an example of in the OpenShiftAuthManager for example, where rather than sending their username:password directly to the server, clients instead have some kind of authorization token that has been provided to them by the server and which they pass back with requests to prove their identity.

vic-ma:
It's working as intended with curl (just had to change `.../recordings/foo` to `.../recordings`).

The code looks good as well, but it just needs a `spotless:apply`.

andrewazores:
It should work with `.../recordings/foo` too, although I suppose it will try to download the JFR binary and dump that to stdout, so `curl` will probably complain. Append a `--output foo.jfr` to the second and third `curl` commands and it should continue normally and download the file.

vic-ma:
Oh, I thought that command was just to see what recordings were in `localhost` (like running `list`), and I didn't see the usual output so I changed it...I'll try again now.

vic-ma:
Yeah, it works with `--output foo.jfr` appended.

PR #241, opened 2020-08-18
Add smoketesting script

andrewazores:

PR #243, opened 2020-08-19
JMX SSL

andrewazores:


andrewazores:
@ebaron could you take a look over and see if this makes sense? What I've tried to do is:

1. generate a self-signed cert
2. copy the system trusted certificate store to a writeable copy
3. import the self-signed cert into the writeable store copy
4. iterate over any certs in mounted volume `/truststore` and import those into the copy
5. use the store copy for the JVM's SSL trust
6. enable SSL on JMX remote connections

At various stages along the way I had various configuration mistakes that caused things to break in various ways, so I'm fairly confident now that the setup I have is indeed enabling SSL on ContainerJFR itself at least, and importing the certificates seems to work although I haven't yet tested importing a certificate for a target that has also secured itself with SSL.

I tested this with a pieced-together setup of this branch, #241 with the vertx-fib-demo version upgraded to 0.3.0, and a very lightly modified rh-jmc-team/container-jfr-web#116 that works with the JMX auth protocol in this branch (X-JMX-Authorization header name changed). This gives me a setup with ContainerJFR and a vertx-fib-demo sample app, both configured with JMX credentials, and with ContainerJFR also securing itself with SSL. I was still able to connect to both ContainerJFR and the sample app as usual. For ContainerJFR with auth credentials enabled (which also enables SSL), I no longer see the "discouraged management agent settings" entry in the automated analysis report, which is extra confirmation that the SSL setup is now working.

andrewazores:
Ah, yea I need to double check that RMI port thing. I hit an issue earlier in this branch where the JVM complained at startup that the port numbers are the same so I changed them, but when I went back to patch up the vertx-fib-demo with what I learned here I got it working without separated port numbers. I'll check if they can be put back here but I don't see why not.

To be honest I didn't fully catch what the separate RMI port is used for exactly, but the two port numbers is just part of the protocol for some purpose. Apparently they can be assigned to the same port, or at least the JVM handles that gracefully for our purposes, so it seems to be fine.

andrewazores:
Seems like the port numbers can still be set to the same value, so I've put it back that way. I also cleaned up the truststores directory handling in `run.sh` a bit and fixed the `entrypoint.sh` cert import so that it works recursively on files in the mounted volume, rather than trying to import anything at the top-level of the directory contents (including other directories).

andrewazores:
I did also test connecting ContainerJFR to a vertx-fib-demo with everything set up for SSL, except without the vertx-fib-demo cert added to ContainerJFR's trust store, and it results in a connection failure with a message along the lines of "non-JRMP server". This ends up in a `404` response status code for the client, so that needs to be handled better, but the connection does fail in a predictable and detectable way when the connection seems to be refused due to lack of SSL trust from ContainerJFR to the target.

PR #245, opened 2020-08-20
Add Check for Upload Handler POST Response

### LINKED ISSUE #237: Recording Upload HTTP Handlers Should Check Server Response's Status Code
`POST /api/v1/targets/:targetId/recordings/:recordingName/upload` and `POST /api/v1/recordings/:recordingName/upload` should also implement a check similar to #224 so that if `GRAFANA_DATASOURCE_URL` points to an existing but invalid server, the user will be notified when an upload is attempted. Right now, an upload just returns the response from the invalid server.

vic-ma:
Fixes #237.

Mostly just the same implementation as #224. I had to remove the response body from the exception message though, since it could contain newlines or other characters not allowed in a status message.

Tests are again just the `shouldDoPost()` test with a modification to the `Answer` and verifications at the end.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingUploadPostHandler.java
@@ -41,6 +41,8 @@
  */
 package com.redhat.rhjmc.containerjfr.net.web.handlers;
 
+import static com.redhat.rhjmc.containerjfr.util.HttpStatusCodeIdentifier.isSuccessCode;
+
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.InvalidPathException;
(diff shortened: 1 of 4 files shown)


vic-ma:
I'm running into a weird problem where writing to the body in `handleAuthenticated()` is causing the failure handler to not be able to set the status code or status message.

If I have `handleAuthenticated()` to just be this one line:
```
throw new HttpStatusException(502, "EXCEPTION MESSAGE");
```

Then it will work as expected; when I curl, I can see `502 EXCEPTION MESSAGE` in the status line and `EXCEPTION MESSAGE` in the body.

But if I write to the body before throwing the exception, so the method looks like this:
```
ctx.response().setChunked(true).write("BODY MESSAGE");
throw new HttpStatusException(502, "EXCEPTION MESSAGE");
```

Then the body of the response will be `BODY MESSAGEEXCEPTION MESSAGE`, which is expected. But the status line will be `200 OK`, which is unexpected.

In the failure handler I tried printing `ctx.response().getStatusCode()` and `ctx.response().getStatusMessage()`, but it is set correctly to `502` and `EXCEPTION MESSAGE`.

If I add this to `handleAuthenticated()`:
```
ctx.response().setStatusCode(502).setStatusMessage("EXCEPTION MESSAGE");
```
Then the status line will be set properly.

If I do the above, and then hardcode the failure handler to set the status code and message to something else, then the response will still use the status code and message from `handleAuthenticated()`.

andrewazores:
Strange. Probably worth figuring out, but not necessary for this PR.

andrewazores (approved):

PR #253, opened 2020-08-26
Use autogenerated SSL cert for WebServer

### LINKED ISSUE #249: (Optionally) use generated self-signed cert to enable WebServer SSL
When generally deployed, ContainerJFR's WebServer uses insecure HTTP/WS connections unless the deploying user configures a specific SSL cert to use and enable HTTPS/WSS.

In the OpenShift scenario this works well because OpenShift allows TLS edge termination, which the container-jfr-operator enables, allowing the connections to be secured with a certificate that the end client presumably already trusts.

Since #243, ContainerJFR generates a self-signed SSL certificate for securing its incoming JMX connections.

By default, this self-signed SSL certificate should also be used for enabling SSL and HTTPS/WSS on the WebServer. It should still be possible to disable this behaviour and force insecure HTTP/WS for use behind TLS edge termination, and it must also remain possible for the user to specify a certificate to be used in place of the generated one.

Question: if the user specifies a WebServer SSL cert, should this also be used for the JMX connection?

andrewazores:
Fixes #249


src/main/extras/app/entrypoint.sh
```diff
@@ -108,35 +117,45 @@ FLAGS=(
...
+    if [ -z "$KEYSTORE_PATH" ] || [ -z "$KEYSTORE_PASS" ]; then
+        generateSslCert
+        banner "Using self-signed SSL certificate"
+
+        KEYSTORE_PATH="$SSL_KEYSTORE"
+        KEYSTORE_PASS="$SSL_KEY_PASS"
+    fi
+
+    FLAGS+=("-Djavax.net.ssl.keyStore=$SSL_KEYSTORE")
+    FLAGS+=("-Djavax.net.ssl.keyStorePassword=$SSL_KEY_PASS")
```
ebaron, line 151: Should these be set to `KEYSTORE_PATH` and `KEYSTORE_PASS`?

andrewazores: I think you're right. Fixed, thanks

PR #254, opened 2020-08-27
HTTP API changes to support recording options

### LINKED ISSUE #250: HTTP API should support recording options
The following proposed changes would enable the web client to show and modify recording options such as toDisk, maxAge and maxSize.

#### New Recordings
Create a new recording on a target JVM
Method: `POST /api/v1/targets/targetId/recordings`
New (optional) request body attributes:
* **toDisk** *bool*
* **maxAge** *integer (seconds)*
* **maxSize** *integer (bytes)*
* **setDefault** *bool*

#### Default recording options
Retrieve or set default recording options that apply to recordings created on any JVM target

Method: `GET /api/v1/recordingOptions`
Response attributes:
 * **toDisk** *bool*
* **maxAge** *integer (seconds)*
* **maxSize** *integer (bytes)*

Method: `POST /api/v1/recordingOptions`
Request body attributes:
 * **toDisk** *bool*
* **maxAge** *integer (seconds)*
* **maxSize** *integer (bytes)*

Alexjsenn:
Fixes #250


Alexjsenn:
One thing that needs attention is the units in which maxAge and maxSize are stored, and what units the API should expect. Following the WS command, I used seconds for age and bytes for size, but after the age is read it it is converted to miliseconds (apparently that is how it is stored internally). At the moment the new endpoints are asymetrical because the PUT endpoint will recieve a seconds value and convert into miliseconds, then the GET endpoint will return it in miliseconds. I am not sure what the best way to deal with this is.

andrewazores:
The units expected in should be the same as the units returned out. Since we're already expecting the units in to be seconds then the units out should be converted to match. The fact that it's stored internally as something other than seconds is an implementation detail which should not be leaked into the API.

andrewazores:
I have unit test failures on this branch, see attached log. There are also spotless style violations:

```
$ mvn spotless:check
[INFO] Scanning for projects...
[INFO] 
[INFO] ------------< com.redhat.rhjmc.containerjfr:container-jfr >-------------
[INFO] Building container-jfr 0.19.0
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- spotless-maven-plugin:1.27.0:check (default-cli) @ container-jfr ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 5.026 s
[INFO] Finished at: 2020-09-03T15:51:20-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:1.27.0:check (default-cli) on project container-jfr: The following files had format violations:
[ERROR]     src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
[ERROR]         @@ -108,7 +108,7 @@
[ERROR]          
[ERROR]          ····························Map<String,·IOptionDescriptor<?>>·TargetRecordingOptions·=
[ERROR]          ····································connection.getService().getAvailableRecordingOptions();
[ERROR]         -···························
[ERROR]         +
[ERROR]          ····························Map<String,·String>·map·=·new·HashMap<String,·String>();
[ERROR]          
[ERROR]          ····························if·(recordingOptions.get("toDisk")·!=·null)·{
[ERROR]     src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/TargetRecordingsPostHandler.java
[ERROR]         @@ -166,8 +166,7 @@
[ERROR]          ····································builder·=·builder.toDisk(Boolean.valueOf(attrs.get("toDisk")));
[ERROR]          ································}
[ERROR]          ································if·(attrs.contains("maxAge"))·{
[ERROR]         -····································builder·=
[ERROR]         -············································builder.maxAge(Long.parseLong(attrs.get("maxAge")));
[ERROR]         +····································builder·=·builder.maxAge(Long.parseLong(attrs.get("maxAge")));
[ERROR]          ································}
[ERROR]          ································if·(attrs.contains("maxSize"))·{
[ERROR]          ····································builder·=·builder.maxSize(Long.parseLong(attrs.get("maxSize")));
[ERROR] Run 'mvn spotless:apply' to fix these violations.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

Needs a `mvn spotless:apply`.
[test.log](https://github.com/rh-jmc-team/container-jfr/files/5170872/test.log)

Alexjsenn:
My bad, I changed one line at the end without noticing, it should work now.

andrewazores:
```
$ curl -vk -X PUT 'maxAge=30&toDisk=true' https://localhost:8181/api/v1/targets/localhost/recordingOptions
* Could not resolve host: maxAge=30&toDisk=true
* Closing connection 0
curl: (6) Could not resolve host: maxAge=30&toDisk=true
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#1)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  3 20:15:22 2020 GMT
*  expire date: Mar  2 20:15:22 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> PUT /api/v1/targets/localhost/recordingOptions HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> 
^C
```

I tried sending a PUT request and never got a response. I don't see anything in the implementation that would send any response or close the connection, either. The handler should set a status code and end the response at the least. The request doesn't show up in the container-jfr logs either, but I think that has the same root cause of the request never being completed.

andrewazores:
Same when using curl's `-d` flag:

```
$ curl -vk -X PUT -d maxAge=30 https://localhost:8181/api/v1/targets/localhost/recordingOptions
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  3 20:20:24 2020 GMT
*  expire date: Mar  2 20:20:24 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> PUT /api/v1/targets/localhost/recordingOptions HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> Content-Length: 9
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 9 out of 9 bytes
^C
```

Alexjsenn:
Ah, I think I was going to ask you if there needed to be any checks or errors it should look out for, should I just always return 201?

andrewazores:
I don't think `201 CREATED` is the right response - nothing here is newly created. Maybe if it was a `POST`. For a `PUT`, `200 OK` seems fine.

Also, I think since we ended up settling on implementing the partial update map, the `PATCH` verb would be more appropriate than `PUT` in the end. My bad on that one - I recently told you on IRC to implement it so that the partial maps are allowed, but earlier on the API proposal issue I said that that could be a later enhancement and that the full-map `PUT` could be done first.

Alexjsenn:
Ok, Ill make those changes!

Alexjsenn:
It should now work, after sending PATCH requests it seems to properly update and reflect it if pulled using GET

andrewazores:
Cool, functionality-wise this seems to all be generally working as expected now. Nice work.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
```diff
@@ -0,0 +1,150 @@
...
+                            RecordingOptionsBuilder builder =
+                                    recordingOptionsBuilderFactory.create(connection.getService());
+                            IConstrainedMap<String> recordingOptions = builder.build();
+
+                            Map<String, IOptionDescriptor<?>> TargetRecordingOptions =
+                                    connection.getService().getAvailableRecordingOptions();
+
+                            Map<String, String> map = new HashMap<String, String>();
+
+                            if (recordingOptions.get("toDisk") != null) {
```
andrewazores, line 114: Can these three if/elses be refactored into a loop over a set of keys? And rather than using constant strings like `toDisk`, could these use the RecordingOptionsCustomizer.OptionKeys?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
```diff
@@ -0,0 +1,150 @@
...
+                                map.put("toDisk", recordingOptions.get("toDisk").toString());
+                            } else {
+                                map.put(
+                                        "toDisk",
+                                        TargetRecordingOptions.get("disk").getDefault().toString());
+                            }
+
+                            String maxAge;
+                            if (recordingOptions.get("maxAge") != null) {
```
andrewazores, line 123: Does the `recordingOptions` have a method like `.containsKey()`? If so, it would be better to use that than `.get() != null`.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
```diff
@@ -0,0 +1,150 @@
...
+    void handleAuthenticated(RoutingContext ctx) throws Exception {
+        Map<String, String> optionMap =
+                connectionManager.executeConnectedTask(
+                        getConnectionDescriptorFromContext(ctx),
+                        connection -> {
+                            RecordingOptionsBuilder builder =
+                                    recordingOptionsBuilderFactory.create(connection.getService());
+                            IConstrainedMap<String> recordingOptions = builder.build();
+
+                            Map<String, IOptionDescriptor<?>> TargetRecordingOptions =
```
andrewazores, line 109: This should be renamed `targetRecordingOptions` (camelCase, not PascalCase)

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
```diff
@@ -0,0 +1,99 @@
...
+import com.redhat.rhjmc.containerjfr.core.RecordingOptionsCustomizer.OptionKey;
+import com.redhat.rhjmc.containerjfr.net.AuthManager;
+
+import io.vertx.core.MultiMap;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+
+class RecordingOptionsPatchHandler extends AbstractAuthenticatedRequestHandler {
+
+    static final String PATH = "/api/v1/targets/:targetId/recordingOptions";
```
andrewazores, line 56: This is duplicated from the `GET` handler, isn't it? One of them should probably reference the other's path here.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    public String path() {
+        return PATH;
+    }
+
+    @Override
+    public boolean isAsync() {
+        return false;
```
andrewazores, line 77: The `handleAuthenticated` here is non-blocking so I think this can be omitted.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        if (attrs.contains("maxAge")) {
+            OptionKey.fromOptionName("maxAge")
+                    .ifPresent(key -> customizer.set(key, attrs.get("maxAge")));
+        }
+        if (attrs.contains("maxSize")) {
+            OptionKey.fromOptionName("maxSize")
+                    .ifPresent(key -> customizer.set(key, attrs.get("maxSize")));
+        }
+        ctx.response().setStatusCode(200);
```
andrewazores, line 96: I wonder if it would make sense to respond with 200 and then send back the updated map, or respond `204` and no response body? If the client doesn't care about the updated value then they can simply ignore it.

Alexjsenn: Yes, returning the updated map is good, do I need to convert it to json when sending it back?

andrewazores: Yes.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
Alexjsenn: the IConstrainedMap does not seem to have that method, the only method that I can tell that serves the purpose is .get

andrewazores: What about a `.keySet().contains()`?

andrewazores: Hmm.

```
	/**
	 * @return A {@link Set set} of keys which are known to be valid. That is, those that currently
	 *         are known to have a {@link IConstraint constraint}. This includes all keys which
	 *         currently have a value, but additional keys may be included. In other words,
	 *         {@link #get(Object)} may return {@code null} for some keys included in this set.
	 */
	Set<K> keySet();
```

Well, nevermind. Just keep using `.get() != null` then.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
Alexjsenn: Should work now!

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
```diff
@@ -0,0 +1,148 @@
...
+import com.redhat.rhjmc.containerjfr.net.AuthManager;
+import com.redhat.rhjmc.containerjfr.net.TargetConnectionManager;
+
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+
+public class RecordingOptionsGetHandler extends AbstractAuthenticatedRequestHandler {
+
+    static final String PATH = "/api/v1/targets/:targetId/recordingOptions";
+    protected final TargetConnectionManager connectionManager;
```
andrewazores, line 68: I think this can be `private`, too.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
```diff
@@ -0,0 +1,123 @@
...
+
+import io.vertx.core.MultiMap;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+
+class RecordingOptionsPatchHandler extends AbstractAuthenticatedRequestHandler {
+
+    static final String PATH = RecordingOptionsGetHandler.PATH;
+    private final RecordingOptionsCustomizer customizer;
+    protected final TargetConnectionManager connectionManager;
```
andrewazores, line 66: `private` here too

src/test/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandlerTest.java
```diff
@@ -0,0 +1,152 @@
...
+    void shouldSetRecordingOptions() throws Exception {
+        Map<String, String> defaultValues =
+                Map.of("toDisk", "true", "maxAge", "50", "maxSize", "32");
+        Mockito.when(recordingOptionsBuilderFactory.create(Mockito.any())).thenReturn(builder);
+        Mockito.when(builder.build()).thenReturn(recordingOptions);
+        Mockito.when(recordingOptions.get("toDisk")).thenReturn(defaultValues.get("toDisk"));
+        Mockito.when(recordingOptions.get("maxAge")).thenReturn(defaultValues.get("maxAge"));
+        Mockito.when(recordingOptions.get("maxSize")).thenReturn(defaultValues.get("maxSize"));
+
+        MultiMap RequestAttrs = MultiMap.caseInsensitiveMultiMap();
```
andrewazores, line 119: should be `requestAttrs`

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
Alexjsenn: I made a similar loop in the PATCH handler, but this one makes less sense because toDisk has to be handled differently than the other two. It would have to be a loop for just two of the options.

andrewazores: How about extracting utility methods for handling plain textual values like `toDisk` and numeric values like the other two?

Alexjsenn: How would that work, how would the utility method know which type it need to check? woudl this still be in a loop?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
```diff
@@ -98,19 +98,17 @@ void handleAuthenticated(RoutingContext ctx) throws Exception {
...
-                            if (attrs.contains("maxAge")) {
-                                OptionKey.fromOptionName("maxAge")
-                                        .ifPresent(key -> customizer.set(key, attrs.get("maxAge")));
-                            }
-                            if (attrs.contains("maxSize")) {
-                                OptionKey.fromOptionName("maxSize")
-                                        .ifPresent(
-                                                key -> customizer.set(key, attrs.get("maxSize")));
+                            String[] keys = {"toDisk", "maxAge", "maxSize"};
+                            for (String key : keys) {
```
andrewazores, line 102: `keys` can be inlined since it isn't used anywhere lower down in the scope. Or, to be extra fancy, you could do:

```
Arrays.asList("toDisk", "maxAge", "maxSize").forEach(key -> {
  if (attrs.contains(key)) {
  ...
  }
...
});
```

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
andrewazores: No loop. Make a utility method like `getNumericOption` and another like `getBooleanOption`.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsPatchHandler.java
Alexjsenn: Ok, that's definitely better!

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
Alexjsenn: I created a helper function for the numeric ones, I don't know if it makes sense for the boolean one because for some reason the name of the option is different between recordingOptionsCustomizer and the target recordingOptions default

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/RecordingOptionsGetHandler.java
```diff
@@ -0,0 +1,147 @@
...
+import com.google.gson.Gson;
+
+import com.redhat.rhjmc.containerjfr.commands.internal.RecordingOptionsBuilderFactory;
+import com.redhat.rhjmc.containerjfr.net.AuthManager;
+import com.redhat.rhjmc.containerjfr.net.TargetConnectionManager;
+
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+
+class RecordingOptionsGetHandler extends AbstractAuthenticatedRequestHandler {
```
andrewazores, line 65: Ah, now that the path includes the `targetId` portion, these handlers should be renamed like `TargetRecordingOptionsVerbHandler`.

andrewazores:
That refactor does look much better, the implementation looks pretty clean now.

I did a basic setup using `sh smoketest.sh` and then inspected the results of interacting with this API using `curl`:

```
$ curl -vk -H "X-JMX-Authorization: Basic $(echo -n smoketest:smoketest | base64)" https://localhost:8181/api/v1/targets/localhost/recordingOptions
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  8 15:58:45 2020 GMT
*  expire date: Mar  7 15:58:45 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> GET /api/v1/targets/localhost/recordingOptions HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> X-JMX-Authorization: Basic c21va2V0ZXN0OnNtb2tldGVzdA==
> 
< HTTP/1.1 200 OK
< content-length: 45
< 
* Connection #0 to host localhost left intact
{"maxAge":"0","toDisk":"false","maxSize":"0"}
```

Seems reasonable - the request was successfully processed and responded to with what seems to be a map of default values.

```
$ curl -X POST -d "recordingName=foo&events=template=Continuous" -vk -H "X-JMX-Authorization: Basic $(echo -n smoketest:smoketest | base64)" https://localhost:8181/api/v1/targets/localhost/recordings
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  8 15:58:45 2020 GMT
*  expire date: Mar  7 15:58:45 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> POST /api/v1/targets/localhost/recordings HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> X-JMX-Authorization: Basic c21va2V0ZXN0OnNtb2tldGVzdA==
> Content-Length: 44
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 44 out of 44 bytes
< HTTP/1.1 201 Created
< location: /foo
< content-length: 394
< 
* Connection #0 to host localhost left intact
{"downloadUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/recordings/foo","reportUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/reports/foo","id":1,"name":"foo","state":"RUNNING","startTime":1599581783107,"duration":0,"continuous":true,"toDisk":true,"maxSize":0,"maxAge":0}
```

```
$ curl -vk -H "X-JMX-Authorization: Basic $(echo -n smoketest:smoketest | base64)" https://localhost:8181/api/v1/targets/localhost/recordings
*   Trying ::1:8181...
* TCP_NODELAY set
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  8 15:58:45 2020 GMT
*  expire date: Mar  7 15:58:45 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> GET /api/v1/targets/localhost/recordings HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.66.0
> Accept: */*
> X-JMX-Authorization: Basic c21va2V0ZXN0OnNtb2tldGVzdA==
> 
< HTTP/1.1 200 OK
< content-length: 396
< 
* Connection #0 to host localhost left intact
[{"downloadUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/recordings/foo","reportUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/reports/foo","id":1,"name":"foo","state":"RUNNING","startTime":1599581783107,"duration":0,"continuous":true,"toDisk":true,"maxSize":0,"maxAge":0}]
```

But, if the previous query said `"toDisk": "false"`, why was my new recording created with `"toDisk": "true"`? This is shown in both the recording creation response and the subsequent recording list query. I think there may be a bug with how the `GET` handler populates its map of values.

Alexjsenn:
So from what I can tell, this problem is stemming from the way the target JVM deals with its own defaults. When you run the GET handler it doesn't find any local defaults, and therefore it returns the target JVM defaults. However, it seems that the target JVM then ignores its own toDisk default and just uses true instead of false. Below I modified the GET handler to also show the raw response from IFlightRecorderService::getAvailableRecordingOptions(). I really have no idea what to do about this.

```curl -vk -H "X-JMX-Authorization: Basic $(echo -n smoketest:smoketest | base64)" https://localhost:8181/api/v1/targets/localhost/recordingOptions
*   Trying ::1:8181...
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  8 19:08:58 2020 GMT
*  expire date: Mar  7 19:08:58 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> GET /api/v1/targets/localhost/recordingOptions HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.69.1
> Accept: */*
> X-JMX-Authorization: Basic c21va2V0ZXN0OnNtb2tldGVzdA==
> 
< HTTP/1.1 200 OK
< content-length: 655
< 
* Connection #0 to host localhost left intact
{"maxAge":"0","toDisk":"false","testing":"{name=OptionInfo [label=Name, default=Recording, constraint=Type(text)], duration=OptionInfo [label=Duration, default=30s[s], constraint=org.openjdk.jmc.common.unit.ComparableConstraint@7d728ed1], maxSize=OptionInfo [label=Max Size, default=0B[B], constraint=org.openjdk.jmc.common.unit.ComparableConstraint@10d28693], maxAge=OptionInfo [label=Max Age, default=0s[s], constraint=org.openjdk.jmc.common.unit.ComparableConstraint@7d728ed1], disk=OptionInfo [label=To disk, default=false, constraint=Type(boolean)], dumpOnExit=OptionInfo [label=Dump on Ex
```
```[asenn@localhost web-client]$ curl -X POST -d "recordingName=foo&events=template=Profiling&duration=10" -vk -H "X-JMX-Authorization: Basic $(echo -n smoketest:smoketest | base64)" https://localhost:8181/api/v1/targets/localhost/recordings
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying ::1:8181...
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  8 19:08:58 2020 GMT
*  expire date: Mar  7 19:08:58 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> POST /api/v1/targets/localhost/recordings HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.69.1
> Accept: */*
> X-JMX-Authorization: Basic c21va2V0ZXN0OnNtb2tldGVzdA==
> Content-Length: 55
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 55 out of 55 bytes
< HTTP/1.1 201 Created
< location: /foo
< content-length: 399
< 
* Connection #0 to host localhost left intact
{"downloadUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/recordings/foo","reportUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/reports/foo","id":1,"name":"foo"} 
```
``` [asenn@localhost web-client]$ curl -vk -H "X-JMX-Authorization: Basic $(echo -n smoketest:smoketest | base64)" https://localhost:8181/api/v1/targets/localhost/recordings
*   Trying ::1:8181...
* Connected to localhost (::1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  8 19:08:58 2020 GMT
*  expire date: Mar  7 19:08:58 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> GET /api/v1/targets/localhost/recordings HTTP/1.1
> Host: localhost:8181
> User-Agent: curl/7.69.1
> Accept: */*
> X-JMX-Authorization: Basic c21va2V0ZXN0OnNtb2tldGVzdA==
> 
< HTTP/1.1 200 OK
< content-length: 401
< 
* Connection #0 to host localhost left intact
[{"downloadUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/recordings/foo","reportUrl":"https://0.0.0.0:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/reports/foo","id":1,"name":"foo","state":"STOPPED","startTime":1599592243708,"duration":10000,"continuous":false,"toDisk":true,"maxSize":0,"maxAge":0}]

**Comment by Alexjsenn** [comment]
When I explicitly set the toDisk to false (which creates the entry in the RecordingOptionsCustomizer), any new recordings get created with toDisk=false

**Comment by andrewazores** [comment]
> However, it seems that the target JVM then ignores its own toDisk default and just uses true instead of false.

This seems suspicious... something we should come back to and re-examine later. After creating the recording we should be seeing the option values actually decided upon by the target JVM, but beforehand when checking the defaults with `getAvailableRecordingOptions` we're somehow communicating through a JMC service implementation, so maybe we're doing something wrong with that service. Perhaps when we create the recordings we should explicitly check the `getAvailableRecordingOptions` map as well and merge its values with the ones from our `Customizer`, similar to what you've implemented here.

Anyway, I think this PR is good to go now. The problem above should be investigated more deeply and handled separately, I think.

**Comment by Alexjsenn** [comment]
Yeah, that makes sense.

File: src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/TargetRecordingOptionsPatchHandler.java [hunk 12]
```diff
@@ -0,0 +1,125 @@
...
+                            Arrays.asList("toDisk", "maxAge", "maxSize")
+                                    .forEach(
+                                            key -> {
+                                                if (attrs.contains(key)) {
+                                                    OptionKey.fromOptionName(key)
+                                                            .ifPresent(
+                                                                    optionKey ->
+                                                                            customizer.set(
+                                                                                    optionKey,
+                                                                                    attrs.get(
```
andrewazores, line 111: Last thing:

```
$ curl -vk -X PATCH -d "maxAge=no&toDisk=123" -H "X-JMX-Authorization: Basic $(echo -n user:pass | base64)" https://0.0.0.0:8181/api/v1/targets/localhost/recordingOptions
*   Trying 0.0.0.0:8181...
* TCP_NODELAY set
* Connected to 0.0.0.0 (127.0.0.1) port 8181 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; O=Red Hat; CN=container-jfr
*  start date: Sep  9 16:27:38 2020 GMT
*  expire date: Mar  8 16:27:38 2021 GMT
*  issuer: C=US; O=Red Hat; CN=container-jfr
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
> PATCH /api/v1/targets/localhost/recordingOptions HTTP/1.1
> Host: 0.0.0.0:8181
> User-Agent: curl/7.66.0
> Accept: */*
> X-JMX-Authorization: Basic dXNlcjpwYXNz
> Content-Length: 20
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 20 out of 20 bytes
< HTTP/1.1 200 OK
< content-length: 45
< 
{"maxAge":"0","toDisk":"false","maxSize":"0"}
```

The options I set are invalid, and the ContainerJFR logs do reflect that, but the response sent is still a `200 OK`. The Customizer needs refactoring to throw exceptions when its validations fail instead of simply logging them, however, that refactor should probably wait until the Command equivalent to this handler is removed so that the Customizer doesn't need to support both cases.

So, for now, please add some basic validations for these three attributes, and if the validation fails respond with some `4xx` status code.

andrewazores: Here's the backend log snippet from the invalid request above:

```
[WARN] java.lang.NumberFormatException: For input string: "no"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Long.parseLong(Long.java:692)
	at java.base/java.lang.Long.parseLong(Long.java:817)
	at com.redhat.rhjmc.containerjfr.core.RecordingOptionsCustomizer$OptionKey$1.acceptThrows(RecordingOptionsCustomizer.java:88)
	at com.redhat.rhjmc.containerjfr.core.RecordingOptionsCustomizer$OptionKey$1.acceptThrows(RecordingOptionsCustomizer.java:85)
	at com.redhat.rhjmc.containerjfr.core.util.CheckedConsumer.accept(CheckedConsumer.java:49)
	at com.redhat.rhjmc.containerjfr.core.RecordingOptionsCustomizer.lambda$apply$0(RecordingOptionsCustomizer.java:67)
	at java.base/java.util.HashMap$Values.forEach(HashMap.java:976)
	at com.redhat.rhjmc.containerjfr.core.RecordingOptionsCustomizer.apply(RecordingOptionsCustomizer.java:67)
	at com.redhat.rhjmc.containerjfr.commands.internal.CommandsInternalModule.lambda$provideRecordingOptionsBuilderFactory$0(CommandsInternalModule.java:172)
	at com.redhat.rhjmc.containerjfr.net.web.handlers.TargetRecordingOptionsPatchHandler.lambda$handleAuthenticated$2(TargetRecordingOptionsPatchHandler.java:117)
	at com.redhat.rhjmc.containerjfr.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:86)
	at com.redhat.rhjmc.containerjfr.net.web.handlers.TargetRecordingOptionsPatchHandler.handleAuthenticated(TargetRecordingOptionsPatchHandler.java:98)
	at com.redhat.rhjmc.containerjfr.net.web.handlers.AbstractAuthenticatedRequestHandler.handle(AbstractAuthenticatedRequestHandler.java:82)
	at com.redhat.rhjmc.containerjfr.net.web.handlers.AbstractAuthenticatedRequestHandler.handle(AbstractAuthenticatedRequestHandler.java:61)
	at io.vertx.ext.web.impl.RouteState.handleContext(RouteState.java:1034)
	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:131)
	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
	at io.vertx.ext.web.handler.impl.BodyHandlerImpl$BHandler.doEnd(BodyHandlerImpl.java:296)
	at io.vertx.ext.web.handler.impl.BodyHandlerImpl$BHandler.end(BodyHandlerImpl.java:276)
	at io.vertx.ext.web.handler.impl.BodyHandlerImpl.lambda$handle$0(BodyHandlerImpl.java:87)
	at io.vertx.core.http.impl.HttpServerRequestImpl.onEnd(HttpServerRequestImpl.java:521)
	at io.vertx.core.http.impl.HttpServerRequestImpl.handleEnd(HttpServerRequestImpl.java:507)
	at io.vertx.core.http.impl.Http1xServerConnection.handleEnd(Http1xServerConnection.java:176)
	at io.vertx.core.http.impl.Http1xServerConnection.handleContent(Http1xServerConnection.java:163)
	at io.vertx.core.http.impl.Http1xServerConnection.handleMessage(Http1xServerConnection.java:140)
	at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:366)
	at io.vertx.core.impl.EventLoopContext.execute(EventLoopContext.java:43)
	at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:229)
	at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:173)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.ChannelInboundHandlerAdapter.channelRead(ChannelInboundHandlerAdapter.java:93)
	at io.netty.handler.codec.http.websocketx.extensions.WebSocketServerExtensionHandler.channelRead(WebSocketServerExtensionHandler.java:101)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103)
	at io.netty.handler.codec.MessageToMessageCodec.channelRead(MessageToMessageCodec.java:111)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:324)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:296)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:271)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1518)
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1267)
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1314)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:501)
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:440)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:163)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:714)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:650)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:576)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:493)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
```

Alexjsenn: Should none of the options be set if any of them are invalid?

andrewazores: Yea, that's probably best. Otherwise it will be difficult to communicate to a client that their request has been only partially accepted.

Alexjsenn: Ok!

Alexjsenn: My solution was very unelegant, if you can think of a more compact way I can redo it!

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/TargetRecordingOptionsPatchHandler.java
```diff
@@ -0,0 +1,144 @@
...
+    public String path() {
+        return PATH;
+    }
+
+    @Override
+    void handleAuthenticated(RoutingContext ctx) throws Exception {
+        MultiMap attrs = ctx.request().formAttributes();
+        if (attrs.contains("toDisk")) {
+            try {
+                Boolean.parseBoolean(attrs.get("toDisk"));
```
andrewazores, line 101: This doesn't quite work as intended - `parseBoolean` is `true` if the input string is `"true"`, case insensitive, and `false` for all other inputs. It doesn't throw.

On a related note, it would be good to exercise these validations with unit tests.

Alexjsenn: I just realized that the POST handler should probably also do validation right?

andrewazores: Yes, it probably should

Alexjsenn: Ok, there are now tests for the validations I added. I also added validation for the boolean value in TargetRecordingPostHandler, I throw a NumberFormatException when the boolean can't be parsed to handle it together with the other exceptions, but I realize this is probably not the best way, Im just not sure what exception I should use.

andrewazores: I agree that `NumberFormatException` seems like not quite the right fit here, although I don't think it's too important anyway - in any case, this will be an uncaught exception that is then later on caught by the `AbstractAuthenticatedRequestHandler` and turned into a response with status `500`, so all the exception type (and message) does is become part of the response status message. That will probably be displayed to end users in the end, but clients (as in, the web-client, operator, etc. - other automated clients) should only really care about the status code.

Maybe just `IllegalArgumentException` is fine, unless you want to create a custom one like `BooleanValueException` or `BooleanFormatException`. That would be nice, but it's probably overkill here.

src/test/java/com/redhat/rhjmc/containerjfr/net/web/handlers/TargetRecordingOptionsPatchHandlerTest.java
```diff
@@ -0,0 +1,178 @@
...
+        Mockito.when(req.formAttributes()).thenReturn(requestAttrs);
+        HttpStatusException ex =
+                Assertions.assertThrows(
+                        HttpStatusException.class, () -> handler.handleAuthenticated(ctx));
+        MatcherAssert.assertThat(ex.getStatusCode(), Matchers.equalTo(400));
+    }
+
+    private static Stream<Map<String, String>> getRequestMaps() {
+        return Stream.of(
+                Map.of("toDisk", "5"), Map.of("maxAge", "true"), Map.of("maxSize", "false"));
```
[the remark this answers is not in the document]
Alexjsenn: Ok, there are more cases for both handlers.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/handlers/TargetRecordingsPostHandler.java
```diff
@@ -162,6 +162,22 @@ void handleAuthenticated(RoutingContext ctx) throws Exception {
                                                     TimeUnit.SECONDS.toMillis(
                                                             Long.parseLong(attrs.get("duration"))));
                                 }
+                                if (attrs.contains("toDisk")) {
+                                    Pattern bool = Pattern.compile("true|false");
+                                    Matcher m = bool.matcher(attrs.get("toDisk"));
+                                    if (!m.find())
+                                        throw new NumberFormatException(
```
andrewazores, line 169: The other handler looks like it directly throws an `HttpStatusException`, but this one is still a `NumberFormatException` - they should probably be consistent with each other

Alexjsenn: I had issues throwing HttpStatusExceptions inside executeConnectedTask, so I changed the PATCH handler to match this one.

andrewazores: What kind of issues?

I think think the `NumberFormatException` is out of place when parsing a boolean.

Alexjsenn: Ok, I went back to the old format, and throw a HttpStatusException for the boolean instead of the NumberFormatException

Alexjsenn:
Also there are still some issues regarding maxAge, because I currently assume maxAge is set in seconds (and the default that the JVM reports back is also in seconds), but it seems that when the value is sent to the JVM, the values that are reported back for the recording info, the value is multiplied by 1000. So when I create a recording with maxAge 1s, when I list recordings it says that recording's maxAge = 1000. I don't know if that is just the convention it uses when sending back the recording information, but its confusing.

andrewazores:
I think that's fine. Internally the JFR implementation within the JVM probably only understands durations as milliseconds, so that's what's seen in the response after creating a recording. We should clean that up later on, but it's a separate concern and out of scope of this PR.

andrewazores (approved):

PR #260, opened 2020-09-08
Set port mapping at pod creation time

### LINKED ISSUE #259: Having Trouble Running `run.sh`
I've been having trouble running `run.sh` for a while. I thought maybe after some time it would fix itself, but it looks like that's not the case. It's definitely on my end though, since commits where it was previously working no longer work for me.

```
[vma@victor-work ~/upstream-cjfr]$ ./run.sh
...
+ podman run --pod container-jfr --mount type=tmpfs,target=/flightrecordings --mount type=tmpfs,target=/templates --mount type=bind,source=/home/vma/upstream-cjfr/truststore,destination=/truststore,relabel=shared,bind-propagation=shared -p 9091:9091 -p 9090:9090 -p 8181:8181 -e CONTAINER_JFR_DISABLE_SSL= -e CONTAINER_JFR_DISABLE_JMX_AUTH= -e CONTAINER_JFR_LOG_LEVEL=ALL -e CONTAINER_JFR_RJMX_USER= -e CONTAINER_JFR_RJMX_PASS= -e CONTAINER_JFR_RJMX_PORT=9091 -e CONTAINER_JFR_CORS_ORIGIN= -e CONTAINER_JFR_WEB_HOST=0.0.0.0 -e CONTAINER_JFR_WEB_PORT=8181 -e CONTAINER_JFR_EXT_WEB_PORT=8181 -e CONTAINER_JFR_LISTEN_HOST=0.0.0.0 -e CONTAINER_JFR_LISTEN_PORT=9090 -e CONTAINER_JFR_EXT_LISTEN_PORT=9090 -e CONTAINER_JFR_AUTH_MANAGER=com.redhat.rhjmc.containerjfr.net.NoopAuthManager -e CONTAINER_JFR_ARCHIVE_PATH=/flightrecordings -e CONTAINER_JFR_TEMPLATE_PATH=/templates -e GRAFANA_DATASOURCE_URL=http://localhost:8080 -e GRAFANA_DASHBOARD_URL=http://localhost:3000/d/CazTnbSMz/ -e USE_LOW_MEM_PRESSURE_STREAMING= -e KEYSTORE_PATH= -e KEYSTORE_PASS= -e KEY_PATH= -e CERT_PATH= --rm -it quay.io/rh-jmc-team/container-jfr:latest
Error: invalid config provided: published or exposed ports must be defined when the pod is created: network cannot be configured when it is shared with a pod
+ cleanup
...
```

The only thing I can think of that has changed is my podman version (through normal fedora updates). With podman 2.0.5, it was not working, and right now I just updated to 2.0.6, which is also not working. From IRC logs, 2.0.2 was definitely working. I think 2.0.5 is the one where it broke, but I'm not sure.

I haven't been able to find older `2.x.x` versions of podman in the fedora repos or elsewhere online so I'm not able to downgrade to test if the podman version is indeed the problem.

`podman ps` and `podman pod list` both return empty.

Integration tests seem to work fine.

andrewazores:
Fixes #259

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -4,6 +4,7 @@ set -x
 set -e
 
 function cleanup() {
+    podman pod kill container-jfr
     podman pod rm container-jfr
 }
 trap cleanup EXIT
@@ -55,7 +56,12 @@ if [ ! -d "$PWD/truststore" ]; then
 fi
 
 if ! podman pod exists container-jfr; then
-    podman pod create --hostname container-jfr --name container-jfr --publish $CONTAINER_JFR_EXT_WEB_PORT
+    podman pod create \
+        --hostname container-jfr \
+        --name container-jfr \
+        --publish $CONTAINER_JFR_RJMX_PORT:$CONTAINER_JFR_RJMX_PORT \
+        --publish $CONTAINER_JFR_EXT_LISTEN_PORT:$CONTAINER_JFR_LISTEN_PORT \
+        --publish $CONTAINER_JFR_EXT_WEB_PORT:$CONTAINER_JFR_WEB_PORT
 fi
 
 podman run \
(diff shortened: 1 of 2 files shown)


Alexjsenn (approved):

vic-ma (approved):
Working with Podman `3.0.4` and `3.0.6` for me.

PR #261, opened 2020-09-09
ITests use HTTP API where applicable

andrewazores:
- Replace Command Channel calls with HTTP equivalents
- Refactor to extract utility method

Avoids use of deprecated WebSocket/Command Channel APIs, replacing with HTTP API equivalents wherever applicable


vic-ma (approved):
Looks good.

PR #266, opened 2020-09-16
Add HTTP API Documentation

### LINKED ISSUE #240: HTTP API documentation
Documentation similar to `COMMANDS.md` should be created, outlining the HTTP API (any concrete class implementing `RequestHandler`). For each such handler the documentation should include:

- a brief synopsis of the purpose of the endpoint
- its HTTP method
- its path and an explanation of any path parameters
- a description of any query parameters, such as the expected format and whether they are required
- a description of any PUT/POST/PATCH request bodies (text format, file upload name, etc.)
- a description of the response body format
- descriptions of any response status codes other than `2xx`

Ideally, the documentation should be split into a main index document and a sub-document for specifically `API v1`, so that the document can be neatly extended later for `v2` and on.

This may initially be done by hand in some some human-readable markup format, but even better would be to implement this by using an API documentation generation tool. Here's one such tool: https://github.com/outofcoffee/vertx-oas

And even better would be to flip the script and define the API surface first, then generate implementation stubs from that, and finally fill in the existing implementations back into the stubs. Here's a document to help get started down that path: https://vertx.io/docs/vertx-web-api-contract/java/#_openapi_3. To create the API document, use a tool like [Swagger Editor](https://editor.swagger.io). However, this is probably a larger, later enhancement for a `v2+` API.

vic-ma:
Fixes #240.

### DIFF
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -275,6 +275,7 @@
     `DELETE /api/v1/recordings/:recordingName`
 
     `recordingName` - The name of the saved recording to delete.
+    Should use percent-encoding.
 
     ###### response
     `200` - No body.
@@ -308,6 +309,7 @@
     `GET /api/v1/recordings/:recordingName`
 
     `recordingName` - The name of the saved recording to get.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is an octet stream consisting of the requested recording.
@@ -450,6 +452,7 @@
     `POST /api/v1/recordings/:recordingName/upload`
 
     `recordingName` - The name of the saved recording to upload.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is the body of the response that Container JFR got
@@ -491,6 +494,7 @@
     `GET /api/v1/reports/:recordingName`
 
     `recordingName` - The name of the recording to get the report for.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is the requested report, as an HTML document.
@@ -526,6 +530,7 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is a JSON array of event type objects.
@@ -567,8 +572,10 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     `recordingName` - The name of the recording to delete.
+    Should use percent-encoding.
 
     ###### response
     `200` - No body.
@@ -602,8 +609,10 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     `recordingName` - The name of the recording to get.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is an octet stream consisting of the requested recording.
@@ -649,8 +658,10 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     `recordingName` - The name of the recording to patch.
+    Should use percent-encoding.
 
     The body must be either `STOP`, to stop the recording,
     or `SAVE`, to save the recording (case insensitive).
@@ -704,6 +715,7 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is a JSON array of recording objects.
@@ -740,10 +752,12 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     **The request must include the following fields:**
 
     `recordingName` - The name of the recording to create.
+    Should use percent-encoding.
 
     `events` - The events configuration for the recording.
     This can be a comma-seperated list of events, with each event having the
@@ -795,8 +809,10 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     `recordingName` - The name of the recording to upload.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is the body from the response that Container JFR got
@@ -840,8 +856,10 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     `recordingName` - The name of the recording to get the report for.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is the requested report, as an HTML document.
@@ -880,6 +898,7 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is the name of the recording.
@@ -913,6 +932,7 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     `templateName` - The name of the template to get.
 
@@ -953,6 +973,7 @@
 
     `targetId` - The location of the target JVM to connect to,
     in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+    Should use percent-encoding.
 
     ###### response
     `200` - The body is a JSON array of template objects.
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -45,7 +45,7 @@
 * #### `AuthPostHandler`
 
     ###### synopsis
-    Attempts user authenticaiton;
+    Attempts user authentication;
     used as a simple way to check a user authentication header.
 
     ###### request
@@ -138,7 +138,7 @@
 * #### `HealthGetHandler`
 
     ###### synopsis
-    Returns whether or not the Grafana datasource and Grafana dashbaord
+    Returns whether or not the Grafana datasource and Grafana dashboard
     that Container JFR is configured with are running properly.
 
     ###### request
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -52,12 +52,23 @@
     `POST /api/v1/auth`
 
     The request should include an `Authorization` header to be checked.
+    The format of this header depends on the auth manager used.
+    Token-based auth managers expect `Authorization: Bearer TOKEN`,
+    while basic credentials-based auth managers expect
+    `Authorization: Basic ${base64(user:pass)}`.
+    The `NoopAuthManager` accepts all authentication requests,
+    regardless of the existence or contents of any
+    `Authorization` header.
+    For more details, see
+    [`README.md`](https://github.com/rh-jmc-team/container-jfr#user-authentication--authorization)).
 
     ###### response
-    `200` - No body. Getting this response means that the header is valid.
+    `200` - No body. Getting this response means that the header is valid
+    and that the user has been successfully authenticated.
 
     `401` - User authentication failed. The body is an error message.
-    Getting this response means that the header is invalid.
+    Getting this response means that the header has an invalid format
+    or the user has not been successfully authenticated.
 
     `500` - There was an unexpected error. The body is an error message.
 
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -477,7 +477,7 @@
     `recordingName` - The name of the recording to get the report for.
 
     ###### response
-    `200` - The body is the requested report.
+    `200` - The body is the requested report, as an HTML document.
 
     `401` - User authentication failed. The body is an error message.
 
@@ -812,7 +812,7 @@
     `recordingName` - The name of the recording to get the report for.
 
     ###### response
-    `200` - The body is the requested report.
+    `200` - The body is the requested report, as an HTML document.
 
     `401` - User authentication failed. The body is an error message.
 
@@ -883,7 +883,7 @@
     `templateType` - The type of the template to get.
 
     ###### response
-    `200` - The body is the requested event template.
+    `200` - The body is the requested event template, as an XML document.
 
     `401` - User authentication failed. The body is an error message.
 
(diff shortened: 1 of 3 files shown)


vic-ma:
Oh whoops I forgot about the new handler Alex added.

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+| Download a report of a recording in persistent storage                    | [`ReportGetHandler`](#ReportGetHandler)                                 |
+| Upload a recording from persistent storage to the Grafana datasource      | [`RecordingUploadPostHandler`](#RecordingUploadPostHandler)             |
+
+
+### Core
+
+* #### `AuthPostHandler`
+
+    ###### synopsis
+    Attempts user authenticaiton;
```
andrewazores, line 48: typo: `authenticaiton`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    used as a simple way to check a user authentication header.
+
+    ###### request
+    `POST /api/v1/auth`
+
+    The request should include an `Authorization` header to be checked.
```
andrewazores, line 54: Should this explain the format of the `Authorization` header, ex. `Basic` vs `Bearer`? I think it's also worth including a short explanation of how the different `AuthManager`s influence the interpretation of this header - `README.md` goes into some detail on it already so the text here can be shorter, or even just refer back to the README.

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### response
+    `200` - No body. Getting this response means that the header is valid.
```
andrewazores, line 57: `that the header is valid and that the user has been successfully authenticated`, maybe?

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    `401` - User authentication failed. The body is an error message.
+    Getting this response means that the header is invalid.
```
andrewazores, line 60: `the header has an invalid format or the user has not been successfully authenticated`?

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    $ curl localhost:8181/api/v1/grafana_datasource_url
+    {"grafanaDatasourceUrl":"http://localhost:8080"}
+    ```
+
+
+* #### `HealthGetHandler`
+
+    ###### synopsis
+    Returns whether or not the Grafana datasource and Grafana dashbaord
+    that Container JFR is configured with are running properly.
```
andrewazores, line 142: It also implicitly tells the client whether ContainerJFR itself is running properly, if a response is received vs if it times out or otherwise fails.

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### synopsis
+    Scans for and returns discoverable target JVMs.
+
+    ###### request
+    `GET /api/v1/targets`
+
+    ###### response
+    `200` - The body consists of a comma-seperated list of targets,
+    enclosed in square brackets.
```
andrewazores, line 208: This `comma-separated list of targets, enclosed in square brackets` can be more concisely described as `a JSON array of target objects`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    <!doctype html><html lang="en-US"><head><meta charset="utf-8"><title>ContainerJFR</title><meta id="appName" name="application-name" content="ContainerJFR"><meta name="viewport" content="width=device-width,initial-scale=1"><link rel="icon" type="image/png" href="https://www.patternfly.org/components/patternfly/dist/img/favicon.ico"><base href="/"><link href="app.css" rel="stylesheet"></head><body><noscript>Enabling JavaScript is required to run this app.</noscript><div id="root"></div><script src="app.bundle.js"></script></body></html>
+    ```
+
+
+### Flight Recorder
+
+* #### `RecordingDeleteHandler`
+
+    ###### synopsis
+    Deletes a recording that was saved to persistent storage.
```
andrewazores, line 257: We use the term "archive" in some other places, particularly in the web-client, so perhaps it should also be included here alongside the more technical "persistent storage"?

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    The format for a target is
+    `{"connectUrl":"$CONNECT_URL","alias":"$ALIAS"}`.
+
+    So a list with no recordings looks like `[]`,
+    a list with one recording looks like `[{...}]`,
+    and a list with two recordings looks like `[{...},{...}]`.
+
+    `401` - User authentication failed. The body is an error message.
+
+    `407` - JMX authentication failed. The body is an error message.
```
andrewazores, line 219: I'm adding this note here but it applies to all `407`s, or maybe this needs a separate explanation. When a `407` is sent, the response will also include an `X-JMX-Authenticate` header, currently always with the value `Basic` but in the future this could also be `Bearer` or other authentication schemes. This indicates to the client the form of authentication that is expected/accepted.

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                     Dload  Upload   Total   Spent    Left  Speed
+    100  942k  100  942k    0     0  61.3M      0 --:--:-- --:--:-- --:--:-- 61.3M
+    ```
+
+
+* #### `RecordingsGetHandler`
+
+    ###### synopsis
+    Returns a list of the recordings that are saved in persistent storage.
```
andrewazores, line 318: Same re: `archive`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### request
+    `GET /api/v1/recordings`
+
+    ###### response
+    `200` - The body consists of a comma-seperated list of recordings,
+    enclosed in square brackets.
```
andrewazores, line 325: Same re: `JSON array`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    ```
+    $ curl localhost:8181/api/v1/recordings
+    [{"downloadUrl":"http://192.168.0.109:8181/api/v1/recordings/localhost_foo_20200903T202547Z.jfr","name":"localhost_foo_20200903T202547Z.jfr","reportUrl":"http://192.168.0.109:8181/api/v1/reports/localhost_foo_20200903T202547Z.jfr"}]
+    ```
+
+
+* #### `RecordingsPostHandler`
+
+    ###### synopsis
+    Uploads a recording from the client to Container JFR's persistent storage.
```
andrewazores, line 353: Same re: `archive`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### request
+    `POST /api/v1/recordings`
+
+    The recording should be uploaded in a form with the name `recording`.
+    The filename of the recording must follow the format that Container JFR
+    uses for recordings it saves itself
```
andrewazores, line 360: I think the format deserves a brief explanation here - for an inexperienced user, the example filename `localhost_foo_20200903T202547Z.jfr` may be non-obvious.

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    (with [`TargetRecordingPatchHandler`](#TargetRecordingPatchHandler)).
+
+    ###### response
+    `200` - The body is `{"name":"$NAME"}`, where `$NAME` is the name of the
+    recording that is now saved in persistent storage.
+    This name will be different from the uploaded recording's filename
+    if another recording with the same name already existed.
```
andrewazores, line 367: Please specify how the filename is mutated when this collision-avoidance is performed

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### request
+    `GET /api/v1/targets/:targetId/events`
+
+    `targetId` - The location of the target JVM to connect to,
+    in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+
+    ###### response
+    `200` - The body consists of a comma-seperated list of event types,
+    enclosed in square brackets.
```
andrewazores, line 470: Same re: `JSON array`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    ###### synopsis
+    Deletes a recording from a target JVM, removing it from the target JVM
+    and freeing the buffer memory used. The recording will be stopped if
+    it is running at the time of deletion.
+
+    ###### request
+    `DELETE /api/v1/targets/:targetId/recordings/:recordingName`
+
+    `targetId` - The location of the target JVM to connect to,
+    in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
```
andrewazores, line 508: This applies to all paths with `:targetId` and `:recordingName` - just to be safe, perhaps we should specify that these path components should be URI-encoded? Otherwise the resulting string-interpolated URL may be invalid or may be interpreted differently than the client intended.

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### request
+    `GET /api/v1/targets/:targetId/recordings`
+
+    `targetId` - The location of the target JVM to connect to,
+    in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+
+    ###### response
+    `200` - The body consists of a comma-seperated list of recordings,
+    enclosed in square brackets.
```
andrewazores, line 637: Same re: `JSON array`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    The format for a recording is
+    `{"downloadUrl":"$DOWNLOAD_URL","name":"$NAME","reportUrl":"$REPORT_URL"}`.
```
andrewazores, line 640: I think this should include all of the attributes of a recording, so including `state`, `startTime`, `duration`, etc., no?

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    `404` - The report could not be found, or the target could not be found.
+    The body is an error message.
+
+    `407` - JMX authentication failed. The body is an error message.
+
+    `500` - There was an unexpected error. The body is an error message.
+
+    ###### example
+    ```
+    [vma@victor-work ~]$ curl localhost:8181/api/v1/targets/localhost/reports/foo --output report.html
```
andrewazores, line 780: I think the shell prompt stuff here `[vma@victor-work ~]` should be removed for consistency with the rest of the document

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+    ###### request
+    `GET /api/v1/targets/:targetId/reports/:recordingName`
+
+    `targetId` - The location of the target JVM to connect to,
+    in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+
+    `recordingName` - The name of the recording to get the report for.
+
+    ###### response
+    `200` - The body is the requested report.
```
andrewazores, line 767: `as an HTML document`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    `targetId` - The location of the target JVM to connect to,
+    in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+
+    `templateName` - The name of the template to get.
+
+    `templateType` - The type of the template to get.
+
+    ###### response
+    `200` - The body is the requested event template.
```
andrewazores, line 834: `as an XML document`

HTTP_API.md
```diff
@@ -0,0 +1,952 @@
...
+
+    ###### request
+    `GET /api/v1/targets/:targetId/templates`
+
+    `targetId` - The location of the target JVM to connect to,
+    in the form of a `service:rmi:jmx://` JMX Service URL, or `hostname:port`.
+
+    ###### response
+    `200` - The body consists of a comma-seperated list of templates,
+    enclosed in square brackets.
```
andrewazores, line 867: Same re: `JSON array`

andrewazores (commented):
Great work overall, I got pretty nit-picky with it since we really need this to be as technically precise as possible but I'm generally quite pleased with how this turned out.

HTTP_API.md
vic-ma: Mm so what you mean by this is that if the handler returns a `500` or if it times out (would timing out also give a 500?), then something is probably wrong with cjfr because it would mean that cjfr isn't sending/receiving HTTP messages properly?

HTTP_API.md
vic-ma: Yeah I wasn't really sure which to use, since I saw "persistent storage" in COMMANDS.md but "archive" elsewhere.
But on second thought, I see that "archive" is the word that's used in the README as well, so maybe we should just skip "persistent storage" altogether? Since anyone wondering about "archive" would be able to find what it is easily that way, and keeping the language consistent with the README is probably best.

HTTP_API.md
vic-ma: Do the `401` need a similar note, or is it only because this is a custom header?

HTTP_API.md
vic-ma: Could you explain a bit what URI-encoded means? I wasn't able to look that up.

HTTP_API.md
vic-ma: Yeah I think I just copied it from the non-targeted version not realizing it's different.

HTTP_API.md
vic-ma: Whoops, yeah that should've been removed. I actually found another one hidden at the end of the example for `TargetTemplatesGetHandler`!

HTTP_API.md
andrewazores: No, if the requesting client gets no response by the end of the timeout period, there would be no response status at all, so no `500` code to interpret. My point here is that if the client gets no response or gets some kind of HTTP(S) request failure, they can also discover information about the status or reachability of ContainerJFR - maybe it is not actually running, or the container is up but somehow the Java process has died or is not responding, or there is a network misconfiguration, etc.

andrewazores: Of course, the same conclusions can also be reached if the client sees such failure modes on any path, not just `/health`, but `/health` in particular is a good candidate for this kind of "liveness probe" (https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#when-should-you-use-a-liveness-probe) for clients to check if the container is running properly.

HTTP_API.md
andrewazores: Sure, using only "archive" sounds fine too.

HTTP_API.md
andrewazores: https://www.w3schools.com/tags/ref_urlencode.ASP / https://en.wikipedia.org/wiki/Percent-encoding

andrewazores: ie clients should not try to request ex. `DELETE /api/v1/targets/service:jmx:rmi:///jndi/rmi://container-jfr:9091/jmxrmi/recordings/foo`, but rather `DELETE /api/v1/targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcontainer-jfr%3A9091%2Fjmxrmi%2Frecordings%2Ffoo/recordings/foo`

HTTP_API.md
andrewazores: Yea, this is a custom header so it needs additional documentation. There is a similar dynamic with the `401 / Authenticate / Authorization` which I guess we should also document, since we use a custom `X-WWW-Authenticate` instead of the standard `WWW-Authenticate`.

andrewazores (approved):

PR #271, opened 2020-09-17
Use status code 427 instead of 407

### LINKED ISSUE #270: Client "Failed to fetch"
When running recent builds with JMX Auth and SSL enabled, the web-client fails to request various resources in Chromium browsers, displaying a notification saying "Request failed: Failed to fetch". In the browser console there is a "ERR_UNEXPECTED_PROXY_AUTH" error message.

This seems to work as expected in Firefox.

Possible cause:

https://stackoverflow.com/questions/11697943/when-should-one-use-connect-and-get-http-methods-at-http-proxy-server/40329026

Or perhaps Chrome responds to the `407` response code this way, whereas Firefox allows the web-client to handle the auth failure.

andrewazores:
Fixes #270

See also rh-jmc-team/container-jfr-web#129


andrewazores:
This can be tested with `CONTAINER_JFR_IMAGE=quay.io/andrewazores/container-jfr:0.21.0 sh smoketest.sh`.

vic-ma:
If I have SSL enabled, what extra steps do I have to take to get the web client working in my browser? Since right now, it doesn't load for me when I turn it on.

vic-ma (approved):

Alexjsenn (approved):

vic-ma (approved):
Works now with built-in web client.

Alexjsenn (approved):

PR #285, opened 2020-10-05
Add Snapshot API V2 handler

### LINKED ISSUE #267: Snapshot API response is not JSON-encoded
See #266

According to our (upcoming) new API docs, the response from the Snapshot HTTP API handler is just the name of the snapshot in plain text. The rest of the API is generally JSON-encoded for `2xx` status responses (or an empty body), so this should be JSON-encoded to match.

andrewazores:
Fixes #267

Encodes response as JSON with the same properties as customized recording creation

### DIFF
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -1169,3 +1169,50 @@
     ```
     $ curl --form "template=@Foo.jfc" localhost:8181/api/v1/templates
     ```
+
+## V2 API
+
+### Quick Reference
+
+| What you want to do                                                       | Which handler you should use                                                |
+| ------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
+| **Recordings in target JVMs**                                             |                                                                             |
+| Create a snapshot recording in a target JVM                               | [`TargetSnapshotPostHandler`](#TargetSnapshotPostHandler)                   |
+
+### Flight Recorder
+
+* #### `TargetSnapshotPostHandler`
+
+    ###### synopsis
+    Creates a recording named `snapshot-n`, where `n` is a sequentially
+    assigned ID, which contains information about all events recorded
+    across all active recordings at the time of invocation.
+
+    ###### request
+    `POST /api/v2/targets/:targetId/snapshot`
+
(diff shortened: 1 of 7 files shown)


andrewazores:
This breaks the anchor links in `HTTP_API.md` since there are now two handlers by the same name, performing generally the same role, but for API V1/V2. IMO the document should probably be split per version to avoid problems like this and prevent it from growing indefinitely. This problem would also most likely be solved by using an automated/generated API specification tool.

HTTP_API.md
```diff
@@ -1169,3 +1169,58 @@
...
     ```
+
+## V2 API
+
+### Quick Reference
+
+| What you want to do                                                       | Which handler you should use                                                |
+| ------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
+| **Recordings in target JVMs**                                             |                                                                             |
+| Create a snapshot recording in a target JVM                               | [`TargetSnapshotPostHandler`](#TargetSnapshotPostHandler)                   |
```
vic-ma, line 1180: >This breaks the anchor links in HTTP_API.md since there are now two handlers by the same name, performing generally the same role, but for API V1/V2. 

I tried clicking on the anchor button next to the header and it looks like github adds a counter for duplicate names. So I tested it using `[...](#TargetSnapshotPostHandler-1)` and it looks like that works.

>IMO the document should probably be split per version to avoid problems like this and prevent it from growing indefinitely.

Yeah, I think that's a good idea too. If V2 had the same amount of handlers as V1, I think that would make the document really hard to use and navigate for whichever version was at the bottom.

vic-ma (commented):
Code looks good and runs as expected; just a few comments for the documentation.

HTTP_API.md
andrewazores: re: anchor counts - cool, that's very neat

re: doc splitting - alright, that's probably a topic for a separate PR. I'll leave it as-is tagged on the end for now and then we can figure out the best way to split it up

vic-ma (approved):

PR #299, opened 2020-10-15
Upload certificate

### LINKED ISSUE #273: Allow clients to upload trusted target SSL certs
Currently the only way for ContainerJFR to gain trust for SSL-enabled targets is for the target's cert to be added to ContainerJFR's `/truststore` directory, which must be done before ContainerJFR startup (ie by mounting the pre-populated volume to the container). It should be possible for ContainerJFR to gain trust for a target/cert at runtime so that ContainerJFR does not need to be restarted to accommodate new targets/certs. This can probably be achieved by a `POST` handler that accepts, validates, and writes certs to `/truststore`.

Maybe useful SO link https://stackoverflow.com/questions/18889058/programmatically-import-ca-trust-cert-into-existing-keystore-file-without-using

Alexjsenn:
Fixes #273 
A few things that may need attention:
* The status code 409 for certificate already exists (not sure if that is the most appropriate)
* I am not sure if the fact that a restart of the JVM is needed should be communicated anywhere (I guess it would be added to the docs and in the future web UI)
* When uploading the certificate, it is in a form with the name cert (it checks for this). This might not be the best name or even the most appropriate way to send it. (current curl command is `curl -X POST -F cert=@PATH/vertx-fib-demo.cer -vk https://localhost:8181/api/v2/certificates`)

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
@@ -42,12 +42,13 @@
 package com.redhat.rhjmc.containerjfr.net.security;
 
 import java.io.InputStream;
+import java.security.cert.Certificate;
 import java.security.cert.CertificateFactory;
 import java.util.Collection;
 
 public class CertificateValidator {
 
-    public Collection parseCertificate(InputStream stream) throws Exception {
+    public Collection<Certificate> parseCertificates(InputStream stream) throws Exception {
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
         return cf.generateCertificates(stream);
     }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
@@ -48,7 +48,8 @@ import java.util.Collection;
 
 public class CertificateValidator {
 
-    public Collection<Certificate> parseCertificates(InputStream stream) throws Exception {
+    public Collection<? extends Certificate> parseCertificates(InputStream stream)
+            throws Exception {
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
         return cf.generateCertificates(stream);
     }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
@@ -47,7 +47,7 @@ import java.security.cert.CertificateFactory;
 
 public class CertificateValidator {
 
-    public Certificate verify(ByteArrayInputStream byteStream) throws Exception {
+    public Certificate parseCertificate(ByteArrayInputStream byteStream) throws Exception {
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
         return cf.generateCertificate(byteStream);
     }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
@@ -41,7 +41,6 @@
  */
 package com.redhat.rhjmc.containerjfr.net.security;
 
-
 import java.io.ByteArrayInputStream;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateFactory;
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
@@ -41,14 +41,14 @@
  */
 package com.redhat.rhjmc.containerjfr.net.security;
 
-import java.io.ByteArrayInputStream;
-import java.security.cert.Certificate;
+import java.io.InputStream;
 import java.security.cert.CertificateFactory;
+import java.util.Collection;
 
 public class CertificateValidator {
 
-    public Certificate parseCertificate(ByteArrayInputStream byteStream) throws Exception {
+    public Collection parseCertificate(InputStream stream) throws Exception {
         CertificateFactory cf = CertificateFactory.getInstance("X.509");
-        return cf.generateCertificate(byteStream);
+        return cf.generateCertificates(stream);
     }
 }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
@@ -39,54 +39,17 @@
  * SOFTWARE.
  * #L%
  */
-package com.redhat.rhjmc.containerjfr.net.web.http.api.v2;
+package com.redhat.rhjmc.containerjfr.net.security;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.util.function.Function;
 
-import javax.inject.Named;
-import javax.inject.Singleton;
+import java.io.ByteArrayInputStream;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
 
-import com.redhat.rhjmc.containerjfr.net.security.CertificateValidator;
-import com.redhat.rhjmc.containerjfr.net.web.http.RequestHandler;
+public class CertificateValidator {
 
-import dagger.Binds;
-import dagger.Module;
-import dagger.Provides;
-import dagger.multibindings.IntoSet;
-
-@Module
-public abstract class HttpApiV2Module {
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindTargetSnapshotPostHandler(TargetSnapshotPostHandler handler);
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindCertificatePostHandler(CertificatePostHandler handler);
-
-    @Provides
-    @Singleton
-    @Named("OutputStreamFunction")
-    static Function<File, FileOutputStream> provideOutputStreamFunction() throws RuntimeException {
-        return (File file) -> {
-            try {
-                return new FileOutputStream(file);
-            } catch (IOException e) {
-                throw new RuntimeException(e);
-            }
-        };
+    public Certificate verify(ByteArrayInputStream byteStream) throws Exception {
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        return cf.generateCertificate(byteStream);
     }
-
-    @Provides
-    static CertificateValidator provideCertificateValidator() {
-        return new CertificateValidator();
-    }
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindCertificatePostBodyHandler(CertificatePostBodyHandler handler);
 }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
@@ -118,7 +118,7 @@ class CertificatePostHandler extends AbstractAuthenticatedRequestHandler {
         }
 
         if (cert == null) {
-            throw new HttpStatusException(400, "No certificate found");
+            throw new HttpStatusException(400, "A file named \"cert\" was not included in the request");
         }
 
         String certPath = fs.pathOf(cert.uploadedFileName()).normalize().toString();
@@ -128,29 +128,26 @@ class CertificatePostHandler extends AbstractAuthenticatedRequestHandler {
         }
 
         String truststoreDir = env.getEnv(TRUSTSTORE_DIR);
-        String filePath = truststoreDir + "/" + cert.fileName();
+        String filePath = fs.pathOf(truststoreDir, cert.fileName()).normalize().toString();
         if (fs.exists(fs.pathOf(filePath))) {
-            throw new HttpStatusException(409, "Certificate already exists");
+            throw new HttpStatusException(409, filePath + " Certificate already exists");
         }
 
-        try {
+        File certFile = new File(filePath);
+
+        try (FileInputStream fis = new FileInputStream(certPath);
+                DataInputStream dis = new DataInputStream(fis);
+                FileOutputStream out = new FileOutputStream(certFile)) {
             CertificateFactory cf = CertificateFactory.getInstance("X.509");
 
-            FileInputStream fis = new FileInputStream(certPath);
-            DataInputStream dis = new DataInputStream(fis);
             byte[] bytes = new byte[dis.available()];
             dis.readFully(bytes);
-            dis.close();
-            fis.close();
+
             ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes);
             Certificate certificate = cf.generateCertificate(bytestream);
             byte[] buf = certificate.getEncoded();
 
-            File certFile = new File(filePath);
-            FileOutputStream out = new FileOutputStream(certFile);
             out.write(buf);
-            out.close();
-
         } catch (Exception e) {
             throw new HttpStatusException(500, e.getMessage());
         }
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
@@ -144,22 +144,20 @@ class CertificatePostHandler extends AbstractAuthenticatedRequestHandler {
         }
 
         String truststoreDir = env.getEnv(TRUSTSTORE_DIR);
-        String filePath = fs.pathOf(truststoreDir, cert.fileName()).normalize().toString();
+        Path filePath = fs.pathOf(truststoreDir, cert.fileName()).normalize();
 
-        if (fs.exists(fs.pathOf(filePath))) {
-            throw new HttpStatusException(409, filePath + " Certificate already exists");
+        if (fs.exists(filePath)) {
+            throw new HttpStatusException(409, filePath.toString() + " Certificate already exists");
         }
-
-        File certFile = new File(filePath);
 
         try (InputStream fis = fs.newInputStream(certPath);
                 DataInputStream dis = new DataInputStream(fis);
-                FileOutputStream out = outputStreamFunction.apply(certFile)) {
+                FileOutputStream out = outputStreamFunction.apply(filePath.toFile())) {
 
             byte[] bytes = new byte[dis.available()];
             dis.readFully(bytes);
             ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes);
-            Certificate certificate = certValidator.verify(bytestream);
+            Certificate certificate = certValidator.parseCertificate(bytestream);
             byte[] buf = certificate.getEncoded();
 
             out.write(buf);
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
@@ -156,11 +156,12 @@ class CertificatePostHandler extends AbstractAuthenticatedRequestHandler {
 
             byte[] bytes = new byte[dis.available()];
             dis.readFully(bytes);
-            ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes);
-            Certificate certificate = certValidator.parseCertificate(bytestream);
-            byte[] buf = certificate.getEncoded();
 
-            out.write(buf);
+            try (ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes)) {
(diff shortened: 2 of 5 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        FileUpload cert = null;
+        for (FileUpload fu : ctx.fileUploads()) {
+            if ("cert".equals(fu.name())) {
+                cert = fu;
+                break;
+            }
+        }
+
+        if (cert == null) {
+            throw new HttpStatusException(400, "No certificate found");
```
andrewazores, line 121: Maybe this should be more direct about the problem, like `No file named \"cert\" uploaded`?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        String certPath = fs.pathOf(cert.uploadedFileName()).normalize().toString();
+
+        if (!env.hasEnv(TRUSTSTORE_DIR)) {
+            throw new HttpStatusException(500, "Truststore directory not set");
+        }
+
+        String truststoreDir = env.getEnv(TRUSTSTORE_DIR);
+        String filePath = truststoreDir + "/" + cert.fileName();
```
andrewazores, line 131: Better to avoid manually constructing paths like this. `fs.pathOf` takes varargs, so you should just be able to do `fs.exists(fs.pathOf(truststoreDir, cert.fileName()))`

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        if (fs.exists(fs.pathOf(filePath))) {
+            throw new HttpStatusException(409, "Certificate already exists");
+        }
+
+        try {
```
andrewazores, line 136: This looks like it can/should be refactored to use try-with-resources:

```
try (DataInputStream dis = new DataInputStream(new FileInputStream(certPath))) {
// do stuff
}
```

This is assuming that these resources are autocloseable, but I would expect they are. Generally anything that is autocloseable should be wrapped in a try-with-resources, since it guarantees that the resource is properly closed, even if an exception is thrown or there is an early return statement or any other exit condition.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,160 @@
...
+        String certPath = fs.pathOf(cert.uploadedFileName()).normalize().toString();
+
+        if (!env.hasEnv(TRUSTSTORE_DIR)) {
+            throw new HttpStatusException(500, "Truststore directory not set");
+        }
+
+        String truststoreDir = env.getEnv(TRUSTSTORE_DIR);
+        String filePath = truststoreDir + "/" + cert.fileName();
+        if (fs.exists(fs.pathOf(filePath))) {
+            throw new HttpStatusException(409, "Certificate already exists");
```
andrewazores, line 133: May as well include the name of the conflicting certificate here, just for ease of debugging if this is showing up in ex. Operator logs or something.

andrewazores (commented):
Is there any refactoring you can do to the main handler here to make it testable, and then include some tests?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
Alexjsenn: How can it distinguish between the file name and the form entry? because cert is the entry, not the file name right?

andrewazores: The file uploads is something like a map of name to file contents, so what I mean here is that the literal string "cert" is not a key in the map. That's what `cert == null` is basically testing here already, right?

Alexjsenn: Yes, exactly, I guess what I thought was confusing was that the key in the map is not the same as the name of the uploaded file, but I guess it makes sense.

andrewazores: No, it's the value in the map that is the name of the file. We should just tell the user which key they missed so that they can correct the response by including it and retrying.

Alexjsenn: Yea, makes sense.

Alexjsenn:
I will look at creating some tests, what do you mean by refactoring to make it testable?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+        try (FileInputStream fis = new FileInputStream(certPath);
+                DataInputStream dis = new DataInputStream(fis);
+                FileOutputStream out = new FileOutputStream(certFile)) {
+            CertificateFactory cf = CertificateFactory.getInstance("X.509");
+
+            byte[] bytes = new byte[dis.available()];
+            dis.readFully(bytes);
+
+            ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes);
```
andrewazores, line 146: I think `ByteArrayInputStream` must also be closed, and is `AutoCloseable`.

andrewazores:
> I will look at creating some tests, what do you mean by refactoring to make it testable?

Can you write a test to exercise your implementation? I think you will run into difficulties writing such a test because the implementation is currently tightly coupled to things like `new FileInputStream(certPath)` - unit tests should not be creating actual file resources, so you would need to create a mock that simulates a `File` well enough to be read by that `FileInputStream`, which is tedious and error-prone. If you could refactor the implementation to use something like a `Provider<FileInputStream>` then the test could also supply the implementation with a mock `FileInputStream`, for example, significantly simplifying the test setup.

Alexjsenn:
There is a strange bug that spotbug finds, that occurs on line 163, where apparently fileOutputStream::write does a null check that is redundant. I am not sure exactly how this would be fixed.

andrewazores:
I've hit that before and last I checked it's actually a bug with SpotBugs itself and try-with-resources.

```
$ rg --ignore-case redundant src
src/main/java/com/redhat/rhjmc/containerjfr/net/NetworkResolver.java
106:    // try-with-resources generates a "redundant" nullcheck in bytecode
107:    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/TargetRecordingGetHandler.java
126:    // try-with-resources generates a "redundant" nullcheck in bytecode
127:    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
```

Looks like a fix has been commited in upstream SpotBugs, but I'm not sure if there's a new release with it yet: https://github.com/spotbugs/spotbugs/issues/756

If not then simply applying the same supression annotation is fine for now.

Alexjsenn:
I think there is a newer version of spotbugs, I would have to update in both core and cjfr right?

andrewazores:
They don't necessarily have to be on the same version but it only makes sense to update both. That's probably for a separate PR though. For now just apply that suppression annotation and then the Spotbugs version can be updated later and the annotations removed again, if that version removes the need for them.

Alexjsenn:
Ok, makes sense!

src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package com.redhat.rhjmc.containerjfr.net.security;
+
+import java.io.ByteArrayInputStream;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+
+public class CertificateValidator {
+
+    public Certificate verify(ByteArrayInputStream byteStream) throws Exception {
```
andrewazores, line 50: `verify` doesn't seem like the right name given the argument types and return type, even though it does reflect what the caller is currently using the method for. `certificateFromBytes` or something might be a more accurate/descriptive name.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        Path certPath = fs.pathOf(cert.uploadedFileName());
+
+        if (!env.hasEnv(TRUSTSTORE_DIR)) {
+            throw new HttpStatusException(500, "Truststore directory not set");
+        }
+
+        String truststoreDir = env.getEnv(TRUSTSTORE_DIR);
+        String filePath = fs.pathOf(truststoreDir, cert.fileName()).normalize().toString();
```
andrewazores, line 147: I think `filePath` can be left as type `Path` here, removing the `.toString()`...

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+        if (fs.exists(fs.pathOf(filePath))) {
```
andrewazores, line 149: ... which would remove the need for the repeated `fs.pathOf` here...

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,172 @@
...
+            throw new HttpStatusException(409, filePath + " Certificate already exists");
+        }
+
+        File certFile = new File(filePath);
```
andrewazores, line 153: ... and here, `filePath.toFile()` would achieve the same thing if you really need it as a `File`.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+            byte[] bytes = new byte[dis.available()];
+            dis.readFully(bytes);
+            ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes);
+            Certificate certificate = certValidator.verify(bytestream);
+            byte[] buf = certificate.getEncoded();
+
+            out.write(buf);
+        } catch (Exception e) {
+            throw new HttpStatusException(500, e.getMessage());
```
andrewazores, line 167: This `catch` clause should probably be removed - the parent `AbstractAuthenticatedRequestHandler` class already has handling for exceptions thrown by subclasses, including general Exceptions which get wrapped into `500`s similar to what has been repeated here. IMO it would be better to allow the abstract parent class to handle this so that the behaviour across all concrete implementations is uniform.

Alexjsenn: Yes, that is definitely better!

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+        try (InputStream fis = fs.newInputStream(certPath);
+                DataInputStream dis = new DataInputStream(fis);
+                FileOutputStream out = outputStreamFunction.apply(filePath.toFile())) {
+
+            byte[] bytes = new byte[dis.available()];
+            dis.readFully(bytes);
+            ByteArrayInputStream bytestream = new ByteArrayInputStream(bytes);
+            Certificate certificate = certValidator.parseCertificate(bytestream);
+            byte[] buf = certificate.getEncoded();
```
andrewazores, line 161: `buf` seems unnecessary - if `parseCertificate` is just being used to validate that `bytestream` represents a well-formed certificate, can't we do `out.write(bytestream)` and avoid potentially re-encoding the `Certificate` back into a byte array, since we already have the byte array that encoded it to begin with?

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,169 @@
...
+
+        if (fs.exists(filePath)) {
+            throw new HttpStatusException(409, filePath.toString() + " Certificate already exists");
+        }
+
+        try (InputStream fis = fs.newInputStream(certPath);
+                DataInputStream dis = new DataInputStream(fis);
+                FileOutputStream out = outputStreamFunction.apply(filePath.toFile())) {
+
+            byte[] bytes = new byte[dis.available()];
```
andrewazores, line 157: Are you sure `dis.available()` tells you the total size of the certificate data in all scenarios? Not all streams implement `available()` in this way - it's more common that this method reports the number of bytes that can be read without blocking on the next invocation of `read()`. If this is the case, then for larger certificates (or various other conditions), this byte buffer will end up undersized and only some initial chunk of the certificate will be read into the buffer, rather than the whole thing.

Why is the `ByteArrayInputStream` wrapping around this buffer needed to begin with? Can't `fis` or `dis` be passed directly to `parseCertificate`? The docs for `generateCertificate` say it accepts a generic `InputStream`, not specifically a `ByteArrayInputStream`.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
andrewazores: I don't think so - if there were some additional garbage bytes then I should hope that parsing the buffer as a certificate should fail.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
Alexjsenn: It gives parsing errors when passing in the InputStream or DataInputStream. However I could add the byte array part into the parseCertificate method.

andrewazores: What "parsing errors"? Does it throw a `CertificateException`, and if so, does the exception say anything informative?

The official docs suggest that passing the `FileInputStream` wrapped in a `BufferedInputStream` should be sufficient: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/cert/CertificateFactory.html

ie.

```
 FileInputStream fis = new FileInputStream(filename);
 BufferedInputStream bis = new BufferedInputStream(fis);

 CertificateFactory cf = CertificateFactory.getInstance("X.509");

 while (bis.available() > 0) {
    Certificate cert = cf.generateCertificate(bis);
    System.out.println(cert.toString());
 }
```

and

```
 FileInputStream fis = new FileInputStream(filename);
 CertificateFactory cf = CertificateFactory.getInstance("X.509");
 Collection c = cf.generateCertificates(fis);
 Iterator i = c.iterator();
 while (i.hasNext()) {
    Certificate cert = (Certificate)i.next();
    System.out.println(cert);
 }
```

It would be much better if we can figure a way to generate the certificate more directly from a stream of the file, without first reading the whole file into an application memory buffer and making a stream out of that. Less bug prone and probably better overall throughput performance.

Alexjsenn: I changed the certificateValidator, it now takes in a InputStream, and outputs a Collection.

andrewazores:
Cool. I just tested this by removing the `vertx-fib-demo.cer` from `truststore` and running `smoketest.sh`, then doing a `GET /api/v1/targets/(service URL for localhost:9093 - the vertx-fib-demo)/recordings` and it failed, as expected. The status code is still a 404 which is not really correct, but that's a bug from before this PR.

Then I did a `POST` as you outlined in the PR body to re-upload the certificate into `truststore`. Just to verify, I retried the `GET` and it still failed the same way. Then I killed the `smoketest.sh` process and restarted it, at which point I saw `entrypoint.sh` processing the `vertx-fib-demo.cer` and adding it to the truststore. Then, trying the `GET` again responded with a success code and an empty recordings array, as expected/usual. So functionally this looks to be working as expected!

The `vertx-fib-demo.cer` looks like it's binary data - does this also work with base64-encoded certificates? This SO QA has some good details:

https://stackoverflow.com/questions/22743415/what-are-the-differences-between-pem-cer-and-der/22743616

So this `.cer` is probably really a `.der`. Can this handler also accept `.pem`?

src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package com.redhat.rhjmc.containerjfr.net.security;
+
+import java.io.InputStream;
+import java.security.cert.CertificateFactory;
+import java.util.Collection;
+
+public class CertificateValidator {
+
+    public Collection parseCertificate(InputStream stream) throws Exception {
```
andrewazores, line 50: Little nitpick: since this now returns a `Collection`, the method name should probably be update to reflect that. ie `parseCertificates`. Also, the raw type `Collection` could probably be written here as `Collection<Certificate>` for a bit of extra type-safety.

andrewazores: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/security/cert/CertificateFactory.html#generateCertificates(java.io.InputStream)

Actually, `Collection<? extends Certificate>` - makes sense as there are subtypes of `Certificate` as well, like `X509Certificate`.

src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
```diff
@@ -0,0 +1,167 @@
...
+        Path filePath = fs.pathOf(truststoreDir, cert.fileName()).normalize();
+
+        if (fs.exists(filePath)) {
+            throw new HttpStatusException(409, filePath.toString() + " Certificate already exists");
+        }
+
+        try (InputStream fis = fs.newInputStream(certPath);
+                FileOutputStream out = outputStreamFunction.apply(filePath.toFile())) {
+
+            Collection certificates = certValidator.parseCertificate(fis);
```
andrewazores, line 156: Likewise, this `Collection` and its `Iterator` are raw types here but are generics and so should be `Collection<? extends Certificate>` and `Iterator<? extends Certificate>`.

Alexjsenn: So according to the documentation for generateCertificates, the certificate must be DER-encoded, but it can be binary or base64. Therefore, since it seems .pem files are base64 DER-encoded, they also work. I tested it and it seemed to not have any issues saving it, as well as loading it in after a restart. However I can't test that the certificate still works for sure.

Alexjsenn:
So according to the documentation for generateCertificates, the certificate must be DER-encoded, but it can be binary or base64. Therefore, since it seems .pem files are base64 DER-encoded, they also work. I tested it and it seemed to not have any issues saving it, as well as loading it in after a restart. However I can't test that the certificate still works for sure.

andrewazores:
If you've generated a base64 certificate then you can modify the github.com/andrewazores/vertx-fib-demo pretty easily to use it. I think it should be as easy as replacing the `src/main/extras/app/resources/vertx-fib-demo.cer` with the new one you generated, then rebuilding the container (`./gradlew --no-daemon jibDockerBuild`) and re-running `smoketest.sh`.

src/main/java/com/redhat/rhjmc/containerjfr/net/security/CertificateValidator.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.InputStream;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+import java.util.Collection;
+
+public class CertificateValidator {
+
+    public Collection<Certificate> parseCertificates(InputStream stream) throws Exception {
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        return cf.generateCertificates(stream);
```
andrewazores, line 53: This is failing to compile for me - this line returns `Collection<? extends Certificate>` but the return type of `parseCertificates` is `Collection<Certificate>`. The method return type should be updated to match the bounded wildcard as we discussed over IRC.

Alexjsenn: Sorry about that, should be fixed.

andrewazores (approved):

PR #304, opened 2020-10-22
Discover Kube targets by Endpoints, not Services

### LINKED ISSUE #282: KubeApiPlatformClient should discover via Endpoints, not Services
The OpenShiftPlatformClient was previously updated to discover targets using Endpoints rather than Services. Services are not preferred since they act as load-balancers in front of a set of target application Pods, and so when connecting to a "target" through a Service, ContainerJFR will actually be connected to just one of the load-balanced Pods according to the load balancer. If a user starts a recording on one "target" and reconnects to it later, they may no longer see that recording since the reconnection attempt could give them a different target JVM with no distinction in the web-client or even API responses.

Endpoints are a Kubernetes concept, not OpenShift-specific, so the KubeApiPlatformClient should be updated to match this implementation.

andrewazores:
Fixes #282

This basically duplicates the existing `OpenShiftPlatformClient` implementation for reuse as the `KubeApiPlatformClient`. These use two different client libraries for the internal API clients, which also means different library implementations for the various Kubernetes types (Endpoints, Subsets, Ports, etc.). So although these types model the same Kubernetes datatypes, the types themselves are distinct hierarchies, and so the final concrete `PlatformClient` implementations are also distinct and do not inherit from each other (unfortunately).

This can be tested by setting the `CONTAINER_JFR_PLATFORM` env var to force the `KubeApiPlatformClient` into use and then running as usual in CRC. This should result in essentially no noticeable difference from running with the standard `OpenShiftPlatformClient`. Compared to the previous `KubeApiPlatformClient` implementation, only ports named `jfr-jmx` or numbered `9091` should cause `ServiceRef`s to be listed in the targets list. If a single `Service` has more than one such port, or if there are multiple `Pod`s (each distinguished by cluster-internal IP address) behind the `Service`, then there should be near-duplicate `ServiceRef`s for each of these, differentiated by port number or by IP address.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/platform/internal/KubeApiPlatformClient.java
@@ -41,6 +41,7 @@
  */
 package com.redhat.rhjmc.containerjfr.platform.internal;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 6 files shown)


ebaron (approved):
Looks good and works as advertised!

PR #308, opened 2020-10-23
Make Changes to Support SLF4J Change in Core Logger

vic-ma:
rh-jmc-team/container-jfr-core#49


andrewazores (commented):
`CONTAINER_JFR_LOG_LEVEL` is still referenced in `run.sh` and `pom.xml`. Looks good otherwise.

andrewazores (approved):

PR #316, opened 2020-10-27
WebServer IO cleanup and enhancements

### LINKED ISSUE #62: Investigate Vert.x webserver recording streaming performance
See previous discussion during review of https://github.com/rh-jmc-team/container-jfr/pull/57 .

The current implementation for streaming of in-memory recordings relies on using an intermediate buffer in container-jfr, which reads data out of the target JVM into a buffer, then asks the Vert.x webserver to copy the contents of that buffer into its internal write queue. Once all of the data has thus been copied out of the remote JVM and into the response write queue, the response begins to be flushed out the network stack. There should be some better way to stream the data through container-jfr and the Vert.x webserver so that less time is spend building up the intermediate copy of the flight recording, perhaps using a Vert.x Pump.

andrewazores:
This PR does some refactoring and cleanup, primary in the two handlers for downloading recordings (active from targets, and archived).

There should be no behavioural changes other than the removal of the `USE_LOW_MEM_PRESSURE_STREAMING` env var, which was never confirmed to have any noticeable effect in practice and IMO isn't worth carrying the maintenance burden or adding to the deployment complexity. Theoretically, I/O throughput for archived recordings should be improved here due to the use of a more direct API call for sending the file contents over the network without first streaming the file contents into our userspace code and back out again.

Closes #62


vic-ma:
Could you rebase and fix the README conflict?

andrewazores:
Done.

vic-ma (approved):

PR #318, opened 2020-10-29
Better error reporting for report generation failures

andrewazores:


vic-ma:
For the HTTP doc, in handlers where `500` is also used for errors we expected, I added that expected error to the beginning of the description (e.g. `ClientUrlGetHandler`). So in this case I think we should add to the beginning of the `500` status code for these handlers a note saying it could mean the report generation used too much memory.

For failures, is the status text maybe too long?
```
< HTTP/1.1 500 java.util.concurrent.ExecutionException: com.redhat.rhjmc.containerjfr.net.internal.reports.SubprocessReportGenerator$ReportGenerationException: [3] The report generation process consumed too much memory and was terminated
```
Would it be possible to have the status text shortened and keep the full error message in the body?

andrewazores:
> For the HTTP doc, in handlers where `500` is also used for errors we expected, I added that expected error to the beginning of the description (e.g. `ClientUrlGetHandler`). So in this case I think we should add to the beginning of the `500` status code for these handlers a note saying it could mean the report generation used too much memory.

Hmm. Okay, I'll add some text giving some examples of things that might cause `500`s for this particular handler, but I don't know if it's worth the effort to try to document every single possible cause of "unexpected internal errors" that can arise during the report generation, subprocess forking, I/O, etc... The client probably shouldn't be trying to handle various subtypes of `500` differently from each other anyway.

> 
> For failures, is the status text maybe too long?
> 
> ```
> < HTTP/1.1 500 java.util.concurrent.ExecutionException: com.redhat.rhjmc.containerjfr.net.internal.reports.SubprocessReportGenerator$ReportGenerationException: [3] The report generation process consumed too much memory and was terminated
> ```
> 
> Would it be possible to have the status text shortened and keep the full error message in the body?

That behaviour generally comes from the WebServer's failureHandler, not something specific to this API handler or the report generation/caching system. It's a good idea to fix this but maybe out of scope of this PR?

vic-ma:
>Hmm. Okay, I'll add some text giving some examples of things that might cause 500s for this particular handler, but I don't know if it's worth the effort to try to document every single possible cause of "unexpected internal errors" that can arise during the report generation, subprocess forking, I/O, etc... The client probably shouldn't be trying to handle various subtypes of 500 differently from each other anyway.

I guess that makes sense. I think for some of the handlers, like when the cause is a missing env var, it might be worth noting, but I agree that in this case it probably doesn't make sense.

>That behaviour generally comes from the WebServer's failureHandler, not something specific to this API handler or the report generation/caching system. It's a good idea to fix this but maybe out of scope of this PR?

Yeah I wasn't sure if it was easily fixable or was because of how WebServer works. In that case it sounds fine then.

vic-ma (approved):

PR #321, opened 2020-11-05
fix difference in truststore env var name

Alexjsenn:
Not sure when it was changed, but the name of the env var for the ssl truststore was changed in entrypoint, and was not updated in the certificatePostHandler.

### DIFF
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -1271,7 +1271,7 @@
 
     `409` - A certificate with the same filename already exists in the truststore directory. The body includes the path where the file already exists.
 
-    `500` - The `TRUSTSTORE_DIR` environment variable is not set, or there is an unexpected error. The body is an error message.
+    `500` - The `SSL_TRUSTSTORE_DIR` environment variable is not set, or there is an unexpected error. The body is an error message.
 
     ###### example
     ```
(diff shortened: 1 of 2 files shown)


andrewazores (dismissed):

andrewazores:
```
$ ag --ignore-dir web-client TRUSTSTORE_DIR
src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v2/CertificatePostHandler.java
78:    private static final String TRUSTSTORE_DIR = "TRUSTSTORE_DIR";
142:        if (!env.hasEnv(TRUSTSTORE_DIR)) {
146:        String truststoreDir = env.getEnv(TRUSTSTORE_DIR);

src/main/extras/app/entrypoint.sh
40:if [ -z "$SSL_TRUSTSTORE_DIR" ]; then
41:    SSL_TRUSTSTORE_DIR="/truststore"
44:export SSL_TRUSTSTORE_DIR
60:    if [ ! -d "$SSL_TRUSTSTORE_DIR" ]; then
61:        banner "$SSL_TRUSTSTORE_DIR does not exist; no certificates to import"
63:    elif [ ! "$(ls -A $SSL_TRUSTSTORE_DIR)" ]; then
64:        banner "$SSL_TRUSTSTORE_DIR is empty; no certificates to import"
68:    for cert in $(find "$SSL_TRUSTSTORE_DIR" -type f); do
168:export SSL_TRUSTSTORE_DIR

HTTP_API.md
1274:    `500` - The `TRUSTSTORE_DIR` environment variable is not set, or there is an unexpected error. The body is an error message.
```

Could you update that line in `HTTP_API.md` while we're at it?

andrewazores (approved):

PR #322, opened 2020-11-06
Do not set exception payload as response status message

### LINKED ISSUE #320: Exceptions caught by WebServer failureHandler should write payloads to body, not status message
SSIA. https://github.com/rh-jmc-team/container-jfr/pull/318#issuecomment-722012930

Related to #317

andrewazores:
Fixes #320

See also https://github.com/rh-jmc-team/container-jfr-web/pull/160


andrewazores:
To test, do something that triggers an exception response. For example, try to create a recording named Foo, then try to create another recording with the same name. Or trigger a report generation and have it fail (ex. set the max heap size very small to cause this immediately). Observe that the custom exception message should still be present in the response body, but the response status message should now be simply the default message corresponding to the response status code (ex. `400 -> Bad Request, 500 -> Internal Error`).

vic-ma (approved):

PR #326, opened 2020-11-10
Remove TTY/TCP execution modes

### LINKED ISSUE #302: Remove interactive and TCP socket modes
Long-running async HTTP/WS mode is the only mode receiving any development attention, and already there are important features missing from the interactive TTY and TCP socket modes. Most of the commands available there have been removed since they are superseded by HTTP API handlers, and for the remaining functionality these modes do have, they provide no mechanism for supplying JMX auth credentials or even nicely reporting auth failures. I don't see it being worth the effort to try to bring these modes up to par, so they should instead be removed.

andrewazores:
Fixes #302 

This removes the old interactive TTY and TCP socket execution modes and all associated configuration, command executors, etc. These modes have not seen any development in some time and do not support things like platform or target auth, and are no longer intended use cases - only the WebSocket execution mode is a targeted use case.


vic-ma:
Looks good, but it looks like it's saying there's a conflict with the web client.

andrewazores:
Rebased.

vic-ma (approved):

PR #331, opened 2020-11-17
Async JDP

andrewazores:
Related to #158

This enables async JDP (rh-jmc-team/container-jfr-core#50) in the `DefaultPlatformClient`, sending a message on the Notification channel whenever a change in the known targets is observed. This also adds a `start` method to the `PlatformClient` interface so that other implementations can be extended with similar async discovery logic (ex. the `OpenShiftPlatformClient` can use "watch" functionality to observe changes to endpoints and notify based on these) or other async behaviours.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
 
   <com.google.dagger.version>2.26</com.google.dagger.version>
 
-  <com.redhat.rhjmc.containerjfr.core.version>0.16.0</com.redhat.rhjmc.containerjfr.core.version>
+  <com.redhat.rhjmc.containerjfr.core.version>0.17.0</com.redhat.rhjmc.containerjfr.core.version>
 
   <org.apache.commons.lang3.version>3.9</org.apache.commons.lang3.version>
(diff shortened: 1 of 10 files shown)


vic-ma:
I tested this by using `run.sh` and then manually running the vertx-fib-demo and stopping it with `podman kill`. Is it okay that the `LOST` notification is a few seconds delayed from when I kill the demo app?

andrewazores:
That sounds like a good way to test it.

The delay in notifications when a target (dis)appears is because of what the JMC JDP client does in the backend - it's actually doing its own polling behind the scenes, but it provides an abstraction for async listening to changes. That's what I wrapped around in `-core` and am building on top of here to send notifications. So, the API looks async, but the reality of the implementation is that it's still doing periodic polling behind the scenes, so the behaviour you see is expected.

Of course, if JMC's JDP client is improved and polls at a different period or somehow gains some true async discovery feature, then we get to benefit from that for free. Or we can implement our own async discovery in `-core` instead, and the application will against benefit from that for free thanks to this abstraction.

vic-ma (approved):

PR #336, opened 2020-11-19
Use custom base image

### LINKED ISSUE #333: Create customized base image
ContainerJFR should have its own customized base image, built on top of the current `ubi8-openjdk` base image used, but with some minor tweaks on top.

andrewazores:
Create a custom base image Containerfile and build script (currently just extends the previous ubi8 base image with no changes) and set pom.xml to use this new image for container builds

Fixes #333


vic-ma:
Should the `TAG` and `BUILDER` env vars be documented in the README?

vic-ma (dismissed):

PR #338, opened 2020-11-19
Create self-signed cert at image build time

### LINKED ISSUE #335: Customized base image should create self-signed certificate in advance
Part of #333 

Similar to #334, `entrypoint.sh` currently creates the self-signed SSL/TLS certificate at container startup time. This takes time and also results in each redeployed instance having a different certificate, requiring clients to re-trust the application, even though it is really the same. This should be done at container base image build time to save startup time and not require clients to re-trust the application, allowing `entrypoint.sh` to simply grab and use the certificate as a fallback if the end user/administrator has not specified their own.

andrewazores:
Based on top of #337

Creates the self-signed SSL certificate at image build time. At runtime, `entrypoint.sh` determines whether this certificate should be used as a fallback, or if the user-supplied certificate can be used instead.

Fixes #335

### DIFF
--- a/base-image/Containerfile
+++ b/base-image/Containerfile
@@ -2,6 +2,8 @@ FROM registry.access.redhat.com/ubi8/openjdk-11:1.3-6
 
 ENV CONF_DIR=/opt/containerjfr.d
 
+ENV PATH=$CONF_DIR:$PATH
+
 RUN mkdir -p $CONF_DIR
 
 ENV SSL_TRUSTSTORE=$CONF_DIR/truststore.p12 \
--- a/base-image/Containerfile
+++ b/base-image/Containerfile
@@ -5,8 +5,10 @@ ENV CONF_DIR=/opt/containerjfr.d
 RUN mkdir -p $CONF_DIR
 
 ENV SSL_TRUSTSTORE=$CONF_DIR/truststore.p12 \
-    SSL_TRUSTSTORE_PASS_FILE=$CONF_DIR/truststore.pass
+    SSL_TRUSTSTORE_PASS_FILE=$CONF_DIR/truststore.pass \
+    SSL_KEYSTORE=$CONF_DIR/keystore.p12
 
 COPY include $CONF_DIR
 
-RUN $CONF_DIR/truststore-setup.sh
+RUN $CONF_DIR/truststore-setup.sh && \
+    $CONF_DIR/generate-cert.sh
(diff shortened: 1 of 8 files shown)


ebaron:
I'm trying to think if there are any security implications to giving everyone the same self-signed cert instead of generating them upon startup. Any thoughts?

andrewazores:
> I'm trying to think if there are any security implications to giving everyone the same self-signed cert instead of generating them upon startup. Any thoughts?

I thought about this too. We should probably defer to a security expert, but I don't know if there's a problem here - the cert isn't used for any kind of authorization, so this shouldn't lead to unprivileged users gaining access to other instances where they shouldn't. It's a very valid question however.

andrewazores:
> I'm trying to test using my own SSL certificate on upstream so I can do the same to test this. I ran the `-genkeypair` command in `generateSslCert()` in `entrypoint.sh`, and then I ran the `-exportcert` command just to verify that the password and keystore are working. But when I run with the keystore and plaintext mounted and env vars set, I get:
> 
> ```
> Exception in thread "main" io.vertx.core.VertxException: java.io.IOException: keystore password was incorrect
> ```
> 
> The plaintext file, I just created with a text editor, and there isn't a blank line under the password or anything.

Is this describing how you tested upstream behaviour first, or are you already testing against this PR? I haven't done the manual SSL config setup in a while myself but I'll try it, too.

vic-ma:
>Is this describing how you tested upstream behaviour first, or are you already testing against this PR? I haven't done the manual SSL config setup in a while myself but I'll try it, too.

This is just me testing it on upstream first to make sure I can do it.

andrewazores:
> > Is this describing how you tested upstream behaviour first, or are you already testing against this PR? I haven't done the manual SSL config setup in a while myself but I'll try it, too.
> 
> This is just me testing it on upstream first to make sure I can do it.

https://github.com/rh-jmc-team/container-jfr/pull/342

This PR should help out. It adds a script that makes it easy to create a self-signed cert keystore setup and a bit of modification to run.sh to use that setup if present. This way you can create a certificate in advance and have it used across run.sh runs. If you remove it then run.sh will fall back to its current runtime generation of its own self-signed cert as usual.

vic-ma (approved):
Tested with CJFR-generated, and manually generated.

The problem I had was thinking that `KEYSTORE_PASS` is a path to a plaintext file containing the password, when really it's just the password itself :P

vic-ma (approved):

andrewazores:
#342 makes things more convenient during development, which is maybe the useful part of what this PR would have done. Otherwise, perhaps it is better after all that any other deployment should generate and use a new self-signed certificate on each new startup, if only as a deterrent from end users simply relying upon that self-signed cert in production. Real deployments should really have specific certificates created for them.

PR #342, opened 2020-11-24
Self-signed SSL cert generation for development testing

andrewazores:


vic-ma:
Do you want me to review this now, since it's marked as draft?

vic-ma:
I seem to be getting `502` "Target SSL Untrusted" when using handlers that require JMX auth.

andrewazores:
If you edit `generate-dev-certs.sh` and remove the `rm` on the last line, then copy `certs/server.ser` into `truststore`, does that fix the problem?

vic-ma:
Yes that fixes it.

vic-ma (approved):

PR #345, opened 2020-11-25
Do not exit early from platform detection

### LINKED ISSUE #344: Manual selection of deployment platform results in PlatformClient not started
When allowing ContainerJFR to detect the runtime platform automatically, the `PlatformClient` implementation is `start()`ed once by the `PlatformModule`. However, if the platform is specified manually with the env var `CONTAINER_JFR_PLATFORM`, then the client is never started. This currently leads to broken async target discovery, since that is the only behaviour enabled when starting the platform clients.

andrewazores:
Use single method exit point to ensure post-selection logic (currently, client starting) is performed for both automatic and manual platform selection

Fixes #344

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -75,6 +75,7 @@ podman run \
     --mount type=tmpfs,target=/templates \
     --mount type=bind,source="$(dirname $0)/truststore",destination=/truststore,relabel=shared,bind-propagation=shared \
     --mount type=bind,source="$(dirname $0)/certs",destination=/certs,relabel=shared,bind-propagation=shared \
+    -e CONTAINER_JFR_PLATFORM=$CONTAINER_JFR_PLATFORM \
     -e CONTAINER_JFR_DISABLE_SSL=$CONTAINER_JFR_DISABLE_SSL \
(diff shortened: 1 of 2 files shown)


vic-ma:
I'm trying to test the bug on upstream, but I don't see an `-e` flag in `run.sh` for the `CONTAINER_JFR_PLATFORM` env var. Is the environment variable only supposed to be used by the operator? And if it is, can I still test this bug/fix anyway by just adding it to `run.sh`?

andrewazores:
I was testing against the Operator when I discovered this bug, since the Operator does explicitly set that env var to avoid the cost of doing the automatic detection at startup. That env var could be added to `run.sh` (I'll add a commit here in a minute to do that), but the only sensible deployment platform to set when running in Docker/Podman is the `DefaultPlatformStrategy`. Still, manually setting that value vs allowing it to do automatic detection should be enough to trigger this bug.

vic-ma (approved):

andrewazores:
Rebased.

vic-ma (approved):

PR #350, opened 2020-12-01
Use Alias in Archived Recording Names if Possible

### LINKED ISSUE #297: Archived recording names don't match up to web-client dropdown
image

When a recording is archived, the generated name does not include the target's alias, which is what is normally used to identify targets in the web-client UI. This makes the archived recording list more difficult or even impossible to accurate determine which target an archived recording came from.

In the image above, the recording was created from the `es.andrewazor.demo.Main` application.

vic-ma:
Fixes #297

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/TargetRecordingPatchSave.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/TargetRecordingPatchSave.java
@@ -58,6 +58,7 @@ import com.redhat.rhjmc.containerjfr.core.sys.Clock;
 import com.redhat.rhjmc.containerjfr.core.sys.FileSystem;
 import com.redhat.rhjmc.containerjfr.net.ConnectionDescriptor;
 import com.redhat.rhjmc.containerjfr.net.TargetConnectionManager;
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/TargetRecordingPatchSave.java
```diff
@@ -116,7 +120,25 @@ private String saveRecording(JFRConnection connection, IRecordingDescriptor desc
         if (recordingName.endsWith(".jfr")) {
             recordingName = recordingName.substring(0, recordingName.length() - 4);
         }
-        String targetName = connection.getHost().replaceAll("[\\._]+", "-");
+
+        String targetName =
+                platformClient.listDiscoverableServices().stream()
```
andrewazores, line 125: This is my only issue - it's unfortunate that this lookup is performed every time. We should implement something that allows us to do this cross-referencing without having to trigger target discovery first. Something as simple as a map from targetIds to known ServiceRefs would fit the description, but I'm not sure if that data structure should be held by the PlatformClient or if we should abstract it somewhere else. Maybe just leave a TODO comment here with some short summary and description of this problem. It could also be filed as an issue of type "chore".

andrewazores (approved):

PR #352, opened 2020-12-03
V2 Response API Format

### LINKED ISSUE #317: Design JSON API response format
Our HTTP response format should include more (meta)data and have a more standardized format. Since we don't have a traditional backing database with things like record IDs, some of the go-to existing standards (like https://jsonapi.org/) may or may not make sense.

The goals are:

1. have a more cohesive success response format, ideally enforced by some abstraction around a type system as present in the Command Channel `ResponseMessage` types, so that all API handlers can produce uniform results with minimal overhead in each handler implementation
2. eliminate need for custom, non-standard HTTP status codes. We already have one such status code (`427`, used for JMX auth failures) and are currently working through a similar case where `502` is now overloaded but clients need to be able to distinguish the cases. This should be achievable in the new JSON response format by encoding information about errors within the response body, including something equivalent to a status code which can be used to represent more specific failure modes.

Here's a made-up example of a success and three failures with responses that generally meet the goals:

```
GET /api/v2/reports/foo.jfr
HTTP 200
{ meta: { status: "OK", type: "text/html" }, data: { report: "<html><p>report as HTML string</p></html>" } }
```

```
GET /api/v2/reports/nope.jfr
HTTP 404
{ meta: { status: "NOT FOUND" } } // if no more specific status message, default to standard HTTP message. data part is optional and probably unnecessary here
```

```
GET /api/v2/reports/noauth.jfr
HTTP 407
{ meta: { status: "AUTH FAIL", type: "text/plain" }, data: { reason: "\"Authorization\" header was not included or supplied invalid credentials" } } // if a more specific status message is available then set it in the metadata, and preferably include a data part with a more informative error for clients to consume
```

```
GET /api/v2/reports/notargetauth.jfr
HTTP 407
{ meta: { status: "JMX AUTH FAIL", type: "text/plain }, data: { reason: "\"X-JMX-Authorization\" header was not included or supplied invalid credentials" } }
```

andrewazores:
Fixes #317

This PR introduces a new `AbstractV2RequestHandler` class, similar to the existing `AbstractAuthenticatedRequestHandler`, for all V2 API handlers to extend. Rather than passing in a Vertx `RoutingContext` directly to the handler and having it directly write to the context response, subclasses of this new abstract handler receive a `RequestParams` object which abstracts out the read-only request properties that the handler may require (such as path parameters, headers, query parameters). Rather than writing to the response directly (which is no longer available through the `RequestParams`), subclasses return an instance of `IntermediateResponse`, which encapsulates the response result body as well as an optional status code (default `200`), additional headers, and optional metadata status message. This overall structure is reminiscent of how Command Channel `(Serializable)Command`s are implemented, with the Command/Handler returning a data structure representing their response, and a higher-level class wrapping this response into some metadata, serializing it, and writing out the final representation to the network.

This structure allows all V2 handlers to maintain a consistent overall response format with each other without each needing to implement it individually, including any additions or changes to the overall response format. This also allows for more information about failures to be provided to clients through the metadata fields, without necessarily having to use custom HTTP status codes and messages.

The `WebServer failureHandler` is also modified to handle the new `ApiException` type in such a way that responses generated from this exception type being thrown follow the same response format as the V2 handlers, but with the metadata, data, and HTTP status code/message set to indicate that this is a failure response. The response body contains a JSON object formatted in the same way as a successful response, allowing clients to parse responses in a uniform way in either case.

This does not affect any V1 handlers, which continue to directly receive a `RoutingContext` and write responses directly, and which do not throw the new `ApiException` type (and so the failure behaviour implemented in the `WebServer` is unchanged).

### DIFF
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -1199,6 +1199,41 @@
 
 ## V2 API
 
+Unless otherwise specified, all V2 API handlers respond to requests with a
+metadata-wrapped and JSON-encoded response format with the following general
+form:
+
+```
+{ "meta:" { "status:" "statusString", "type:" "mime/type"}, "data:" { "someKey:" someValue } }
+```
+
+`statusString` will always be `OK` if the response status code is `200`. If the
+response status code is `4xx` or `5xx` then `statusString` may match the HTTP
+response status message, or it may provide more detail specific to the handler
+and the exact cause of failure.
+
+`type` is a MIME type specifier (ex. `application/json`, `text/plain`) which
+hints to the client how the response `data` is formatted. The `data` key will
+always be included and the value will always be an object, but the contents of
+the object vary by handler.
+
+`someKey` may be named anything. The most common name on successful requests is
+`result`, but this may vary by handler. The associated value may be anything and
+varies by handler. It may be a plain text string, or a next object, or an array.
+
+The response format on failure is similar:
+
+```
+{"meta":{"type":"text/plain","status":"Authentication Failure"},"data":{"reason":"JMX Authentication Failure"}}
+```
+
+Any response with a `4xx` or `5xx` status code will have `someKey` replaced by
+`reason` as explained above, and the value of `reason` will be a plain text
+string containing some additional detail about the failure.
+
+The handler-specific descriptions below describe how each handler populates the
+`type` and `data` fields. The general formats follow as above.
+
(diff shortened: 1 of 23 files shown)


vic-ma (commented):
Everything else looks good.

vic-ma (approved):

PR #363, opened 2020-12-17
Remove String.format from logger calls to improve performance

### LINKED ISSUE #358: Don't use String.format inside log statements
See also https://github.com/rh-jmc-team/container-jfr-core/issues/52
See http://www.slf4j.org/faq.html#logging_performance

Alexjsenn:
Fixes #358 
There are two or three statements where I left the String.format call because they had several arguments, and it probably doesn't make sense to create functions for that exact number of objects in cjfr-core.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/ContainerJfr.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/ContainerJfr.java
@@ -62,10 +62,7 @@ class ContainerJfr {
 
         logger.trace(String.format("env: %s", environment.getEnv().toString()));
 
-        logger.info(
-                String.format(
-                        "%s started.",
-                        System.getProperty("java.rmi.server.hostname", "container-jfr")));
+        logger.info("{} started.", System.getProperty("java.rmi.server.hostname", "container-jfr"));
 
         Client client = DaggerContainerJfr_Client.builder().build();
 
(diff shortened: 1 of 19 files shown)


andrewazores:
> they had several arguments, and it probably doesn't make sense to create functions for that exact number of objects in cjfr-core.

[varargs](https://docs.oracle.com/javase/8/docs/technotes/guides/language/varargs.html)?

Alexjsenn:
> > they had several arguments, and it probably doesn't make sense to create functions for that exact number of objects in cjfr-core.
> 
> [varargs](https://docs.oracle.com/javase/8/docs/technotes/guides/language/varargs.html)? I believe the documentation specifies that this is how SLF4J does it - there are `logger.log(String msg)`, `logger.log(String msg, Object a)`, `logger.log(String msg, Object a, Object b)`, and `logger.log(String msg, Object... args)` methods defined. We can do the same.

Should I add these methods for all of the logger message types even though some may not be used?

andrewazores:
They may not be used _right now_... IMO it only makes sense to make all of them consistent that way.

Alexjsenn:
yes, it makes sense, I added all the methods in cjfr-core, and changed the last string.format calls.

andrewazores:
Looks good. I manually rebased this and tested it myself and everything seems to continue working as expected. Please rebase this PR and I'll merge it.

andrewazores (approved):

PR #367, opened 2020-12-22
Add guice and plexus-utils dependencies

ebaron:
These dependencies are currently brought in with JMC/Eclipse and need to specified explicitly after https://github.com/rh-jmc-team/container-jfr-core/pull/56.


andrewazores:
Is `guice` available for downstream product builds, but `dagger` isn't?

ebaron:
It looks like some form of dagger is there, I'm just not sure if it's usable yet. I wouldn't worry about trying to replace it with something else just yet.

andrewazores (approved):

PR #378, opened 2020-12-24
Add isAsync() for All Handlers

### LINKED ISSUE #312: Investigate whether HTTP handlers are all correctly async
https://github.com/rh-jmc-team/container-jfr/pull/310#pullrequestreview-515995054

The `isAsync` method on `RequestHandler`s returns `true` by default, but some handlers override this to return `false`. This value is used by the `WebServer` to determine how to install the handler into the `Router` - if `isAsync == true` then the handler is installed normally, otherwise it is installed as a blocking handler. Any handlers which are not themselves implemented async should have `isAsync == false` so that Vertx runs the handler on a separate worker thread (perhaps with some small exceptions like `GrafanaDashboardUrlGetHandler` which simply returns a String read from env).

The goal in investigating/fixing this issue is not to reimplement handlers to add/remove async pieces of their implementations but simply to check behaviours and update `isAsync` to match.

vic-ma:
Fixes #312 

Handler | Async? | Notes
-- | -- | --
`AuthPostHandler` | No | 
`ClientUrlGetHandler` | No | A unit test expects it to be async.
`GrafanaDashboardUrlGetHandler` | No | 
`GrafanaDatasourceUrlGetHandler` | No | 
`RecordingDeleteHandler` | No | 
`RecordingGetHandler` | Yes | `ctx.vertx().filesystem().exists()` is async.
`RecordingsGetHandler` | No | 
`RecordingsPostHandler` | Yes | `validateRecording()` and `saveRecording()` are async, but this handler already has `isAsync()` set to `false`.
`RecordingUploadPostHandler` | No | Already has `isAsync()` set to `false`.
`ReportGetHandler` | No | Already has `isAsync()` set to `false`.
`TargetEventsGetHandler` | No | 
`TargetRecordingDeleteHandler` | No | 
`TargetRecordingGetHandler` | No | Already has `isAsync()` set to `false`.
`TargetRecordingOptionsGetHandler` | No | Already has `isAsync()` set to `false`.
`TargetRecordingOptionsPatchHandler` | No | 
`TargetRecordingPatchHandler` | No | Already has `isAsync()` set to `false`.
`TargetRecordingsGetHandler` | No | 
`TargetRecordingsPostHandler` | No | ~~Already has `isAsync()` set to `false`.~~
`TargetRecordingUploadPostHandler` | No | Already has `isAsync()` set to `false`.
`TargetReportGetHandler` | No | Already has `isAsync()` set to `false`.
`TargetsGetHandler` | No | 
`TargetSnapshotPostHandler` | No | 
`TargetTemplateGetHandler` | No | 
`TargetTemplatesGetHandler` | No | 
`TemplateDeleteHandler` | No | 
`TemplatesPostHandler` | No | 
`CertificatePostHandler` (V2) | No |  Already has `isAsync()` set to `false`.
`TargetEventsSearchGetHandler` (V2) | No |  Already has `isAsync()` set to `false`. 
`TargetRecordingOptionsListGetHandler` (V2) | No |  Already has `isAsync()` set to `false`.
`TargetSnapshotPostHandler` (V2) | No | 

`AuthPostHandler`, `GrafanaDashboardUrlGetHandler`, and `GrafanaDatasourceUrlGetHandler` are not async, but they're all simple handlers, so I've set their `isAsync()` values to `true` anyway, as suggested.

`ClientUrlGetHandler` doesn't seem to be async from what I can tell, but there's a unit test that expects it to be async, so I didn't change it.

`RecordingsPostHandler` seems to have async helper functions, but it already had its `isAsync()` set to `false`, so I didn't change it.

**Edit:** `TargetRecordingsPostHandler` didn't actually have `isAsync()` set to `false` already, but it does appear to be synchronous.

### DIFF
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/AuthPostHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/AuthPostHandler.java
@@ -72,6 +72,13 @@ class AuthPostHandler extends AbstractAuthenticatedRequestHandler {
         return basePath() + "auth";
     }
 
+    // This handler is not async, but it's simple enough that it doesn't need
+    // to be run in a seperate worker thread.
+    @Override
+    public boolean isAsync() {
+        return true;
+    }
+
     @Override
     public void handleAuthenticated(RoutingContext ctx) throws Exception {
         ctx.response().setStatusCode(200);
--- a/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/ClientUrlGetHandler.java
+++ b/src/main/java/com/redhat/rhjmc/containerjfr/net/web/http/api/v1/ClientUrlGetHandler.java
@@ -89,6 +89,11 @@ class ClientUrlGetHandler implements RequestHandler {
     }
 
     @Override
+    public boolean isAsync() {
+        return true;
+    }
+
+    @Override
     public void handle(RoutingContext ctx) {
         ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, HttpMimeType.JSON.mime());
         try {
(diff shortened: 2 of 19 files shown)


andrewazores (approved):
Nice work. I read over all of your notes and glanced over the implementations and I agree with all of your findings. Please also prepare a PR to backport this to the `v1` branch.

PR #391, opened 2021-01-12
Replace io.kubernetes:client-java with io.fabric8:kubernetes-client

### LINKED ISSUE #377: Replace io.kubernetes:client-java with io.fabric8:kubernetes-client
We already pull in io.fabric8:kubernetes-client as a dependency of io.fabric8:openshift-client, so we may as well use it. This will eliminate a dependency we have to ship as well. This also needs to be done in the v1 branch.

andrewazores:
Fixes #377

This removes the io.kubernetes:client-java dependency and explicitly adds io.fabric8:kubernetes-client instead, which I believe was already present as a transitive dependency of io.fabric8:openshift-client, which was used for the `OpenShiftPlatformClient` and related classes.

The implementation of `OpenShiftPlatformClient` was more or less copied into `KubeApiPlatformClient`, with the internal `fabric8 OpenShiftClient` replaced with the more generic `fabric8 KubernetesClient`, then the `OpenShiftPlatformClient` removed. The `OpenShiftPlatformStrategy` and `KubeApiPlatformStrategy` are now nearly identical as well, although for OpenShift there is still a specific `AuthManager` used, and the platform detection check is whether the client can list Routes as opposed to Endpoints.

The changes to tests mirror the changes above - the `OpenShiftPlatformClient` implementation overwrote the previous `KubeApiPlatformClient` with minor adjustments to make it Kubernetes-generic, and the OpenShiftPlatformClientTest was removed.

I have tested this in OpenShift (CRC) and everything seems to work as before. I don't have any other Kubernetes environments to test this in, however.


ebaron (approved):

ebaron:
Looks good! Will you be doing this for v1 as well?

andrewazores:
Yes, I'll prepare the backport and (hopefully) open the PR for it tomorrow.

PR #393, opened 2021-01-14
Replace io.kubernetes:client-java with io.fabric8:kubernetes-client (#391)

andrewazores:
This is a backport of #391, also including some changes from #348,
namely that Endpoints discovery is used rather than Services, but the
async/notifications parts have been removed as those are not in v1.


ebaron (approved):
Looks good! Thanks for taking care of this so quickly

PR #414, opened 2021-02-10
[v1] Disable non-websocket execution modes

andrewazores:
Force WebSocket execution mode hardcoded at startup, while leaving other modes' code paths intact to reduce invasiveness of change


ebaron (approved):
Looks good!

PR #416, opened 2021-02-12
Automatic recording rules

### LINKED ISSUE #355: Automatic recording rules/triggers

andrewazores:
Closes #355

https://github.com/rh-jmc-team/container-jfr/pull/356#issuecomment-777087839

`POST /api/v2/rules`

- [x] When a rule with the same name as an existing rule is created, it will overwrite that rule. Maybe this should be prevented and the user should have to delete the existing rule first. Or there could be a note in the response that says that a rule was overwritten.

- [x] The numerical parameters don't give an error if they are set to negative numbers. They get set to zero when this happens, but I think ideally negative numbers shouldn't be allowed. This would also more in line with TargetRecordingsPostHandler, which rejects negatives.

- [x] When a required parameter is missing, the response is a 500 with "data":{"reason":null}, resulting from an NPE. It would be helpful if the code explicitly checked the given parameters, and let the user know if any were missing.

- [x] Some stuff, like setting the numerical parameters to letters, or setting name to !- or ////////, will cause exceptions, and the response doesn't specifically say that X parameter was set to an invalid value. But I guess this is sort of intentional, right? Since the alternative would be to check every parameter and write the error messages into the response, which would take a lot of extra code.

`DELETE /api/v2/rules/:ruleName`

- [x] Deleting a non-existent rule, as well as an existing one, gives a 200. It would be helpful if non-existent-rule deletions gave a 404 instead, to indicate to the user that the given rule was not found and that no deletion has occurred. Or they could make a note of it in the result, if a 404 is inappropriate here.

- [x] If a rule is already running, and the user would like to stop that rule (delete the active recording and stop the PeriodicArchiver), is there any way for them to do this besides killing the target container? Should deleting a rule maybe be able to do this? Or maybe this could be the job of a new PATCH STOP handler?

- [x] When both the username and password are missing, the error message only says that the username is missing. It would be better if both missing parameters were mentioned.

- [x] Maybe the response could specify if credentials were found and deleted, or if there were no credentials to be deleted? Not super important, but might be helpful.

- [x] When a rule for a target is running, and I delete the credentials for that target, the archiver is still able to archive recordings, despite the fact that JMX auth is required for this, and the saved JMX auth credentials have been deleted. I see that this is because in the code, the credentials are given to the PeriodicArchiver in the constructor and saved to the object, not retreived from the CredentialsManager each time it is used. This is more efficient, but I'm wondering if it makes sense from the user's point of view. If they deleted some credentials, would they expect for CJFR to not have a copy of those credentials stored anywhere, anymore? Or at least not be able to use it?

General

- [x] Deleting the recording created by an automatic rule causes an exception whenever the archiver tries to archive the recording (expectedly). I imagine there's no good way to protect that recording from being deleting, but maybe there is a way to stop the archiver if the recording does get deleted? It does seem unlikely that a user would accidentally delete the automatic recording, though.

- [x] When a target that has a rule running is killed, I get a few exceptions from the archiver trying to archive the recording, before the target is discovered as LOST and the archiver is stopped. This is because I set a low archival period for testing, but obviously could also happen with a longer period, if the archiver just so happens to archive in the in-between time. Also, when this happens, an archived recording will be pruned, but a new one will not be made to replace it, so the end result is one less archived recording than the amount specified by preservedArchives. I don't know how big of a deal any of this is.


andrewazores:
@vic-ma finally, this first pass at the automatic rules/triggers API addition is ready. You have already reviewed all of the contents of this PR piece by piece, anyway. There are still some outstanding TODOs/FIXMEs in the code that are to be addressed, but I will do those as separate PRs later on and see if I can get @jan-law or @hareetd to start contributing patches or reviews toward that effort as well.

andrewazores:
Minor rebase for #521. Didn't really need much.

src/main/java/io/cryostat/net/web/http/api/v2/RulesPostHandler.java
```diff
@@ -0,0 +1,208 @@
...
+        if (rawMime == null) {
+            throw new ApiException(415, "Bad content type: null");
+        }
+        String firstMime = rawMime.split(";")[0];
+        HttpMimeType mime = HttpMimeType.fromString(firstMime);
+        if (mime == null) {
+            throw new ApiException(415, "Bad content type: " + rawMime);
+        }
+        switch (mime) {
+            case MULTIPART_FORM:
```
vic-ma, line 119: What's the purpose of this empty case?

andrewazores: It's a fall-through. Both multipart forms and urlencoded forms can be handled the same.

vic-ma:
For creating a new rule, when using JSON format, negative numbers don't get caught. Is that something that should be addressed or added as a todo?

Otherwise everything looks good.

andrewazores:
> For creating a new rule, when using JSON format, negative numbers don't get caught. Is that something that should be addressed or added as a todo?
> 
> Otherwise everything looks good.

Another good catch. I think addressing it as a new minor bugfix issue is sufficient. Thanks for all the review attention on this.

vic-ma (approved):

PR #436, opened 2021-04-14
[draft] Vertx 4 fixes

andrewazores:
https://github.com/rh-jmc-team/container-jfr/pull/426


andrewazores:
This breaks the React web-client client-side router navigation. Visiting ex. `https://localhost:8181/` prior to the Vert.x upgrade would fall through to the `WebClientAssetsGetHandler` and a `sendFile()` of `index.html`. After the upgrade, this is handled by the `StaticAssetsGetHandler` rather than passing down the priority list to the `WebClientAssets` handler, and the `StaticAssets` handler sees the request for `/` and redirects the browser to `/index.html/`.

https://github.com/vert-x3/vertx-web/issues/1896

PR #463, opened 2021-04-20
Tree struct backend API

Alexjsenn:
Right now this is a very basic implementation that is only working in default platform client.

### DIFF
--- a/src/main/java/io/cryostat/net/AbstractNode.java
+++ b/src/main/java/io/cryostat/net/AbstractNode.java
@@ -52,9 +52,20 @@ public abstract class AbstractNode {
         this.labels = labels;
     }
 
+    protected NodeType getNodeType() {
+        return this.nodeType;
+    }
+
+    protected Map<String, String> getLabels() {
+        return this.labels;
+    }
+
     public enum NodeType {
         NAMESPACE,
         DEPLOYMENT,
+        DEPLOYMENTCONFIG,
+        REPLICASET,
+        REPLICATIONCONTROLLER,
         POD,
         CONTAINER,
         ENDPOINT;
--- a/src/main/java/io/cryostat/net/AbstractNode.java
+++ b/src/main/java/io/cryostat/net/AbstractNode.java
@@ -56,6 +56,7 @@ public abstract class AbstractNode {
         NAMESPACE,
         DEPLOYMENT,
         POD,
-        CONTAINER;
+        CONTAINER,
+        ENDPOINT;
     }
 }
--- a/src/main/java/io/cryostat/net/AbstractNode.java
+++ b/src/main/java/io/cryostat/net/AbstractNode.java
@@ -35,67 +35,27 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package io.cryostat.net.web.http.api.v2;
+package io.cryostat.net;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.util.function.Function;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 
-import javax.inject.Named;
-import javax.inject.Singleton;
+public abstract class AbstractNode {
+    protected NodeType nodeType;
+    protected Map<String, String> labels;
 
-import io.cryostat.net.security.CertificateValidator;
-import io.cryostat.net.web.http.RequestHandler;
-
-import dagger.Binds;
-import dagger.Module;
-import dagger.Provides;
-import dagger.multibindings.IntoSet;
-
-@Module
-public abstract class HttpApiV2Module {
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindTargetSnapshotPostHandler(TargetSnapshotPostHandler handler);
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindCertificatePostHandler(CertificatePostHandler handler);
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindTargetRecordingOptionsListGetHandler(
-            TargetRecordingOptionsListGetHandler handler);
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindTargetEventsSearchGetHandler(TargetEventsSearchGetHandler handler);
-
-    @Binds
-    @IntoSet
-    abstract RequestHandler bindTargetEnvironmentGetHandler(TargetEnvironmentGetHandler handler);
-
-    @Provides
-    @Singleton
-    @Named("OutputStreamFunction")
-    static Function<File, FileOutputStream> provideOutputStreamFunction() throws RuntimeException {
-        return (File file) -> {
-            try {
-                return new FileOutputStream(file);
-            } catch (IOException e) {
-                throw new RuntimeException(e);
-            }
-        };
+    protected AbstractNode(NodeType nodeType, Map<String, String> labels) {
+        this.nodeType = nodeType;
+        this.labels = labels;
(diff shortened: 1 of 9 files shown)


Alexjsenn:
Regarding tthe algorithm to populate the tree given we only have the reverse relationships (child -> parent), the current implementation may not be optimal for two reasons:
1. It assumes the structure of deployment -> replicaset -> pod -> endpoint, and so although it can "skip" levels, it might not contemplate certain scenarios
2. It might not be the most efficient in large deployments because of the possibly unecessary creation of nodes for pods and deployments that aren't necessary (a better solution could be to only create the pod/deployment nodes if they appear as parents, but it would likely mean increasing the amount of client calls, which I don't know how it compares in efficiency)

andrewazores:
This looks like a fair start, but you're right that there are some ways it seems non-optimal.

For 1), I think that hierarchy is mostly fair to assume in Kubernetes - there won't be an Endpoint that belongs directly to a Namespace, for example. However, I don't know if the `Deployment -> ReplicaSet -> Pod` relationship is always correct - what about `DeploymentConfig`s and `ReplicationController`s? I don't think the algorithm would currently notice these.

For 2), I wouldn't really worry about the cost of creating `Node` instances. Those are quite small and so allocating them is cheap, and because it's a tree structure the GC should be able to clean them up pretty quickly as well (although excess allocations may lead to heap fragmentation, but probably not to a degree that matters for any realistically-sized Kubernetes namespace). Client API requests are the resource that costs us here because each of those is a network request that has 2x latency plus whatever API response time, which is what will add up if we aren't being careful.

Also regarding 2) in a sense, I don't think we want our representation of the namespace to include any "empty environments", ex. intermediate nodes that have no eventual leaf (grand)children. So for example, a Deployment where the ReplicaSet is scaled to 0 - there will be 0 Pods and therefore 0 Endpoints, and so the ReplicaSet and the Deployment should not be included in our API response.

Generally, the algorithm we have here looks like this:

```
1. Define the hierarchy as Namespace -> Deployment -> ReplicaSet -> Pod -> Endpoint
2. Loop over Namespaces and create Deployments, noting the Deployment names
3. Loop over Deployments to create ReplicaSets, noting the ReplicaSet names
4. Repeat for Pods and Endpoints
5. Inspect Endpoint ownerRef back-references and chase them back upward
6. Add an Endpoint as a child of a Pod iff the Endpoint's ownerRef matches a Pod's name
7. Add a Pod as a child of a Deployment iff the Pod's ownerRef matches a Deployment's name
8. Repeat for ReplicaSets and Deployments
```

Point 1) could be addressed here, but it would require some special-casing in the iteration steps to detect which path to go down, which seems fragile and complex, but may not be easily avoidable in any case since there simply are two different type hierarchy possibilities. Point 2) is not too bad - we only make one client API request for each level of the hierarchy, which seems like it's probably the minimal amount if there is no API request we can make to retrieve all of the data at once anyway. However, for the latter part of 2), it seems like we would need to do a final recursive pass over the tree to prune it, removing any branches that do not terminate in TargetNode leaves.

Here's another possible idea for an algorithm:

```
1. Define the hierarchy as Namespace -> Deployment -> ReplicaSet -> Pod -> Endpoint
2. Query API for Endpoints. Loop over result for Endpoints.
3. Create an Endpoint Node for each Endpoint. Loop over ownerRefs for the current Endpoint, filtered by Pod type.
4. Create a Node for each Pod (Endpoint's ownerRef) and immediately assign the Endpoint as a child of the Pod. Add the Pods to a separate List/Set for easily looking them up later without traversing the tree.
5. Query API for Pods, filter by names found in previous step. Loop over these Pods.
6. Loop over ownerRefs for the Pod.
7. Create a Node for each ReplicaSet and immediately assign the Pod as a child of the ReplicaSet. Add the ReplicaSets to a separate List/Set for easily looking them up later without traversing the tree.
8. Query API for ReplicaSets, filter by names found in previous step. Loop over these ReplicaSets.
8. etc...
```

The cost in terms of Node allocations is approximately the same here, but I think we end up spending less time traversing the tree since the parent->child relationships are established as we go, rather than having to reconstruct the tree from the hierarchy-level-lists after the fact. The number of API requests is also the same. And, since at each hierarchy level we filter the API results by the relevant results found in the prior step, we don't need to do a pruning pass at the end - the tree should already only contain branches that terminate in TargetNode leaves.

What do you think? Does what I've described make sense? It's more or less taking what you already have and inverting it, and adding a bit of filtering and early-link-creation so that a little extra work is done on each step rather than doing a large block of work at the end.

Alexjsenn:
So Ill start by making one distinction about the current algorithm, which is that as it goes back up the tree, it only adds pods/deployments that have children, so in fact the current tree is already "pruned" because only pods that have endpoints are added to deployments, and only deployments that have children are added to the namespace (and replicaSets don't even appear as you suggested). 

Regarding the algorithm you suggest, I think it would infact save a lot of node creation, because you would not have to create nodes that won't be used (which are being created in my current algorithm). However, something I am not sure about is the efficiency of filtering the incoming list of pods with a list of pod environmentNodes. It would either be two for loops that are nested or streams, which I don't understand indepth enough to know how efficient they are. Did you have something in mind for this step?

A scenario I am not sure would ever happen is if an ownerRef name/kind is not found as a resource in the client call. If so my current algorithm adds the object (with an endpoint down the branch) to the root (namespace) because it probably makes sense for them to still appear somewhere. Does this sound allright?

Alexjsenn:
As for the DeploymentConfig and ReplicationControllers, I had looked at them but because I hadn't encountered them I didn't add them yet. In both my current algorithm and your suggesting there would be no problem to add these resources, they only require their own lists in the same way as deployments and pods.

Alexjsenn:
Oh, one last detail. Currently because you suggested it probably makes sense to not show replicaSets, I don't add them to a list and treat them in the same way, but instead look up relevant ones (ones that are part of a path with an endpoint) by name through the api call. This might not be ideal and I could definitely add a list and populate it once if that is more efficient.

andrewazores:
> So Ill start by making one distinction about the current algorithm, which is that as it goes back up the tree, it only adds pods/deployments that have children, so in fact the current tree is already "pruned" because only pods that have endpoints are added to deployments, and only deployments that have children are added to the namespace (and replicaSets don't even appear as you suggested).

Ah okay, that's good then.

> 
> Regarding the algorithm you suggest, I think it would infact save a lot of node creation, because you would not have to create nodes that won't be used (which are being created in my current algorithm). However, something I am not sure about is the efficiency of filtering the incoming list of pods with a list of pod environmentNodes. It would either be two for loops that are nested or streams, which I don't understand indepth enough to know how efficient they are. Did you have something in mind for this step?

I don't think the efficiency of either the loops or streams needs to be really considered too heavily here. Again, for a realistically-sized namespace, the actual size of these lists/sets is going to be fairly small, like in the magnitude of tens of entries. That's quite small, even an O(n^3) algorithm is going to be fast here compared to the network calls. So, I wouldn't worry much about it. We can profile ourselves and investigate how to optimize performance later if it appears to be an issue :-) .

It would really just have to look something like this:

```
...
List<Pod> pods = new ArrayList<>();
endpoints.forEach(endpoint -> {
  Pod pod = new Pod(endpoint.owner.name);
  pod.addChild(endpoint);
  pods.add(pod);
});

List<Deployment> deployments = new ArrayList<>();
client.pods().inNamespace(namespace).list().getItems().stream()
  .filter(pod -> pods.stream().anyMatch(p -> Objects.equals(p.name, pod.name)))
  .forEach(pod -> {
    Deployment deployment = new Deployment(pod.owner.name);
    deployment.addChild(pod);
    deployments.add(deployment);
  });

...
```

> 
> A scenario I am not sure would ever happen is if an ownerRef name/kind is not found as a resource in the client call. If so my current algorithm adds the object (with an endpoint down the branch) to the root (namespace) because it probably makes sense for them to still appear somewhere. Does this sound allright?

PR #464, opened 2021-04-21
[v1] Rename ContainerJFR (and all variants) to Cryostat

andrewazores:


smoketest.sh
```diff
@@ -3,27 +3,27 @@
...
         --rm -d quay.io/andrewazores/vertx-fib-demo:0.4.0
 }
 
 function runJfrDatasource() {
     podman run \
         --name jfr-datasource \
-        --pod container-jfr \
-        --rm -d quay.io/rh-jmc-team/jfr-datasource:0.0.1
+        --pod cryostat \
+        --rm -d quay.io/cryostat/jfr-datasource:0.0.1
```
ebaron, line 26: This should be 1.0.0-BETA6.

smoketest.sh
```diff
@@ -53,7 +53,7 @@ function configureGrafanaDashboard() {
     local TEMP="$(mktemp -d)"
     pushd "$TEMP"
     echo '{"overwrite":false,"dashboard":' > dashboard.json
-    curl https://raw.githubusercontent.com/rh-jmc-team/jfr-datasource/master/dashboards/dashboard.json >> dashboard.json
+    curl https://raw.githubusercontent.com/cryostat/jfr-datasource/master/dashboards/dashboard.json >> dashboard.json
```
ebaron, line 56: This should be `cryostatio` I think.

ebaron (commented):
Looks good, just noticed a couple things.

ebaron (approved):

Issue #354, opened 2020-12-10
Notifications on all "state" changes


andrewazores:
All backend "state" changes (not all of these are truly backend state, but some kind of action on a target) should all be indicated by an event on the Notifications channel so that other client instances are informed of the changes without having to poll.

Examples:
1. Active recording created/deleted
2. Active recording archived
3. Archived recording created/deleted
4. Event template uploaded/deleted

jan-law:
The corresponding frontend issue is closed (https://github.com/cryostatio/cryostat-web/issues/165). Can this issue be closed too?

andrewazores:
Yes, I think so. Thanks.

PR #470, opened 2021-04-27
Backend state changes

Josh-Matsuoka:
Hi,

This PR addresses https://github.com/cryostatio/cryostat/issues/354 , sending events on the Notification channel when recordings are created/deleted, archived, and templates are uploaded/deleted. I've adjusted the tests accordingly.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
@@ -113,10 +113,6 @@ class RecordingDeleteHandler extends AbstractAuthenticatedRequestHandler {
                                     throw new HttpStatusException(404, recordingName);
                                 }
                                 fs.deleteIfExists(path);
-                            } catch (IOException e) {
-                                throw new HttpStatusException(500, e.getMessage(), e);
-                            } finally {
-                                reportService.delete(recordingName);
                                 notificationFactory
                                         .createBuilder()
                                         .metaCategory(NOTIFICATION_CATEGORY)
@@ -124,6 +120,10 @@ class RecordingDeleteHandler extends AbstractAuthenticatedRequestHandler {
                                         .message(Map.of("recording", recordingName))
                                         .build()
                                         .send();
+                            } catch (IOException e) {
+                                throw new HttpStatusException(500, e.getMessage(), e);
+                            } finally {
+                                reportService.delete(recordingName);
                             }
                             ctx.response().setStatusCode(200);
                             ctx.response().end();
--- a/src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
@@ -39,15 +39,18 @@ package io.cryostat.net.web.http.api.v1;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Map;
 
 import javax.inject.Inject;
 import javax.inject.Named;
 
 import io.cryostat.MainModule;
 import io.cryostat.core.sys.FileSystem;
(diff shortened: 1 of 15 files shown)


src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
```diff
@@ -110,6 +117,13 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
                                 throw new HttpStatusException(500, e.getMessage(), e);
                             } finally {
                                 reportService.delete(recordingName);
+                                notificationFactory
+                                        .createBuilder()
+                                        .metaCategory(NOTIFICATION_CATEGORY)
+                                        .metaType(HttpMimeType.JSON)
+                                        .message(Map.of("recording", recordingName))
+                                        .build()
+                                        .send();
```
andrewazores, line 126: I think this should be a few lines higher up in the main body of the `try` - if the deletion fails and an exception is thrown, then a notification should probably not be emitted.

andrewazores:
Looks good otherwise!

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java
```diff
@@ -81,6 +82,8 @@
     private final Path savedRecordingsPath;
     private final Gson gson;
     private final Logger logger;
+    private final NotificationFactory notificationFactory;
+    private static final String NOTIFICATION_CATEGORY = "RecordingArchived";
```
andrewazores, line 86: Oh, I think this category should be renamed - `RecordingArchived` sounds like an in-memory/active recording was copied into archives, but this handler allows users to re-upload previously archived recordings from their workstation into Cryostat.

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
```diff
@@ -103,6 +110,13 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
                     if (descriptor.isPresent()) {
                         connection.getService().close(descriptor.get());
                         reportService.delete(connectionDescriptor, recordingName);
+                        notificationFactory
+                                .createBuilder()
+                                .metaCategory(NOTIFICATION_CATEGORY)
+                                .metaType(HttpMimeType.JSON)
+                                .message(Map.of("recording", recordingName))
```
andrewazores, line 117: Maybe this payload should also include the `targetId` (really a JMX service URL) of the target that "owns" the named recording?

Josh-Matsuoka: Makes sense, I've added this to the target-oriented handlers' payloads

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java
Josh-Matsuoka: makes sense, I've renamed this now

src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
Josh-Matsuoka: makes sense, I moved this and adjusted the test accordingly.

andrewazores (approved):

andrewazores:
@Josh-Matsuoka do you have any interest in looking at https://github.com/cryostatio/cryostat-web/issues/165 ? That's the corresponding frontend issue. This would simply entail adding notification handlers for each of these new notification types emitted and either using the notification payload to update the view, or if that's infeasible then to use the notification as a trigger to poll for fresh data.

Josh-Matsuoka:
I can take a look at that, sure

PR #476, opened 2021-04-30
Use latest OpenJDK UBI image

ebaron:


andrewazores:
Same comment as here: https://github.com/cryostatio/cryostat-base/pull/1#issuecomment-830358244

andrewazores (approved):

PR #489, opened 2021-05-31
483 View in Grafana not working with smoketest.sh

### LINKED ISSUE #483: View in Grafana not working with smoketest.sh
When running smoketest.sh, selecting "View in Grafana..." fails with a 512 error.

```
SEVERE: Exception thrown
io.vertx.ext.web.handler.impl.HttpStatusException: Server Error (512)
```
Screenshot_2021-05-25 Recordings(1)

If I visit the JFR Datasource URL at `http://0.0.0.0:8080`, I instead see a Fibonacci demo. Perhaps there's a conflict with port numbers?
Screenshot_2021-05-25 Fibonacci Demo

jan-law:
Fixes #483, where both the JFR datasource and the fibonacci demos attempted to bind to port 8080. Resolved by setting the fibonacci demos to listen at port 8081.

Question: I tried to update my branch with ```git pull upstream main``` and now the PR includes the changes made in commit ```43713a3``` as part of my PR. How do I fix that?

### DIFF
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -15,24 +15,27 @@ function runCryostat() {
 function runDemoApps() {
     podman run \
         --name vertx-fib-demo-1 \
+        --env HTTP_PORT=8081 \
         --env JMX_PORT=9093 \
         --pod cryostat \
-        --rm -d quay.io/andrewazores/vertx-fib-demo:0.6.0
+        --rm -d quay.io/andrewazores/vertx-fib-demo:0.7.0
 
     podman run \
         --name vertx-fib-demo-2 \
+        --env HTTP_PORT=8081 \
         --env JMX_PORT=9094 \
         --env USE_AUTH=true \
         --pod cryostat \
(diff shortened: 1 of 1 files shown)


andrewazores:
> Question: I tried to update my branch with git pull upstream main and now the PR includes the changes made in commit 43713a3 as part of my PR. How do I fix that?

You can do a `git reset --soft HEAD^` locally, which will undo your commit and move the changes into the staging area.You can then use`git reset`, `git add`, `git checkout`, etc. to pare down the patch to just what you intend to include. Then `git commit` and force-push the new commit.

In the future, doing a `git fetch upstream && git rebase -i upstream/main` might be preferable to a `git pull` - doing the straight pull will merge the changes into your branch we see here, making PR review more confusing etc. I believe there is a way to configure git to do rebases when pulling rather than merges, too.

ebaron (approved):
Looks good!

PR #491, opened 2021-06-03
Add a badge to README for Google Group

ebaron:
This PR adds a badge to the README linking to our Google Group. What do you think of this idea?


andrewazores:
Nice, I like it. I don't think I was even aware that there were badges for Google Groups, but there are badges for everything now, so makes sense!

I guess we should probably put this in -web and -operator as well, since those might also be common first points of contact.

andrewazores (approved):

PR #498, opened 2021-06-04
Query POM for core version in CI

ebaron:
Backport of #493 for v1 branch.


andrewazores (approved):

andrewazores:
That's an old intermittent test failure: https://github.com/cryostatio/cryostat/issues/132

The fix (#330) is contained in some preparation from tthe switch from a two-way command channel to a one-way notification channel, so it was never backported to v1.

PR #503, opened 2021-06-07
Use older 2.x Surefire/Failsafe test runner plugins

### LINKED ISSUE #502: Some unit test failures are unreported, or tests do not run
Some unit tests are evidently not run by Maven failsafe and/or JUnit or for some other reason and are observed to be failing when run manually. @hareetd could you document these specific failures here since you have the first-hand report data?

Maybe relevant: https://stackoverflow.com/questions/3365628/junit-tests-pass-in-eclipse-but-fail-in-maven-surefire

andrewazores:
3.x  preview releases are not correctly running JUnit 5 parameterized tests in all scenarios

This PR is expected to fail CI checks, since it exposes two manually-discovered unit test failures which are currently silently lost with the Surefire/JUnit configuration in use. @hareetd has a correction for those two test cases.

Fixes #502

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,8 +32,8 @@
   <cryostat.itest.podName>cryostat-itests</cryostat.itest.podName>
 
   <org.apache.maven.plugins.compiler.version>3.8.1</org.apache.maven.plugins.compiler.version>
-  <org.apache.maven.plugins.surefire.version>3.0.0-M4</org.apache.maven.plugins.surefire.version>
-  <org.apache.maven.plugins.failsafe.version>3.0.0-M5</org.apache.maven.plugins.failsafe.version>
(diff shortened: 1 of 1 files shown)


hareetd (approved):
Looks good! I'm currently working on fixing the test cases; one of the two worked with a simple removal of the null test cases as previously discussed but the other revealed more complicated errors.

andrewazores:
Cleanly rebased on top of @hareetd 's fix for the hidden failures. This should now run and pass all unit tests.

andrewazores:
Before @hareetd 's last PR in #512 we have 639 tests run/passed. With that PR it becomes 664 run/passed, which is continued with this PR.

hareetd (approved):

Issue #461, opened 2021-04-20
Add custom JFR events to self


andrewazores:
There should be application-specific JFR events added for things like:

- HTTP request (incl. path, HTTP method, status code, duration)
- WebSocket message emitted (incl. duration, data about type of message)
- JMX connection opened/closed/failed

jan-law:
@hareetd Summary of what we covered:
- Re-did Andrew's WebServerRequestEvent demo from this afternoon
- Hareet will start on JMX connection events and I will start on WebSocket message events
- Will post comments about adding potential events on this issue as part of an ongoing process

andrewazores:
Sounds good.

For each of your event implementations, please add "Related #461" to the PR body so they all get linked back to this issue. We'll add them incrementally, one-by-one, so that the overall review process is easiest to manage for everyone.

For whoever first creates the event template XML document, just drop it in `src/main/extras` to begin with.

jan-law:
`OpenShiftAuthManager.java`, lines 89-97 from `validateToken(Supplier<String> tokenProvider)`, will log an exception but doesn't notify the user that the token authentication failed. Would it make sense to add an event here? 
There's also an unfinished `todo` that might change how the exceptions are handled?

```
try (OpenShiftClient authClient =
                                    new DefaultOpenShiftClient(
                                            new OpenShiftConfigBuilder()
                                                    .withOauthToken(token)
                                                    .build())) {
    // only an authenticated user should be allowed to list routes
    // in the namespace
    // TODO find a better way to authenticate tokens
    authClient.routes().inNamespace(getNamespace()).list();
    return true;
} catch (KubernetesClientException e) {
    logger.info(e);
} catch (Exception e) {
    logger.error(e);
}
```

andrewazores:
Sure, an event to track auth API requests probably makes sense, including information about if the request failed.

The `todo` here is more about the "list namespaces" check, which currently matches the permissions model we expect in a deployment, but does so indirectly. Ideally there would be some auth API method we could call and pass the token to with a set of permissions and just receive a result directly informing us whether the supplied token represents a user who has all of the required permissions, like "user can list things in this namespace" and "user can read/write files to PersistentVolumes attached to Pods in this namespace", etc.

However, you are correct that any exception thrown when this request is made will simply be logged and swallowed. The user will get an auth failure because of the `return false` below, but there will be no information about *why* their token was rejected (are they lacking permissions? Or is the auth server down?).

[...]

PR #505, opened 2021-06-08
461 add custom jfr events to self

jan-law:
Related (#461)

@hareetd and I recreated this WebServerRequestEvent together by following @andrewazores' example.

(From now on, I'll be creating a separate PR for each event)

### DIFF
--- a/src/main/extras/app/Cryostat.xml
+++ b/src/main/extras/app/Cryostat.xml
@@ -2,1199 +2,1199 @@
      Recommended way to edit .jfc files is to use Java Mission Control,
      see Window -> Flight Recorder Template Manager.
 -->
-<configuration version="2.0" label="Cryostat" description="Low overhead configuration safe for continuous use in production environments, typically less than 1 % overhead." provider="Oracle"> 
+<configuration version="2.0" label="Cryostat" description="Low overhead configuration safe for continuous use in production environments, typically less than 1 % overhead." provider="Oracle">
  <event name="io.cyostat.net.web.WebServer.WebServerRequestEvent">
-<setting name="enabled">
-    true
-</setting>
+  <setting name="enabled">
+   true
+  </setting>
  </event>
- <event name="jdk.ThreadAllocationStatistics"> 
+ <event name="jdk.ThreadAllocationStatistics">
   <setting name="enabled">
    true
-  </setting> 
+  </setting>
   <setting name="period">
    everyChunk
-  </setting> 
- </event> 
- <event name="jdk.ClassLoadingStatistics"> 
+  </setting>
+ </event>
+ <event name="jdk.ClassLoadingStatistics">
   <setting name="enabled">
    true
-  </setting> 
+  </setting>
   <setting name="period">
    1000 ms
-  </setting> 
- </event> 
- <event name="jdk.ClassLoaderStatistics"> 
+  </setting>
+ </event>
+ <event name="jdk.ClassLoaderStatistics">
   <setting name="enabled">
    true
-  </setting> 
+  </setting>
   <setting name="period">
    everyChunk
-  </setting> 
- </event> 
- <event name="jdk.JavaThreadStatistics"> 
+  </setting>
+ </event>
+ <event name="jdk.JavaThreadStatistics">
   <setting name="enabled">
    true
-  </setting> 
+  </setting>
   <setting name="period">
    1000 ms
-  </setting> 
- </event> 
- <event name="jdk.ThreadStart"> 
+  </setting>
+ </event>
+ <event name="jdk.ThreadStart">
   <setting name="enabled">
    true
-  </setting> 
+  </setting>
   <setting name="stackTrace">
    true
-  </setting> 
- </event> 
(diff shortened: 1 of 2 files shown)


src/main/extras/app/Cryostat.xml
```diff
@@ -0,0 +1,1200 @@
+<?xml version="1.0" encoding="UTF-8"?><!--
+     Recommended way to edit .jfc files is to use Java Mission Control,
+     see Window -> Flight Recorder Template Manager.
+-->
+<configuration version="2.0" label="Cryostat" description="Low overhead configuration safe for continuous use in production environments, typically less than 1 % overhead." provider="Oracle">
```
andrewazores, line 5: Could you replace the `description` here with some short blurb about this being used for Cryostat self-profiling?

Also, please update the `provider` to Cryostat.

jan-law:
@andrewazores Since we are not using the event fields (`remoteAddr`, `method`, `path`, `statusCode`) in the rest of our code, SpotBugs outputs a `URF_UNREAD_FIELD` error for all four fields. Are we missing anything that will specify that this class is an event, or should we add a `@SuppressFBWarnings` annotation?

andrewazores:
> @andrewazores Since we are not using the event fields (`remoteAddr`, `method`, `path`, `statusCode`) in the rest of our code, SpotBugs outputs a `URF_UNREAD_FIELD` error for all four fields. Are we missing anything that will specify that this class is an event, or should we add a `@SuppressFBWarnings` annotation?

Ah, that's annoying. I think the `@SuppressFBWarnings` is needed here.

andrewazores:
Looking good so far.

Please read over `Part II` and `Part III` in this doc: https://docs.oracle.com/en/java/javase/14/jfapi/event-threshold.html#GUID-7EAD62F2-1769-4486-9A61-29F40D121E1F (@hareetd too)

I didn't cover it in my quick demo initially, but `evt.shouldCommit()` is important here. Some of the other annotations might be nice to include, ex. `@Name`, `@Category` (maybe a Cryostat category is appropriate). `@StackTrace` is also interesting - we probably don't care about stack traces for most of the event types we might be recording, however for webserver requests in particular it might make sense to give context for request failures. This could be a setting we specify in the template, however.

hareetd:
We read over the docs but we're unsure about what the threshold should be for our usage of `evt.shouldCommit()`. The docs say "it's recommended to set a threshold if an operation occurs frequently and outliers are of greatest concern", so we're assuming that we only care about recording webserver requests which timeout?

Based on this we've decided to set the threshold to be the timeout duration specified here:

https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/web/http/generic/TimeoutHandler.java#L50

However, we don't think this is right because of how rare it would be for a request to take this long, meaning such events would rarely, if ever, be recorded.

andrewazores:
I think the default threshold can simply be 0ms, ie always record this event. This can be set in templates instead - we might have different templates with different values for this setting. `shouldCommit()` would allow this event to be conditionally recorded depending on those different template settings, I think.

src/main/java/io/cryostat/net/web/WebServer.java
```diff
@@ -203,21 +208,56 @@ public void start() throws FlightRecorderException, SocketException, UnknownHost
...
+                                        evt.end();
+                                        if (evt.shouldCommit()) {
+                                            evt.commit();
+                                        }
+                                    });
                     router.handle(req);
                 });
     }
 
+    @Name("io.cryostat.next.web.WebServer.WebServerRequest")
```
andrewazores, line 238: typo: `next`

src/main/java/io/cryostat/net/web/WebServer.java
```diff
@@ -203,21 +208,56 @@ public void start() throws FlightRecorderException, SocketException, UnknownHost
...
     }
 
+    @Name("io.cryostat.net.web.WebServer.WebServerRequest")
+    @Label("Web Server Request")
+    @Category("Cryostat")
+    @SuppressFBWarnings(
+            value = "URF_UNREAD_FIELD",
+            justification = "The event fields are recorded with JFR instead of accessed directly")
+    public static class WebServerRequest extends Event {
+        String remoteAddr;
```
andrewazores, line 245: In testing this manually using the web UI, it looks like the remote address contains the same host portion each time but often (each request, even?) changes port numbers. This suggests that it might actually be more space efficient to split the `String remoteAddr` (`"127.0.0.1:1234"`) into a `String host` (`"127.0.0.1"`) and an `int port` (`1234`) fields. This way, the commonly re-appearing host strings can be interned effectively too. This would also make it easier for somebody analyzing the recording to sort/filter by requesting client, since the request host/IP is a distinct field.

andrewazores (approved):

andrewazores:
@jan-law looks good to me. Are you planning to make any further changes? If not, I will merge this now.

jan-law:
> @jan-law looks good to me. Are you planning to make any further changes? If not, I will merge this now.

Good to go, ready to merge

PR #506, opened 2021-06-09
[autorules] Rules deletion enhancements

andrewazores:
This addresses the following two items in the ongoing autorules PR #416:

> Deleting a non-existent rule, as well as an existing one, gives a 200. It would be helpful if non-existent-rule deletions gave a 404 instead, to indicate to the user that the given rule was not found and that no deletion has occurred. Or they could make a note of it in the result, if a 404 is inappropriate here.

`DELETE` on a rule now responds with either `200` or `404`.

> If a rule is already running, and the user would like to stop that rule (delete the active recording and stop the PeriodicArchiver), is there any way for them to do this besides killing the target container? Should deleting a rule maybe be able to do this? Or maybe this could be the job of a new PATCH STOP handler?

`DELETE`ing a rule definition which is "active", in the sense that it has been triggered and a PeriodicArchiver exists due to that rule definition, will now also stop and remove such archiver tasks. The in-memory active recording is left intact on the target since the user may still wish to see that - they can follow up with a `DELETE` or `PATCH stop` on the recording resource itself if they want it removed/stopped, since it's ambiguous which behaviour would be desired when deleting a rule.

### DIFF
--- a/rulestest.sh
+++ b/rulestest.sh
@@ -61,3 +61,9 @@ podman run \
     --env HTTP_PORT=8081 \
     --env JMX_PORT=9094 \
     --rm -d quay.io/andrewazores/vertx-fib-demo:0.7.0
+
+sleep 5
+echo "Deleting rule definition"
+curl -k \
+    -X DELETE \
+    "$CRYOSTAT_HOST/api/v2/rules/Default_Rule"
--- a/rulestest.sh
+++ b/rulestest.sh
@@ -13,7 +13,7 @@ if [ -z "$CRYOSTAT_HOST" ]; then
     CRYOSTAT_HOST="https://0.0.0.0:8181"
 fi
 
-TARGET_CONTAINER=vertx-fib-demo-1
+TARGET_CONTAINER=vertx-fib-demo-2
 
 echo "Killing $TARGET_CONTAINER container"
 podman kill $TARGET_CONTAINER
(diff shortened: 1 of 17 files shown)


vic-ma (approved):

rulestest.sh
```diff
@@ -58,6 +63,12 @@ echo "Restarting $TARGET_CONTAINER"
 podman run \
     --name $TARGET_CONTAINER \
     --pod cryostat \
-    --env JMX_PORT=9093 \
-    --env USE_AUTH=true \
```
vic-ma, line 62: Sorry, just noticed this now, but is getting rid of `USE_AUTH` intentional?

rulestest.sh
```diff
@@ -58,6 +63,12 @@ echo "Restarting $TARGET_CONTAINER"
...
-    --rm -d quay.io/andrewazores/vertx-fib-demo:0.6.0
+    --env HTTP_PORT=8081 \
+    --env JMX_PORT=9094 \
+    --rm -d quay.io/andrewazores/vertx-fib-demo:0.7.0
+
+sleep 5
+echo "Deleting rule definition"
```
vic-ma, line 71: Now that I think about it, maybe this shouldn't delete the rule definition? Since then the periodic archiver doesn't get to run, which I think is probably more important than excersising the deletion function. Alhough, maybe to test both, the `archivalPeriodSeconds` for the rule could just be made to something small like 5 seconds, and then there could be a `sleep 30` before the deletion call? I usually change it to 5 seconds anyway to see the archiver working, though I suppose it is a little unrealistic.

andrewazores: Sure - `rulestest.sh` is just meant as a temporary testing aid while the autorules branch is in progress, and it'll be removed before autorules is merged into main. Whatever behaviours and workflows make sense to test will/should be included in AutoRulesIT instead. I'll open a small PR to restore rulestest.sh back to what it was before this PR.

rulestest.sh
andrewazores: See other comment

PR #507, opened 2021-06-09
Add WebSocket Message Emitted Event

jan-law:
Related #461 

What do you think about the level of detail needed for recording WebSocket events? I decided to include the outgoing json to record more information rather than less:

Screenshot from 2021-06-09 14-53-17

Also, when is it generally appropriate to add a threshold setting for events? I added a threshold setting in this case because WebSocket messages that take a long time to send may indicate a problem.

### DIFF
--- a/src/main/java/io/cryostat/messaging/MessagingServer.java
+++ b/src/main/java/io/cryostat/messaging/MessagingServer.java
@@ -166,14 +166,8 @@ public class MessagingServer implements AutoCloseable {
     public void writeMessage(WsMessage message) {
         String json = gson.toJson(message);
         logger.info("Outgoing WS message: {}", json);
-
         synchronized (connections) {
-            connections
-                    .keySet()
-                    .forEach(
-                            c -> {
-                                c.writeMessage(json);
-                            });
+            connections.keySet().forEach(c -> c.writeMessage(json));
         }
     }
 
--- a/src/main/java/io/cryostat/messaging/MessagingServer.java
+++ b/src/main/java/io/cryostat/messaging/MessagingServer.java
@@ -200,7 +200,7 @@ public class MessagingServer implements AutoCloseable {
             this.message = message;
         }
 
-        public void SetConnection(WsClient connection) {
+        public void setConnection(WsClient connection) {
             this.connection = connection;
         }
     }
--- a/src/main/java/io/cryostat/messaging/MessagingServer.java
+++ b/src/main/java/io/cryostat/messaging/MessagingServer.java
@@ -181,7 +181,9 @@ public class MessagingServer implements AutoCloseable {
                                 evt.begin();
                                 c.writeMessage(json);
                                 evt.end();
-                                evt.commit();
+                                if (evt.shouldCommit()) {
+                                    evt.commit();
+                                }
                             });
         }
     }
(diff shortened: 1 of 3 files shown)


andrewazores:
> What do you think about the level of detail needed for recording WebSocket events? I decided to include the outgoing json to record more information rather than less:

This is definitely too much data to be packed into a recording event like this. The JSON responses may be arbitrarily large and messages may be written arbitrarily often, so we don't want to ram this much data into the JFR buffer this way - this can degrade Cryostat's performance.

I don't think we want to know all of the JSON data so much as the metadata about the connection. The `WsConnection` field reference you added to this event can't actually be serialized - JFR doesn't know how to deal with that. I think instead the event should contain the remote address/host and IP that are being written to and other relevant metadata about the connection itself, and not necessarily the message content (maybe just the length of the message?).

It may be cleanest to do this by inserting the event within the `WsClient` class, rather than in the messaging server.

> Also, when is it generally appropriate to add a threshold setting for events? I added a threshold setting in this case because WebSocket messages that take a long time to send may indicate a problem.

That's hard to give a concrete universal answer to, which is why this is best left as a configurable setting in a template. Defaulting it to always (0ms) is fine for now - we'll worry about tweaking the exact settings of the template(s) later, when we have all of our events defined and get some sample data.

src/main/java/io/cryostat/messaging/MessagingServer.java
```diff
@@ -166,8 +166,14 @@ public String readMessage() {
...
         logger.info("Outgoing WS message: {}", json);
+
         synchronized (connections) {
-            connections.keySet().forEach(c -> c.writeMessage(json));
+            connections
+                    .keySet()
+                    .forEach(
+                            c -> {
+                                c.writeMessage(json);
+                            });
```
andrewazores, line 176: I think the changes to this file can be removed from the patch now, right?

src/main/java/io/cryostat/messaging/WsClient.java
```diff
@@ -83,13 +88,46 @@ String readMessage() {
...
             try {
+                WsMessageEmitted evt =
+                        new WsMessageEmitted(
+                                sws.remoteAddress().host(),
+                                sws.remoteAddress().port(),
+                                sws.uri(),
+                                message.length());
+                evt.begin();
+
                 this.sws.writeTextMessage(message);
```
andrewazores, line 99: If `writeTextMessage` here throws an Exception, the event object will be instantiated and `begin()`'d, but never committed. I think we would want to capture some info about the exception - even just a boolean flag indicating whether an exception occurred here or not might be sufficient.

The pattern to employ here looks roughly like this:

```
WsMessageEmitted evt = new WsMessageEmitted(...);
evt.begin();
try {
  sws.writeTextMesage(message);
} catch (Exception e) {
  logger.warn(e);
  evt.setExceptionStatus(true); // assuming we default the state to false when the event is created
} finally {
  evt.end();
  if (evt.shouldCommit()) {
    evt.commit();
  }
}
```

This way the event is guaranteed to always be initialized, and always check if it should commit and then commit if so, whether an exception occurs or not. This is a good pattern to follow in general if there is any chance at all that an exception or error can occur between event instantiation and event commit - and there effectively always is, since something like an OutOfMemoryError could happen at any point and at any time.

andrewazores:
Code looks good. Please rebase this so that the Cryostat.xml template definition just updates the file that now already exists in `main`.

andrewazores (approved):

PR #511, opened 2021-06-14
[autorules] Mention all missing form fields on credentials POST

andrewazores:
This addresses the following deficiency in the auto-rules draft PR #416:

> When both the username and password are missing, the error message only says that the username is missing. It would be better if both missing parameters were mentioned.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/TargetCredentialsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/TargetCredentialsPostHandler.java
@@ -41,14 +41,16 @@ import java.io.IOException;
 
 import javax.inject.Inject;
 
+import com.google.gson.Gson;
+
+import org.apache.commons.lang3.StringUtils;
+
 import io.cryostat.configuration.CredentialsManager;
 import io.cryostat.core.log.Logger;
 import io.cryostat.core.net.Credentials;
 import io.cryostat.net.AuthManager;
 import io.cryostat.net.web.http.HttpMimeType;
 import io.cryostat.net.web.http.api.ApiVersion;
-
-import com.google.gson.Gson;
 import io.vertx.core.http.HttpMethod;
 
 class TargetCredentialsPostHandler extends AbstractV2RequestHandler<Void> {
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/net/web/http/api/v2/TargetCredentialsPostHandler.java
```diff
@@ -105,17 +104,19 @@ public boolean isOrdered() {
...
+        String username = params.getFormAttributes().get("username");
+        String password = params.getFormAttributes().get("password");
+
+        if (username == null || password == null) {
+            StringBuilder sb = new StringBuilder();
+            if (username == null) {
+                sb.append("username is required.");
+            }
+            if (password == null) {
+                sb.append(" password is required.");
```
vic-ma, line 116: I think username/password should be capitalized.

andrewazores: I left them lowercased here since that reflects the form attribute name, although perhaps I should make that more explicit in the message?

andrewazores: Although, form attributes are actually case-insensitive - I suppose it doesn't really matter in that case. I can capitalize these messages if you do prefer.

andrewazores: Okay, scratch that - from what I can determine, the spec doesn't say if form parameters are case insensitive, so they should probably be assumed to be case-sensitive.

In that case my preference would be to leave the attribute names lower case in the error message, but to include a bit more detail in the message to indicate that we're talking about a form attribute of the given name being required. What do you think?

vic-ma: Hmm I thought capitalized would be better since there are periods, but that's a good point. In that case maybe the attributes should be quoted? It looks like that's how it's done for the RulesPostHandler:
```
{"meta":{"type":"text/plain","status":"Bad Request"},"data":{"reason":"\"targetAlias\" cannot be blank, was \"null\""}}
```

vic-ma: Just saw your last comment. Yes I think lower case makes sense, and probably with quotes it should be obvious enough.

vic-ma (approved):

PR #516, opened 2021-06-16
[autorules] Credentials deletion enhancements

andrewazores:
This addresses the following review deficiencies from #416 :

> Maybe the response could specify if credentials were found and deleted, or if there were no credentials to be deleted? Not super important, but might be helpful.

~~(TODO):~~
> When a rule for a target is running, and I delete the credentials for that target, the archiver is still able to archive recordings, despite the fact that JMX auth is required for this, and the saved JMX auth credentials have been deleted. I see that this is because in the code, the credentials are given to the PeriodicArchiver in the constructor and saved to the object, not retreived from the CredentialsManager each time it is used. This is more efficient, but I'm wondering if it makes sense from the user's point of view. If they deleted some credentials, would they expect for CJFR to not have a copy of those credentials stored anywhere, anymore? Or at least not be able to use it?

### DIFF
--- a/rulestest.sh
+++ b/rulestest.sh
@@ -42,7 +42,7 @@ curl -k \
     -F targetAlias="es.andrewazor.demo.Main" \
     -F description="This is a test rule" \
     -F eventSpecifier="template=Continuous,type=TARGET" \
-    -F archivalPeriodSeconds="60" \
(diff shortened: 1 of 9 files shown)


andrewazores:
@vic-ma this is ready

vic-ma (approved):

Issue #520, opened 2021-06-17
[Epic] Explore non-JMX connection options


andrewazores:
This is a high-level issue representing the desire to find non-JMX connection options. The frontrunner idea here is a Cryostat agent attached to the target JVMs, which may use a pull-based HTTP API or more likely some HTTP push protocol. This would be a complement to JMX for users/deployments where JMX is undesirable and attaching a JVM agent is feasible.

- [x] #1309

andrewazores:
Pushing to 2.2.0.

andrewazores:
See https://github.com/cryostatio/cryostat-agent. This isn't complete yet but underway.

Issue #495, opened 2021-06-04
API changes for ServiceRef/batch operations


ebaron:
Let's discuss how to modify the HTTP API to support some new use-cases:
- Syncing recordings with operator (https://github.com/cryostatio/cryostat-operator/issues/205)
- Operations on Deployments/ReplicaSets/Containers (related: https://github.com/cryostatio/cryostat/issues/421)

In both cases, we've determined that the target ID being a JMX service URL is insufficient. For the operator syncing case, it's not possible to efficiently map a JMX service URL to a Pod. For the operations on Deployments/ReplicaSets, there would be multiple JMX service URLs per object. We need some other way to refer to a Deployment/ReplicaSet in the HTTP API.

andrewazores:
As it is, the Operator does its own independent discovery of targets, right? But the implementation mirrors the way Cryostat does it, so the FlightRecorder objects and the Cryostat API targets list happen to agree. If I understand the syncing and ID issue correctly, having Cryostat assign a unique ID that it can map back to a Deployment/Pod/Container still doesn't help the Operator, because the Operator wouldn't know about those IDs - it just knows ex. the Deployment's name, labels, the namespace it's in, etc.

Is there some way we can reconcile this difference? Perhaps if we can come up with a way to generate the IDs from the resource type and some specific fields, then both the Operator and Cryostat can agree on the IDs as well, without the Operator having to query Cryostat for target discovery and then maintain that result as internal state for mapping back any new/changed CRs. Does that make sense? ex. an ID for a Deployment could look like `deployment-${hash(deployment.namespace, deployment.name)}`.

ebaron:
Perhaps we could use literals instead of a hash. Then it's trivial to map the IDs back to their objects. We can use underscores as separators as they're not valid in Kubernetes names [1].

Something like:
`[kind]_[namespace]_[name](_containerName)?`

We could make the container name optional if we need to refer to a specific container. The container name is mandatory and must be unique within the Pod. [2]

[1] https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
[2] https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core

andrewazores:
[...]

Issue #281, opened 2020-09-30
DefaultPlatformClient should support discovery by hsperfdata


andrewazores:
Currently, the DefaultPlatformClient only supports discovery of targets by using JDP. This should be extended to also include targets from the `sun.jvmstat.monitor` API. JMC does this in `LocalJVMToolkit`, for reference, which may be adaptable to our platform client.

Issue #280, opened 2020-09-30
Additional fields for ServiceRefs


andrewazores:
Currently, ServiceRefs only have a Service URL and an Alias. This should be extended with additional information to help users identify and distinguish targets. Any field additions must be generic enough to be applicable to any platform, even if the value would be optional/null/blank.

Examples:

- container/pod name
- list of labels. In Kubernetes/OpenShift, these could be populated with Kubernetes/OpenShift labels. In Docker/Podman it could include the pod ID, hostname, pod config labels, etc.
- list of annotations. Similar to above
- start time
- PID (probably only useful in a local baremetal JVM case, but still applicable anywhere)

PR #521, opened 2021-06-17
Loosen ServiceRef JMXServiceURL coupling

andrewazores:
Related to #520 #495 #280 #281

Use (absolute) URIs in most ServiceRef API touch-points rather than JMXServiceURLs to reduce coupling and ease future explorations for non-JMX connections. JMXServiceURLs are still required in other places, namely the TargetConnectionManager only knows how to deal with JMXServiceURLs.


ebaron (approved):
Looks good!

PR #492, opened 2021-06-03
feat(platform): deployment scenario tree view

Alexjsenn:
Continuing from #463 

Depends on #573
Fixes #421
Related to #495

PR #523, opened 2021-06-17
Additional fields for ServiceRefs

### LINKED ISSUE #280: Additional fields for ServiceRefs

andrewazores:
Rebased on top of #538

Fixes #280

Related to #495

Sample output from `http :8181/api/v1/targets | jq`:
```json
[
  {
    "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9096/jmxrmi",
    "alias": "/deployments/quarkus-run.jar",
    "labels": {},
    "annotations": {
      "platform": {},
      "cryostat": {
        "HOST": "cryostat",
        "PORT": "9096",
        "JAVA_MAIN": "/deployments/quarkus-run.jar"
      }
    }
  },
  {
    "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9095/jmxrmi",
    "labels": {},
    "annotations": {
      "platform": {},
      "cryostat": {
        "HOST": "cryostat",
        "PORT": "9095",
        "JAVA_MAIN": "es.andrewazor.demo.Main"
      }
    }
  }
]
```

### DIFF
--- a/src/main/java/io/cryostat/platform/ServiceRef.java
+++ b/src/main/java/io/cryostat/platform/ServiceRef.java
@@ -41,7 +41,6 @@ import java.net.URI;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 
 import com.google.gson.annotations.SerializedName;
@@ -182,36 +181,15 @@ public class ServiceRef {
     }
 
     public enum AnnotationKey {
-        HOST("host"),
-        PORT("port"),
-        JAVA_MAIN("javaMain"),
-        PID("pid"),
-        START_TIME("startTime"),
-        NAMESPACE("namespace"),
-        SERVICE_NAME("serviceName"),
-        CONTAINER_NAME("containerName"),
-        POD_NAME("podName"),
+        HOST,
+        PORT,
+        JAVA_MAIN,
+        PID,
+        START_TIME,
+        NAMESPACE,
+        SERVICE_NAME,
+        CONTAINER_NAME,
+        POD_NAME,
         ;
-
-        public static final String PREFIX = "target.cryostat.io";
-
-        private final String key;
-
-        AnnotationKey(String key) {
-            this.key = String.format("%s/%s", PREFIX, key);
-        }
-
-        public String getKey() {
-            return key;
-        }
-
-        public AnnotationKey fromString(String s) {
-            for (AnnotationKey ak : values()) {
-                if (Objects.equals(ak.getKey(), s)) {
-                    return ak;
-                }
-            }
-            return null;
-        }
     }
 }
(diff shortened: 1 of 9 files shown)


src/main/java/io/cryostat/platform/ServiceRef.java
```diff
@@ -63,18 +69,111 @@ public URI getServiceUri() {
...
+    }
+
+    public Map<String, String> getLabels() {
+        return Collections.unmodifiableMap(labels);
+    }
+
+    public void setPlatformAnnotations(Map<String, String> annotations) {
+        this.annotations.platform.clear();
+        this.annotations.platform.putAll(annotations);
+    }
```
ebaron, line 84: Is this implemented yet? I assume this would contain Kubernetes annotations?

andrewazores: Yes exactly, but no it is not implemented yet for Kubernetes (or any platform). The `labels` field of the ServiceRef is also intended to contain labels provided by the platform but is not yet implemented anywhere.

src/main/java/io/cryostat/platform/internal/KubeApiPlatformClient.java
```diff
@@ -171,13 +173,25 @@ private boolean isCompatiblePort(EndpointPort port) {
...
+                                                addr.getTargetRef().getName());
+                                serviceRef.setCryostatAnnotations(
+                                        Map.of(
+                                                AnnotationKey.HOST, addr.getIp(),
+                                                AnnotationKey.PORT,
+                                                        Integer.toString(port.getPort()),
+                                                AnnotationKey.NAMESPACE,
+                                                        addr.getTargetRef().getNamespace(),
+                                                AnnotationKey.SERVICE_NAME,
+                                                        addr.getTargetRef().getName()));
```
ebaron, line 193: This should be `POD_NAME`.

andrewazores: Right, done.

src/main/java/io/cryostat/platform/ServiceRef.java
andrewazores: For context, the reason it isn't implemented anywhere is because in this PR, nothing in Cryostat knows about Pods, or any other Kubernetes resource than Endpoints really. Once this is merged and #492 is rebased on top, then we'll have a pretty clean way to integrate these two things together and apply the labels/annotations from the target Pod to these ServiceRef fields.

ebaron (approved):
Looks good!

PR #524, opened 2021-06-18
Add JMX connection status events

hareetd:
Related #461 

Add custom JFR events for monitoring JMX connection status (opened, closed and failed). 

Example output:

Screenshot from 2021-06-18 13-34-14

### DIFF
--- a/src/main/extras/app/Cryostat.xml
+++ b/src/main/extras/app/Cryostat.xml
@@ -3,6 +3,30 @@
      see Window -> Flight Recorder Template Manager.
 -->
 <configuration version="2.0" label="Cryostat" description="Continuous self-profiling for Cryostat" provider="Cryostat">
+ <event name="io.cryostat.net.TargetConnectionManager.JMXConnectionOpened">
+  <setting name="enabled">
+   true
+  </setting>
+  <setting name="threshold">
+   0 ms
+  </setting>
+ </event>
+ <event name="io.cryostat.net.TargetConnectionManager.JMXConnectionClosed">
+  <setting name="enabled">
+   true
+  </setting>
+  <setting name="threshold">
+   0 ms
+  </setting>
+ </event>
+ <event name="io.cryostat.net.TargetConnectionManager.JMXConnectionFailed">
+  <setting name="enabled">
+   true
+  </setting>
+  <setting name="threshold">
+   0 ms
+  </setting>
+ </event>
  <event name="io.cyostat.net.web.WebServer.WebServerRequestEvent">
   <setting name="enabled">
    true
--- a/src/main/extras/app/Cryostat.xml
+++ b/src/main/extras/app/Cryostat.xml
@@ -19,14 +19,6 @@
    0 ms
   </setting>
  </event>
- <event name="io.cryostat.net.TargetConnectionManager.JMXConnectionFailed">
-  <setting name="enabled">
-   true
-  </setting>
-  <setting name="threshold">
-   0 ms
-  </setting>
- </event>
  <event name="io.cyostat.net.web.WebServer.WebServerRequestEvent">
   <setting name="enabled">
    true
--- a/src/main/java/io/cryostat/net/TargetConnectionManager.java
+++ b/src/main/java/io/cryostat/net/TargetConnectionManager.java
@@ -90,7 +90,8 @@ public class TargetConnectionManager {
                                     public void onRemoval(
                                             ConnectionDescriptor descriptor,
                                             JFRConnection connection,
-                                            RemovalCause cause) {
+                                            RemovalCause cause)
+                                            throws RuntimeException {
                                         if (descriptor == null) {
                                             logger.warn(
                                                     "Connection eviction triggered with null descriptor");
@@ -109,8 +110,9 @@ public class TargetConnectionManager {
                                         evt.begin();
                                         try {
                                             connection.close();
-                                        } catch (Exception e) {
+                                        } catch (RuntimeException e) {
                                             evt.setExceptionThrown(true);
+                                            throw e;
                                         } finally {
                                             evt.end();
                                             if (evt.shouldCommit()) {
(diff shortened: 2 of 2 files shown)


src/main/java/io/cryostat/net/TargetConnectionManager.java
```diff
@@ -164,16 +176,80 @@ private JFRConnection attemptConnectAsHostPortPair(ConnectionDescriptor connecti
...
+                    jfrConnectionToolkit
+                            .get()
+                            .connect(
+                                    url,
+                                    credentials.orElse(null),
+                                    Collections.singletonList(
+                                            () -> this.connections.invalidate(cacheKey)));
+            evtOpened.end();
+            evtFailed.end();
+            if (evtOpened.shouldCommit()) {
```
andrewazores, line 196: I would move this into the `finally` block, so that the Opened event is always recorded even if the connection open actually fails. Then, in the exception handler block, either information about the exception can be optionally included in the Opened event, or the Failed event you've created can be used instead. I think in that case the Failed event doesn't need to be begin()/end()ed - it can just be an instantaneous event that occurs at the end of the timespan of an Opened event.

However, to be more consistent, I think including information about the exception in the Opened event might be the better way to go. This would better match @jan-law 's recent event addition(s).

src/main/java/io/cryostat/net/TargetConnectionManager.java
```diff
@@ -164,16 +176,80 @@ private JFRConnection attemptConnectAsHostPortPair(ConnectionDescriptor connecti
...
     }
+
+    @Name("io.cryostat.net.TargetConnectionManager.JMXConnectionOpened")
+    @Label("JMX Connection Status")
+    @Category("Cryostat")
+    @SuppressFBWarnings(
+            value = "URF_UNREAD_FIELD",
+            justification = "The event fields are recorded with JFR instead of accessed directly")
+    public static class JMXConnectionOpened extends Event {
+        String JMXServiceURL;
```
andrewazores, line 221: Variable/field names like this should be named starting with lowercase letters, so in this case `jmxServiceUrl` or maybe `jmxServiceURL`.

However, since #521, I would also recommend changing this to `serviceUri` or something instead to be a bit more future-proof. Same goes for the same-named field in the Closed event.

src/main/java/io/cryostat/net/TargetConnectionManager.java
hareetd: I followed what @jan-law did and added an `exceptionThrown` boolean field to my `JMXConnectionOpened` event but on second thought, would a boolean named `connectionFailed` be more appropriate?

andrewazores: I don't think it matters too much - if the connection failed it's because an exception was thrown, so they seem pretty equally valid to me. Just for consistency's sake again I would think leaving it as `exceptionThrown` is perhaps better.

src/main/java/io/cryostat/net/TargetConnectionManager.java
```diff
@@ -99,7 +104,14 @@ public void onRemoval(
                                         logger.info(
                                                 "Removing cached connection for {}",
                                                 descriptor.getTargetId());
+                                        JMXConnectionClosed evt =
+                                                new JMXConnectionClosed(descriptor.getTargetId());
+                                        evt.begin();
                                         connection.close();
```
andrewazores, line 110: `connection.close()` may still throw an (unchecked) exception here, so wrapping this in `try/catch/finally` to ensure the event is recorded would be best. The exception can simply be rethrown after it is caught and the failure flagged in the event.

src/main/extras/app/Cryostat.xml
```diff
@@ -3,6 +3,30 @@
...
+ </event>
+ <event name="io.cryostat.net.TargetConnectionManager.JMXConnectionClosed">
+  <setting name="enabled">
+   true
+  </setting>
+  <setting name="threshold">
+   0 ms
+  </setting>
+ </event>
+ <event name="io.cryostat.net.TargetConnectionManager.JMXConnectionFailed">
```
andrewazores, line 22: This can be removed now that the event definition has been removed.

src/main/java/io/cryostat/net/TargetConnectionManager.java
hareetd: I don't think an exception is thrown judging from the `JFRConnection` implementation in `cryostat.core`:

https://github.com/cryostatio/cryostat-core/blob/main/src/main/java/io/cryostat/core/net/JFRConnection.java#L199

From what I can tell, the exception is handled and not thrown upon closing.

andrewazores: That's an internal implementation detail of -core, which is true right now but could change in the future.

hareetd: I flagged the failure in the event but when I tried to throw an exception I got the error `Exception Exception is not compatible with throws clause in RemovalListener<ConnectionDescriptor,JFRConnection>.onRemoval(ConnectionDescriptor, JFRConnection, RemovalCause)`.

It seems you can't throw a checked exception if the original method being overridden doesn't throw an exception as well, but for our purposes simply flagging the failure should be enough, right?

andrewazores: We don't want to modify any control flow or other behaviour while adding these events, so swallowing the exception and not re-throwing it is not desirable. If you change your catch clause to catch and rethrow `RuntimeException` instead of simply `Exception` then I think it will do what we need.

src/main/java/io/cryostat/net/TargetConnectionManager.java
```diff
@@ -90,7 +90,8 @@
                                     public void onRemoval(
                                             ConnectionDescriptor descriptor,
                                             JFRConnection connection,
-                                            RemovalCause cause) {
+                                            RemovalCause cause)
+                                            throws RuntimeException {
```
andrewazores, line 94: I don't think this line is required, is it? You shouldn't need to declare that you throw unchecked exceptions.

hareetd: You're right, I'll go head and change that.

andrewazores (approved):

PR #526, opened 2021-06-21
Conditionally disable web-client testing on minimal builds

### LINKED ISSUE #525: Conditionally disable ClientAssetsIT on minimal builds
Currently, `mvn -Dcryostat.minimal=true verify` will always fail because the ClientAssetsIT expects to find the web-client index.html, but in a minimal build the web-client is not built and this asset is missing. This is the expected behaviour, so the integration test should be configured to match.

hareetd:
Disable web-client testing, specifically `ClientAssetsIT`, when running integration tests on minimal builds.

Fixes #525

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -575,6 +575,20 @@
     <properties>
       <cryostat.imageVersion>${project.version}-minimal</cryostat.imageVersion>
     </properties>
+    <build>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-failsafe-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


andrewazores:
Nice, that's pretty simple and does the job well. If it's feasible I would like to see how this could work using the `@Disabled` annotation and its variations applied to the `ClientAssetsIT` class itself too - this way the tests will still show up in the test runner output, but they will be marked as skipped because of the minimal build profile. That isn't too large of a concern however, so if that ends up being impractical, I'm fine to accept this patch as-is.

hareetd:
Sounds good, I'll take a look into that.

andrewazores (approved):

PR #528, opened 2021-06-22
Configure jfr-datasource/grafana-dashboard in itests

andrewazores:
This PR adds configuration to the integration test pod setup to include jfr-datasource and grafana-dashboard, so that the integration test environment more closely resembles a realistic "full" Cryostat deployment. Some existing integration tests have been lightly updated to adjust for this.

Since it's fairly related to this change I have also updated the jfr-datasource and grafana-dashboard versions used in smoketest.sh, and I have also added configuration to the pom.xml to add convenient properties for skipping unit tests only or skipping integration tests only.


hareetd (approved):
Looks good to me!

jan-law:
LGTM!

PR #531, opened 2021-06-22
InterleavedExternalTargetRequestsIT waits and retries if too many targets observed in setup

andrewazores:
JDP occasionally discovers too many targets. This commit allows the integration test to wait and retry if this scenario is observed, in case JDP settles and removes spurious targets after some delay. If not then the setup will still fail.


jan-law:
I ran `mvn clean verify -DskipUTs=true` a couple times and only saw this error once. Is this related?
```
[ERROR] testOtherContainersFound  Time elapsed: 0.009 s  <<< FAILURE!
java.lang.AssertionError: 

Expected: <10>
     but: was <9>
	at itest.InterleavedExternalTargetRequestsIT.testOtherContainersFound(InterleavedExternalTargetRequestsIT.java:183)
```

andrewazores:
That's definitely related. I'm not sure how or why the JDP-detected list of targets would have passed the setup step and then increased again between that and the execution of `testOtherContainersFound`, however. I'll see if there's anything more I can do to make the test less flaky, but it seems like this is probably just something we have to deal with when using JDP multicast discovery, or else it's a bug in the JMC JDP discovery client and so the bug is down in our dependency stack.

andrewazores:
@jan-law I think I've done as much as I reasonably can at this level to try to smooth out any JDP-related jitter in the reported number of discovered targets. If it's failing for the reasons I suspect then it's still probably technically possible for it to fail with these changes, but the probability should be lower.

jan-law (approved):
Looks good to me!

PR #532, opened 2021-06-23
More itest refactoring, cleanup, slight stabilization

andrewazores:
This refactors the TestBase class into bases/StandardSelfTest, creates a new bases/ExternalTargetsTest, and sets the InterleavedExternalTargetRequestsIT and AutoRulesIT to extend ExternalTargetsTest. This is mostly just to clean up the test class and remove common setup/teardown logic from test logic, and it also allows AutoRulesIT to do the same timed-retry-polling on teardown as InterleavedExternalTargetRequestsIT already does, which may also help to stabilize JDP discovery-related intermittent failures.


jan-law (approved):
Looks good to me!

PR #537, opened 2021-06-23
Rename websocket notification channel URL path

### LINKED ISSUE #478: /api/v1/command should be renamed /api/v1/notifications
The URL for the websocket notification channel still uses the path `/api/v1/command`. The channel is now used for one-way notifications only, so the path should be renamed. Since it's part of the published 1.0 API, the existing path should still be exposed and result in a successful connection if clients attempt to use it.

hareetd:
Change the path from `api/v1/command` to `api/v1/notifications` since the websocket channel is now used for one-way notifications instead of commands. 

Fixes #478


andrewazores:
How does this handle the `Since it's part of the published 1.0 API, the existing path should still be exposed and result in a successful connection if clients attempt to use it.` requirement from the original issue?

hareetd:
Looks like I misunderstood the issue. Breaking that requirement down:

1. `the existing path should still be exposed`: does this mean the ClientUrlGetHandler should still return the path `api/v1/command`, or perhaps return both possible paths?

2. `result in a successful connection if clients attempt to use it`: I guess this means I should include functionality in the code such that both paths "work"

I'm also a bit confused because I thought with the deprecation of Commands we would want to remove any mention of them. Are we waiting to do that until the 2.0 API is published and completely replaces the 1.0 API?

andrewazores:
> the existing path should still be exposed: does this mean the ClientUrlGetHandler should still return the path api/v1/command, or perhaps return both possible paths?

It can just return the new path, as you've done. This is what clients are supposed to use to check what the notification channel URL is, so it should tell them what the newest/preferred place to go is.

> result in a successful connection if clients attempt to use it: I guess this means I should include functionality in the code such that both paths "work"

Right. Even though we want to stop referring to the notification channel as a command channel, since that is now a misleading name, it should also be cheap and easy for us to still support clients that try to connect on the old path. If a redirect (ex. HTTP 301) can be used here then that's probably ideal.

> I'm just a bit confused because I thought with the deprecation of Commands we would want to remove any mention of them. Are we waiting to do that until the 2.0 API is published and completely replaces the 1.0 API?

The Commands removal definitely won't be going into `v1`, since we already have a published 1.0.0 release, and so any further 1.x.y releases should avoid changing or removing anything that would break any potential end users' scripts that exist and were built against the 1.0.0 API. Breaking changes like that should only arrive in 2.0.0 (or any other major version release). That said, where it's possible, it would also be nice of us to make migration from 1.0->2.0 as easy as possible, so that could mean doing things like HTTP 301 redirects on renamed paths, so that any third-party clients/scripts will continue to work but users will (hopefully) see the 301 status and realize that they're using something that is changed/deprecated and they should prepare for it to be removed or otherwise more significantly altered (in a breaking way) in a future release.

andrewazores:
On the other hand, since `/api/v1/commands` in 1.0.0 implements an interactive command channel, and `/api/v1/notifications` in 2.0.0 implements a push notification channel, perhaps `/api/v1/commands` should not redirect the client but instead respond with a `410 GONE` response, so that the clients fail to connect and hopefully do so in an informative way for the end user.

hareetd:
Thanks for the explanation, it cleared up any misconceptions I had. 

I think a `410 GONE` would make more sense as it would (hopefully) funnel clients towards getting the new path from the updated handler. I'll go ahead and implement your other suggestions as well; will let you know if I have any other questions.

hareetd:
@andrewazores (not sure if you get notified for comments on draft PRs)

I'm almost done but had a few questions first:

1. I decided to rename `ClientUrlGetHandler` to `NotificationsUrlGetHandler`, instead of `NotificationsGetHandler` as you previously suggested, because it matches the `GrafanaDashboardUrlGetHandler` convention and I think it's a more accurate name. Is this fine or should I change it to `NotificationsGetHandler`?

2. For the HTTP_API.md, should I also change the synopsis for the updated handler, and if so, what to? 

3. Should I add a new test in `MessagingServerTest` to check if `api/v1/command` rejects the websocket connection with a `410 GONE` status code? I'm leaning towards no since all the test would check is an `if` statement.

andrewazores:
Yes, I still get notified on drafts.

1. No, what you've done makes sense and sounds good.
2. I think just removing/changing mentions of "command channel" to "notification channel" or similar should be fine?
3. A test probably does make sense even though it seems so simple. Doing it as an integration-test might be even better here.

hareetd:
I've finished writing up the integration test; had a few difficulties extracting the HTTP status code from the thrown exception(s) so I settled with comparing the error message instead.

src/main/java/io/cryostat/messaging/MessagingServer.java
```diff
@@ -95,9 +95,13 @@ public void start() throws SocketException, UnknownHostException {
 
         server.websocketHandler(
                 (sws) -> {
-                    if (!"/api/v1/command".equals(sws.path())) {
+                    if (!"/api/v1/command".equals(sws.path())
+                            && !"/api/v1/notifications".equals(sws.path())) {
                         sws.reject(404);
                         return;
+                    } else if (!"/api/v1/notifications".equals(sws.path())) {
```
andrewazores, line 102: Isn't this effectively equivalent to `if ("/api/v1/notifications".equals(sws.path())` since it's an `else if` of the prior condition? I think as-is it's a little bit more confusing to read through than it needs to be, ie if someone else comes along later trying to figure out when exactly the `410` response is sent.

In other words,

```
if path is "/api/v1/command" then reject 410
else if path is not "/api/v1/notifications" then reject 404
```

seems to me like it would be more readable. What do you think?

hareetd: Definitely more readable, will change.

src/test/java/itest/MessagingServerIT.java
```diff
@@ -0,0 +1,113 @@
...
+                    if (ar.failed()) {
+                        future.completeExceptionally(ar.cause());
+                        return;
+                    }
+                    WebSocket ws = ar.result();
+
+                    ws.handler(
+                            m -> {
+                                JsonObject resp = m.toJsonObject();
+                                String commandName = resp.getString("commandName");
```
andrewazores, line 76: I think we can simplify this test setup a bit. We don't actually expect the messaging server to act upon the command request, so there isn't really a need to send it a well-formed command message and check that we got a well-formed response that matches what we requested, right? The connection rejection will happen before any of that occurs, so the test case doesn't need to be set up to send nice messages or process responses.

hareetd: Sounds good, so for future reference, tests should only consist of what we actually expect them to execute and not the full range of possibilities?

andrewazores: Well, there's a balance. The full range of possibilities is almost literally any behaviour, after all, but in the test cases you wrote, you (very reasonably) handled a lot of those with "if any error occurs, just throw the exception and move on", for example. In this case, what we're interested in is that we get a 410 response, and any other behaviour is incorrect from the point of view of this test. There isn't too much need to be specific about behaviours that are unexpected/incorrect - so long as the behaviour is reported with enough detail to debug the test failure. In this particular example, the test just failing and saying "I expect a 410 response but instead, I got a 200" (or whatever else) seems sufficient - no need to go into "I expect a 410 response, but got a 200, but also the 200 was for the correct command response"

hareetd: Ah makes sense, thank you.

src/test/java/itest/MessagingServerIT.java
```diff
@@ -0,0 +1,81 @@
...
+            throws InterruptedException, ExecutionException, TimeoutException {
+        CompletableFuture<JsonObject> future = new CompletableFuture<>();
+
+        Utils.HTTP_CLIENT.webSocket(
+                DEPRECATED_COMMAND_PATH,
+                ar -> {
+                    if (ar.failed()) {
+                        future.completeExceptionally(ar.cause());
+                        return;
+                    }
```
andrewazores, line 66: Just in case, it would be best if the `future` was still completed here in an `else`. Otherwise, if the request doesn't fail, the test only proceeds once the `.get()` call lower down times out. You can just complete it with `null` in this scenario.

src/test/java/itest/MessagingServerIT.java
```diff
@@ -0,0 +1,81 @@
...
+        return future;
+    }
+
+    @Test
+    public void shouldRejectDeprecatedCommandPathWith410StatusCode() {
+        try {
+            sendMessage("ping").get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            String errorMessage = "Websocket connection attempt returned HTTP status code 410";
+            MatcherAssert.assertThat(e.getCause().getMessage(), Matchers.equalTo(errorMessage));
```
andrewazores, line 78: When you're writing a test that expects an exception to occur, it's best to write it like this:

```
SomeException ex = Assertions.assertThrows(SomeException.class, () -> doSomething());
MatcherAssert.assertThat(ex, Matchers.hasSomeProperties());
```

This way, the test will immediately fail if the expected exception is *not* thrown. Simply using a `try/catch` the typical/intuitive way means that if no exception is thrown, the test will skip the `catch` clause and silently pass, never hitting any assertions along the way, possibly resulting in a false positive.

andrewazores (commented):
Looks good! Again, I will hold off on approving/merging this until there is a corresponding change in -web. Are you interested in taking a look at that yourself? It should be a pretty small patch.

hareetd:
Sure, I can do that.

It looks like `NotificationChannel.service.tsx` in -web still references `clienturl`; I can open an issue and take care of that.

Were there any other corresponding changes you had in mind?

andrewazores:
No, that's all that comes to mind.

hareetd:
Just double-checking, but the merge conflict is because the web-client is pointing to a different commit than the one on the main branch, right?

andrewazores:
Yes, something like that. I think the easiest way to resolve it would be to `git fetch upstream`, then `git rebase -i upstream/main` and drop your commit where you updated the web-client commit. This should remove the conflict. Then you can update the web-client and commit that again, and hopefully that would apply cleanly.

hareetd:
Looks like it worked. I guess in the future it's best to rebase at the end to prevent merge conflicts? I've only been rebasing my branches right before opening the PR.

andrewazores:
Usually you don't need to do a rebase before merging - if you do, you'll see the merge conflict status like it was here. This time it only happened because git couldn't figure out how to apply the different changes to the web-client from both of the changed branches. If your PR can be replayed on top of main, and there are either no conflicts or git can figure out how to resolve them automatically, then the pre-merge rebase is typically unnecessary. However, for a submodule commit hash change, there's no way to figure out how to merge those two changes - it has to just be one or the other, so we see the conflicting status.

andrewazores (approved):

PR #538, opened 2021-06-24
New rule definitions trigger on existing targets

andrewazores:
With this change, adding a new rule definition causes that rule to become immediately activated for any matching targets that are already known. Prior to this change adding a new rule definition would only affect targets that (re)appear after the rule was added.

Example:

1. Run smoketest.sh
2.
```
$ http --verify=false -f https://0.0.0.0:8181/api/v2/rules name=foorule description="Rule for testing" targetAlias=es.andrewazor.demo.Main eventSpecifier="template=Continuous"
HTTP/1.1 201 Created
content-encoding: gzip
content-length: 94
location: /api/v2/rules/foorule

{"meta":{"type":"text/plain","status":"Created"},"data":{"result":"foorule"}}
```

3. Check that the vertx sample app listening on 9093 has a new recording named auto_foorule. The other vertx apps on 9094 and 9095 require credentials, so they will not have matching recordings, but Cryostat's logs should reveal that it attempted to start recordings on them and failed due to the lack of credentials.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
@@ -39,7 +39,6 @@ package io.cryostat.net.web.http.api.v1;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
-import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
@@ -48,26 +47,22 @@ import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Provider;
 
-import org.openjdk.jmc.common.unit.IConstrainedMap;
 import org.openjdk.jmc.common.unit.QuantityConversionException;
-import org.openjdk.jmc.flightrecorder.configuration.events.EventOptionID;
 import org.openjdk.jmc.flightrecorder.configuration.recording.RecordingOptionsBuilder;
-import org.openjdk.jmc.rjmx.services.jfr.IEventTypeInfo;
 import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
 
-import io.cryostat.commands.internal.EventOptionsBuilder;
 import io.cryostat.commands.internal.RecordingOptionsBuilderFactory;
 import io.cryostat.core.net.JFRConnection;
-import io.cryostat.core.templates.Template;
 import io.cryostat.core.templates.TemplateType;
 import io.cryostat.jmc.serialization.HyperlinkedSerializableRecordingDescriptor;
-import io.cryostat.messaging.notifications.NotificationFactory;
 import io.cryostat.net.AuthManager;
+import io.cryostat.net.ConnectionDescriptor;
 import io.cryostat.net.TargetConnectionManager;
 import io.cryostat.net.web.WebServer;
 import io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler;
 import io.cryostat.net.web.http.HttpMimeType;
 import io.cryostat.net.web.http.api.ApiVersion;
+import io.cryostat.recordings.RecordingCreationHelper;
 
 import com.google.gson.Gson;
 import io.vertx.core.MultiMap;
@@ -76,45 +71,31 @@ import io.vertx.core.http.HttpMethod;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 
 public class TargetRecordingsPostHandler extends AbstractAuthenticatedRequestHandler {
 
-    // TODO refactor this to use the RecordingCreationHelper after PR #486 is merged
-    public static final Template ALL_EVENTS_TEMPLATE =
-            new Template(
-                    "ALL",
-                    "Enable all available events in the target JVM, with default option values. This will be very expensive and is intended primarily for testing Cryostat's own capabilities.",
-                    "Cryostat",
-                    TemplateType.TARGET);
-
-    private static final Pattern TEMPLATE_PATTERN =
-            Pattern.compile("^template=([\\w]+)(?:,type=([\\w]+))?$");
-
     static final String PATH = "targets/:targetId/recordings";
     private final TargetConnectionManager targetConnectionManager;
+    private final RecordingCreationHelper recordingCreationHelper;
     private final RecordingOptionsBuilderFactory recordingOptionsBuilderFactory;
-    private final EventOptionsBuilder.Factory eventOptionsBuilderFactory;
     private final Provider<WebServer> webServerProvider;
     private final Gson gson;
-    private final NotificationFactory notificationFactory;
-    private static final String NOTIFICATION_CATEGORY = "RecordingCreated";
 
     @Inject
     TargetRecordingsPostHandler(
             AuthManager auth,
             TargetConnectionManager targetConnectionManager,
+            RecordingCreationHelper recordingCreationHelper,
             RecordingOptionsBuilderFactory recordingOptionsBuilderFactory,
-            EventOptionsBuilder.Factory eventOptionsBuilderFactory,
             Provider<WebServer> webServerProvider,
-            Gson gson,
-            NotificationFactory notificationFactory) {
+            Gson gson) {
         super(auth);
         this.targetConnectionManager = targetConnectionManager;
+        this.recordingCreationHelper = recordingCreationHelper;
         this.recordingOptionsBuilderFactory = recordingOptionsBuilderFactory;
-        this.eventOptionsBuilderFactory = eventOptionsBuilderFactory;
         this.webServerProvider = webServerProvider;
         this.gson = gson;
-        this.notificationFactory = notificationFactory;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/io/cryostat/rules/RuleProcessor.java
```diff
@@ -120,91 +118,84 @@ public synchronized void disable() {
     public synchronized void onEvent(Event<RuleEvent, Rule> event) {
         switch (event.getEventType()) {
             case ADDED:
+                platformClient.listDiscoverableServices().stream()
+                        .filter(serviceRef -> registry.applies(event.getPayload(), serviceRef))
+                        .forEach(serviceRef -> activate(event.getPayload(), serviceRef));
                 // FIXME the processor should also be able to apply new rules to targets that have
```
jan-law, line 124: What else would need to be changed in order to address the `fixme`?

andrewazores: Ah - nothing. Whoops.

jan-law:
Checking my understanding: Is the purpose of the `RecordingCreationHelper` to create a new recording, aka apply a new rule, to matching targets without going through the web client?

andrewazores:
Pretty much, yes. The API handler for creating recordings used to process the POST request and form and then also handle all of the actual logic of creating the recording. It made sense to extract that out to somewhere reusable so that automatic rules can perform the same logic, but without having to do an HTTP request to ourselves and deal with a nested request/response (client -> cryostat -> cryostat -> client).

src/test/java/io/cryostat/rules/RuleProcessorTest.java
```diff
@@ -122,25 +125,28 @@ void testDisableRemovesProcessorAsDiscoveryListener() {
...
-                                Mockito.when(res.failed()).thenReturn(false);
-                                Mockito.when(res.result()).thenReturn(response);
-                                ((Handler<AsyncResult>) invocation.getArgument(1)).handle(res);
-                                return null;
-                            }
-                        })
-                .when(request)
-                .sendMultipartForm(Mockito.any(), Mockito.any());
+        RecordingOptionsBuilder recordingOptionsBuilder =
+                Mockito.mock(RecordingOptionsBuilder.class);
```
jan-law, line 129: Quick question, why are `recordingOptionsBuilder` and `recordingOptions` created with `Mockito.mock` instead of `@Mock`?

andrewazores: `@Mock` applies to fields, whereas here `Mockito.mock` is creating a mock that is scoped within the test method. The variable could be pulled up to the field level, but that kind of pollutes the scope if the field isn't actually used in all/most methods.

andrewazores:
Rebased, the only conflict was just imports in the MainModule.

src/main/java/io/cryostat/recordings/RecordingCreationHelper.java
```diff
@@ -0,0 +1,194 @@
...
+                                        "Recording with name \"%s\" already exists",
+                                        recordingName));
+                    }
+                    IRecordingDescriptor desc =
+                            connection
+                                    .getService()
+                                    .start(
+                                            recordingOptions,
+                                            enableEvents(connection, templateName, templateType));
+                    notificationFactory
```
jan-law, line 104: Since automatic rules isn't part of the web client yet, who receives these notifications?

andrewazores: Nobody, really. You could observe them using `websocat` if you wanted to. Otherwise, the notification will flow through to the messaging server and then be essentially dropped as there are no connected clients to write to.

andrewazores: Oh wait - this is a notification from the `RecordingCreationHelper`. So this is a notification with category `RecordingCreated`, which the web-client does have handling for - it's used to refresh the recordings list view. If you're connected to Cryostat via the web-client and are looking at the recordings list view for a target, and Cryostat creates a recording on that target due to some new Rule being added (at this point this probably means you just did that yourself manually with an API request), you'll see the web-client refresh itself to see the new recording appear.

andrewazores: Just to clarify, what this means is that from any notifications channel client's perspective, any `RecordingCreated` notification is indistinguishable. You could connect using the web-client and observe that it receives these notifications:

1. when the same web-client instance performs an API request to create a recording
2. when a _different_ web-client instance connected to the same backend performs an API request to create a recording
3. when the Operator performs an API request to create a recording (assuming this Cryostat backend is deployed by an Operator)
4. when some user, yourself or otherwise, uses curl/httpie to perform an API request to create a recording
5. when Cryostat itself decides to create a recording, because of automatic rule processing or hypothetically any other reason

etc. This is generally how the whole notifications channel system works - Cryostat emits messages when it does certain things, regardless of the reason that thing was done, and any clients who happen to be listening at that time are pushed a notification to inform them of the change.

jan-law (commented):
Looks good! Will wait until @hareetd has had a chance to review this thoroughly

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
```diff
@@ -150,18 +131,11 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
         }
 
         try {
-            Optional<HyperlinkedSerializableRecordingDescriptor> descriptor =
+            ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
+            HyperlinkedSerializableRecordingDescriptor linkedDescriptor =
```
hareetd, line 135: This is for my own understanding; if this `HyperlinkedSerializableRecordingDescriptor ` instance is no longer `Optional<>` does that mean the `linkedDescriptor` can never be `null`? What implications does that have on the code flow, if any?

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
```diff
@@ -185,57 +159,34 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
...
-                    linkedDescriptor -> {
-                        ctx.response().setStatusCode(201);
-                        ctx.response().putHeader(HttpHeaders.LOCATION, "/" + recordingName);
-                        ctx.response()
-                                .putHeader(HttpHeaders.CONTENT_TYPE, HttpMimeType.JSON.mime());
-                        ctx.response().end(gson.toJson(linkedDescriptor));
-                    },
-                    () -> {
-                        throw new HttpStatusException(
-                                500, "Unexpected failure to create recording");
```
hareetd, line 237: This exception for when a recording fails unexpectedly seem to have been removed and hasn't been extracted out to the `RecordingCreationHelper` either. Is this because it isn't possible for a recording to fail for any reason other than those covered by the exceptions that were kept in?

src/main/java/io/cryostat/rules/RuleProcessor.java
```diff
@@ -120,91 +118,82 @@ public synchronized void disable() {
...
+        if (rule.getPreservedArchives() <= 0 || rule.getArchivalPeriodSeconds() <= 0) {
+            return;
+        }
+        tasks.put(
+                Pair.of(serviceRef, rule),
+                scheduler.scheduleAtFixedRate(
+                        periodicArchiverFactory.create(
+                                serviceRef, credentialsManager, rule, this::archivalFailureHandler),
+                        rule.getArchivalPeriodSeconds(),
+                        rule.getArchivalPeriodSeconds(),
```
hareetd, line 170: Does an archival period greater than 0 mean a rule is activated (i.e. recording started) once every period, after the previous recording (of the same rule) has been stopped and archived?

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
andrewazores: It doesn't guarantee that the `descriptor` is non-null, but it does signify that it isn't expected to be. If `getConnectionDescriptorFromContext` returns null then the `targetConnectionManager.executeConnectedTask` call will fail on the null descriptor, resulting in a NullPointerException that is eventually turned into an HTTP 500 response. The same thing would happen if the `descriptor` is non-null but contains a null `targetId` somehow (which shouldn't happen in practice - how/why would the path parameter be a null reference?).

The reason this simplification was done is because the implementation lower down used to make the call to create the recording, then query for it back (`getRecordingByName`). This was always expected to return back the descriptor for the just-created recording, but it returns an `Optional` since it's generally possible that the given name doesn't match any existing recording. In this refactor, `getRecordingByName` is no longer used - `startRecording` always returned the `descriptor` we needed, so we may as well just use that instead of re-querying for it immediately afterward.

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
andrewazores: The `ifPresentOrElse` handling relates back to the simplification of `Optional` above. If the `Optional` was somehow (not sure in what scenarios it might realistically arise) empty, then this exception would be thrown to signify that unexpected failure. Without the `Optional` above, any unexpected exception will already become an HTTP 500 by the WebServer's failureHandler, so there is no need for an equivalent `throw` anymore.

src/main/java/io/cryostat/rules/RuleProcessor.java
andrewazores: Recordings are always started immediately using the template specified by the rule. The recording is never stopped, but periodic copies of it are made and preserved. `archivalPeriodSeconds` dictates how often these copies are made, `preservedArchives` dictates how many copies remain in storage (the oldest ones are pruned from storage when the limit is reached).

hareetd (approved):
Looks good!

PR #539, opened 2021-06-24
Validate JSON document during auto-rules creation

### LINKED ISSUE #530: Auto-Rules JSON creation does not validate numeric values
https://github.com/cryostatio/cryostat/pull/416#issuecomment-865365655

When POSTing a rule definition, all of the same validations should be performed on the input whether it is supplied as a form or a a JSON document. Currently, the JSON document pathway does not validate numeric inputs the same way that the form pathway does.

jan-law:
Fixes #530 

@andrewazores I found a `JsonDeserializer` that can reuse `requireNonNegative()` and I have some questions about how to implement it. 

`private final Gson gson` needs to be modified in order to register the `Type Adapter` for `IntegerDeserializer`. Why is `gson` `final`?

`IntegerDeserializer` gets passed JSON values without their keys, which means I'm not sure how to pass the correct `Attribute` to `requireNonNegative()`.

`IntegerDeserializer` only works if our numeric values are of type `Integer`, not `int`. How will this affect performance?

Here's the output as of now:
```
$ vi testRule.json

{
   "name": "test",
   "description": "AutoRulesIT automated rule",
   "eventSpecifier": "template=Continuous,type=TARGET",
   "targetAlias": "es.andrewazor.demo.Main",
   "archivalPeriodSeconds": 60,
   "preservedArchives": -10,
   "maxAgeSeconds": 60
}

$ curl -v --insecure -H "Content-Type: application/json"  -d @testRule.json https://0.0.0.0:8181/api/v2/rules/
...
< HTTP/1.1 400 Bad Request
< content-length: 131
< 
{"meta":{"type":"text/plain","status":"Bad Request"},"data":{"reason":"\"archivalPeriodSeconds\" cannot be negative, was \"-10\""}}
```

Some of the other approaches I've thought of include parsing `params.getBody()` before calling `gson.fromJson(...)` or using a `Rule.Builder` similar to the JSON form pathway. I'd like to avoid parsing the JSON more than once though, and am not sure how to efficiently map key/value pairs from the `params.getBody()` string to the `Rule.Builder`.

### DIFF
--- a/src/main/java/io/cryostat/rules/Rule.java
+++ b/src/main/java/io/cryostat/rules/Rule.java
@@ -56,18 +56,16 @@ public class Rule {
     private final int maxSizeBytes;
 
     Rule(Builder builder) {
-        this.name = sanitizeRuleName(requireNonBlank(builder.name, Attribute.NAME));
+        this.name = sanitizeRuleName(builder.name);
         this.description = builder.description == null ? "" : builder.description;
-        this.targetAlias = requireNonBlank(builder.targetAlias, Attribute.TARGET_ALIAS);
-        this.eventSpecifier = requireNonBlank(builder.eventSpecifier, Attribute.EVENT_SPECIFIER);
-        this.archivalPeriodSeconds =
-                requireNonNegative(
-                        builder.archivalPeriodSeconds, Attribute.ARCHIVAL_PERIOD_SECONDS);
-        this.preservedArchives =
-                requireNonNegative(builder.preservedArchives, Attribute.PRESERVED_ARCHIVES);
+        this.targetAlias = builder.targetAlias;
+        this.eventSpecifier = builder.eventSpecifier;
+        this.archivalPeriodSeconds = builder.archivalPeriodSeconds;
+        this.preservedArchives = builder.preservedArchives;
         this.maxAgeSeconds =
                 builder.maxAgeSeconds > 0 ? builder.maxAgeSeconds : this.archivalPeriodSeconds;
         this.maxSizeBytes = builder.maxSizeBytes;
+        this.validate();
     }
 
     public String getName() {
@@ -109,7 +107,7 @@ public class Rule {
 
     static String sanitizeRuleName(String name) {
         // FIXME this is not robust
-        return name.replaceAll("\\s", "_");
+        return requireNonBlank(name, Attribute.NAME).replaceAll("\\s", "_");
     }
 
     private static String requireNonBlank(String s, Attribute attr) {
@@ -126,6 +124,15 @@ public class Rule {
                     String.format("\"%s\" cannot be negative, was \"%d\"", attr, i));
         }
         return i;
+    }
+
+    public void validate() throws IllegalArgumentException {
+
+        requireNonBlank(this.name, Attribute.NAME);
+        requireNonBlank(this.targetAlias, Attribute.TARGET_ALIAS);
+        requireNonBlank(this.eventSpecifier, Attribute.EVENT_SPECIFIER);
+        requireNonNegative(this.archivalPeriodSeconds, Attribute.ARCHIVAL_PERIOD_SECONDS);
+        requireNonNegative(this.preservedArchives, Attribute.PRESERVED_ARCHIVES);
     }
 
     @Override
--- a/src/main/java/io/cryostat/rules/Rule.java
+++ b/src/main/java/io/cryostat/rules/Rule.java
@@ -37,6 +37,10 @@
  */
 package io.cryostat.rules;
 
+import java.util.function.Function;
+
+import com.google.gson.JsonObject;
+import io.vertx.core.MultiMap;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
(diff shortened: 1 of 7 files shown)


andrewazores:
> private final Gson gson needs to be modified in order to register the Type Adapter for IntegerDeserializer. Why is gson final?

The gson instance, and generally all other constructor-assigned fields that are provided by dependency injection, because the field reference should never need to be overwritten. Changing the field reference implies that the injected dependency is no longer being used, which violates the purpose of injecting the common dependency. In the case of the Gson instance, it's because the MainModule provides a single instance so that we have uniform behaviour of (de)serialization across the application.

> IntegerDeserializer gets passed JSON values without their keys, which means I'm not sure how to pass the correct Attribute to requireNonNegative().

This makes me think this may not be the correct deserializer to use, then. If I'm understanding correctly, this is what Gson will use when deserializing something in JSON that looks like an integer value, so that you could instead map that to a different type than the Java Integer.

> IntegerDeserializer only works if our numeric values are of type Integer, not int. How will this affect performance?

It's probably not of much concern.

> I'd like to avoid parsing the JSON more than once though, and am not sure how to efficiently map key/value pairs from the params.getBody() string to the Rule.Builder.

Another possible path along this line of thought is to extract the validation that currently occurs in the Rule constructor (which is called by the Builder.build()) into a method like Rule.validate(). The Rule constructor can then call validate() at the end, after it has assigned all of its own fields. In the JSON parsing pathway of the handler, the handler can call rule.validate() on the Rule instance that is created by Gson.

However, the cleanest approach may be to use a deserializer/adapter, so that Gson itself knows how to deal with invalid Rule JSON definitions, and we don't have to ensure we call rule.validate() everywhere one may be deserialized. Our existing `GsonJmxServiceUrlAdapter` could be a good reference for this.

andrewazores:
> Another possible path along this line of thought is to extract the validation that currently occurs in the Rule constructor (which is called by the Builder.build()) into a method like Rule.validate(). The Rule constructor can then call validate() at the end, after it has assigned all of its own fields. In the JSON parsing pathway of the handler, the handler can call rule.validate() on the Rule instance that is created by Gson.

> However, the cleanest approach may be to use a deserializer/adapter, so that Gson itself knows how to deal with invalid Rule JSON definitions, and we don't have to ensure we call rule.validate() everywhere one may be deserialized. Our existing GsonJmxServiceUrlAdapter could be a good reference for this.

You know, now that I've said this, maybe the correct approach is to combine them. There should be a `Rule#validate()` method which the Rule constructor can call to ensure its own field validity, and this can also be called by the Gson adapter after deserializing the Rule JSON.

andrewazores:
This is probably for later enhancements in this same area:

https://stackoverflow.com/questions/21626690/gson-optional-and-required-fields

We could use an approach like this to create annotations and register adapters/deserializers to perform validations, like ensuring deserialized reference-type fields are non-null, or numeric types are positive, etc.

jan-law:
`gson` will successfully deserialize `-0` as `0`. Should we leave this as is?

andrewazores:
That's fine.

```
$ jshell
|  Welcome to JShell -- Version 11.0.11
|  For an introduction type: /help intro

jshell> -0 == 0
$1 ==> true
```

src/main/java/io/cryostat/util/GsonRuleAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.gson.JsonObject;
+
+public class GsonRuleAdapter implements JsonDeserializer<Rule> {
+
+    @Override
+    public Rule deserialize(JsonElement json, Type typeOf, JsonDeserializationContext context)
+            throws IllegalArgumentException {
+
+        JsonObject jsonObject = json.getAsJsonObject();
+        Rule rule = new Gson().fromJson(jsonObject, Rule.class);
```
jan-law: I'd like to sanitize the `name` field, except `sanitizeRuleName()` is private and also has a `fixme`. What will the sanitizer ideally look like? Could it be public somehow to do something like this?

```
public Rule deserialize(JsonElement json, Type typeOf, JsonDeserializationContext context)
            throws IllegalArgumentException {

        JsonObject jsonObject = json.getAsJsonObject();

        String name = Rule.Attribute.NAME.getSerialKey();
        jsonObject.add(name, Rule.sanitizeRuleName(jsonObject.get(name)));

        Rule rule = new Gson().fromJson(jsonObject, Rule.class);
        rule.validate();
        return rule;
    }
```

andrewazores: Sure, something like that would definitely work. There's no hard requirement for the method to be (package-)private if there is another good place that needs access to it, and this deserializer is a good place.

The FIXME is about doing more than just replacing whitespace characters with underscores, which is all the sanitization currently does. It should probably also disallow `.jfr` from appearing within a rule name for example.

Alternatively, the rule name validation could be done by the POST handler that processes the user's request to create a Rule definition. If the form or JSON document provided contains an invalid rule name then a `400` response is sent, ideally with some information about why the rule name failed validation. Then it's up to the user/client to try again with a valid name, rather than for the Rule implementation to try to sanitize the name it's given and turn it into something acceptable.

andrewazores: Also, btw, it works fine for testing right now, but I don't think a Gson adapter should invoke `new Gson().fromJson()` to perform the actual deserialization. Ideally the adapter should perform this deserialization on its own, or at worst there could be a static Gson instance that is shared and reused for this purpose.

src/main/java/io/cryostat/util/RuleDeserializer.java
```diff
@@ -61,9 +61,57 @@ public Rule deserialize(JsonElement json, Type typeOf, JsonDeserializationContex
...
-        return rule;
+        builder = setOptionalInt(builder, Rule.Attribute.ARCHIVAL_PERIOD_SECONDS, jsonObject);
+        builder = setOptionalInt(builder, Rule.Attribute.PRESERVED_ARCHIVES, jsonObject);
+        builder = setOptionalInt(builder, Rule.Attribute.MAX_AGE_SECONDS, jsonObject);
+        builder = setOptionalInt(builder, Rule.Attribute.MAX_SIZE_BYTES, jsonObject);
+
+        return builder.build();
+    }
+
+    private static Rule.Builder setOptionalInt(
```
jan-law, line 92: Would it be useful if we refactored `setOptionalInt()` from `RulesPostHandler.java` to avoid duplicating code here? `setOptionalInt()` would need to handle both `RequestParameter`s or `JsonObject`s

andrewazores: Yes, refactoring to avoid duplication would be very valuable. It isn't immediately obvious how to make it handle both `RequestParameters` and `JsonObject`s as a single method, but perhaps it can be done as an overloaded method (one signature for `RequestParameters` and another signature for `JsonObject`), which both call another common method that basically just contains the `switch` expression that is currently in the middle of the method body.

jan-law: Assuming the best place to put `setOptionalInt(JsonObject jsonObject)` is in `RulesPostHandler`, how can I make `setOptionalInt(JsonObject jsonObject)` visible to `RuleDeserializer` without making `RulesPostHandler` public?

andrewazores: That probably wouldn't be the best place, actually - API handlers should ideally be pretty small and just deal with processing request stuff - grabbing values from path and query params, headers, etc. - and writing responses that are computed elsewhere. Any other deeper business logic should be refactored out for reuse, so that nothing has dependencies on API handlers (except maybe other API handlers when they're very closely related.

jan-law: How about in `Rule.java`? That would move the logic closer to `Builder` and `requireNonNegative`

andrewazores: Sure, that sounds pretty reasonable

jan-law: `RequestParameters` is only visible in `api/v2/`. What's the best way to pass those parameters to `setOptionalInt()` in `Rule.java`?

andrewazores: All you really need from the `RequestParameters` object is the form attributes, right? Maybe you can change the implementation in `Rule` to just take a `MultiMap` instead of the whole `RequestParameters`?

jan-law: Thanks! That looks better now.

src/test/java/io/cryostat/net/web/http/api/v2/RulesPostHandlerTest.java
```diff
@@ -220,8 +220,31 @@ void throwsIfOptionalIntegerAttributesNonInteger(String val) {
...
+            headers.set(HttpHeaders.CONTENT_TYPE, HttpMimeType.JSON.mime());
+
+            String invalidRule =
+                    "{\"name\":\"Invalid_Rule\","
+                            + "\"description\":\"AutoRulesIT automated rule\","
+                            + "\"eventSpecifier\":\"template=Continuous,type=TARGET\","
+                            + "\"targetAlias\":\"es.andrewazor.demo.Main\","
+                            + "\"archivalPeriodSeconds\":"
+                            + val
+                            + "}";
```
andrewazores, line 237: For easier readability, you could construct this String (and the one like it in the next test too) like:

```
String invalidRule = gson.toJson(Map.of(
  "name": "Invalid_Rule",
  ...
));
```

This way you don't need to deal with manually writing in the JSON formatting with braces, colons, commas, and escaped quotes.

andrewazores: You could probably also use `JsonObject` the same way that you do in the `AutoRulesIT` test you added - `toString()` probably does what you need already.

src/main/java/io/cryostat/rules/Rule.java
```diff
@@ -219,4 +230,117 @@ public String toString() {
             return getSerialKey();
         }
     }
+
+    public static Rule buildRule(MultiMap formAttributes) {
```
andrewazores, line 234: This is just food for thought, I'm happy enough with this as is - what do you think about moving this into the Builder class, renaming it, and returning a Builder instance? So instead of callers doing ex. `Rule.buildRule(requestParams.getFormAttributes())`, they would do `Rule.Builder builder = Rule.Builder.from(requestParams.getFormAttributes())`. They could then still mutate the builder's state if that's needed for any reason and rely on the type safety that the Rule.Builder provides instead of trying to make the Builder do things by adding values to the JsonObject/MultiMap ahead of time. Then calling `build()` on it would give the `Rule` instance as usual.

src/main/java/io/cryostat/rules/Rule.java
```diff
@@ -219,4 +230,117 @@ public String toString() {
...
+                                        .getAsString());
+        builder = setOptionalInt(builder, Rule.Attribute.ARCHIVAL_PERIOD_SECONDS, jsonObj);
+        builder = setOptionalInt(builder, Rule.Attribute.PRESERVED_ARCHIVES, jsonObj);
+        builder = setOptionalInt(builder, Rule.Attribute.MAX_AGE_SECONDS, jsonObj);
+        builder = setOptionalInt(builder, Rule.Attribute.MAX_SIZE_BYTES, jsonObj);
+
+        return builder.build();
+    }
+
+    public static Rule.Builder setOptionalInt(
```
andrewazores, line 273: This part I'm a little more curious about refactoring actually, now that I see it. It takes a `Rule.Builder` as its first parameter and then returns that same `Rule.Builder` instance at the end so that it can be used as a fluent API, but it also lives within the `Rule` class. Maybe this should be moved inside the `Rule.Builder` and made non-static, and instead of taking a `Rule.Builder` reference as the first parameter, it can just operate on `this`?

src/main/java/io/cryostat/rules/Rule.java
jan-law: Good idea, this approach is more flexible. I can add that in

src/main/java/io/cryostat/rules/Rule.java
jan-law: Would it look similar to this format?
```
public static class Builder {
  ...
 private setOptionalInt(Rule.Attribute key, MultiMap formAttributes)
            throws IllegalArgumentException {
   ...
   this = fn.apply(value);
}
...
}
```

andrewazores: You can't assign to `this`, but I think what you intend is that the last line of that method would just be `return fn.apply(value)`. The `fn` comes from your `selectAttribute` method, which returns a Function/method reference to setter-methods from the Builder itself, which all return a reference to their Builder instance.

`selectAttribute` could also be moved into the `Builder` class, by the way.

andrewazores (approved):

andrewazores (approved):

jan-law:
I'm done making changes. PR is ready to merge

PR #541, opened 2021-06-25
Add Cryostat template to image

andrewazores:
Related to #461

This allows the Cryostat.xml (rename cryostat.jfc) template to be listed as an available TARGET template and used identically to the standard default.jfc and profile.jfc templates/configurations.

https://user-images.githubusercontent.com/3787464/123456221-207f6380-d5d2-11eb-92a8-3df30a8b26d4.mp4


jan-law (commented):
Looks good, will hold back on approving until after Hareet looks at it

hareetd (approved):
Looks good!

PR #544, opened 2021-06-29
feat(auto-rules): JavaScript Rule Match Expressions

### LINKED ISSUE #413: Document automatic rules API
Automatic rules are being implemented piece-by-piece across multiple PRs, and the API will be in flux for some time. Once it is settled, or close to it, API docs must be updated before the new API can be included in a release.

See #355

### LINKED ISSUE #540: Auto rules should offer matching on more than mainClass/targetAlias
Rules should be able to match targets by more criteria than simply "same alias".

Examples of rule matches that should be possible:

- [ ] "all targets"
- [ ] "target has connectUrl 'service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi'"
- [ ] "target has alias Foo"
- [ ] "target has main class io.cryostat.Cryostat"
- [ ] "target is listening on port 9999"
- [ ] "target has label app=myproject"
- [ ] "target has annotation cryostat.io/target=true"
- [ ] "target has main class io.cryostat.Cryostat AND target has annotation cryostat.io/profiling=true"
- [ ] "target has main class io.cryostat.Cryostat OR target has main class com.example.App"

Related: #280 #495 #413 #523

andrewazores:
Depends on #523

Fixes #540
Closes #413


src/main/java/io/cryostat/rules/RuleMatcher.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    @Name("io.cryostat.rules.RuleMatcher.BindingsCreationEvent")
+    @Label("Rule Binding Creation")
+    @Category("Cryostat")
+    @SuppressFBWarnings(
+            value = "URF_UNREAD_FIELD",
+            justification = "The event fields are recorded with JFR instead of accessed directly")
+    public static class BindingsCreationEvent extends Event {}
```
ebaron, line 138: Are these events meant to be expanded in the future? Right now it seems like the BindingsCreationEvent always occurs at the same time as the RuleAppliesEvent, but doesn't provide any additional info.

ebaron (commented):
Looking good! Just a couple of comments.

src/main/java/io/cryostat/rules/MatchExpressionValidator.java
```diff
@@ -0,0 +1,254 @@
...
+import jdk.nashorn.api.tree.WhileLoopTree;
+import jdk.nashorn.api.tree.WithTree;
+import jdk.nashorn.api.tree.YieldTree;
+
+public class MatchExpressionValidator {
+
+    private final Parser parser = Parser.create();
+
+    private final TreeVisitor<Void, String> treeVisitor =
+            new SimpleTreeVisitorES5_1<Void, String>() {
```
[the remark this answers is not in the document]
andrewazores: I'm anonymously extending `SimpleTreeVisitorES5_1`, which is a concrete class implementing the `TreeVisitor` interface, and only overriding the visitor methods that we want to disallow. All other visitor methods are implemented by the `SimpleTreeVisitorES5_1` base class and inherited by this anonymous extension class, which implicitly pass those AST nodes by default. So, any visitor methods for node types not overridden by this class are implicitly passed, and all the ones I have explicitly named by overriding the relevant visitors are failed. Do you mean to replace this with a class where all of the visitor methods are directly implemented by the class and there is a composition delegate `SimpleTreeVisitor` field for passing, so that all the AST node types/visitor methods are enumerated in the implementation, rather than having the implicit passes that I have here?

src/main/java/io/cryostat/rules/RuleMatcher.java
andrewazores: The `RuleAppliesEvent` starts strictly before and ends strictly after the `BindingsCreationEvent`, with a 1:1 relationship. I meant for these events to be used for profiling how long Nashorn evaluation of match expressions takes, and also to see how much of that time is actually evaluation by the rule engine vs. time spent in Cryostat preparing the bindings to pass to the engine. The time taken by Nashorn evaluation is roughly the duration of the `RuleAppliesEvent` minus the duration of the `BindingsCreationEvent`.

src/main/java/io/cryostat/rules/RuleMatcher.java
```diff
@@ -0,0 +1,139 @@
...
+class RuleMatcher {
+
+    private final ScriptEngine scriptEngine = new ScriptEngineManager().getEngineByName("nashorn");
+
+    public boolean applies(Rule rule, ServiceRef serviceRef) throws ScriptException {
+        RuleAppliesEvent evt = new RuleAppliesEvent(rule.getName());
+        try {
+            evt.begin();
+            Object result =
+                    this.scriptEngine.eval(rule.getMatchExpression(), createBindings(serviceRef));
```
jan-law, line 65: Why does `createBindings()` need to be called each time the script engine wants to compare a `matchExpression` to the `serviceRef`'s annotations?

andrewazores: That's a good question/observation. The bindings need to be created to take the full `ServiceRef` structure and slightly simplify it before passing it into the script engine (ie turn `AnnotationKeys` just into Strings), and both the `matchExpression` and the `serviceRef` have constant state that doesn't change over time, so the `Bindings` object would be a good candidate for caching/memoization. I'm not sure how expensive it actually is to create the bindings however, so it may not be worth the overhead of adding a cache and dealing with cache invalidation when `ServiceRefs` disappear. The `BindingsCreationEvent` could help to profile this and figure out if it's worthwhile.

andrewazores: We could also simply cache the result of whether a given rule's `matchExpression` applies to a given `ServiceRef`, instead of caching the `Binding`, and make it even faster by bypassing the script engine entirely, actually. If it is worth doing any caching then this would be the smarter way to go about it. This still incurs extra overhead in needing to do cache invalidation when `Rule` definitions are deleted or `ServiceRef`s disappear, though.

andrewazores: ... but then again, how often do we actually re-check a Rule against a ServiceRef? When a new Rule is added that rule alone is checked against all known ServiceRefs at the time, and vice-versa when a new ServiceRef appears it is checked against all of the Rules. So in either scenario, these are all new/unique combinations of Rule/ServiceRef, so the cache would be empty for all of these. Would we end up needing to re-evaluate it?

So maybe caching the ServiceRef -> Binding does make the most sense, since at least that would help in the case of a new Rule appearing. After the first one, we already have the Binding for each ServiceRef (assuming the set of ServiceRefs hasn't changed), so we can pass the same binding to the script engine to evaluate against the new rule matchExpression.

andrewazores: Anyway, tl;dr, my hunch is that the majority of the time is spent actually setting up the script engine, passing data to it, executing it, and marshalling the results back into Java, and comparatively little time is spent creating the Binding object. If that's true then it probably isn't worth caching the Bindings.

src/main/java/io/cryostat/rules/MatchExpressionValidator.java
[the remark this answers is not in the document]
ebaron: Right, so that nothing is implicitly passed. I'm thinking it might be slightly less error prone to maintain, but I'll leave it up to you.

src/main/java/io/cryostat/rules/RuleMatcher.java
ebaron: Ah, I understand now. I hadn't considered it being a duration thing.

src/main/java/io/cryostat/rules/MatchExpressionValidator.java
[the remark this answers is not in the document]
andrewazores: Okay, that's fair enough. I'll refactor it to do as you suggest. It's also probably worth pulling it into its own class, too.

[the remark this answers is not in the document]
andrewazores: Hmm - this doesn't seem to work as expected, actually. The `TreeVisitor` only seems to have one of its methods called, and then execution resumes at the line after `cut.accept(treeVisitor, rule.getMatchExpression())`. I expected its methods to be called repeatedly as the parser proceeded, but that doesn't seem to be how it works. The first node type passed is the CompilationUnit, so it will pass everything after that point. I also tried iterating over `cut.getSourceElements()`, but this just gives me a list of AST children of the CompilationUnit, which can also be other generally acceptable abstract node types. So to get the "explicit passing to a delegate" piece working, it seems like I would need to recursively descend the whole AST myself - doesn't seem worth it.

I'll extract the TreeVisitor to a top-level class of its own at least, but I'll leave it with the original implementation of only observing fail-worthy AST nodes.

src/main/java/io/cryostat/rules/MatchExpressionValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    private final Parser parser = Parser.create();
+
+    private final TreeVisitor<Void, String> treeVisitor = new MatchExpressionTreeVisitor();
+
+    String validate(Rule rule) {
+        CompilationUnitTree cut = parser.parse(rule.getName(), rule.getMatchExpression(), null);
+        if (cut == null) {
+            throw new IllegalMatchExpressionException();
+        }
+        cut.accept(treeVisitor, rule.getMatchExpression());
```
hareetd, line 55: If the match expression were to be improperly formatted, would `cut.accept(...)` throw an `IllegalMatchExpressionException` due to one of the overridden visit methods in the `MatchExpressionTreeVisitor` being called?

andrewazores: Probably - it would depend on what "improperly formatted" looks like. `MatchExpressionValidatorTest` exercises various failure modes.

src/main/java/io/cryostat/rules/RuleMatcher.java
hareetd, line 65: Does `scripEngine.eval(...)` return an `Object` because we don't know whether the script we passed in actually evaluates to a `boolean` (i.e. it could be executing other types of code)?

src/test/java/io/cryostat/rules/RuleProcessorTest.java
```diff
@@ -238,7 +238,7 @@ void testTaskCancellationOnFailure() throws Exception {
                 new Rule.Builder()
                         .name("Test Rule")
                         .description("Automated unit test rule")
-                        .targetAlias("com.example.App")
+                        .matchExpression("com.example.App")
```
hareetd, line 241: Is it optional for the target alias match expression to be of the form `target.alias == 'com.example.App'`?

andrewazores: Oh, I missed updating that spot. Thanks for catching it.

hareetd: No problem

src/main/java/io/cryostat/rules/MatchExpressionValidator.java
```diff
@@ -0,0 +1,58 @@
...
+import jdk.nashorn.api.tree.Parser;
+import jdk.nashorn.api.tree.TreeVisitor;
+
+public class MatchExpressionValidator {
+
+    private final Parser parser = Parser.create();
+
+    private final TreeVisitor<Void, String> treeVisitor = new MatchExpressionTreeVisitor();
+
+    String validate(Rule rule) {
```
hareetd, line 50: This is for my own understanding of Java, but why doesn't the `validate` method need to state `throws Exception? The reason I'm asking is that in the corresponding unit test, either an `IllegalMatchExpressionException` or a `NashornException` is possible when calling this method.

andrewazores: `IllegalMatchExpressionException extends RuntimeException`. Java has a distinction between checked and unchecked exceptions. Checked exceptions must be declared thrown and explicitly caught somewhere, but unchecked exceptions don't. NullPointerException is another example of an unchecked exception you've probably seen before - it would really, really suck if NPEs always had to be accounted for with try/catch blocks or throws declarations, because then nearly every line of Java code would need exception handling.

hareetd: Ah makes sense, thanks.

andrewazores: I would say in this case that it probably would make sense for the exception we throw to be a checked kind, since the `TreeVisitor` interface doesn't declare that the methods throw checked exceptions, I can't implement the interface and have mine change the method signatures by adding a new checked exception declaration.

I suppose I could work around this by having the `TreeVisitor` throw an unchecked type, then have the `validator` catch and rethrow those as checked exceptions. That's probably a good idea, actually.

src/test/java/io/cryostat/rules/RuleMatcherTest.java
```diff
@@ -0,0 +1,256 @@
...
+        void targetShouldHaveLabels() {
+            Map<String, String> labels =
+                    (Map<String, String>)
+                            (((Map<String, Object>) bindings.get("target")).get("labels"));
+            MatcherAssert.assertThat(labels, Matchers.equalTo(RuleMatcherTest.this.labels));
+        }
+
+        @Test
+        void annotationsShouldHaveExpectedFields() {
+            Bindings bindings = ruleMatcher.createBindings(serviceRef);
```
hareetd, line 139: I think this is unnecessary because it's just a duplicate of the class field `bindings`.

ebaron (dismissed):

hareetd (dismissed):
Reviewed the recent commits as well, looks good to go.

jan-law (dismissed):

andrewazores:
Cleanly rebased, approving reviews were all dismissed so I will merge regardless once status checks pass.

ebaron (approved):

PR #550, opened 2021-07-06
ci(release-drafter): config for Release Drafter

andrewazores:
Configuration for Release Drafter GitHub Action


andrewazores:
The `update_release_draft` check is failing because it doesn't see the contents of the `.github/release-drafter.yml` file, for some reason - is it expecting that to already exist in `main`?

ebaron:
That seems like a likely culprit.

ebaron (approved):

Issue #543, opened 2021-06-25
Automatic Rules PeriodicArchiver should not use HTTP requests to self


andrewazores:
```
        // FIXME using an HTTP request to localhost here works well enough, but is needlessly
        // complex. The API handler targeted here should be refactored to extract the logic that
        // creates the recording from the logic that simply figures out the recording parameters
        // from the POST form, path param, and headers. Then the handler should consume the API
        // exposed by this refactored chunk, and this refactored chunk can also be consumed here
        // rather than firing HTTP requests to ourselves
```

andrewazores:
The `RecordingCreationHelper` from #538 could be useful.

jan-law:
Will start on this after #538 is merged

jan-law:
Where should I put the extracted logic to save or delete recordings? 
I'd like to reuse `private Optional<IRecordingDescriptor> getDescriptorByName` from `RecordingCreationHelper`. Class names like `RecordingDeletionHelper` or `RecordingSaveHelper` would be more appropriate for helper functions that save or delete recordings though.

andrewazores:
You could probably put it into `RecordingCreationHelper` and just rename that class to something a little more general.

jan-law:
Why does cryostat need to supply credentials when making an http request to itself? Should the refactored `PeriodicArchiver` validate stored credentials before archiving/pruning recordings?

```
PeriodicArchiver.java, line 133:

this.webClient
   .patch(path)
   .putHeaders(headersFactory.apply(credentialsManager.getCredentials(serviceRef)))
...
```

andrewazores:
> Why does cryostat need to supply credentials when making an http request to itself?

In this case, Cryostat is acting as a web client to itself, so it has to go through all the usual communications protocols that any client does. If you chase through what is actually injected here it comes from the `RulesModule`, and this function attaches JMX Authorization credentials as a header if needed/if provided. So the `PeriodicArchiver` checks if the `CredentialsManager` has any stored credentials for the given target, and if it does, those credentials are included along with the request to archive the recording - which entails opening a connection to the target JVM and copying the JFR data from it. If the target requires credentials and none are provided with the API request to Cryostat, then Cryostat will try to connect to the target without credentials, and the target will reject the connection attempt.

[...]

PR #551, opened 2021-07-06
fix(auto-rules): PeriodicArchiver scans archives on startup

### LINKED ISSUE #542: Automatic Rules PeriodicArchiver should scan existing archives on startup
```
        // FIXME this needs to be populated at startup by scanning the existing archived recordings,
        // in case we have been restarted and already previously processed archival for this rule
```

hareetd:
By scanning existing archives on startup, this fix allows pruning of old recordings from previous calls to `PeriodicArchiver` for a given rule. 

Fixes #542 

Depends on #557


hareetd:
I requested a review so I know if I'm on the right track before I start updating the corresponding unit tests.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -184,4 +200,19 @@ void performArchival() throws InterruptedException, ExecutionException {
                         });
         return future;
     }
+
+    JsonArray getArchivedRecordings() throws InterruptedException, ExecutionException {
+        HttpRequest<Buffer> req = this.webClient.get("api/v1/recordings");
```
andrewazores, line 205: We should avoid using HTTP requests to ourselves here - see #543 .

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -97,6 +102,17 @@ public void run() {
         logger.trace("PeriodicArchiver for {} running", rule.getRecordingName());
 
         try {
+            JsonArray response = getArchivedRecordings();
+            Iterator<Object> it = response.iterator();
+            
+            while (it.hasNext()) {
+                JsonObject entry = (JsonObject) it.next();
+                String fileName = entry.getString("name");
+                if (fileName.contains(this.rule.getRecordingName())) {
```
andrewazores, line 111: I think we want a more robust check than `fileName.contains(this.rule.getRecordingName())` - this could lead to a lot of false positive matches if my rule has a really short/common/simple name, or if the rule name happens to equal or be a substring of a target's alias, etc.

andrewazores: There is the `auto_` prefix here that seems to make this less likely, but I think it's still possible we see false positives.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
hareetd: Alright, so I should refactor the `RecordingsGetHandler` in such a way that the JSON array can be obtained without using an HTTP request? From a cursory glance at the handler, it uses `FileSystem` to store the data we need, so it looks like I could add and expose a method which extracts the `FileSystem` logic.

andrewazores: Right. The logic should be extracted into something similar to the `RecordingCreationHelper`, for example called the `RecordingArchiveHelper`, which the `RecordingsGetHandler` and the `PeriodicArchiver` can both use for doing things related to adding/listing/deleting files from the archives. The `RecordingArchiveHelper` would take the existing logic from the `RecordingsGetHandler` that creates files in the `FileSystem` and generates their names, and that same logic but in reverse can be used to check the contents of the `FileSystem` and see if anything has a name that matches the format the `RecordingArchiveHelper` uses.

hareetd: Initially, when I saw the changes in @jan-law 's PR, I was thinking I would only have to implement listing archived files in my `RecordingArchiveHelper`, since adding and deleting files in the `PeriodicArchiver` has already been handled by the `RecordingHelper` from that PR.

However, I noticed that the `RecordingHelper` functions are also used by the recording handlers only concerning the target JVM such as `TargetRecordingDeleteHandler`, `TargetRecordingsPostHandler`, etc. 

Would it be worthwhile to keep the logic separate for archive vs. target recordings, i.e. have a `RecordingArchiveHelper` which the `PeriodicArchiver` and the archive related handlers (`TargetRecordingPatchSave`, `RecordingsGetHandler`, and `RecordingDeleteHandler`) all use for adding/listing/deleting files from the archives? 

Then the `RecordingHelper` (potentially renamed to `RecordingTargetHelper`) would no longer be used by the `PeriodicArchiver`, instead by recordings only involving the target?

andrewazores: Sure, that does sound like a distinction probably worth making.

jan-law: @hareetd I can move my related helper functions `saveRecording` ,`deleteRecording` and `writeRecordingToDestination` into a new file `RecordingArchiveHelper`, and also rename `RecordingHelper` to `RecordingTargetHelper`. Then you can move your helper functions into `RecordingArchiveHelper`

hareetd: @jan-law Thank you, that sounds like it'll save me time re-writing the saving/deleting of the archive files. I'm just wondering how it'll work in terms of me getting the changes into my branch in a clean way (i.e. I still want my PR to be clear about its original intention) @andrewazores?

jan-law: @andrewazores since `public Optional<IRecordingDescriptor> getDescriptorByName` is used by both `RecordingTargetHelper` and `RecordingArchiveHelper`, should we make the function static? I think the function output only depends on its arguments.

andrewazores: Rebase your PR on top of Janelle's branch instead of `main` :-) you can then edit your PR's text with "Depends on" and link it to hers that way.

andrewazores: > @andrewazores since `public Optional<IRecordingDescriptor> getDescriptorByName` is used by both `RecordingTargetHelper` and `RecordingArchiveHelper`, should we make the function static? I think the function output only depends on its arguments.

Maybe each of the Helpers should have their own implementation of this function, returning a descriptor (if any) from their respective domains. If some consumer needs to get a descriptor from either domain by name then it can simply use both Helpers and take from them in whichever order of preference makes sense.

hareetd:
@andrewazores After adding Janelle's fork as a remote repo (in order to get her branch), I'm assuming I'd have to wait until she's pushed the required changes before rebasing on top of her branch?

andrewazores:
> @andrewazores After adding Janelle's fork as a remote repo (in order to get her branch), I'm assuming I'd have to wait until she's pushed the required changes before rebasing on top of her branch?

You'd have to rebase after each change she makes to her branch anyway, up until eventually her branch is merged into main, at which point you would rebase onto main as well. If her branch doesn't yet contain the changes you need then you can still rebase on top of it for now, and then re-rebase later after her branch is updated with the changes.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
hareetd: I noticed from the HTTP API docs that archived recordings are saved with the following filename format:

[$TARGET_NAME]_[$RECORDING_NAME]_[$DATE]T[$TIME]Z[.$COUNTER][.jfr]

In my opinion, the target name and recording name together can be used to accurately determine a recording as belonging to a specific rule, using a Pattern regex for example.

So now my question is what exactly is TARGET_NAME? Judging from the code used to fulfill a PATCH save request here: 

https://github.com/cryostatio/cryostat/blob/98924a38e879d1284dd4c57369951383f71abf32/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchSave.java#L139

it's not simply the `Rule.targetAlias` field so I would have to "calculate" it the same way. This seems redundant however, so maybe I should implement the map as described in the TODO in the link above? That way the map can be used in the `PeriodicArchiver` and when doing a PATCH save. Which class should I store this map in?

andrewazores: That map would help avoid lookups/calculations but it would also need to be repopulated if the Cryostat instance goes down and comes back up later and has pre-existing archives contents, so I don't think it will end up helping you in the end, here.

The targetName part of the saved file is indeed somewhat ambiguous and can't always be mapped exactly back to a target uniquely. We could change this behaviour and always save the file using the service URI of a target instead, which would make them unique/unambiguous.

Another possibility would be to subdivide the archives. Currently all archived recording files are dumped into a flat directory. If we instead created directories within the archive corresponding uniquely to targets/ServiceRefs, then archived files could be placed into those directories and identified as unambiguously belonging to a target as well.

ex. right now we have things like:

```
archive/target1alias_recording1.jfr
archive/target2_recording1.jfr
```

This could be made into:

```
archive/$hashOfTarget1ServiceURI/target1alias_recording1.jfr
archive/$hashOfTarget2ServiceURI/target2_recording.jfr
```

This way we keep the human-friendly filenames but also have a machine-friendly way of grouping and identifying them.

hareetd: Sounds good, I'll take a look into modifying the archive directory structure.

hareetd:
Before I make more changes, I'm going to rebase on to upstream/main now that Janelle's PR has been merged. During this interactive rebase, I noticed my commit history contains Janelle's commits first (up to a certain point before the branch got merged and deleted) and then my commits (due to the rebasing I did). 

Since upstream//main already contains all of Janelle's commits, should I drop her commits from my branch, keeping only my commits, thereby minimizing merge conflicts during the rebase?

andrewazores:
> Before I make more changes, I'm going to rebase on to upstream/main now that Janelle's PR has been merged. During this interactive rebase, I noticed my commit history contains Janelle's commits first (up to a certain point before the branch got merged and deleted) and then my commits (due to the rebasing I did).
> 
> Since upstream/main already contains all of Janelle's commits, should I drop her commits from my branch, keeping only my commits, thereby minimizing merge conflicts during the rebase?

Yes exactly, you should drop all of the commits in your branch that are from the old base. Those changes (roughly) are now in `main`, so your branch should not contain any commits that redo the same changes over again, or else you'll have some strange conflicts to try to resolve where the commits probably just end up empty. git doesn't know what happened to those commits because they got squashed into one larger commit, so when it uses the commit hashes to determine the rebase, it doesn't know that the first series of commits in your branch represent logic changes that are already in the new base.

hareetd:
Nice, that makes sense. This a good way for me to more comfortable with Git.

hareetd:
Could you take a look at just the `RecordingArchiveHelper` and `PeriodicArchiver` to see if my changes make sense?

andrewazores:
> Edit: Also, for the various archived recording handlers (GET, DELETE, etc.) I'm thinking of changing them so that the target service URI is now a request parameter, allowing me to properly index searches in the restructured directory. Does that seem right?

You mean to add the target's service URI as a path parameter on the request? That would be a breaking API change, so it shouldn't be done to the existing V1 handlers.

If I'm understanding correctly, the concern is that ex. `DELETE /api/v1/recordings/:recordingName` doesn't necessarily correlate the request to a target's service URI, since the recording file name may not contain the service URI to be parsed out and used for reference.

This is no different from the current state of the archives really, though - right now we might have `archives/targetA_foo.jfr, archives/targetB_foo.jfr`, but with the subdirectory reorganization here it would be `archives/targetA_hash/targetA_foo.jfr, archives/targetB_hash/targetB_foo.jfr`. So, there's just one additional level of directories to search through to find the matching recording file, unless the client can tell us what `targetA` or `targetA_hash` are. But the problem with that is that the client does not know those values either, since the client can only be presumed to know about `targetA_foo.jfr` because they did a `GET /api/v1/recordings`, and got back the bare list of recording names.

I do think that along with subdividing the archives by target it would make sense to mirror that change in the API, but it would need to be done in V2 handlers. `PATCH 'save' /api/v2/targets/:targetId/recordings/:recordingName` could create an archived recording at `archives/targetId_hash/recordingName_timestamp.jfr`, for example, and then `DELETE /api/v2/targets/:targetId/archives/:filename` could be used for deleting that recording, or `GET` to download it, etc.

If we had that then the unstructured archives used by the V1 handlers could be loosened up further and not require uploads to follow any filename convention, and this could be used to allow users to upload arbitrary JFR files for analysis using jfr-datasource/grafana. They can already do this, really, but the filename convention is a minor barrier that makes that a bit annoying.

Anyway, getting on a tangent. For the scope of this issue/PR, I don't think we want any V1 API behaviour to be changed - everything done in here should be an internal implementation detail. If it sets us up for V2 (or V3) enhancements later like what I outlined above, then great.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -121,8 +138,50 @@ public void deleteRecording(ConnectionDescriptor connectionDescriptor, String re
                 });
     }
 
+    public List<Map<String, String>> getRecordings() throws Exception {
```
andrewazores, line 141: This probably shouldn't return a List of Maps, but instead a List of some more constrained type. The Maps used here always have the same set of keys, so this could/should instead be some class with fields named like those keys, and getters to access them.

Right now I could do `recordingArchiveHelper.getRecordings().get("nonsense key")` and this always returns `null`, but it should just be a compile-time error instead.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -121,8 +138,50 @@ public void deleteRecording(ConnectionDescriptor connectionDescriptor, String re
...
+        if (!fs.isDirectory(recordingsPath)) {
+            throw new ArchivePathException(recordingsPath.toString(), "is not a directory");
+        }
+        WebServer webServer = webServerProvider.get();
+        List<String> files = this.fs.listDirectoryChildren(recordingsPath);
+        return files.stream()
+                .map(
+                        file -> {
+                            String serviceUriHash = file.split("/")[0];
+                            String recordingName = file.split("/")[1];
```
andrewazores, line 157: Can this be done using the `Path` class rather than `split("/")`? Something like `serviceUriHash = Path.of(file).getParent()`, `recordingName = Path.of(file).getFileName()`.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -121,8 +138,50 @@ public void deleteRecording(ConnectionDescriptor connectionDescriptor, String re
...
+                        })
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
+    }
+
     private String writeRecordingToDestination(
             JFRConnection connection, IRecordingDescriptor descriptor) throws Exception {
+        URI serviceUri = URIUtil.convert(connection.getJMXURL());
+        Path specificRecordingsPath =
+                recordingsPath.resolve(String.format("%d", serviceUri.hashCode()));
```
andrewazores, line 183: `String.format("%d")` seems a little overkill here - `Integer.toString()` might be a better fit.

andrewazores: Or, we could use something other than the direct `int hashCode()` of the serviceUri. It could instead be the base64 encoding of the serviceUri, for example, which is already naturally a String and contains only characters which are legal for filenames/directories, and as an added bonus can be decoded back into a serviceUri (unlike the hashcode, where that information is lost).

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -146,4 +181,17 @@ public void run() {
         }
         return future;
     }
+
+    public Future<JsonArray> getArchivedRecordings() throws Exception {
+
+        CompletableFuture<JsonArray> future = new CompletableFuture<>();
```
andrewazores, line 187: I'm not sure why this needs to return a `Future` - can't `run()` above just directly call `recordingArchiveHelper.getRecordings()`? Then above the direct return type (currently `List<Map<String, String>>`, but as I said in an earlier comment, should be `List<SomeOtherType>`) can be used, rather than having to go through this JSON encode/decode cycle on top.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -91,6 +97,34 @@ public void run() {
...
+            // with any previously archived recordings for this rule.
+            if (previousRecordings.isEmpty()) {
+                JsonArray archivedRecordings = getArchivedRecordings().get();
+                Iterator<Object> it = archivedRecordings.iterator();
+
+                URI serviceUri = serviceRef.getServiceUri();
+                Pattern recordingFilenamePattern =
+                        Pattern.compile(
+                                String.format(
+                                        "([A-Za-z\\d-]*)_(%s)_([\\d]*T[\\d]*Z)(\\.[\\d]+)?",
```
andrewazores, line 113: Instead of building the regex pattern with the serviceUri inserted into the middle of it, this can be a constant pattern with a capture group where the serviceUri is expected. Then the captured value of that group can be pulled out to compare to the expected serviceUri, or hashed, or whatever else is needed.

andrewazores: It looks like this pattern probably originally came from: https://github.com/cryostatio/cryostat/blob/af47f3f5b08c673cb1df7cae8c739db5b2f597b8/src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java#L76

The matcher and capture group stuff can also be seen in use with that pattern here:

https://github.com/cryostatio/cryostat/blob/af47f3f5b08c673cb1df7cae8c739db5b2f597b8/src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java#L164

https://github.com/cryostatio/cryostat/blob/af47f3f5b08c673cb1df7cae8c739db5b2f597b8/src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java#L169

hareetd:
Alright, thanks for the thorough explanation, makes sense.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
hareetd: I'll switch to base64 encoding then, seems to have more pros then the direct hashing, especially the decoding.

andrewazores: https://stackoverflow.com/questions/3945541/base64-encoding-safe-for-filenames

Modified Base64 might be okay, but this discussion makes a good point about using ex. Base32 instead.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
hareetd: For the `Future`, I was following the advice here: 

https://github.com/cryostatio/cryostat/pull/557#discussion_r665415792

Should I change it to return a non-Future object? 

The separate method does seem redundant though, especially with the JSON overhead so I'll change that.

andrewazores: Oh, my bad - I think when I wrote that I was thinking that the method belonged to one of the `Helper`s. Since those are part of an internal API, it makes sense for them to be public and to return a `Future` to indicate to callers that these are blocking operations. This `getArchivedRecordings` method and others within the `PeriodicArchiver` are not really part of an internal API, so it's okay for them to be non-`public`, and in that case certainly don't need to return `Future`s or otherwise be marked as long-running operations if they are just internal methods within the class.

andrewazores: So the actual advice I meant to share in that comment would mean that in this case, `RecordingArchiveHelper#getRecordings` should be `public`, and should have a return type of `Future<List<ArchivedRecordingInfo>>` (or whatever that type to replace the `Map<String, String>` gets named).

src/main/java/io/cryostat/rules/PeriodicArchiver.java
hareetd: Just to clarify, the regex pattern has the recording name (in this case, `auto_$(Rule.name)`) inserted into the middle, not the serviceUri. The serviceUri hash is compared directly inside the loop, and not with the matcher, since the serviceUri is not part of the overall filename but just the parent directory.

Capturing seems more robust, I'll take a look at those lines, thanks.

andrewazores: Ah right, makes sense. In any case, I think the capture group way is more conventional and so probably easier to read/maintain in the future. As an added bonus it's only one pattern that needs to be compiled at application startup, rather than a new pattern to be compiled on every request, so there should be a (very slight) performance benefit too.

hareetd: Sounds good, I'll implement the capture group regardless.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
hareetd: No worries, I'll go ahead and remove the unnecessary `Future` stuff from the `PeriodicArchiver`.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
```diff
@@ -102,34 +94,24 @@ public boolean isAsync() {
...
+        try {
+            recordingArchiveHelper.deleteRecording(recordingName).get();
+            notificationFactory
+                    .createBuilder()
+                    .metaCategory(NOTIFICATION_CATEGORY)
+                    .metaType(HttpMimeType.JSON)
+                    .message(Map.of("recording", recordingName))
+                    .build()
+                    .send();
+            ctx.response().setStatusCode(200);
```
andrewazores, line 106: Just a style thing, but `200` is already the default status code assumed by the `IntermediateResponse` if you don't set one, so explicitly setting it to `200` isn't required but also isn't harmful.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
```diff
@@ -102,34 +94,24 @@ public boolean isAsync() {
...
-                            }
-                            ctx.response().setStatusCode(200);
-                            ctx.response().end();
-                        },
-                        () -> {
-                            throw new HttpStatusException(404, recordingName);
-                        });
+        try {
+            recordingArchiveHelper.deleteRecording(recordingName).get();
+            notificationFactory
```
andrewazores, line 99: Would it make sense to also extract this notification emission into the archive helper's `deleteRecording` method?

src/main/java/io/cryostat/net/web/http/api/v1/RecordingGetHandler.java
```diff
@@ -86,28 +85,18 @@ public boolean isAsync() {
...
-                                        .setStatusMessage(
-                                                String.format(
-                                                        "Recording \"%s\" not found",
-                                                        recordingName));
-                                ctx.response().end();
-                            }
-                        });
+        try {
+            Path archivedRecording = recordingArchiveHelper.getRecordingPath(recordingName).get();
+            ctx.response().sendFile(archivedRecording.normalize().toAbsolutePath().toString());
```
andrewazores, line 90: Should `getRecordingPath` perhaps already return a normalized, absolute path?

src/main/java/io/cryostat/net/web/http/api/v1/RecordingUploadPostHandler.java
```diff
@@ -140,9 +135,17 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
...
-        Path recordingPath =
-                getRecordingPath(recordingName)
-                        .orElseThrow(() -> new HttpStatusException(404, recordingName));
+        Path recordingPath = null;
+        try {
+            recordingPath = recordingArchiveHelper.getRecordingPath(recordingName).get();
+        } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof RecordingNotFoundException) {
+                throw new HttpStatusException(404, cause.getMessage(), cause);
```
andrewazores, line 144: I think it would be better to pass `e.getMessage()`/`e` into the `HttpStatusException` constructor rather than `cause`, so we aren't losing a layer of context information, right?

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsGetHandler.java
```diff
@@ -111,48 +88,12 @@ public boolean isAsync() {
...
-            throw new HttpStatusException(
-                    501,
-                    String.format(
-                            "Archive path %s does not exist", savedRecordingsPath.toString()));
+        try {
+            List<ArchivedRecordingInfo> result = recordingArchiveHelper.getRecordings().get();
+            ctx.response().setStatusCode(200);
+            ctx.response().end(gson.toJson(result));
+        } catch (ArchivePathException e) {
+            throw new HttpStatusException(501, e.getMessage());
```
andrewazores, line 96: Should `e` be passed to the exception constructor as well to establish the causal link between them?

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java
```diff
@@ -174,12 +178,18 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
                         ? 0
                         : Integer.parseInt(m.group(4).substring(1));
 
+        ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
+        Base32 base32 = new Base32();
```
andrewazores, line 182: `base32` looks like it could be extracted to a field or constant - doesn't seem like it needs to be instantiated every time.

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchSave.java
```diff
@@ -85,8 +86,13 @@ void handle(RoutingContext ctx, ConnectionDescriptor connectionDescriptor) throw
                                     connectionDescriptor.getTargetId()))
                     .build()
                     .send();
-        } catch (RecordingNotFoundException e) {
-            throw new HttpStatusException(404, e);
+        } catch (ExecutionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof RecordingNotFoundException) {
+                throw new HttpStatusException(404, cause.getMessage(), cause);
```
andrewazores, line 92: Same here re. `e` vs `cause`

src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
hareetd: I'll remove it then, looks cleaner that way.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingDeleteHandler.java
hareetd: Sure, I'll do the same for the patch save.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,216 @@
...
+                                } else {
+                                    throw new RecordingNotFoundException(recordingName);
+                                }
+                            });
+            future.complete(saveName);
+        } catch (RecordingNotFoundException e) {
+            future.completeExceptionally(e);
+        }
+
+        return future;
```
andrewazores, line 134: This method could be simplified somewhat since the `executeConnectedTask` actually runs synchronously, but the way you've written it lends itself well to the connected tasks possibly being done asynchronously as well, so I'm okay with leaving it as-is.

However, I think it's a bit awkward for consumers of this internal API to receive a `Future` but also have to catch the generic `Exception` when calling the method. I think it would be less surprising if the `catch` block in this method caught all of those `Exceptions, removing it from the method signature, and pushing them into the `future.completeExceptionally` instead. Does that make sense?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,216 @@
...
+    }
+
+    public Future<Void> deleteRecording(String recordingName) throws Exception {
+
+        CompletableFuture<Void> future = new CompletableFuture<>();
+
+        try {
+            Path archivedRecording = getRecordingPath(recordingName).get();
```
andrewazores, line 142: Maybe the returned `Future` could relay this Path to the caller as well - they may or may not need it, but it's a piece of potentially useful information.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,216 @@
...
+        CompletableFuture<Path> future = new CompletableFuture<>();
 
-        targetConnectionManager.executeConnectedTask(
-                connectionDescriptor,
-                connection -> {
-                    Optional<IRecordingDescriptor> descriptor =
-                            this.getDescriptorByName(connection, recordingName);
+        try {
+            Path archivedRecording = null;
+            Boolean recordingFound = false;
```
andrewazores, line 208: Can this be `boolean` (lowercase `b`, primitive type) instead?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,216 @@
...
+                    if (recordingName.equals(file)) {
+                        archivedRecording = recordingsPath.resolve(subdirectory + "/" + file);
+                        recordingFound = true;
+                        break;
                     }
-                    return null;
-                });
+                }
+
+                if (recordingFound) {
```
andrewazores, line 222: It's easily abused so usually best avoided, but in this case a labelled break:

(see https://docs.oracle.com/javase/tutorial/java/nutsandbolts/branch.html)

could be used instead of this `recordingFound` flag to break out of the outer loop from the inner loop.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,216 @@
...
+                        this.fs.listDirectoryChildren(recordingsPath.resolve(subdirectory));
 
-                    if (descriptor.isPresent()) {
-                        connection.getService().close(descriptor.get());
-                        reportService.delete(connectionDescriptor, recordingName);
-                    } else {
-                        throw new RecordingNotFoundException(recordingName);
+                for (String file : files) {
+                    if (recordingName.equals(file)) {
+                        archivedRecording = recordingsPath.resolve(subdirectory + "/" + file);
```
andrewazores, line 216: This part doesn't look right: `recordingsPath.resolve(subdirectory + "/" + file)`

Would this be equivalent to `recordingsPath.resolve(subdirectory).resolve(file)`?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
andrewazores: Alternatively you could extract this piece of logic out into a small helper method, and use `return` to break the control flow out from the inner loop back up to the callsite.

src/main/java/io/cryostat/rules/ArchivedRecordingInfo.java
```diff
@@ -0,0 +1,102 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package io.cryostat.rules;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+public class ArchivedRecordingInfo {
```
andrewazores, line 43: This class looks a lot like the existing `io.cryostat.jmc.serialization.SavedRecordingDescriptor` class, which is apparently now unused. This class probably replaces it, so the `SavedRecordingDescriptor` could be deleted now.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -91,59 +97,71 @@ public void run() {
         logger.trace("PeriodicArchiver for {} running", rule.getRecordingName());
 
         try {
-            while (this.previousRecordings.size() > this.rule.getPreservedArchives() - 1) {
-                pruneArchive(this.previousRecordings.remove()).get();
+            // If there are no previous recordings, either this is the first time this rule is being
+            // archived or the Cryostat instance was restarted. Since it could be the latter,
+            // populate the array with any previously archived recordings for this rule.
+            if (previousRecordings.isEmpty()) {
+                Base32 base32 = new Base32();
```
andrewazores, line 104: Extract from here as well.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -91,59 +97,71 @@ public void run() {
...
+                String serviceUri = serviceRef.getServiceUri().toString();
+                List<ArchivedRecordingInfo> archivedRecordings =
+                        recordingArchiveHelper.getRecordings().get();
+                Iterator<ArchivedRecordingInfo> it = archivedRecordings.iterator();
```
andrewazores, line 108: Why is this `Iterator` used instead of a simple for-each loop?

src/main/java/io/cryostat/util/ArchivedRecordingInfoSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.reflect.Type;
+
+import io.cryostat.rules.ArchivedRecordingInfo;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+
+public class ArchivedRecordingInfoSerializer implements JsonSerializer<ArchivedRecordingInfo> {
```
andrewazores, line 49: What does this do that the default Gson serialization for the ArchivedRecordingInfo doesn't? It looks like it drops the `encodedServiceUri` from the serialized representation - if that's the only goal, then simply adding the `transient` keyword to the field declaration in the ArchivedRecordingInfo class would instruct Gson to ignore it in serialization as well, and this class wouldn't be needed.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -235,25 +235,8 @@
...
-                        break;
-                    }
-                }
-
-                if (recordingFound) {
-                    break;
-                }
-            }
+            Path archivedRecording = searchSubdirectories(subdirectories, recordingsPath, recordingName);
             if (!(fs.exists(archivedRecording)
```
andrewazores, line 240: Should there be a null-check here too?

src/main/java/io/cryostat/net/web/http/api/v1/RecordingGetHandler.java
```diff
@@ -87,7 +87,7 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
         String recordingName = ctx.pathParam("recordingName");
         try {
             Path archivedRecording = recordingArchiveHelper.getRecordingPath(recordingName).get();
-            ctx.response().sendFile(archivedRecording.normalize().toAbsolutePath().toString());
+            ctx.response().sendFile(archivedRecording.toString());
```
andrewazores, line 90: And here - what if the returned path is null?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,244 @@
...
+    private Path searchSubdirectories(List<String> subdirectories, Path parent, String recordingName) throws Exception {
+        Path archivedRecording = null;
+        for (String subdirectory : subdirectories) {
+            List<String> files =
+                    this.fs.listDirectoryChildren(parent.resolve(subdirectory));
+
+            for (String file : files) {
+                if (recordingName.equals(file)) {
+                    archivedRecording = parent.resolve(subdirectory).resolve(file).normalize().toAbsolutePath();
+                    return archivedRecording;
```
andrewazores, line 262: No need to assign the `archivedRecording` value and then also immediately return it here - you can just directly return the found value.

In fact, you could remove the `archivedRecording` local variable altogether, and simply return the result here. If you get to the end of the method without having returned anything then you can either `return null` or `throw RecordingNotFoundException`.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingGetHandler.java
hareetd: I'll resolve the null case inside the `getRecordingPath` function.

hareetd:
I've implemented all the suggestions above, thank you.

I realized that I had misunderstood the archived report caching process so I added a call to `RecordingArchiveHelper.getRecordingPath(...)` in `ArchivedRecordingReportCache` to fix the issue. However, this resulted in a circular dependency in Dagger because the `RecordingArchiveHelper` calls `ReportService.delete(...)` when deleting an archived recording, where `ReportService` depends on `ArchivedRecordingReportCache`.

In order to resolve this, I extracted the logic for deleting archived recording reports into the `RecordingArchiveHelper`, which allowed me to remove the `ReportService` dependency from the class. However, I was wondering whether it's worth it to make a new class called `ReportArchiveHelper` and put the logic in there?

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java
```diff
@@ -174,12 +182,17 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
                         ? 0
                         : Integer.parseInt(m.group(4).substring(1));
 
+        ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
```
andrewazores, line 185: How does this work? `getConnectionDescriptorFromContext` relies on retrieving the `targetId` path parameter, but this API handler doesn't have such a path parameter. I think you'll end up with a `ConnectionDescriptor` with a null targetId and null credentials, no?

andrewazores:
> I've implemented all the suggestions above, thank you.
> 
> I realized that I had misunderstood the archived report caching process so I added a call to `RecordingArchiveHelper.getRecordingPath(...)` in `ArchivedRecordingReportCache` to fix the issue.

I'm not sure I follow. What was the issue and how did calling `getRecordingPath` fix it?

> However, this resulted in a circular dependency in Dagger because the `RecordingArchiveHelper` calls `ReportService.delete(...)` after deleting an archived recording, where `ReportService` depends on `ArchivedRecordingReportCache`.
> 
> In order to resolve this, I extracted the logic for deleting archived recording reports into the `RecordingArchiveHelper`, which allows me to remove the `ReportService` dependency from the class. However, I was wondering whether it's worth it to make a new class called `ReportArchiveHelper` and put the logic in there?

That's probably a good idea, but I think it can be left for a later refactor/cleanup after we get the functionality hammered out here.

You also likely could have resolved the circular dependency using `dagger.Lazy`, but extracting common logic to the helpers is also worthwhile to avoid duplication.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java
hareetd: So the `ConnectionDescriptor` fields are populated using the path parameters, which is why this approach works for `TargetRecordingPatchSave` but not here, right? 

In that case, how would I organize the directory structure since here the recording is being uploaded directly from the client? I believe the only target identifier I have is `TARGET_NAME` in the file format `[$TARGET_NAME]_[$RECORDING_NAME]_[$DATE]T[$TIME]Z[.$COUNTER][.jfr]`.

hareetd: To add on, my changes to the other handlers are all centered around recordings being organized in base 32 encoded `ServiceUri` (i.e. `targetId`) directories.

andrewazores: Right, it works elsewhere in handlers that do have a `targetId` path parameter, but it won't work here unfortunately.

Can you add a special-cased subdirectory for these client re-uploaded recordings? The actual subdirectories would always have names that are base32-encoded ServiceURIs (or maybe base32-encoded host:port pairs), so simply naming the directory an unencoded `"unlabelled"` or something could work, I think. You can upload into that and also merge those into the results when scanning the filesystem for all archived recordings.

At some point in a separate PR we should add a V2 handler that allows for re-uploads of recordings that does have a targetId path parameter so that the uploads can be associated with a particular target. Then we can also make it possible to query for archived recordings by target, and then the web-client can display archived recordings per-target as well.

hareetd:
> I'm not sure I follow. What was the issue and how did calling `getRecordingPath` fix it?
>
The issue was that `ReportIT` was failing. This was because the way archived recording report caching works is that first the path to the archived recording is retrieved and a separate destination path for the report created. Both paths are then passed into the  `SubprocessReportGenerator.exec(...)` method, which dumps a report to the destination path.

Since I had changed the directory structure for archived recordings in the other handlers, report caching was failing since it still used the old way of getting the archived recording path, meaning no archived recording was being found. I fixed this by calling `getRecordingPath`, which uses the new, base 32 encoded `ServiceUri` directory structure to search.  
> 
> That's probably a good idea, but I think it can be left for a later refactor/cleanup after we get the functionality hammered out here.
> 
> You also likely could have resolved the circular dependency using `dagger.Lazy`, but extracting common logic to the helpers is also worthwhile to avoid duplication.

Yeah, the goal was to avoid duplication but good to know `dagger.Lazy` works as well.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java
hareetd: That's a good idea and it's a simple add-on; I'll go ahead and do that. 

As for the V2 handler, I can get started on that if you'd like once I finish this PR and have helped Janelle with the API IT coverage.

andrewazores:
Awesome!

```diff
diff --git a/run.sh b/run.sh
index 3cfa1e89..b2c673d6 100755
--- a/run.sh
+++ b/run.sh
@@ -63,7 +63,7 @@ fi
 podman run \
     --pod cryostat \
     --mount type=tmpfs,target=/opt/cryostat.d/conf.d \
-    --mount type=tmpfs,target=/opt/cryostat.d/recordings.d \
+    --mount type=bind,source="$(dirname $0)/recordings",destination=/opt/cryostat.d/recordings.d,relabel=shared,bind-propagation=shared \
     --mount type=tmpfs,target=/opt/cryostat.d/templates.d \
     --mount type=bind,source="$(dirname $0)/truststore",destination=/truststore,relabel=shared,bind-propagation=shared \
     --mount type=bind,source="$(dirname $0)/certs",destination=/certs,relabel=shared,bind-propagation=shared \
```

This minor configuration change lets me easily view the contents of the archive directory.

Then I did:
```
https -f POST :8181/api/v2/rules Content-Type:"multipart/form-data" name=aRule matchExpression="target.alias=='es.andrewazor.demo.Main' && target.annotations.cryostat.PORT==9093" eventSpecifier=template=Continuous archivalPeriodSeconds=10 preservedArchives=3
```

and let it run for a bit, and also manually re-uploaded two recordings into the archives from the web interface.

```
$ tree recordings/
recordings/
├── ONSXE5TJMNSTU2TNPA5HE3LJHIXS6L3KNZSGSL3SNVUTULZPMNZHS33TORQXIORZGA4TGL3KNV4HE3LJ
│   ├── es-andrewazor-demo-Main_auto_aRule_20210806T192754Z.jfr
│   ├── es-andrewazor-demo-Main_auto_aRule_20210806T192804Z.jfr
│   ├── es-andrewazor-demo-Main_auto_aRule_20210806T192814Z.jfr
│   └── es-andrewazor-demo-Main_foo_20210806T192644Z.jfr
└── unlabelled
    ├── es-andrewazor-demo-Main_auto_aRule_20210806T192414Z.jfr
    └── es-andrewazor-demo-Main_ongoing_20210806T191756Z.jfr

2 directories, 6 files
```

Then I stopped and restarted `sh smoketest.sh` and did a `https :8181/api/v1/recordings` and got:

```json
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 239
content-type: text/plain

[
    {
        "downloadUrl": "https://0.0.0.0:8181/api/v1/recordings/es-andrewazor-demo-Main_auto_aRule_20210806T192804Z.jfr",
        "name": "es-andrewazor-demo-Main_auto_aRule_20210806T192804Z.jfr",
        "reportUrl": "https://0.0.0.0:8181/api/v1/reports/es-andrewazor-demo-Main_auto_aRule_20210806T192804Z.jfr"
    },
    {
        "downloadUrl": "https://0.0.0.0:8181/api/v1/recordings/es-andrewazor-demo-Main_foo_20210806T192644Z.jfr",
        "name": "es-andrewazor-demo-Main_foo_20210806T192644Z.jfr",
        "reportUrl": "https://0.0.0.0:8181/api/v1/reports/es-andrewazor-demo-Main_foo_20210806T192644Z.jfr"
    },
    {
        "downloadUrl": "https://0.0.0.0:8181/api/v1/recordings/es-andrewazor-demo-Main_auto_aRule_20210806T192754Z.jfr",
        "name": "es-andrewazor-demo-Main_auto_aRule_20210806T192754Z.jfr",
        "reportUrl": "https://0.0.0.0:8181/api/v1/reports/es-andrewazor-demo-Main_auto_aRule_20210806T192754Z.jfr"
    },
    {
        "downloadUrl": "https://0.0.0.0:8181/api/v1/recordings/es-andrewazor-demo-Main_auto_aRule_20210806T192814Z.jfr",
        "name": "es-andrewazor-demo-Main_auto_aRule_20210806T192814Z.jfr",
        "reportUrl": "https://0.0.0.0:8181/api/v1/reports/es-andrewazor-demo-Main_auto_aRule_20210806T192814Z.jfr"
    },
    {
        "downloadUrl": "https://0.0.0.0:8181/api/v1/recordings/es-andrewazor-demo-Main_auto_aRule_20210806T192414Z.jfr",
        "name": "es-andrewazor-demo-Main_auto_aRule_20210806T192414Z.jfr",
        "reportUrl": "https://0.0.0.0:8181/api/v1/reports/es-andrewazor-demo-Main_auto_aRule_20210806T192414Z.jfr"
    },
    {
        "downloadUrl": "https://0.0.0.0:8181/api/v1/recordings/es-andrewazor-demo-Main_ongoing_20210806T191756Z.jfr",
        "name": "es-andrewazor-demo-Main_ongoing_20210806T191756Z.jfr",
        "reportUrl": "https://0.0.0.0:8181/api/v1/reports/es-andrewazor-demo-Main_ongoing_20210806T191756Z.jfr"
    }
]
```

Thanks for all your hard work on this @hareetd , this looks great.

hareetd:
Awesome, it's nice to see the changes visualized in that manner. 

And no problem, it was a great learning experience. Thanks for your detailed reviews/help @andrewazores!

src/main/java/io/cryostat/net/reports/ArchivedRecordingReportCache.java
```diff
@@ -41,111 +41,84 @@
...
+            Path saveFile =
+                    subprocessReportGeneratorProvider
+                            .get()
+                            .exec(
+                                    archivedRecording,
+                                    dest,
+                                    Duration.ofMillis(TimeoutHandler.TIMEOUT_MS))
+                            .get();
+            f.complete(saveFile);
+        } catch (ExecutionException | InterruptedException e) {
```
andrewazores, line 104: I think this can be removed/merged with the `catch (Exception e)` below. Or at least, the `fs.deleteIfExists(dest)` should be performed here as well.

src/main/java/io/cryostat/net/web/http/api/v1/RecordingGetHandler.java
```diff
@@ -86,28 +85,15 @@ public boolean isAsync() {
...
-                                                String.format(
-                                                        "Recording \"%s\" not found",
-                                                        recordingName));
-                                ctx.response().end();
-                            }
-                        });
+        try {
+            Path archivedRecording = recordingArchiveHelper.getRecordingPath(recordingName).get();
+            ctx.response().sendFile(archivedRecording.toString());
+            ctx.response().end();
```
andrewazores, line 91: I think `sendFile` already ends the response, so this can be removed.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,270 @@
...
+                                    "recording",
+                                    saveName,
+                                    "target",
+                                    connectionDescriptor.getTargetId()))
+                    .build()
+                    .send();
+        } catch (RecordingNotFoundException
+                | IOException
+                | URISyntaxException
+                | FlightRecorderException e) {
```
andrewazores, line 160: Can this `catch` be merged with the one below?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -39,90 +39,270 @@
...
+
+        try {
+            List<String> subdirectories = this.fs.listDirectoryChildren(archivedRecordingsPath);
+            Path archivedRecording =
+                    searchSubdirectories(subdirectories, archivedRecordingsPath, recordingName);
+            if (!(archivedRecording != null
+                    && fs.exists(archivedRecording)
+                    && fs.isRegularFile(archivedRecording)
+                    && fs.isReadable(archivedRecording))) {
+                throw new RecordingNotFoundException("archives", recordingName);
```
andrewazores, line 270: `RecordingNotFoundException` will be thrown here if the path exists but is either not `regularFile` (so is a directory, probably) or is unreadable. Should this be a differently named exception in this case, perhaps? Or at least, should the specific reason for the failure (not a file, or not readable) be included in the exception?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
hareetd: I was thinking to keep the specific `Exceptions` separate from the generic `Exception` because they'll be wrapped inside an `ExecutionException` but I guess that doesn't matter since the caller can just use `ExceptionUtils.getRootCause(e)` to determine the specific `Exception` thrown.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
hareetd: Good catch, the `ArchivePathException` class I created would be more appropriate here.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
andrewazores: In both branches the Future is completed exceptionally, so the caller would get an `ExecutionException` on the `.get()` call in either case. So I think collapsing these two clauses together actually won't change anything for callers.

andrewazores:
I see a test failure with this stack trace in the Cryostat logs, locally:

```
Caused by: java.lang.ClassCastException: class java.io.IOException cannot be cast to class io.cryostat.net.reports.SubprocessReportGenerator$ReportGenerationException (java.io.IOException is in module java.base of loader 'bootstrap'; io.cryostat.net.reports.SubprocessReportGenerator$ReportGenerationException is in unnamed module of loader 'app')
	at io.cryostat.net.web.http.api.v1.TargetReportGetHandler.targetRecordingNotFound(TargetReportGetHandler.java:124)
	at io.cryostat.net.web.http.api.v1.TargetReportGetHandler.handleAuthenticated(TargetReportGetHandler.java:114)
	at io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler.handle(AbstractAuthenticatedRequestHandler.java:85)
	at io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler.handle(AbstractAuthenticatedRequestHandler.java:62)
```

I suspect this is probably the same thing happening and causing the CI run to hang.

hareetd:
Thanks, I'll take a look into that. What's weird is that when I run tests locally sometimes they still pass. Also, when they do fail it's not necessarily the exact same tests failing.

andrewazores:
Yea, I'm seeing the same thing - some flaky itest results right now.

andrewazores:
#634

andrewazores:
> Edit: On that note, during an interactive rebase is it fine if I make any changes I see fit when resolving conflicts or should I stick to only pressing the "Accept current changes", "Accept incoming changes", "Accept both changes", etc. commands? I'm worried I might mess something up regarding the commit history if I move stuff around or add/delete too much.

It's best to just resolve the conflicts during the rebase so that each commit in the series can re-apply. If there are any further adjustments required - in this example, updating tests because a method was added to an interface in the branch you're rebasing onto - it's usually easiest/cleanest to do as you did and add that as an additional commit on top. The more you do to adjust the commits in series during conflict resolution, the more likely those changes are to cause further conflicts with the next commits in the series.

andrewazores (approved):

PR #552, opened 2021-07-06
ci(mergify): remove post_check action

andrewazores:
post_check action is apparently a premium/paid feature, Remove it from the config


ebaron (approved):
Too bad

PR #557, opened 2021-07-06
refactor(auto-rules): Periodic archiver should not use http req to self

### LINKED ISSUE #543: Automatic Rules PeriodicArchiver should not use HTTP requests to self

jan-law:
Fixes #543 

I renamed `recordingCreationHelper` to `recordingHelper` as it now contains helper functions to save and delete recordings. These helper functions are used in both `PeriodicArchiver.java` and `TargetRecordingPatchSave.java`.

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingHelper.java
@@ -37,10 +37,17 @@
  */
 package io.cryostat.recordings;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import javax.inject.Named;
 
 import org.openjdk.jmc.common.unit.IConstrainedMap;
 import org.openjdk.jmc.flightrecorder.configuration.events.EventOptionID;
@@ -48,18 +55,24 @@ import org.openjdk.jmc.flightrecorder.configuration.recording.RecordingOptionsBu
 import org.openjdk.jmc.rjmx.services.jfr.IEventTypeInfo;
 import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
 
+import io.cryostat.MainModule;
 import io.cryostat.commands.internal.EventOptionsBuilder;
 import io.cryostat.core.net.JFRConnection;
+import io.cryostat.core.sys.Clock;
+import io.cryostat.core.sys.FileSystem;
 import io.cryostat.core.templates.TemplateType;
 import io.cryostat.messaging.notifications.NotificationFactory;
 import io.cryostat.net.ConnectionDescriptor;
 import io.cryostat.net.TargetConnectionManager;
 import io.cryostat.net.web.http.HttpMimeType;
+import io.cryostat.platform.PlatformClient;
+import io.cryostat.util.URIUtil;
 
+import io.vertx.ext.web.handler.impl.HttpStatusException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 
-public class RecordingCreationHelper {
+public class RecordingHelper {
 
     private static final String NOTIFICATION_CATEGORY = "RecordingCreated";
 
@@ -69,13 +82,25 @@ public class RecordingCreationHelper {
     private final TargetConnectionManager targetConnectionManager;
     private final EventOptionsBuilder.Factory eventOptionsBuilderFactory;
     private final NotificationFactory notificationFactory;
+    private final FileSystem fs;
+    private final Path recordingsPath;
+    private final Clock clock;
+    private final PlatformClient platformClient;
 
-    RecordingCreationHelper(
+    RecordingHelper(
+            FileSystem fs,
+            @Named(MainModule.RECORDINGS_PATH) Path recordingsPath,
             TargetConnectionManager targetConnectionManager,
             EventOptionsBuilder.Factory eventOptionsBuilderFactory,
+            Clock clock,
+            PlatformClient platformClient,
             NotificationFactory notificationFactory) {
+        this.fs = fs;
+        this.recordingsPath = recordingsPath;
         this.targetConnectionManager = targetConnectionManager;
         this.eventOptionsBuilderFactory = eventOptionsBuilderFactory;
+        this.clock = clock;
+        this.platformClient = platformClient;
         this.notificationFactory = notificationFactory;
     }
 
@@ -117,6 +142,30 @@ public class RecordingCreationHelper {
                 });
     }
 
+    public String saveRecording(ConnectionDescriptor connectionDescriptor, String recordingName)
+            throws Exception {
+
+        String saveName =
+                targetConnectionManager.executeConnectedTask(
+                        connectionDescriptor,
+                        connection -> {
+                            Optional<IRecordingDescriptor> descriptor =
+                                    getDescriptorByName(connection, recordingName);
+
+                            if (descriptor.isPresent()) {
+                                return writeRecordingToDestination(connection, descriptor.get());
+                            } else {
+                                throw new HttpStatusException(
+                                        404,
+                                        String.format(
+                                                "Recording with name \"%s\" not found",
+                                                recordingName));
+                            }
+                        });
+
+        return saveName;
+    }
+
     public static Pair<String, TemplateType> parseEventSpecifierToTemplate(String eventSpecifier)
             throws IllegalArgumentException {
         if (TEMPLATE_PATTERN.matcher(eventSpecifier).matches()) {
@@ -133,7 +182,7 @@ public class RecordingCreationHelper {
         throw new IllegalArgumentException(eventSpecifier);
     }
 
-    private Optional<IRecordingDescriptor> getDescriptorByName(
+    public Optional<IRecordingDescriptor> getDescriptorByName(
             JFRConnection connection, String recordingName) throws Exception {
         return connection.getService().getAvailableRecordings().stream()
                 .filter(recording -> recording.getName().equals(recordingName))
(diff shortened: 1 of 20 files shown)


andrewazores:
@hareetd the RecordingHelper change sounds very similar to what we were discussing for your PR draft.

jan-law:
How do I add a release type to this pull request?
```
Error: No release type found in pull request title "refactor(auto-rules) Periodic archiver should not use http req to self".
```

andrewazores:
The PR title has to follow the format:verb(area): some messageI think the check here failed because the colon is missing.

hareetd:
@andrewazores Yeah, it does. I'll try to make the changes in my PR stay logically consistent with what @jan-law has done in the `RecordingHelper`.

src/main/java/io/cryostat/recordings/RecordingHelper.java
```diff
@@ -117,6 +146,52 @@ public IRecordingDescriptor startRecording(
...
+                targetConnectionManager.executeConnectedTask(
+                        connectionDescriptor,
+                        connection -> {
+                            Optional<IRecordingDescriptor> descriptor =
+                                    getDescriptorByName(connection, recordingName);
+
+                            if (descriptor.isPresent()) {
+                                return writeRecordingToDestination(connection, descriptor.get());
+                            } else {
+                                throw new HttpStatusException(
```
andrewazores, line 162: This is more of a style thing relating to loose coupling or encapsulation, but I think the `RecordingHelper` should avoid throwing `HttpStatusException`s. Here, for example, this should be something like a `RecordingNotFoundException`. The `TargetRecordingPatchSaveHandler` should then catch those and rethrow them as `HttpStatusException 404`.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -105,50 +98,30 @@ public void run() {
...
                                 String.format(
                                         "/api/v1/targets/%s/recordings/%s",
                                         URLEncodedUtils.formatSegments(
                                                 serviceRef.getServiceUri().toString()),
-                                        URLEncodedUtils.formatSegments(rule.getRecordingName())))
+                                        URLEncodedUtils.formatSegments()))
                         .normalize()
                         .toString();
+
         this.logger.trace("PATCH \"save\" {}", path);
```
andrewazores, line 121: This should be removed, or moved back to the `PATCH` handler.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -105,50 +98,30 @@ public void run() {
...
-        // creates the recording from the logic that simply figures out the recording parameters
-        // from the POST form, path param, and headers. Then the handler should consume the API
-        // exposed by this refactored chunk, and this refactored chunk can also be consumed here
-        // rather than firing HTTP requests to ourselves
+    void performArchival() throws InterruptedException, ExecutionException, Exception {
 
+        String recordingName = rule.getRecordingName();
+        ConnectionDescriptor connectionDescriptor =
+                new ConnectionDescriptor(serviceRef, credentialsManager.getCredentials(serviceRef));
         String path =
```
andrewazores, line 111: Do we still need this path?

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -161,27 +134,18 @@ void performArchival() throws InterruptedException, ExecutionException {
                         .normalize()
                         .toString();
         this.logger.trace("DELETE {}", path);
```
andrewazores, line 136: Likewise, this should be moved back to the Handler or removed.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
jan-law: The path is only used for the logger message above. Since we're moving the log message, `path` can be removed too

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -105,50 +98,30 @@ public void run() {
...
-                                return;
-                            }
-                            future.complete(resp.bodyAsString());
-                        });
-        this.previousRecordings.add(future.get());
+        String saveName = recordingHelper.saveRecording(connectionDescriptor, recordingName);
+        this.previousRecordings.add(saveName);
     }
 
     Future<Boolean> pruneArchive(String recordingName) {
```
andrewazores, line 127: This should probably be public (anything not `private` in here should most likely be `public`).

Also, everything `public` in here is some kind of blocking operation - either filesystem access or network access, or both. So, I would say either every method should return a `Future` or none should. I think I would prefer them all to return `Future`s so that it's made clear that these are all potentially long operations and whoever is calling them should be doing so asynchronously/on some worker thread.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
jan-law: Should we keep `logger.trace("Pruning {}", recordingName);`? `performArchival()` doesn't have a similar log message

andrewazores (approved):
Looks great other than those minor style things above.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
andrewazores: I think that can also be removed, or moved to the PerioicArchiver.

jan-law: ^ I'll remove it for consistency

src/main/java/io/cryostat/recordings/RecordingHelper.java
jan-law: How should  `PeriodicArchiver.run()` handle `RecordingNotFoundException` from the helper functions and `Exception` from `targetConnectionManager.execute`? 
`run()` throws an `HttpStatusException` but I realized the archiver isn't part of the HTTP API. Should I use the `failureNotifier`?

andrewazores: Yea, I don't think `run` should throw `HttpStatusException` either. But being that it's there because `PeriodicArchiver implements Runnable`, `run` can't be declared to throw any checked Exception type either. Any exceptions that occur need to be handled some other way than by (re)throwing. The purpose of the `failureNotifier` is so that the `RuleProcessor` can cancel the scheduled future runs of `PeriodicArchiver`s if they encounter failures, so that they don't continually retry and fail indefinitely. `RecordingNotFoundException` sounds like a good reason to stop running. The general `Exception` from `targetConnectionManager.execute` is harder to say - it seems like it might depend on what the exact problem is/was. Most likely, failures from `execute` will be because the target credentials were rejected, or the network connection could not be opened. If the credentials are rejected then it seems like stopping makes sense. If the network connection fails then it might make sense to just skip the current execution but not cancel altogether, because network problems are often transitory.

jan-law: > it seems like it might depend on what the exact problem is/was

How can I determine what caused `Exception`?

andrewazores: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/exception/ExceptionUtils.html#hasCause-java.lang.Throwable-java.lang.Class-

You could do something like `if (ExceptionUtils.hasCause(ex, ConnectionException.class))`, for example. `AbstractAuthenticatedRequestHandler` does a few similar checks so that API handlers have consistent behaviour in various generalized failure cases like JMX connection errors, you could reference that for some inspiration.

jan-law: What do you think about only applying the `failureNotifier` for `RecordingNotFound` and credentials-related exceptions? Other `Exception`s will get logged and the scheduler will retry.

```
} catch (InterruptedException | ExecutionException | RecordingNotFoundException e) {

            logger.error(e);
            failureNotifier.apply(Pair.of(serviceRef, rule));
        } catch (Exception e) {
            logger.error(e);

            if (ExceptionUtils.hasCause(e, SecurityException.class)
                    || ExceptionUtils.hasCause(e, SaslException.class)) {

                failureNotifier.apply(Pair.of(serviceRef, rule));
            }
        }
```

andrewazores: Seems to make sense to me.

jan-law: Thanks! PR is ready to go

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -102,86 +96,55 @@ public void run() {
             }
 
             performArchival();
-        } catch (InterruptedException | ExecutionException e) {
+        } catch (InterruptedException | ExecutionException | RecordingNotFoundException e) {
```
andrewazores, line 99: I think this can be refactored - each of these exception types could be moved down into the general `catch (Exception)` block and added to the `if (hasCause(SomeException.class))` handling, and it would behave the same, right?

src/main/java/io/cryostat/rules/PeriodicArchiver.java
```diff
@@ -102,86 +96,55 @@ public void run() {
             }
 
             performArchival();
```
andrewazores, line 98: Since `performArchival` returns a `Future`, this should call ex. `.get()` on that `Future` so that the call blocks until the result is completed.

src/main/java/io/cryostat/rules/PeriodicArchiver.java
jan-law: Good idea, will do

andrewazores (approved):

PR #559, opened 2021-07-08
ci(log): print podman logs on CI failure

andrewazores:
If there are itest failures, the log output of the cryostat container itself will be printed out in the workflow action run to aid troubleshooting. As it is now, only the itest runner's side of the output is seen, but not the cryostat backend's internal logging, which would likely contain additional details.


ebaron (approved):
Good idea. The log gets created for me locally and the action makes sense.

PR #560, opened 2021-07-08
build(jib): Update-maven-jib-plugin

### LINKED ISSUE #529: Update maven jib plugin
3.0.0 is currently used, but 3.1.1 is available.

https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/CHANGELOG.md

The classpath stuff in the `CHANGED` section sounds potentially useful to add in to our `entrypoint.sh`.

jan-law:
Resolves #529

Upgrades maven jib plugin to version 3.1.1 and uses the new JVM argument files `/app/jib-classpath-file` and `/app/jib-main-class-file` to set the custom container entrypoint.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -519,7 +519,6 @@
           <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
           <entrypoint>${cryostat.entrypoint}</entrypoint>
           <mainClass>${cryostat.mainClass}</mainClass>
-          <extraClasspath>${env.CLASSPATH}</extraClasspath>
           <ports>
             <port>${cryostat.webPort}</port>
             <port>${cryostat.listenPort}</port>
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
   <org.codehaus.mojo.exec.plugin.version>3.0.0</org.codehaus.mojo.exec.plugin.version>
   <com.mycila.license.maven.plugin.version>4.0</com.mycila.license.maven.plugin.version>
   <org.owasp.dependency.check.version>6.1.5</org.owasp.dependency.check.version>
-  <com.google.cloud.tools.jib.maven.plugin.version>3.0.0</com.google.cloud.tools.jib.maven.plugin.version>
+  <com.google.cloud.tools.jib.maven.plugin.version>3.1.1</com.google.cloud.tools.jib.maven.plugin.version>
 
   <com.google.dagger.version>2.34.1</com.google.dagger.version>
   <com.google.dagger.compiler.version>2.26</com.google.dagger.compiler.version>
@@ -517,6 +517,8 @@
           <format>OCI</format>
           <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
           <entrypoint>${cryostat.entrypoint}</entrypoint>
+          <mainClass>io.cryostat.Cryostat</mainClass>
+          <extraClasspath>${env.CLASSPATH}</extraClasspath>
           <ports>
             <port>${cryostat.webPort}</port>
             <port>${cryostat.listenPort}</port>
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,7 @@
   <cryostat.minimal>false</cryostat.minimal>
   <cryostat.imageStream>quay.io/cryostat/cryostat</cryostat.imageStream>
   <cryostat.entrypoint>/app/entrypoint.sh</cryostat.entrypoint>
+  <cryostat.mainClass>io.cryostat.Cryostat</cryostat.mainClass>
   <cryostat.rjmxPort>9091</cryostat.rjmxPort>
   <cryostat.webPort>8181</cryostat.webPort>
   <cryostat.listenPort>9090</cryostat.listenPort>
@@ -249,7 +250,7 @@
       <artifactId>exec-maven-plugin</artifactId>
       <version>${org.codehaus.mojo.exec.plugin.version}</version>
       <configuration>
-        <mainClass>io.cryostat.Cryostat</mainClass>
+        <mainClass>${cryostat.mainClass}</mainClass>
         <additionalClasspathElements>
           <additionalClasspathElement>${project.build.directory}/assets/app/resources</additionalClasspathElement>
         </additionalClasspathElements>
@@ -517,7 +518,7 @@
           <format>OCI</format>
           <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
           <entrypoint>${cryostat.entrypoint}</entrypoint>
-          <mainClass>io.cryostat.Cryostat</mainClass>
+          <mainClass>${cryostat.mainClass}</mainClass>
           <extraClasspath>${env.CLASSPATH}</extraClasspath>
           <ports>
             <port>${cryostat.webPort}</port>
--- a/src/main/extras/app/entrypoint.sh
+++ b/src/main/extras/app/entrypoint.sh
@@ -162,6 +162,6 @@ export SSL_TRUSTSTORE_DIR
 set -x
 exec java \
     "${FLAGS[@]}" \
-    -cp "$CLASSPATH" \
-    io.cryostat.Cryostat \
+    -cp @/app/jib-classpath-file \
+    @/app/jib-main-class-file \
     "$@"
--- a/src/main/extras/app/entrypoint.sh
+++ b/src/main/extras/app/entrypoint.sh
@@ -150,7 +150,7 @@ if [ -n "$CRYOSTAT_JUL_CONFIG" ]; then
     FLAGS+=("-Djava.util.logging.config.file=$CRYOSTAT_JUL_CONFIG")
 fi
 
-CLASSPATH="/app/resources:/app/classes:/app/libs/*"
+CLASSPATH="$( cat /app/jib-classpath-file )"
 if [ -n "CRYOSTAT_CLIENTLIB_PATH" ]; then
     CLASSPATH="$CLASSPATH:$CRYOSTAT_CLIENTLIB_PATH/*"
 fi
@@ -162,6 +162,6 @@ export SSL_TRUSTSTORE_DIR
 set -x
 exec java \
     "${FLAGS[@]}" \
-    -cp @/app/jib-classpath-file \
+    -cp "$CLASSPATH" \
     @/app/jib-main-class-file \
     "$@"


jan-law:
If we already know the main class is `io.cryostat.Cryostat` and the runtime classpath is inside the `CLASSPATH` env var, how do Jib's JVM argument files provide a more reliable way of knowing these paths?

[Jib readme - Custom Container Entrypoint](https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/README.md#custom-container-entrypoint):
> you will want to have a way inside a shell script to reliably know the default runtime classpath and the main class that Jib would use by default. To help this, Jib >= 3.1 creates two JVM argument files under /app (the default app root) inside the built image.

andrewazores:
The mainClass is definitely known ahead of time so that isn't really buying us anything, although it is nice to avoid hardcoding things (ex. if we change the name of the class, or change the package, etc.). Similar reasoning for the classpath - using Jib's generated file doesn't gain us anything directly right now, but it might make maintenance a little easier if we do change anything.

pom.xml
```diff
@@ -502,6 +502,8 @@
           <format>OCI</format>
           <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
           <entrypoint>${cryostat.entrypoint}</entrypoint>
+          <mainClass>io.cryostat.Cryostat</mainClass>
+          <extraClasspath>${env.CLASSPATH}</extraClasspath>
```
andrewazores, line 506: I'm not sure about this line - if we don't add this, will the Jib-generated classpath file still exist and be populated with default values? If so then I think that would be sufficient. Adding arbitrary things to that classpath based on the host/builder's environment doesn't seem necessary right now.

pom.xml
```diff
@@ -502,6 +502,8 @@
           <format>OCI</format>
           <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
           <entrypoint>${cryostat.entrypoint}</entrypoint>
+          <mainClass>io.cryostat.Cryostat</mainClass>
```
andrewazores, line 505: There's also already a `mainClass` reference in the `exec-maven-plugin` configuration. Could this value be extracted to a property and referenced in both of these instances?

src/main/extras/app/entrypoint.sh
```diff
@@ -162,6 +162,6 @@ export SSL_TRUSTSTORE_DIR
 set -x
 exec java \
     "${FLAGS[@]}" \
-    -cp "$CLASSPATH" \
-    io.cryostat.Cryostat \
+    -cp @/app/jib-classpath-file \
```
andrewazores, line 165: I think we want the Jib classpath file included *and* at least part of the `$CLASSPATH` env var, since the `$CLASSPATH` currently includes the `$CRYOSTAT_CLIENTLIB_PATH`. Just replacing `$CLASSPATH` with the file seems like it would exclude the clientlib directory from the classpath used in the end. Can this be included in the pom.xml config for Jib, maybe, so that the file content also includes the clientlib path? If not then I think joining the clientlib path with the Jib file together would work.

jan-law: How can I check that my last commit contains the clientlib path in `/app/jib-classpath-file`?

pom.xml
```diff
@@ -504,7 +504,7 @@
           <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
           <entrypoint>${cryostat.entrypoint}</entrypoint>
           <mainClass>${cryostat.mainClass}</mainClass>
-          <extraClasspath>${env.CLASSPATH}</extraClasspath>
+          <extraClasspath>${env.CLASSPATH}:${env.CRYOSTAT_CLIENTLIB_PATH}</extraClasspath>
```
andrewazores, line 507: I think this should just be a hardcoded string (defined in a Maven property maybe, or else just hardcoded here). We want/need this to be very repeatable across different build environments, whether that's your laptop or my desktop or the GitHub Actions CI builder, so all of these inputs should be consistent and not come from the builder's environment.

Basically, consider that the environment witnessed by the builder reading the `pom.xml` is certainly a different environment from that witnessed by the Cryostat instance running inside the built image later on.

src/main/extras/app/entrypoint.sh
andrewazores: I like to use dive: https://github.com/wagoodman/dive

ex. `dive podman://quay.io/cryostat/cryostat:latest`

This lets you see the contents within an image by listing directories and filenames/permissions/etc, although I'm not sure it will show you the contents of specific files. You could use `podman export` for that if not.

andrewazores: Otherwise, a cheap hack would be to just modify the `entrypoint.sh` to do something like `cat /app/jib-classpath-file` so that the file contents get dumped to stdout when you run the container.

pom.xml
jan-law: If environment variables that need to be consistent across all builder environments are hardcoded, why are some properties defined in `pom.xml` and others are defined in the shell files?

andrewazores: By "the shell files", do you mean `entrypoint.sh`? Or do you mean things like `smoketest.sh` and `run.sh`?

`pom.xml` is only used by Maven for building the product (compiling Java, building web-client, assembling the OCI image) and running integration tests against that product. The various Maven properties defined and used here are all part of the build system, not part of the Cryostat runtime, and AFAIK none of it currently depends on the environment variables present on the host (builder, who runs mvn) machine. Some of those properties and configurations may be used to set environment variables that are passed on to the OCI container during integration testing, for example, but this is a different and later stage than building the image.

`entrypoint.sh` is one of the pieces that gets put inside that OCI image and is what is initially run within the container when it's started, so it's part of the runtime and is completely distinct/independent of the build environment. It checks the runtime environment and does some setup (ex. SSL/TLS cert import) before starting the actual Cryostat Java process.

`smoketest.sh`/`run.sh` set environment variables that are provided to the OCI container at runtime, and those environment variables end up read by `entrypoint.sh` or the Cryostat Java process within the container.

andrewazores: To give a more concrete illustration of why I'm against setting `${env.CLASSPATH}` here, this would mean that I can do:

`CLASSPATH=/home/aazores/workspace/cryostat/someExtraClasspath mvn package`

and this would produce a Cryostat image for me where Jib generates a classpath file that contains the string `/home/aazores/workspace/cryostat/someExtraClasspath` as part of the application's runtime classpath. But for anyone else who wants to run this image, that's a meaningless path. Even worse, for *myself* it's a meaningless path, because I set it using an absolute path for my development machine, but this path definitely doesn't existing within the Cryostat image. If I change that to some path that does exist in the image it still doesn't do what it sounds like, because Jib will add that path to the classpath, but Maven/Jib doesn't actually copy the contents of that directory into the image, and so at runtime Cryostat will add that path to the classpath but there will still be nothing in there. Or there might be things in there, but not what I thought there would be at build time, because my build environment is in an independent time/place from the runtime environment Cryostat finds itself in.

If I want to actually include things into the classpath I would not only need to set this `CLASSPATH` env var, I would also need to configure Jib to include the directory into the built image (that's what `src/main/extras` is). But if that's the case, then that path is well-known within the pom.xml already since it has to be there for the Jib plugin to include it, so defining it by env var means either it can/will be overridden by an incorrect value, or it will be empty by default even though the correct value is already known.

src/main/extras/app/entrypoint.sh
jan-law: There's an open [feature request](https://github.com/wagoodman/dive/issues/336) to view files in Dive.

In the meantime, `podman export` shows that `${env.CLASSPATH}:/clientlib` gets placed in front of Jib's computed classpath: [github gist link](https://gist.githubusercontent.com/jan-law/336111fcf6980397054bb997a8dc548c/raw/fc8c4cfef88208a2b7122dd916c0aad8ffec012a/jib-classpath-file)

andrewazores: `${env.CLASSPATH}:/clientlib`

That looks very wrong to me - I would have thought that it would at least be expanded out, not inserted as a literal string into the classpath. Does `java -cp` even accept that? It's either going to cause the command to choke and fail to start, or it will be ignored and may as well be removed.

andrewazores: Is it possible to do `java -cp @/app/jib-classpath-file:/something:/another/directory`? ie, to use the `@` syntax to read from a file and also to append to it using the typical semicolon separator to directly add additional paths?

`entrypoint.sh` currently does this:

```
CLASSPATH="/app/resources:/app/classes:/app/libs/*"
if [ -n "CRYOSTAT_CLIENTLIB_PATH" ]; then
    CLASSPATH="$CLASSPATH:$CRYOSTAT_CLIENTLIB_PATH/*"
fi
```

so I wonder if this can just be replaced by:

```
CLASSPATH="@/app/jib-classpath-file"
if [ -n "CRYOSTAT_CLIENTLIB_PATH" ]; then
    CLASSPATH="$CLASSPATH:$CRYOSTAT_CLIENTLIB_PATH/*"
fi
```

this allows the `$CRYOSTAT_CLIENTLIB_PATH` to still be set at runtime, rather than requiring it to be hardcoded to `/clientlib`.

pom.xml
jan-law: Thank you, I was wondering why `run.sh` and `smoketest.sh` set environment variables. This makes more sense now.

From the github gist above, it looks like Jib's computed classpath is the same as the classpath defined in `entrypoint.sh` (`/app/resources:/app/classes:/app/libs/*`), which means I don't think I need to include `{env.CLASSPATH}` to `<extraClasspath>`.

andrewazores: Makes sense. In the other conversation thread above I also explained that I don't think the `clientLibPath` needs to be added to `extraClasspath` either. Doing it that way requires the path to be determined at build time only, which removes some flexibility for end users who are running/deploying Cryostat. It isn't a hard requirement that that path is configurable, but I think it makes some sense to allow it to be. It could save an end user some time and effort if they want to build their own Cryostat using a different base image which already has a directory containing some client libraries, for example.

src/main/extras/app/entrypoint.sh
jan-law: `entrypoint.sh`:
```
CLASSPATH="@/app/jib-classpath-file"
if [ -n "CRYOSTAT_CLIENTLIB_PATH" ]; then
    CLASSPATH="$CLASSPATH:$CRYOSTAT_CLIENTLIB_PATH/*"
fi
...

exec java \
    "${FLAGS[@]}" \
    -cp "$CLASSPATH" \
    @/app/jib-main-class-file \
    "$@"
```
Output: `Error: could not open `/app/jib-classpath-file:/clientlib/*'`

I also tried
`echo "$CRYOSTAT_CLIENTLIB_PATH/*" >> @/app/jib-classpath-file`
which gives me
`/app/entrypoint.sh: line 158: @/app/jib-classpath-file: No such file or directory`

pom.xml
jan-law: Resolving this conversation as Jib generates the same `CLASSPATH` by default

src/main/extras/app/entrypoint.sh
andrewazores: That `echo` line should work, but you would have to remove the `@`. That's special syntax that the `java` command uses to know that the argument passed to the `-cp` flag is a file containing text defining the classpath, rather than the argument simply being the text of the classpath itself.

I think you would also need to prepend the `:` separator, so that the clientlib path defined by the env var is correctly appended/separated from the pre-existing file contents.

jan-law: > if [ -n "$CRYOSTAT_CLIENTLIB_PATH" ]; then
>   echo -n ":$CRYOSTAT_CLIENTLIB_PATH/*" >> /app/jib-classpath-file
> fi
> 
> exec java -cp @/app/jib-classpath-file

Hmm, I don't have permissions to edit the file:
`/app/entrypoint.sh: line 155: /app/jib-classpath-file: Permission denied`

I'll also take a look at Jib's github issues and let you know if I find anything.

andrewazores: Ah. That's unfortunate. You could try this then:

```
CLASSPATH="$( cat /app/jib-classpath-file )"
if [ -n "CRYOSTAT_CLIENTLIB_PATH" ]; then
    CLASSPATH="$CLASSPATH:$CRYOSTAT_CLIENTLIB_PATH/*"
fi

exec java -cp "$CLASSPATH"
```

Presumably the uid has access to at least read the jib-classpath-file, so this should work around lacking write permissions for it.

If you have read permissions you could also read and copy the contents to a different, new file. Then conditionally append the additional clientlib path to that new file, and then pass the path to the new file with the `java -cp @newfile`.

jan-law: >CLASSPATH="$( cat /app/jib-classpath-file )"

This approach works:
```
+ exec java ...(flags)... -cp '/app/resources:/app/classes:/app/libs/cryostat-core-2.3.0.jar:
...(more jars)...
:/app/libs/jsoup-1.13.1.jar:/clientlib/*' @/app/jib-main-class-file
```

I don't have copy permissions to create a new file, so I'll stick to the above solution.
`cp: cannot create regular file '/app/extended-classpath-file': Permission denied`

jan-law:
Cleaned up git history by removing commits that don't contribute to our solution

src/main/extras/app/entrypoint.sh
andrewazores: The uid running inside the container probably just doesn't have write permissions inside `/app`, but you could probably write inside of `/tmp` instead, or somewhere else. Anyway, just reading the generated contents into an env var seems fine. It seems pretty unlikely that the classpath length would exceed the ~32k character limit of an env var. If it ever does (!) we'll find a way to do it by file copy/append instead, again.

andrewazores (commented):
Looks good - just rebase this onto latest main and I'll approve/merge it. Thanks!

andrewazores (approved):

PR #562, opened 2021-07-09
docs(auto-rules): update HTTP_API.md

andrewazores:
Related to #413

### DIFF
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -1532,6 +1532,65 @@ The handler-specific descriptions below describe how each handler populates the
 
 ### Stored Target Credentials
 
+* #### `TargetCredentialsPostHandler`
+
+    ##### synopsis
+    Creates stored credentials for a given target. These are used for automated
+    rules processing - if a Target JVM requires JMX authentication, Cryostat
+    will use stored credentials when attempting to open JMX connections to the
+    target. These are retained in Cryostat's memory only and not persisted to
+    disk.
+
+    ##### request
+    `POST /api/v2/targets/:targetId/credentials`
+
+    The request should be an HTTP form with the attributes `"username"` and
+    `"password"`. Both are required.
+
+    ##### response
+    `200` - The result is null. The request was processed successfully and the
+    credentials were stored, potentially overriding previous credentials for the
+    same target.
+
+    `400` - `"username"` and/or `"password"` were not provided.
+
+    `401` - User authentication failed. The reason is an error message.
+    There will be an `X-WWW-Authenticate: $SCHEME` header that indicates
+    the authentication scheme that is used.
+
+    `500` - There was an unexpected error.
+
+    ##### example
+    ```
+    $ curl -F username=user -F password=pass http://0.0.0.0:8181/api/v2/targets/localhost/credentials
+    {"meta":{"type":"text/plain","status":"OK"},"data":{"result":null}}
+    ```
+
+* #### `TargetCredentialsDeleteHandler`
+
+    ##### synopsis
+    Deletes stored credentials for a given target.
+
+    ##### request
+    `DELETE /api/v2/targets/:targetId/credentials`
+
+    ##### response
+    `200` - The result is null. The request was processed successfully and the
+    credentials were deleted.
+
+    `404` - The target had no stored credentials.
+
+    `401` - User authentication failed. The reason is an error message.
+    There will be an `X-WWW-Authenticate: $SCHEME` header that indicates
+    the authentication scheme that is used.
+
+    `500` - There was an unexpected error.
+
+    ##### example
+    ```
+    $ curl -X DELETE http://0.0.0.0:8181/api/v2/targets/localhost/credentials
+    {"meta":{"type":"text/plain","status":"OK"},"data":{"result":null}}
+    ```
 
 ### Security
 
--- a/HTTP_API.md
+++ b/HTTP_API.md
@@ -1234,14 +1234,22 @@ The handler-specific descriptions below describe how each handler populates the
 
 | What you want to do                                                       | Which handler you should use                                                    |
 | ------------------------------------------------------------------------- | --------------------------------------------------------------------------------|
-| **Recordings in target JVMs**                                             |                                                                                 |
+| **Recordings in Target JVMs**                                             |                                                                                 |
 | Search event types that can be produced by a target JVM                   | [`TargetEventsSearchGetHandler`](#TargetEventsSearchGetHandler)                 |
 | Get a list of recording options for a target JVM                          | [`TargetRecordingOptionsListGetHandler`](#TargetRecordingOptionsListGetHandler) |
 | Create a snapshot recording in a target JVM                               | [`TargetSnapshotPostHandler`](#TargetSnapshotPostHandler-1)                     |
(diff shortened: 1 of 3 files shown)


andrewazores:
Neat, #559 coming in handy right on time.

The failure looks unrelated to this PR, but here's what appears to be the culprit:

```
 SEVERE: Exception thrown
Caused by: org.openjdk.jmc.rjmx.services.jfr.FlightRecorderException: Could not create a recording!
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderServiceV2.start(FlightRecorderServiceV2.java:168)
	at io.cryostat.recordings.RecordingTargetHelper.lambda$startRecording$0(RecordingTargetHelper.java:101)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:128)
	at io.cryostat.recordings.RecordingTargetHelper.startRecording(RecordingTargetHelper.java:89)
	at io.cryostat.net.web.http.api.v1.TargetRecordingsPostHandler.lambda$handleAuthenticated$0(TargetRecordingsPostHandler.java:166)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:128)
	at io.cryostat.net.web.http.api.v1.TargetRecordingsPostHandler.handleAuthenticated(TargetRecordingsPostHandler.java:136)
	at io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler.handle(AbstractAuthenticatedRequestHandler.java:83)
	at io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler.handle(AbstractAuthenticatedRequestHandler.java:62)
Caused by: java.io.IOException: The client has been closed.
	at java.management/com.sun.jmx.remote.internal.ClientCommunicatorAdmin.restart(ClientCommunicatorAdmin.java:99)
	at java.management/com.sun.jmx.remote.internal.ClientCommunicatorAdmin.gotIOException(ClientCommunicatorAdmin.java:59)
	at java.management.rmi/javax.management.remote.rmi.RMIConnector$RMIClientCommunicatorAdmin.gotIOException(RMIConnector.java:1497)
	at java.management.rmi/javax.management.remote.rmi.RMIConnector$RemoteMBeanServerConnection.invoke(RMIConnector.java:1027)
	at org.openjdk.jmc.rjmx.internal.MCMBeanServerConnection.invoke(MCMBeanServerConnection.java:259)
	at org.openjdk.jmc.rjmx.ConnectionToolkit.invokeOperation(ConnectionToolkit.java:196)
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderCommunicationHelperV2.invokeJfrOperation(FlightRecorderCommunicationHelperV2.java:91)
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderCommunicationHelperV2.invokeOperation(FlightRecorderCommunicationHelperV2.java:82)
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderServiceV2.start(FlightRecorderServiceV2.java:166)
	... 15 more
```

HTTP_API.md
```diff
@@ -1368,6 +1376,222 @@ The handler-specific descriptions below describe how each handler populates the
...
+* #### `RuleGetHandler`
+
+    ##### synopsis
+    Get a JSON document describing a rule definition with the given name.
+
+    ##### request
+    `GET /api/v2/rules/:name`
+
+    ##### response
+    `200` - The result is a JSON string representing the rule defintion.
```
hareetd, line 1494: "Definition" is misspelled.

HTTP_API.md
```diff
@@ -1368,6 +1376,222 @@ The handler-specific descriptions below describe how each handler populates the
...
+
+    `401` - User authentication failed. The reason is an error message.
+    There will be an `X-WWW-Authenticate: $SCHEME` header that indicates
+    the authentication scheme that is used.
+
+    `500` - There was an unexpected error.
+
+    ##### example
+    ```
+    $ curl -F username=user -F password=pass http://0.0.0.0:8181/api/v2/targets/localhost/credentials
```
hareetd, line 1565: I think this should say "curl -X POST ..."

HTTP_API.md
```diff
@@ -1368,6 +1376,222 @@ The handler-specific descriptions below describe how each handler populates the
...
+
+    `401` - User authentication failed. The reason is an error message.
+    There will be an `X-WWW-Authenticate: $SCHEME` header that indicates
+    the authentication scheme that is used.
+
+    `404` - No rule with the given name exists.
+
+    `500`- An unexpected IOException occurred while deleting the rule
+    definition.
+
```
hareetd, line 1478: For consistency, the `RulesPostHandler`, `TargetCredentialsPostHandler`, and `TargetCredentialsDeleteHandler` should probably have the same explanation for an HTTP 500 response since (from what I can tell) all four handlers give such a response only upon catching an `IOException`. Unless, unlike this handler, these 3 handlers can have other unexpected errors that aren't explicitly stated.

andrewazores: > Unless, unlike this handler, these 3 handlers can have other unexpected errors that aren't explicitly stated.

That's generally the case for any `500` error - those are what you get if there is some unexpected internal error. I'm not really happy with the fact that other parts of the document list any specific reasons for `500`s, actually, because almost by definition this list is non-exhaustive, so including a partial list of reasons could be misleading for readers.

> all four handlers give such a response only upon catching an IOException

That's from a current reading of the explicit handling of an IOException, yes. But if the implementation changes at all and say a new NullPointerException becomes possible, that would also result in a 500 but would almost certainly not be explicitly handled and turned into a 500 by the handler - it would fall through to the WebServer failureHandler.

HTTP_API.md
andrewazores: `-F` implies the `POST` verb already. I can add it if you'd prefer, for consistency, but I think it's a little over-verbose.

hareetd: No, that's fine, your decision makes sense.

hareetd:
Looks good, will leave approval to @jan-law.

HTTP_API.md
jan-law, line 1565: If I try the following with cryostat smoketest, I get a `427`. How do I properly store credentials?
```
$ curl -v --insecure -F username=(omitted) -F password=(omitted) https://0.0.0.0:8181/api/v2/targets/cryostat:9091/credentials 

HTTP/1.1 200 OK

$ curl -v --insecure https://0.0.0.0:8181/api/v2/targets/cryostat:9091/recordingOptionsList

HTTP/1.1 427 Client Error (427)
```

andrewazores: You appear to have stored them correctly, but the stored credentials are only (currently) used into automated rules processing. When an external client (like your curl) makes a request, it is still required to provide credentials. I tried to document that in the synopsis for the endpoint:

```
Creates stored credentials for a given target. These are used for automated
rules processing - if a Target JVM requires JMX authentication, Cryostat
will use stored credentials when attempting to open JMX connections to the
target.
```

jan-law: I'm happy with that explanation. If the stored credentials change in the future to support other endpoints we can always update the description.

jan-law (approved):

Issue #474, opened 2021-04-29
Integration test API coverage


andrewazores:
Currently we have lots of unit test coverage, including for things like API request handlers, but we should also have a battery of integration tests to truly exercise the API surface and ensure that handlers are properly registered with the webserver and behave as expected in "real" conditions (not with mocks injected).

jan-law:
@hareetd as discussed, we can work on this together by submitting related PRs for each handler

jan-law:
How thorough should the integration tests be? For example, the `RulesPostHandler` unit test contains invalid mime tests that aren’t tested in `AutoRulesIT`. There are also other scenarios that would result in a 4xx or 5xx error that aren't covered in either test. 
Although it’s probably not feasible to cover all the 5xx errors, should we aim to cover all the 4xx situations with known causes?

andrewazores:
I think that sounds reasonable. Any `100-3xx` status codes defined by the handler (including any implicit `200`s that come from just returning intermediate responses in V2 handlers) should be exercised for sure, and any explicitly defined `4xx` codes as well. The generic `4xx`s like `401` and `427` could probably be added to some testing base class akin to the `StandardSelfTest` we already have, so that that behaviour is checked consistently in all tests where the handler is expected to require authorization (which is almost all of them).

If you can find any case where you can force a `500` to occur in an integration test then I would say that should actually be filed as a bug against Cryostat itself, and either fixed or handled to turn it into a `4xx` or whatever else is appropriate.

jan-law:
As the number of itests grow, how would you recommend organizing the tests?

`AutoRulesIT.java` already contains 6 tests for adding or deleting rules, adding or deleting credentials and various form validations. I thought of grouping tests for one handler in a single file, except some handlers are related. For example, you have to create a rule before deleting it.

Some of my tests can also reuse the same code with `@BeforeAll` only if all of the tests in one file use the same rule definition:
```
static MultiMap testRule = MultiMap.caseInsensitiveMultiMap();

    @BeforeAll
    static void setup() throws Exception {
        testRule.add("name", "Auto Rule");
...
```

andrewazores:
[...]

PR #571, opened 2021-07-13
test(auto-rules): Add POST rule itests

jan-law:
Related #474 

Adds integration tests to cover `4xx` status codes defined by `RulesPostHandler.java`.

### DIFF
--- a/src/test/java/itest/RulesPostFormIT.java
+++ b/src/test/java/itest/RulesPostFormIT.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package itest;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


jan-law:
Note: When using `sendForm()` to `POST` a new rule, Vertx will set the `Content-Type` header to `application/x-www-form-urlencoded`, unless you specify `multipart/form-data`. Any attempts to overwrite the `Content-Type` with an invalid value will be overridden with `application/x-www-form-urlencoded`.

```
void sendForm(MultiMap body,
              Handler<AsyncResult<HttpResponse<T>>> handler)

Like send(Handler) but with an HTTP request body multimap encoded as form and the content type set to application/x-www-form-urlencoded.
When the content type header is previously set to multipart/form-data it will be used instead.

Parameters:
body - the body
```

However, `POST`ing a rule with JSON allows me to test invalid mimes, as Vertx won't check the  `Content-Type` header when calling `sendJsonObject()`.
```
default Future<HttpResponse<T>> sendJsonObject(JsonObject body)

Parameters:
body - the body
```

src/test/java/itest/RulesPostFormIT.java
```diff
@@ -0,0 +1,149 @@
...
+                .post("/api/v2/rules")
+                .putHeader(HttpHeaders.CONTENT_TYPE.toString(), HttpMimeType.MULTIPART_FORM.mime())
+                .sendForm(
+                        null,
+                        ar -> {
+                            if (assertRequestStatus(ar, response)) {
+                                MatcherAssert.assertThat(
+                                        ar.result().statusCode(), Matchers.equalTo(400));
+                            }
+                        });
```
andrewazores, line 82: There should be a `response.get()` at the end here to ensure that the test method waits for the request execution to complete before proceeding to the next test.

andrewazores: IMO it's better to have the request handler complete the future and perform the assertion on the future value at the end. Here's a snippet from the `AutoRulesIT` showing that pattern:

```java
        CompletableFuture<JsonObject> preRules = new CompletableFuture<>();
        webClient
                .get("/api/v2/rules")
                .send(
                        ar -> {
                            if (assertRequestStatus(ar, preRules)) {
                                preRules.complete(ar.result().bodyAsJsonObject());
                            }
                        });
        JsonObject expectedPreRules =
                new JsonObject(
                        Map.of(
                                "meta", Map.of("type", HttpMimeType.JSON.mime(), "status", "OK"),
                                "data", Map.of("result", List.of())));
        MatcherAssert.assertThat(preRules.get(), Matchers.equalTo(expectedPreRules));
```

Reason being that this pattern ensures that the request is fired and completes, either successfully or not, before the test executor moves on to the next method or proceeds any further down the method. If the assertion is performed asynchronously within the response handler, then any failure mode where a response is simply never received will implicitly pass because the assertion is never executed either.

jan-law: Does that mean I also need `response.get()` after every request, including ones that return an error code?

src/test/java/itest/RulesPostFormIT.java
```diff
@@ -0,0 +1,149 @@
...
+
+        webClient
+                .post("/api/v2/rules")
+                .sendForm(
+                        testRule,
+                        ar -> {
+                            if (assertRequestStatus(ar, postResponse)) {
+                                MatcherAssert.assertThat(
+                                        ar.result().statusCode(), Matchers.equalTo(201));
+                                postResponse.complete(ar.result().bodyAsJsonObject());
```
andrewazores, line 99: Similar to the previous comment - I would complete the future with the statusCode, and then perform the assertion on that value a few lines down where you do `postResponse.get()`. The `bodyAsJsonObject` value that completes the Future here doesn't seem to be used.

src/test/java/itest/RulesPostFormIT.java
jan-law: Sorry, didn't see your last comment before I wrote the previous one. Makes sense, I'll get started on refactoring

src/test/java/itest/RulesPostFormIT.java
```diff
@@ -0,0 +1,149 @@
...
+        postResponse.get();
+
+        webClient
+                .post("/api/v2/rules")
+                .sendForm(
+                        testRule,
+                        ar -> {
+                            if (assertRequestStatus(ar, duplicatePostResponse)) {
+                                MatcherAssert.assertThat(
+                                        ar.result().statusCode(), Matchers.equalTo(409));
```
andrewazores, line 111: I think this works but not for the reason you think it does - check the implementation of `assertRequestStatus`.

src/test/java/itest/RulesPostFormIT.java
```diff
@@ -0,0 +1,173 @@
...
+                        ar -> {
+                            assertRequestStatus(ar, duplicatePostResponse);
+                        });
+
+        ExecutionException ex =
+                Assertions.assertThrows(
+                        ExecutionException.class, () -> duplicatePostResponse.get());
+        MatcherAssert.assertThat(ex.getCause().getMessage(), Matchers.equalTo("Conflict"));
+
+        // clean up rule before running next test
```
andrewazores, line 133: Just to be sure, cleanup like this should be in a `finally` block, to ensure that it's at least attempted even if something goes wrong above.

andrewazores:
Looks good otherwise.

andrewazores (approved):

PR #574, opened 2021-07-14
fix(auto-rules): Check for null json before creating rule

### LINKED ISSUE #570: Posting empty JSON rule returns HTTP 500
Adding a new rule without a JSON object and `Content-Type: application/json` results in an `HTTP 500` error. The error should be handled or thrown as an `HTTP 400` error.

```
$ curl --insecure -X POST -H "Content-Type: application/json" https://0.0.0.0:8181/api/v2/rules/

{"meta":{"type":"text/plain","status":"Internal Server Error"},"data":{"reason":null}}
```

jan-law:
Fixes #570

Cause of the error is from `RulesPostHandler line 130`:
- `gson.fromJson` returns `null` without throwing an exception
- `null` is passed to `addRule()` which throws an `IOException`/`500`
```
public <T> T fromJson(String json, Type typeOfT) throws JsonSyntaxException {
    if (json == null) {
      return null;
    }
    StringReader reader = new StringReader(json);
    T target = (T) fromJson(reader, typeOfT);
    return target;
  }

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/RulesPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/RulesPostHandler.java
@@ -130,7 +130,7 @@ class RulesPostHandler extends AbstractV2RequestHandler<String> {
                     rule = gson.fromJson(params.getBody(), Rule.class);
 
                     if (rule == null) {
-                        throw new IllegalArgumentException("No JSON found");
+                        throw new IllegalArgumentException("POST body was null");
                     }
                 } catch (IllegalArgumentException | JsonSyntaxException e) {
                     throw new ApiException(400, e);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/io/cryostat/net/web/http/api/v2/RulesPostHandler.java [hunk 1]
```diff
@@ -127,6 +127,10 @@ public boolean isOrdered() {
             case JSON:
                 try {
                     rule = gson.fromJson(params.getBody(), Rule.class);
+
+                    if (rule == null) {
+                        throw new IllegalArgumentException("No JSON found");
```
andrewazores, line 132: I think the message here could be a little more precise. Something like `POST body was null`, that indicates exactly to the user/client what their mistake was.

andrewazores (approved):

andrewazores:
@jan-law could you rebase this? I'll re-approve and merge it after.

jan-law:
Thanks for the reminder, rebased now.

andrewazores (approved):

PR #587, opened 2021-07-19
fix(api): set Content-Type header on all responses

### LINKED ISSUE #586: API responses should contain Content-Type headers

andrewazores:
Fixes #586

### DIFF
--- a/src/test/java/io/cryostat/net/web/http/api/v1/TargetsGetHandlerTest.java
+++ b/src/test/java/io/cryostat/net/web/http/api/v1/TargetsGetHandlerTest.java
@@ -52,6 +52,7 @@ import io.cryostat.util.URIUtil;
 
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+import io.vertx.core.http.HttpHeaders;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.ext.web.RoutingContext;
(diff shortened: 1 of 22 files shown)


src/test/java/io/cryostat/net/web/http/api/v1/TargetsGetHandlerTest.java
```diff
@@ -131,5 +135,6 @@ public JMXServiceURL answer(InvocationOnMock args) throws Throwable {
                 gson.fromJson(
                         responseCaptor.getValue(), new TypeToken<List<ServiceRef>>() {}.getType());
         MatcherAssert.assertThat(result, Matchers.equalTo(targets));
+        Mockito.verify(resp).putHeader(HttpHeaders.CONTENT_TYPE, "application/json");
```
jan-law, line 138: I noticed only some tests call `verify(resp).putHeader()`. When is it appropriate to `verify()` that `putHeader()` was called?

andrewazores: For V1 handlers, the behaviour is split between two places - for non-default values it's done by the specific implementation, otherwise it's done by the parent `AbstractAuthenticatedRequestHandler` in a different method. Since V1 tests generally directly invoke the `handleAuthenticated` method the default behaviour provided by the abstract parent class is not executed, so the test doesn't check for it either.

I should add a test that the parent class does do this default behaviour, though.

hareetd (dismissed):

PR #591, opened 2021-07-22
test(recordings): Add itests for target recordings POST handler

jan-law:
Related #474 

Tests the `4xx` errors handled by the `TargetRecordingsPostHandler`.

### DIFF
--- a/src/test/java/itest/RecordingsPostIT.java
+++ b/src/test/java/itest/RecordingsPostIT.java
@@ -39,23 +39,25 @@ package itest;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
 
 import io.vertx.core.MultiMap;
-import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
-import itest.bases.StandardSelfTest;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
-
+import itest.bases.StandardSelfTest;
+import itest.util.Podman;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public class RecordingsPostIT extends StandardSelfTest {
-    static final String TARGET_ID = "localhost";
+
+    static final String jmxServiceUrl =
+            String.format("service:jmx:rmi:///jndi/rmi://%s:9091/jmxrmi", Podman.POD_NAME);
+    static final String jmxServiceUrlEncoded = jmxServiceUrl.replaceAll("/", "%2F");
+    static final String requestUrl =
+            String.format("/api/v1/targets/%s/recordings", jmxServiceUrlEncoded);
     static final String TEST_RECORDING_NAME = "workflow_itest";
-    static final String REQ_URL = String.format("/api/v1/targets/%s/recordings", TARGET_ID);
 
     @Test
     public void testPostRecordingThrowsOnEmptyRecordingName() throws Exception {
@@ -67,11 +69,11 @@ public class RecordingsPostIT extends StandardSelfTest {
         form.add("events", "template=ALL");
 
         webClient
-                .post(REQ_URL)
+                .post(requestUrl)
                 .sendForm(
                         form,
                         ar -> {
-                            if(assertRequestStatus(ar, response)) {
+                            if (assertRequestStatus(ar, response)) {
                                 response.complete(ar.result().bodyAsJsonObject());
                             }
                         });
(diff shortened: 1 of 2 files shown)


jan-law:
@andrewazores I tried following the `// create an in-memory recording` request in `RecordingWorkflowIT.java` to create this first test. Instead of dumping the `future`, I send an empty `recordingName` parameter and expect a `400` error.

I receive an HTTP `500` error instead of `400`. I also get a `500` if I include the username and password for `localhost` in the form. What am I missing?

```
[ERROR] Failures: 
[ERROR]   RecordingsPostIT.testPostRecordingThrowsOnEmptyRecordingName:80 
Expected: <400>
     but: was <500>
```

jan-law:
> If you're getting a 500 then there should be some exception stack trace in the log that you can use to trace what went wrong.

I see a `400` thrown, then the `POST` returns a `500` without including a stack trace. Here's the full log: 
https://gist.github.com/jan-law/3e21991479dc3fd80317ad7d8d1cc875

> As in JMX credentials? Those need to be passed in request headers, not form attributes.

If I try either the smoketest credentials or the localhost user/pass with the following format, I get the same log as above.
```
.putHeader(
                        "X-JMX-Authorization",
                        "Basic " + Base64.getUrlEncoder().encodeToString("someUsername:somePassword".getBytes()))
```

andrewazores:
Hmm. I'm not sure why the logs aren't more informative - that's something to look into as well.

Anyway, I suspect that the `localhost` targetId might be the root cause here. If you're trying to get Cryostat to connect to itself then that should be `localhost:0` (the `:0` is important) at least, but it would be better to use a full `service:jmx:rmi://` service URL instead of a bare host:port pair.

jan-law:
Both 'targetId's give me the same log output, please see below.
> If you're trying to get Cryostat to connect to itself then that should be `localhost:0` 

Code:
```
    static final String TARGET_ID = "localhost:0";
    static final String REQ_URL = String.format("/api/v1/targets/%s/recordings", TARGET_ID);
```
Log output:
```
Jul 22, 2021 2:56:20 PM io.cryostat.core.log.Logger warn
WARNING: Exception thrown
io.vertx.ext.web.handler.impl.HttpStatusException: Bad Request

Jul 22, 2021 2:56:20 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:54158): POST /api/v1/targets/localhost:0/recordings 500 22ms
```
> it would be better to use a full `service:jmx:rmi://` service URL instead of a bare host:port pair.

Code:
```
    static final String jmxServiceUrl =
            String.format("service:jmx:rmi:///jndi/rmi://%s:9091/jmxrmi", Podman.POD_NAME);
    static final String jmxServiceUrlEncoded = jmxServiceUrl.replaceAll("/", "%2F");
    static final String REQ_URL = String.format("/api/v1/targets/%s/recordings", jmxServiceUrlEncoded);
```
Log output:
```
Jul 22, 2021 3:01:03 PM io.cryostat.core.log.Logger warn
WARNING: Exception thrown
io.vertx.ext.web.handler.impl.HttpStatusException: Bad Request

Jul 22, 2021 3:01:03 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:54214): POST /api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat-itests:9091%2Fjmxrmi/recordings 500 34ms
```

andrewazores:
Traced down the problem, I'll file a separate bug and fix for it.

jan-law:
> Traced down the problem, I'll file a separate bug and fix for it.

Thank you!

src/test/java/itest/RecordingsPostIT.java
```diff
@@ -0,0 +1,86 @@
...
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class RecordingsPostIT extends StandardSelfTest {
+
+    static final String jmxServiceUrl =
+            String.format("service:jmx:rmi:///jndi/rmi://%s:9091/jmxrmi", Podman.POD_NAME);
+    static final String jmxServiceUrlEncoded = jmxServiceUrl.replaceAll("/", "%2F");
```
andrewazores, line 57: I think you can swap in `SELF_REFERENCE_TARGET_ID` from the `StandardSelfTest` base class here.

src/test/java/itest/TargetRecordingsPostIT.java
```diff
@@ -0,0 +1,183 @@
...
+                        });
+        ExecutionException ex =
+                Assertions.assertThrows(ExecutionException.class, () -> response.get());
+        MatcherAssert.assertThat(
+                ((HttpStatusException) ex.getCause()).getStatusCode(), Matchers.equalTo(400));
+        MatcherAssert.assertThat(ex.getCause().getMessage(), Matchers.equalTo("Bad Request"));
+    }
+
+    @Test
+    public void testPostRecordingThrowsOnEmptyEvents() throws Exception {
```
andrewazores, line 118: This test and the one before are both named "empty", but one explicitly adds an empty form attribute and the other simply neglects to add it at all. Should these both do the same thing? And/or should both of these cases be tested?

jan-law: Good point. I can test both cases for consistency.

andrewazores (approved):

PR #597, opened 2021-07-22
fix(recordings): Throw 404 when deleting non-existent recording

### LINKED ISSUE #596: Deleting a non-existent recording returns HTTP 500
```
> DELETE /api/v1/targets/localhost:0/recordings/nonExistentRecording
...
< HTTP/1.1 500 Internal Server Error
< content-type: text/plain
< content-length: 111
< 
Internal Server Error caused by RecordingNotFoundException: No recording with name "nonExistentRecording" found
```
A `404` error code would be more appropriate.

jan-law:
Fixes #596 

When attempting to delete a non-existent recording, the `TargetRecordingDeleteHandler` will throw an HTTP `404` status code instead of an HTTP `500` error. The error message is the same as before.

```
> DELETE /api/v1/targets/localhost:0/recordings/nonExistentRecording HTTP/1.1
> Host: 0.0.0.0:8181
> User-Agent: curl/7.71.1
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< content-type: text/plain
< content-length: 99
< 
Not Found caused by RecordingNotFoundException: No recording with name "nonExistentRecording" found
```

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
@@ -48,7 +48,7 @@ import io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler;
 import io.cryostat.net.web.http.HttpMimeType;
 import io.cryostat.net.web.http.api.ApiVersion;
 import io.cryostat.recordings.RecordingArchiveHelper;
-
+import io.cryostat.recordings.RecordingNotFoundException;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
@@ -96,6 +96,8 @@ class TargetRecordingDeleteHandler extends AbstractAuthenticatedRequestHandler {
 
         try {
             recordingArchiveHelper.deleteRecording(connectionDescriptor, recordingName);
+        } catch (RecordingNotFoundException e) {
+            throw new HttpStatusException(404, e);
         } catch (Exception e) {
             throw new HttpStatusException(500, e);
         }
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
@@ -49,6 +49,7 @@ import io.cryostat.net.web.http.HttpMimeType;
 import io.cryostat.net.web.http.api.ApiVersion;
 import io.cryostat.recordings.RecordingArchiveHelper;
 import io.cryostat.recordings.RecordingNotFoundException;
+
 import io.vertx.core.http.HttpMethod;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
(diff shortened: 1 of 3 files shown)


jan-law:
I'll add an integration test in a separate PR.

andrewazores:
Could you update the unit test as well?

jan-law:
> Could you update the unit test as well?

I'd like to write something like this except the `RecordingNotFoundException()` constructor is private:
```       
Mockito.doThrow(new RecordingNotFoundException()).when(recordingArchiveHelper)
   .deleteRecording(Mockito.any(), Mockito.anyString());
```
What's the advantage of having a private custom exception here instead of a public exception such as a [NotFoundException](https://docs.oracle.com/javaee/7/api/javax/ws/rs/NotFoundException.html)? 

I also found some answers on [stack overflow](https://stackoverflow.com/questions/6901723/should-exceptions-always-be-public) and am not sure which of their scenarios apply here.

andrewazores:
You can make the RecordingNotFoundException public.

`NotFoundException` as you've linked wasn't originally used because although the name seems applicable, it isn't quite an expected exception type. It's in the Jax-RS package, unlike any other exception we use internally, and we aren't really using Jax-RS either. So, although this would mostly be handled internally and hidden away from end clients, it would still be a bit messy to use it IMO.

andrewazores:
Also take note that those StackOverflow answers are talking about the exception type itself being public, which our `RecordingNotFoundException` is - that's not necessarily the same thing as making the constructor public. In this case I don't really see any harm, though.

andrewazores (approved):

andrewazores:
I think this was probably a regression due to the RecordingHelper refactors, but could you check if this bug applies to v1 as well?

jan-law:
> I think this was probably a regression due to the RecordingHelper refactors, but could you check if this bug applies to v1 as well?

v1 looks okay:
```
~/cryostat (v1) [1]> git pull upstream v1
~/cryostat (v1)> mvn clean package && sh smoketest.sh
~/cryostat (v1)> curl -v --insecure -X DELETE https://0.0.0.0:8181/api/v1/targets/localhost:0/recordings/nonExistentRecording
...
> DELETE /api/v1/targets/localhost:0/recordings/nonExistentRecording HTTP/1.1
...
< HTTP/1.1 404 No recording with name "nonExistentRecording" found
< content-type: text/plain
< content-length: 51
< 
No recording with name "nonExistentRecording" found
```

andrewazores:
Could you backport the unit test addition to v1?

jan-law:
> Could you backport the unit test addition to v1?

The `v1` [TargetRecordingDeleteHandlerTest](https://github.com/cryostatio/cryostat/blob/v1/src/test/java/io/cryostat/net/web/handlers/TargetRecordingDeleteHandlerTest.java) has a test `shouldHandleRecordingNotFound()`. 

After refactoring the `PeriodicArchiver`, I moved this test to [RecordingArchiveHelperTest](https://github.com/cryostatio/cryostat/blob/main/src/test/java/io/cryostat/recordings/RecordingArchiveHelperTest.java) line 484 because the new helper function `deleteRecording()`contained the logic that the `v1` `shouldHandleRecordingNotFound()` was testing.

The unit test I added in this PR checks that the `targetRecordingDeleteHandler` catches the `RecordingNotFoundException` from `recordingArchiveHelper.deleteRecording()`. In `v1`, is the code is already covered by the unit tests because the logic in the handler throws an HTTP `404` without throwing an intermediate `RecordingNotFoundException` ?

andrewazores:
Makes sense, looks like it already was well covered before the refactor/in v1.

PR #599, opened 2021-07-23
fix(permissions): API permissions model using client token

### LINKED ISSUE #554: Use OAuth token for Kubernetes client
The fabric8 Kubernetes client will use the mounted service account by default. Currently when deployed by the operator, the Cryostat pod uses the same service account used by the operator.

It would be more secure to use the provided OAuth token given by the user when authenticating. Then we know that the user will only be able to access pods that they have permission from Kubernetes to access.

I think we would also then be able to use the default service account for the Cryostat deployment when created by the operator.

### LINKED ISSUE #575: Add kubernetes-server-mock test dependency
The unit tests for the `KubernetesApiPlatformClient` are lacking, mostly because it's currently quite a pain to set up all the mocks. There is a mock Kubernetes server that could help out a lot: https://itnext.io/mock-kubernetes-api-server-in-java-using-fabric8-kubernetes-mock-server-81a75cf6c47c

I think this might be best left for a follow-up PR.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/492#issuecomment-881006771_

andrewazores:
Fixes #554
Fixes #575

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -201,6 +201,18 @@
     <version>${com.github.spotbugs.version}</version>
     <scope>provided</scope>
   </dependency>
+  <dependency>
+    <groupId>io.fabric8</groupId>
+    <artifactId>kubernetes-server-mock</artifactId>
+    <version>${io.fabric8.client.version}</version>
+    <scope>test</scope>
+  </dependency>
+  <dependency>
+    <groupId>io.fabric8</groupId>
+    <artifactId>openshift-server-mock</artifactId>
+    <version>${io.fabric8.client.version}</version>
+    <scope>test</scope>
+  </dependency>
 </dependencies>
 
 <build>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <org.apache.commons.validator.version>1.7</org.apache.commons.validator.version>
   <org.apache.commons.io.version>2.8.0</org.apache.commons.io.version>
   <org.apache.httpcomponents.version>4.5.13</org.apache.httpcomponents.version>
-  <io.fabric8.client.version>5.6.0</io.fabric8.client.version>
+  <io.fabric8.client.version>5.4.1</io.fabric8.client.version>
   <io.vertx.web.version>3.9.7</io.vertx.web.version>
   <org.slf4j.version>1.7.30</org.slf4j.version>
   <com.google.code.gson.version>2.8.6</com.google.code.gson.version>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <org.apache.commons.validator.version>1.7</org.apache.commons.validator.version>
   <org.apache.commons.io.version>2.8.0</org.apache.commons.io.version>
   <org.apache.httpcomponents.version>4.5.13</org.apache.httpcomponents.version>
-  <io.fabric8.client.version>5.3.0</io.fabric8.client.version>
+  <io.fabric8.client.version>5.6.0</io.fabric8.client.version>
   <io.vertx.web.version>3.9.7</io.vertx.web.version>
   <org.slf4j.version>1.7.30</org.slf4j.version>
   <com.google.code.gson.version>2.8.6</com.google.code.gson.version>
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -61,6 +61,8 @@ import io.cryostat.net.security.ResourceType;
 import io.cryostat.net.security.ResourceVerb;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.fabric8.kubernetes.api.model.authentication.TokenReview;
+import io.fabric8.kubernetes.api.model.authentication.TokenReviewBuilder;
 import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReview;
 import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReviewBuilder;
 import io.fabric8.kubernetes.client.Config;
@@ -99,7 +101,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
             return CompletableFuture.completedFuture(false);
         }
         if (resourceActions.isEmpty()) {
-            return CompletableFuture.completedFuture(true);
+            return reviewToken(token);
         }
 
         try (OpenShiftClient client = clientProvider.apply(token)) {
@@ -118,6 +120,22 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
             // was thrown on allOf().get() above
             return CompletableFuture.completedFuture(true);
         } catch (KubernetesClientException | ExecutionException e) {
+            logger.info(e);
+            return CompletableFuture.failedFuture(e);
+        } catch (Exception e) {
+            logger.error(e);
+            return CompletableFuture.failedFuture(e);
+        }
+    }
+
+    Future<Boolean> reviewToken(String token) {
+        try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
+            TokenReview review =
+                    new TokenReviewBuilder().withNewSpec().withToken(token).endSpec().build();
+            review = client.tokenReviews().create(review);
+            Boolean authenticated = review.getStatus().getAuthenticated();
+            return CompletableFuture.completedFuture(authenticated != null && authenticated);
+        } catch (KubernetesClientException e) {
             logger.info(e);
             return CompletableFuture.failedFuture(e);
         } catch (Exception e) {
@@ -225,6 +243,17 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
             justification = "Kubernetes namespace file path is well-known and absolute")
     private String getNamespace() throws IOException {
         return fs.readFile(Paths.get(Config.KUBERNETES_NAMESPACE_PATH))
+                .lines()
+                .filter(StringUtils::isNotBlank)
+                .findFirst()
+                .get();
+    }
+
+    @SuppressFBWarnings(
+            value = "DMI_HARDCODED_ABSOLUTE_FILENAME",
+            justification = "Kubernetes serviceaccount file path is well-known and absolute")
+    private String getServiceAccountToken() throws IOException {
+        return fs.readFile(Paths.get(Config.KUBERNETES_SERVICE_ACCOUNT_TOKEN_PATH))
                 .lines()
                 .filter(StringUtils::isNotBlank)
                 .findFirst()
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -47,40 +47,42 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-import org.apache.commons.lang3.StringUtils;
-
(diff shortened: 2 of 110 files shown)


andrewazores:
@ebaron ready for review

pom.xml
```diff
@@ -67,7 +67,7 @@
   <org.apache.commons.validator.version>1.7</org.apache.commons.validator.version>
   <org.apache.commons.io.version>2.8.0</org.apache.commons.io.version>
   <org.apache.httpcomponents.version>4.5.13</org.apache.httpcomponents.version>
-  <io.fabric8.client.version>5.3.0</io.fabric8.client.version>
+  <io.fabric8.client.version>5.6.0</io.fabric8.client.version>
```
ebaron, line 70: Does 5.4.1 contain the fix you needed?

andrewazores: I think so - I believe this was the issue for the specific bug I encountered: https://github.com/fabric8io/kubernetes-client/issues/2529

That's labelled with the 5.4.0 milestone, so I'll test out 5.4.1 but I expect it should do it. I suppose you ask because that version is already available for downstream?

ebaron: Right. Should make things a bit easier for us.

andrewazores: https://github.com/fabric8io/kubernetes-client/issues/2265

https://github.com/fabric8io/kubernetes-client/issues/2264

are also relevant.

When I change the fabric8 client version in the pom.xml to 5.4.1, the server-mock unit tests I added in the last couple of commits fail. They pass with 5.3.0, 5.3.1, 5.5.0, and 5.6.0 however. I haven't found any bug reports that indicate why this might be however, but because of the discrepancy between versions running the same unit tests I'm suspecting it might be another bug in the resource implementation in the kubernetes-client.

The unit test failure in 5.4.1 looks like this:


```
[ERROR] shouldValidateTokenWithSufficientPermissions  Time elapsed: 2.445 s  <<< ERROR!
java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: http://localhost:37341/apis/authorization.k8s.io/v1/selfsubjectacce
ssreviews. Message: SelfSubjectAccessReview is invalid. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=null, message=Required value: metadata is required, reason=ValueRequired, additi
onalProperties={})], group=null, kind=SelfSubjectAccessReview, name=null, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=SelfSubjectAccessReview is invalid, metadata=null, reason=Invalid, status=Fail
ure, additionalProperties={}).
        at io.cryostat.net.OpenShiftAuthManagerTest.shouldValidateTokenWithSufficientPermissions(OpenShiftAuthManagerTest.java:142)
Caused by: java.util.concurrent.ExecutionException: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: http://localhost:37341/apis/authorization.k8s.io/v1/selfsubjectaccessreviews. Message: SelfSubjec
tAccessReview is invalid. Received status: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=null, message=Required value: metadata is required, reason=ValueRequired, additionalProperties={})], group=nul
l, kind=SelfSubjectAccessReview, name=null, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=SelfSubjectAccessReview is invalid, metadata=null, reason=Invalid, status=Failure, additionalProperties={}).
        at io.cryostat.net.OpenShiftAuthManagerTest.shouldValidateTokenWithSufficientPermissions(OpenShiftAuthManagerTest.java:142)
Caused by: io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: POST at: http://localhost:37341/apis/authorization.k8s.io/v1/selfsubjectaccessreviews. Message: SelfSubjectAccessReview is invalid. Received status
: Status(apiVersion=v1, code=422, details=StatusDetails(causes=[StatusCause(field=null, message=Required value: metadata is required, reason=ValueRequired, additionalProperties={})], group=null, kind=SelfSubjectAccessReview, name=nul
l, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=SelfSubjectAccessReview is invalid, metadata=null, reason=Invalid, status=Failure, additionalProperties={}).
        at io.cryostat.net.OpenShiftAuthManagerTest.shouldValidateTokenWithSufficientPermissions(OpenShiftAuthManagerTest.java:142)
```

This is very similar to the failed results I saw when running with 5.3.1 against an actual CRC cluster. So, I wonder if 5.4.0/5.4.1 fixed the implementation in kubernetes-client but didn't fix the behaviour of the kubernetes-server-mock, or something along those lines...

ebaron: Maybe the SelfSubjectAccessReview needs to be given a name?

andrewazores: Tried that just now, no luck.

ebaron: I figured it out. There is an OkHttp Interceptor added in the DefaultOpenShiftClient to retry requests that fail authorization by finding a token or asking the server for a new one:
https://github.com/fabric8io/kubernetes-client/blob/7e747ca61c0097562622eb53b9f008cc6d5e82fb/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java

There's a special case for RBAC review objects since these can return 200 or 201 even when unauthorized. These are set to always retry. As a consequence, I believe each SelfSubjectAccessReview will be sent twice unfortunately. This request doesn't get resent if there is no username/password or token in the OpenShiftConfig object, which should be the case in our tests, but not in practice.

The difference we're seeing between 5.4.1 and 5.6.0 is due to this PR: https://github.com/fabric8io/kubernetes-client/pull/3158. The test was failing because the OpenShiftClient found a token to use within our ~/.kube/config. If you rename that file, the test will pass.

src/test/java/io/cryostat/net/OpenShiftAuthManagerTest.java
```diff
@@ -0,0 +1,324 @@
...
+            "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
+
+    OpenShiftAuthManager mgr;
+    @Mock FileSystem fs;
+    @Mock Logger logger;
+    OpenShiftClient client;
+    OpenShiftMockServer server;
+    TokenProvider tokenProvider;
+    Gson gson = MainModule.provideGson(logger);
+
```
ebaron, line 98: Explicitly telling the OpenShiftClient not to read ~/.kube/config makes the tests work as expected with 5.4.1.

```suggestion

    @BeforeAll
    static void disableKubeConfig() {
        // FIXME Disable reading ~/.kube/config. Remove once updated to 5.5.0 or newer.
        System.setProperty(Config.KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY, "false");
    }

```

src/test/java/io/cryostat/net/OpenShiftAuthManagerTest.java
```diff
@@ -0,0 +1,324 @@
...
+import io.cryostat.core.sys.FileSystem;
+import io.cryostat.net.OpenShiftAuthManager.PermissionDeniedException;
+import io.cryostat.net.security.ResourceAction;
+import io.cryostat.net.security.ResourceType;
+import io.cryostat.net.security.ResourceVerb;
+
+import com.google.gson.Gson;
+import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReview;
+import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReviewBuilder;
+import io.fabric8.openshift.client.OpenShiftClient;
```
ebaron, line 62: New import for below suggestion:

```suggestion
import io.fabric8.kubernetes.client.Config;
import io.fabric8.openshift.client.OpenShiftClient;
```

src/test/java/io/cryostat/net/OpenShiftAuthManagerTest.java
```diff
@@ -0,0 +1,324 @@
...
+import io.fabric8.openshift.client.server.mock.EnableOpenShiftMockClient;
+import io.fabric8.openshift.client.server.mock.OpenShiftMockServer;
+import io.fabric8.openshift.client.server.mock.OpenShiftMockServerExtension;
+import okhttp3.mockwebserver.RecordedRequest;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
```
ebaron, line 71: New import for below suggestion:

```suggestion
import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.BeforeEach;
```

ebaron (commented):
Looks good! I still have some reservations about tying the permissions model to the operator's Kubernetes client. For example, `create flightrecorders` isn't currently a valid action for a user to take. This also ends up creating a hard dependency on the operator. I think we can revisit and fine-tune the permissions later on though.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -78,65 +92,97 @@ public AuthenticationScheme getScheme() {
...
         String token = tokenProvider.get();
         if (StringUtils.isBlank(token)) {
             return CompletableFuture.completedFuture(false);
         }
-        return CompletableFuture.supplyAsync(
-                        () -> {
-                            AuthRequest evt = new AuthRequest();
-                            evt.begin();
+        if (resourceActions.isEmpty()) {
+            return CompletableFuture.completedFuture(true);
```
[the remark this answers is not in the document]
andrewazores: Good catch, this initially just happened to map to requests like `GET /notifications_url`, but there are now and always was the possibility of other requests that don't necessarily have a ResourceAction but should still require authentication.

If there is no associated resource model, how should the RBAC permissions for the given token be checked? Should this instead check for a permission that should normally always be granted to users - something again like the previous "list routes"?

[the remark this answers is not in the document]
andrewazores:
```diff
@@ -99,7 +100,8 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
             return CompletableFuture.completedFuture(false);
         }
         if (resourceActions.isEmpty()) {
-            return CompletableFuture.completedFuture(true);
+            resourceActions = EnumSet.copyOf(resourceActions);
+            resourceActions.add(ResourceAction.READ_TARGET);
         }
 
         try (OpenShiftClient client = clientProvider.apply(token)) {
```

Or even just this, it's a pretty simple workaround and asserts that the user provided a valid token and has at least a bare-minimum permission that would be expected. If the user doesn't have READ_TARGET then they can't really do anything else with Cryostat, either.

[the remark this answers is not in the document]
ebaron: True. There's also this API: https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/

You pass it the token and it will return if it authenticated and who it belongs to.

andrewazores:
I've added a `tokenreviews` query to check the user's supplied token, using the service account token to perform the check, if there are no specific Kubernetes resource permissions required for a given authenticated API request. This means the operator clusterrole (or whatever serviceaccount Cryostat runs as) needs `create` on `tokenreviews` in group `authentication.k8s.io`.

ebaron (approved):
Looks good. I tested it locally with the corresponding changes in the operator. I attempted to log in with a fake token to see what would happen and there was no visual indicator in the UI, but there was a `401 Unauthorized` in the log. I'm not sure if this differs at all from the current behaviour.

The operator changes are pretty much done, just need to add some more tests. I think these PRs should be merged at the same time to prevent breakage.

Depends on https://github.com/cryostatio/cryostat-operator/issues/221

andrewazores:
> I attempted to log in with a fake token to see what would happen and there was no visual indicator in the UI, but there was a 401 Unauthorized in the log. I'm not sure if this differs at all from the current behaviour.

I believe that's normal, although now that you mention it, there probably should be some indication that the request was made and rejected.

PR #602, opened 2021-07-26
fix(templates): Throw 400 when posting unexpected file upload name

### LINKED ISSUE #601: Posting an invalid XML file as a template results in HTTP 200
If I post an XML file that contains some text that isn't XML, `TemplatesPostHandler` will return HTTP 200. 

The `LocalStorageTemplateService` does not add the invalid file to storage.

See comment [here](https://github.com/cryostatio/cryostat/issues/474#issuecomment-885929234) for more info.

jan-law:
Fixes #601 

When posting a file upload with a form attribute that doesn't match the name `template`, the `TemplatesPostHandler` will return an HTTP `400`.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TemplatesPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TemplatesPostHandler.java
@@ -110,7 +110,10 @@ class TemplatesPostHandler extends AbstractAuthenticatedRequestHandler {
                 Path path = fs.pathOf(u.uploadedFileName());
                 try (InputStream is = fs.newInputStream(path)) {
                     if (!"template".equals(u.name())) {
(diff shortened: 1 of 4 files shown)


jan-law:
The unit tests `shouldSendNotifcationOnTemplateDeletion` and `shouldProcessGoodRequest` are failing because the requests include "unused" file uploads. My patch is stricter than the original implementation and doesn't allow extra form attributes in the request. Can I safely remove the following lines from those tests?

```
        FileUpload upload2 = Mockito.mock(FileUpload.class);
        Mockito.when(upload2.name()).thenReturn("unused");
        Mockito.when(upload2.uploadedFileName()).thenReturn("/file-uploads/wxyz-9999");
```

andrewazores:
That seems fine.

andrewazores (dismissed):

andrewazores (approved):

PR #605, opened 2021-07-26
fix(certificates): Empty cert uploads throw 400 instead of 500

### LINKED ISSUE #603: Posting empty file as SSL certificate returns HTTP 500
Related #474 
>If you can find any case where you can force a 500 to occur in an integration test then I would say that should actually be filed as a bug

This should return an HTTP `400` instead of `500`.
```
$ curl -F cert=@empty.cer https://0.0.0.0:8181/api/v2/certificates/
{"meta":{"type":"text/plain","status":"Internal Server Error"},"data":{"reason":"Internal Server Error"}}
```
The cryostat logs show that an Exception was thrown and it was not of type `IOException` or `CertificateEncodingException`. The `rootCauseMessage()` is also `Internal Server Error`.
```
SEVERE: Exception thrown
Caused by: io.cryostat.net.web.http.api.v2.ApiException: Internal Server Error
```

jan-law:
Fixes #603

Source of bug: Inside the `try-with-resources` block, the handler throws an `ApiException(500, "No certificates found")` when it finds an empty file. If I replace the `500` with `400`, I still receive a `500` because the `try-with-resources` block catches all `Exception`s and re-throws them as `500`.

This fix throws a `FileNotFoundException` instead of an `ApiException`, which gets caught and rethrown as an `ApiException` with status code `400`.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/CertificatePostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/CertificatePostHandler.java
@@ -38,6 +38,7 @@
 package io.cryostat.net.web.http.api.v2;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -154,7 +155,7 @@ class CertificatePostHandler extends AbstractV2RequestHandler<Path> {
             Collection<? extends Certificate> certificates = certValidator.parseCertificates(fis);
 
             if (certificates.isEmpty()) {
-                throw new ApiException(500, "No certificates found");
+                throw new FileNotFoundException("No certificates found");
             }
 
             try (FileOutputStream out = outputStreamFunction.apply(filePath.toFile())) {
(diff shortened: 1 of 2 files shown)


jan-law:
I noticed the following catch blocks all throw `ApiException(500, e.getMessage(), e);`. What is the purpose behind catching `IOException` and `CertificateEncodingException` first before catching everything else in `Exception`?
```
catch (IOException ioe) {
            throw new ApiException(500, ioe.getMessage(), ioe);
        } catch (CertificateEncodingException cee) {
            throw new ApiException(500, cee.getMessage(), cee);
        } catch (Exception e) {
            throw new ApiException(500, e.getMessage(), e);
        }
```

andrewazores:
> I noticed the following catch blocks all throw `ApiException(500, e.getMessage(), e);`. What is the purpose behind catching `IOException` and `CertificateEncodingException` first before catching everything else in `Exception`?
> 
> ```
> catch (IOException ioe) {
>             throw new ApiException(500, ioe.getMessage(), ioe);
>         } catch (CertificateEncodingException cee) {
>             throw new ApiException(500, cee.getMessage(), cee);
>         } catch (Exception e) {
>             throw new ApiException(500, e.getMessage(), e);
>         }
> ```

In this case it seems pretty redundant. It would be effectively the same to just have the `catch Exception` block.

andrewazores (approved):

PR #608, opened 2021-07-27
test(certificates): Add SSL certificate itests

jan-law:
Related #474 

Tests for null file uploads and conflicting file names that already exist in the truststore.

Question: If the `/truststore` directory is not mounted for integration tests, how can I test the following in `CertificatePostHandler` line 151?
```
if (fs.exists(filePath)) {
            throw new ApiException(409, filePath.toString() + " Certificate already exists");
        }
```


andrewazores:
`/truststore` doesn't need to be mounted necessarily, like it's done in `run.sh` - this would allow the contents to be written to the host filesystem, but that's more than what's really needed for itest purposes. As long as the Cryostat instance under test is able to store the certificate for its single-run lifetime then that would be sufficient. So, a tmpfs mount would fit the bill. In `pom.xml` we do already mount tmpfs to some directories, ex:

```xml
              <argument>--mount</argument>
              <argument>type=tmpfs,target=/opt/cryostat.d/recordings.d</argument>
```

Adding another itest setup argument like this to mount the certificates storage directory (and possibly to set an environment variable telling Cryostat where that directory is within its container at runtime) should do the trick.

jan-law:
I've finished making changes, ready for review

andrewazores (dismissed):

andrewazores:
Rebase please - I don't expect any conflicts between this commit and the last, but just to be sure.

PR #615, opened 2021-07-29
test(report): Add report itests

jan-law:
Related #474 

Tests the HTTP `200` and `404` scenarios for the `TargetReportGetHandler` and `ReportGetHandler`. 

`TargetReportIT` downloads the report directly from the JVM whereas `ReportIT` also makes one more call to save the recording to archive before getting the report from the archive.

Update: All tests are passing
~~Summary of test failures:~~
~~1) `TargetReportGetHandler` throws `500` instead of `404` when getting a non-existent recording (should pass after #614 is merged)~~
~~2) When the Jsoup HTML parser parses the report, it returns a list of `Unexpected Token` errors. See review below~~

### DIFF
--- a/src/test/java/itest/ReportIT.java
+++ b/src/test/java/itest/ReportIT.java
@@ -127,17 +127,29 @@ public class ReportIT extends StandardSelfTest {
 
         } finally {
             // Clean up recording
-            CompletableFuture<JsonObject> deleteResponse = new CompletableFuture<>();
+            CompletableFuture<JsonObject> deleteRecResponse = new CompletableFuture<>();
+            webClient
+                    .delete(String.format("%s/%s", RECORDING_REQ_URL, TEST_RECORDING_NAME))
+                    .send(
+                            ar -> {
+                                if (assertRequestStatus(ar, deleteRecResponse)) {
+                                    deleteRecResponse.complete(ar.result().bodyAsJsonObject());
+                                }
+                            });
+
+            MatcherAssert.assertThat(deleteRecResponse.get(), Matchers.equalTo(null));
+
+            CompletableFuture<JsonObject> deleteArchivedRecResp = new CompletableFuture<>();
             webClient
                     .delete(String.format("/api/v1/recordings/%s", savedRecordingName.get()))
                     .send(
                             ar -> {
-                                if (assertRequestStatus(ar, deleteResponse)) {
-                                    deleteResponse.complete(ar.result().bodyAsJsonObject());
+                                if (assertRequestStatus(ar, deleteArchivedRecResp)) {
+                                    deleteArchivedRecResp.complete(ar.result().bodyAsJsonObject());
                                 }
                             });
 
-            MatcherAssert.assertThat(deleteResponse.get(), Matchers.equalTo(null));
+            MatcherAssert.assertThat(deleteArchivedRecResp.get(), Matchers.equalTo(null));
         }
     }
 
--- a/src/test/java/itest/ReportIT.java
+++ b/src/test/java/itest/ReportIT.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


src/test/java/itest/ReportIT.java
```diff
@@ -0,0 +1,160 @@
...
+                                            ar.result().statusCode(), Matchers.equalTo(200));
+                                    MatcherAssert.assertThat(
+                                            ar.result()
+                                                    .getHeader(HttpHeaders.CONTENT_TYPE.toString()),
+                                            Matchers.equalTo(HttpMimeType.HTML.mime()));
+                                    getResponse.complete(ar.result().bodyAsBuffer());
+                                }
+                            });
+
+            getResponse.get();
```
jan-law, line 126: Since I'm only checking the content-type header, what validations for the response body would be appropriate here?

jan-law: I found this `TODO` in `RecordingWorkflowIT`. I can parse the response body in both report itest files and `RecordingWorkflowIT` to verify the response is valid HTML.
```
// verify that reports can be downloaded successfully and yield non-empty HTML documents
            // (TODO: verify response body is a valid HTML document)
            String reportUrl = recordingInfo.getString("reportUrl");
            Path reportPath =
                    downloadFileAbs(reportUrl, TEST_RECORDING_NAME + "_report", ".html")
                            .get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
            MatcherAssert.assertThat(reportPath.toFile().length(), Matchers.greaterThan(0L));
```

src/test/java/itest/ReportIT.java
```diff
@@ -0,0 +1,178 @@
...
+                                            Matchers.equalTo(HttpMimeType.HTML.mime()));
+                                    getResponse.complete(ar.result().bodyAsString());
+                                }
+                            });
+
+            int MAX_ERRORS_TRACKED = 100;
+            Parser parser = Parser.htmlParser().setTrackErrors(MAX_ERRORS_TRACKED);
+            Jsoup.parse(getResponse.get(), "", parser);
+            System.out.println(parser.getErrors());
+            MatcherAssert.assertThat(parser.getErrors().isEmpty(), Matchers.equalTo(true));
```
jan-law, line 132: `parser.getErrors()` should return an empty list if the report is a valid HTML document. I receive a list of errors instead. If I try to curl the same request, the HTML report renders correctly in my browser:
https://gist.github.com/jan-law/458bd4208344ba6cbe968d28d20df4ed

 I'm not sure if I'm passing the report response into the parser correctly, or even if `Jsoup` is an appropriate parser to use.
```
[INFO] Running itest.ReportIT
HTTP 404: Not Found
[9921: Unexpected token [EndTag] when in state [InBody], 10182: Unexpected token [EndTag] when in state [InBody], 13475: Unexpected token [EndTag] when in state [InBody], 15188: Unexpected token [EndTag] when in state [InBody], 17321: Unexpected token [EndTag] when in state [InBody], 21740: Unexpected token [EndTag] when in state [InBody], 26326: Unexpected token [EndTag] when in state [InBody], 32291: Unexpected token [EndTag] when in state [InBody], 35701: Unexpected token [EndTag] when in state [InBody], 41053: Unexpected token [EndTag] when in state [InBody], 42571: Unexpected token [EndTag] when in state [InBody], 45912: Unexpected token [EndTag] when in state [InBody], 48293: Unexpected token [EndTag] when in state [InBody], 57950: Unexpected token [EndTag] when in state [InBody], 62565: Unexpected token [EndTag] when in state [InBody], 72199: Unexpected token [EndTag] when in state [InBody], 74122: Unexpected token [EndTag] when in state [InBody], 87230: Unexpected token [EndTag] when in state [InBody], 90204: Unexpected token [EndTag] when in state [InBody], 108079: Unexpected token [EndTag] when in state [InBody], 115641: Unexpected token [EndTag] when in state [InBody]]
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 6.045 s <<< FAILURE! - in itest.ReportIT
[ERROR] testGetReportShouldSendFile  Time elapsed: 4.181 s  <<< FAILURE!
java.lang.AssertionError: 

Expected: <true>
     but: was <false>
	at itest.ReportIT.testGetReportShouldSendFile(ReportIT.java:134)
```

src/test/java/itest/ReportIT.java
andrewazores: > Since I'm only checking the content-type header, what validations for the response body would be appropriate here?

You could just complete the Future with the response header(s) and perform the assertions after/on the `getResponse.get()`, if the purpose of this test is just to check the Content-Type.

It might make sense to use JSoup to do other assertions about the response however, beyond just checking the Content-Type header. It doesn't need to be and probably shouldn't be too in-depth, since the exact format of the rules analysis document can change between JMC versions, but some basic checks that the received content really is a parseable HTML document with some general overall structure would be fine. Very similar to what the ClientAssetsIT does.

src/test/java/itest/ReportIT.java
```diff
@@ -0,0 +1,160 @@
...
+            webClient
+                    .delete(String.format("/api/v1/recordings/%s", savedRecordingName.get()))
+                    .send(
+                            ar -> {
+                                if (assertRequestStatus(ar, deleteResponse)) {
+                                    deleteResponse.complete(ar.result().bodyAsJsonObject());
+                                }
+                            });
+
+            MatcherAssert.assertThat(deleteResponse.get(), Matchers.equalTo(null));
```
andrewazores: I think this cleanup should delete the active and archived recordings as well, right?

src/test/java/itest/ReportIT.java
andrewazores: It does look like you're using the correct JSoup HTML parser . What is the value of `getResponse.get()` at that point? Since it looks like there was a `404` first then the response body probably was not an HTML document, so the parser would choke on whatever it was instead - maybe just the text `Not Found`, for example. The parser errors output looks odd for that case since it seems to imply that it did parse a `<body>` or something - not sure.

jan-law: Printing `getResponse.get()` gives me an HTML file similar to the gist above. If I run the tests with `@TestMethodOrder`, the `404` appears after the parse error, which makes me think the `404` is from `testGetReportThrowsWithNonExistentRecordingName()`
```
[INFO] Running itest.ReportIT
<!--
   Copyright (c) 2018, Oracle and/or its affiliates. All rights reserved.
   Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
-->
<html lang="en">
<head>
<base target="_blank">
<meta http-equiv="X-UA-Compatible" content="IE=edge" />
<title>Automated Analysis Result Overview</title>
...
[9921: Unexpected token [EndTag] when in state [InBody], 
...
HTTP 404: Not Found
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.052 s
```

andrewazores: Strange. What if you get rid of the `setTrackErrors` stuff and just use the plain `Jsoup.parse(file, "UTF-8")` like ClientAssetsIT does? Presumably, if the parser fails to parse the document, then the other test assertions would fail later on - things like simply checking that the parsed Document has a `<head>` and a `<body>`, again like the ClientAssetsIT does.

src/test/java/itest/ReportIT.java
jan-law: Both recordings should be cleaned up now

src/test/java/itest/ReportIT.java
jan-law: Thanks, calling `Jsoup.parse(file, "UTF-8");` works well. I wrote the response into a temporary file in `test/resources` and deleted the file during cleanup.

andrewazores (approved):

PR #616, opened 2021-07-30
test(snapshot): Add POST snapshot itests

jan-law:
Related #474 

Tests a successful snapshot `POST` and a `404` scenario for both `v1` and `v2` versions of `TargetSnapshotPostHandler`.

### DIFF
--- a/src/test/java/itest/SnapshotIT.java
+++ b/src/test/java/itest/SnapshotIT.java
@@ -39,26 +39,48 @@ package itest;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 
 import io.cryostat.net.web.http.HttpMimeType;
 
 import io.vertx.core.MultiMap;
 import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
 import itest.bases.StandardSelfTest;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public class SnapshotIT extends StandardSelfTest {
     static final String TEST_RECORDING_NAME = "someRecording";
-    static final String TARGET_REQ_URL =
-            String.format("/api/v1/targets/%s", SELF_REFERENCE_TARGET_ID);
+    static final String V1_TARGET_REQ_URL =
+            String.format("/api/v1/targets%s", SELF_REFERENCE_TARGET_ID);
+    static final String V2_TARGET_REQ_URL =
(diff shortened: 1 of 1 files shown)


src/test/java/itest/SnapshotIT.java
```diff
@@ -0,0 +1,305 @@
...
+                    snapshotResponse.get(), Matchers.containsString(expectedDownloadUrl));
+            MatcherAssert.assertThat(
+                    snapshotResponse.get(), Matchers.containsString(expectedReportUrl));
+            MatcherAssert.assertThat(
+                    snapshotResponse.get(), Matchers.containsString(expectedRecordingId));
+            MatcherAssert.assertThat(
+                    snapshotResponse.get(), Matchers.containsString(expectedRecordingState));
+            MatcherAssert.assertThat(snapshotResponse.get(), Matchers.containsString("startTime"));
+            MatcherAssert.assertThat(
+                    snapshotResponse.get(), Matchers.containsString(expectedRecordingOptions));
```
jan-law, line 254: Any ideas to make this section (and the `final String`s above) easier to read?

src/test/java/itest/SnapshotIT.java
```diff
@@ -0,0 +1,299 @@
...
+import org.junit.jupiter.api.Test;
+
+public class SnapshotIT extends StandardSelfTest {
+    static final String TEST_RECORDING_NAME = "someRecording";
+    static final String V1_TARGET_REQ_URL =
+            String.format("/api/v1/targets%s", SELF_REFERENCE_TARGET_ID);
+    static final String V2_TARGET_REQ_URL =
+            String.format("/api/v2/targets%s", SELF_REFERENCE_TARGET_ID);
+    static final String SNAPSHOT_NAME_V1 = "snapshot-2";
+    static final String SNAPSHOT_NAME_V2 = "snapshot-4";
```
jan-law, line 65: When running all the integration tests with `mvn verify`, I can't guarantee that my snapshots will be the second and fourth recordings to be created. How else can I verify the snapshot names when I receive the response body?

jan-law: Or would it be enough to check that the recording name contains the string `snapshot` without checking the id?

andrewazores: The success response from creating a snapshot should tell you what the name of the new snapshot is, so you could retain that value from the response and use it for cleanup or whatever else you need to do with it.

andrewazores: As you point out, the actual name of that snapshot isn't precisely determinable in advance, so simply checking that the name fits an expected format (using a regex?) seems sufficient.

src/test/java/itest/SnapshotIT.java
andrewazores: Since what you're really trying to do is verify the structure of a JSON object response, you can remodel that as a Java Map instead of a series of Strings. Example here: https://github.com/cryostatio/cryostat/blob/443928f21124cfba2b94cb1ed0d045834f81c3d6/src/test/java/itest/AutoRulesIT.java#L239

andrewazores (dismissed):

PR #619, opened 2021-08-04
test(recordings): Verify file contents from RecordingWorkflowIT

jan-law:
Related #474 

Addresses the `TODO`s in `RecordingWorkflowIT`:
```
// verify that reports can be downloaded successfully and yield non-empty HTML documents
// (TODO: verify response body is a valid HTML document)
```
```
// verify in-memory and saved recordings can be downloaded successfully and yield
// non-empty recording binaries (TODO: better verification of file content), and that
// the fully completed in-memory recording is larger than the saved partial copy
```

### DIFF
--- a/src/test/java/itest/RecordingWorkflowIT.java
+++ b/src/test/java/itest/RecordingWorkflowIT.java
@@ -39,6 +39,7 @@ package itest;
 
 import java.io.File;
 import java.nio.file.Path;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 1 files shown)


jan-law:
What kinds of verification would be useful for the jfr files?

andrewazores:
> What kinds of verification would be useful for the jfr files?

JMC has a JFR file parser that we can probably use here. I think checking that the JFR binary is parseable and that the resulting list of events is non-empty is probably enough.

If you target Cryostat itself and use the Cryostat event template then I think you can guarantee that the recording should contain some events.

jan-law:
> if there's already a JFR parser available then may as well use that to parse both and check that the parsed list of events is longer

I found a JFR parser from [jdk.jfr.consumer.RecordingFile](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/consumer/RecordingFile.html#readAllEvents(java.nio.file.Path)) that returns a list of events as I couldn't find the parser in -core.

andrewazores:
> > if there's already a JFR parser available then may as well use that to parse both and check that the parsed list of events is longer
> 
> I found a JFR parser from [jdk.jfr.consumer.RecordingFile](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/consumer/RecordingFile.html#readAllEvents(java.nio.file.Path)) that returns a list of events as I couldn't find the parser in -core.

Nice, that's even better.

andrewazores (approved):

PR #624, opened 2021-08-04
test(recordings): itest for loading a recording to jfr-datasource

jan-law:
Related #474 

Addresses the `TODO` in `UploadRecordingIT`:
```
TODO, this test needs to be updated to actually trigger a file to be uploaded to
jfr-datasource/grafana-dashboard now that the integration tests have these additional
containers configured
```

### DIFF
--- a/src/test/java/itest/UploadRecordingIT.java
+++ b/src/test/java/itest/UploadRecordingIT.java
@@ -67,7 +67,7 @@ public class UploadRecordingIT extends StandardSelfTest {
 
     @BeforeAll
     public static void createRecording() throws Exception {
-        CompletableFuture<JsonObject> dumpRespFuture = new CompletableFuture<>();
+        CompletableFuture<JsonObject> dumpPostResponse = new CompletableFuture<>();
         MultiMap form = MultiMap.caseInsensitiveMultiMap();
         form.add("recordingName", RECORDING_NAME);
         form.add("duration", "5");
@@ -77,11 +77,13 @@ public class UploadRecordingIT extends StandardSelfTest {
                 .sendForm(
                         form,
                         ar -> {
-                            if (assertRequestStatus(ar, dumpRespFuture)) {
-                                dumpRespFuture.complete(ar.result().bodyAsJsonObject());
+                            if (assertRequestStatus(ar, dumpPostResponse)) {
+                                MatcherAssert.assertThat(
+                                        ar.result().statusCode(), Matchers.equalTo(201));
+                                dumpPostResponse.complete(ar.result().bodyAsJsonObject());
                             }
                         });
-        dumpRespFuture.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        dumpPostResponse.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
     }
 
     @AfterAll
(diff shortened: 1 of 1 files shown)


jan-law:
Am I missing any additional cleanup or validations after the recording gets uploaded to Grafana?

andrewazores:
There is no additional cleanup to do. You could add some tests querying the datasource directly: https://github.com/cryostatio/jfr-datasource/#readme 

ex. a `GET /list` to verify that the uploaded file is available for selection, and a `POST /query` to get some converted JSON data.

src/test/java/itest/UploadRecordingIT.java
```diff
@@ -110,5 +119,84 @@ public void shouldUploadRecordingToGrafana() throws Exception {
...
+                                        ar.result().statusCode(), Matchers.equalTo(200));
+                                MatcherAssert.assertThat(
+                                        ar.result().getHeader(HttpHeaders.CONTENT_TYPE.toString()),
+                                        Matchers.equalTo(HttpMimeType.JSON.mime()));
+                                getRespFuture.complete(ar.result().bodyAsString());
+                            }
+                        });
+
+        MatcherAssert.assertThat(
+                getRespFuture.get(), Matchers.equalTo(String.format("%s\n", RECORDING_NAME)));
```
andrewazores, line 140: I would prefer to not check for the `\n` character - either do `getRespFuture().get().trim()` to get a "bare" name string, or perhaps `Arrays.asList(getRespFuture().get().split("[\\s]"))` to turn the result into a list of names.

src/test/java/itest/UploadRecordingIT.java
```diff
@@ -92,26 +98,104 @@ public static void deleteRecording() throws Exception {
...
+
+        CompletableFuture<JsonArray> queryRespFuture = new CompletableFuture<>();
+
+        Calendar lastMinute = Calendar.getInstance();
+        lastMinute.add(Calendar.MINUTE, -1);
+        final DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS");
+        final String END = dateFormat.format(Calendar.getInstance().getTime());
+        final String START = dateFormat.format(lastMinute.getTime());
+
+        JsonObject query =
```
jan-law, line 151: I tried writing a query body with the docs I found below. How can I find the correct `panelId` and `targets` values?
This query gives me an empty array.

https://github.com/grafana/simple-json-datasource#query-api
https://oznetnerd.com/2018/04/17/writing-a-grafana-backend-using-the-simple-json-datasource-flask/#query-endpoint-deep-dive

andrewazores: I'm not entirely sure how those queries work, to be honest, but I think you can find some inspiration in the jfr-datasource tests:

https://github.com/cryostatio/jfr-datasource/blob/main/src/test/java/io/cryostat/jfr/datasource/tests/DatasourceTest.java#L106

https://github.com/cryostatio/jfr-datasource/blob/main/src/test/resources/query.timeseries.input.txt

jan-law: I think I have the correct query format because I'm getting a non-empty response. 
```
{"target": "jdk.CPULoad.machineTotal", "datapoints": []}
```
The `datapoints` array is empty, though. Making a couple `GET` requests to the server and using `Thread.sleep(5000)` before uploading the recording to Grafana also doesn't add any datapoints. Any ideas?

jan-law: Interesting, the PR run seems to capture data:
```
Expected: <[{"datapoints":[],"target":"jdk.CPULoad.machineTotal"}]>
     but: was <[{"target":"jdk.CPULoad.machineTotal","datapoints":[[0.9300000071525574,1628278180851],[0.9845361113548279,1628278181823]]}]>
	at itest.UploadRecordingIT.shouldUploadRecordingToGrafana(UploadRecordingIT.java:204)
```
Looking into it.

jan-law:
@hareetd , do you remember what the status of this PR was? I think the issue was that the local integration tests ran too fast to pick up any data points, while the Github CI occasionally found data points when uploading the recording. After that, I'm not sure if you discovered anything else

hareetd:
>  I think the issue was that the local integration tests ran too fast to pick up any data points, while the Github CI occasionally found data points when uploading the recording.

Yes, that's it. 

> After that, I'm not sure if you discovered anything else

I can't remember what exactly, but I tried a few different things to ensure the data points were retrieved, but no luck. I was stuck for a few days but ended up working on something else and forgot about this PR.

tthvo:
Accidentally commit an old sub-module :D Removed it now and rebased on main.

andrewazores:
See #1139 for why mergify just leaves that comment and review there. Maintainers can force-merge it anyway or just dismiss mergify's review so it isn't a big deal, just an annoyance.

andrewazores:
@Mergifyio refresh

andrewazores (changes requested):
Testing something

andrewazores (commented):
Testing something else

andrewazores:
@Mergifyio refresh

tthvo:
I am not sure we can have a fixed expected `datapoints` as each new recording is different. To tackle this, I switched test to check:
- `datapoints` array must be non-null.
- If `datapoints` is not empty, check if each child array has 2 non-null elements. And 2nd element (timestamp) must be >= 0.

src/test/java/itest/UploadRecordingIT.java
tthvo: On this issue again...

A pretty surprising root cause behind is the time frame passed in the query is out of range. However, I have to set the time range to expand forward 1 month to get the data points. Quite odd... 

Still investigating...

tthvo: Actually, 24 hours is enough. This might be due to timezone difference.

tthvo: Nice! It was timezone issue. I adjusted it back to UTC-0 instead of local timezone.

tthvo:
Ready for review :D The old issue where `datapoints` are not picked up was due to timezone mismatch, so adjusting back time frame using UTC-0 worked as expected.

andrewazores (dismissed):

andrewazores (approved):

Issue #263, opened 2020-09-11
target JVMs report a different default "toDisk" value from actual value used in new recordings


Alexjsenn:
The default recording options can be retrieved from a target JVM using `IFlightRecorderService::getAvailableRecordingOptions()`. The map returned specifies the default values for different recording options. While testing (using the cjfr instance itself as a targetJVM) the JVM reports that the default value for `toDisk` is `false`. However, after creating a recording without specifying `toDisk` (which we assume would mean the JVM would fall back on its default), the recording is created with `toDisk=true`. 
Given this inconsistency, the `toDisk` checkbox in the new recordings page of the web-client has been set to default to true, thus requiring the user to specifically disable the option if they wish to.

tthvo:
Hey @andrewazores, is there still something to do with issue? Just saw it mention in code:

https://github.com/cryostatio/cryostat-web/blob/ee99ad168f707e4342444589a466f9abb0ca6599/src/app/CreateRecording/CustomRecordingForm.tsx#L234

PR #626, opened 2021-08-05
test(recordings): Add recording options itests

jan-law:
Related #474 

Tests successful scenarios for the `TargetRecordingOptionsGetHandler`, `TargetRecordingOptionsPatchHandler`, and the `TargetRecordingOptionsListGetHandler`.

### DIFF
--- a/src/test/java/itest/TargetRecordingOptionsIT.java
+++ b/src/test/java/itest/TargetRecordingOptionsIT.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package itest;
+
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


andrewazores:
Tests look good. Do you still observe #263? Either way, I think it would be worth writing some tests to capture that behaviour.

jan-law:
> Do you still observe #263?

Yes, new recordings are created with `toDisk=true`.
Should recordings write to disk by default? If you'd like, I could change the default options to return `toDisk=true`.

andrewazores:
The default should be whatever the target JVM says it is, which according to [docs](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/Recording.html#isToDisk()), seems it should be `true`. #263 seems to reflect a bug in JMC code, then, where the recording options map contains `false` when it would be expected to be `true` or perhaps `null` (unset, as the docs say).

src/test/java/itest/TargetRecordingOptionsIT.java
```diff
@@ -0,0 +1,266 @@
...
+                                        ar.result().statusCode(), Matchers.equalTo(200));
+                                MatcherAssert.assertThat(
+                                        ar.result().getHeader(HttpHeaders.CONTENT_TYPE.toString()),
+                                        Matchers.equalTo(HttpMimeType.JSON.mime()));
+                                getResponse.complete(ar.result().bodyAsJsonObject());
+                            }
+                        });
+
+        JsonObject expectedGetResponse =
+                new JsonObject(Map.of("maxAge", 0, "toDisk", false, "maxSize", 0));
```
andrewazores, line 158: Could you put a `FIXME` here with a link to #263? That's still something we should probably investigate further, although I am fairly convinced that it's just a JMC bug with how those OptionDescriptors are created or how the toDisk setting is read from the JVM.

andrewazores (commented):
Looks good otherwise.

andrewazores (approved):

jan-law:
`RecordingWorkflowIT` failed at the part where it expects the in-memory jfr file to be larger than the saved jfr file. Why would it fail this time but pass most of the time? I'm re-running the tests to see what happens
```
Error:  Failures: 
Error:    RecordingWorkflowIT.testWorkflow:222 
Expected: a value greater than <9919>
but: <8305> was less than <9919>
```

jan-law:
Hmm, now `AutoRulesIT` is failing because I forgot to reset the default recording options. Let me fix that

```
Error:  Failures: 
Error:    AutoRulesIT.testNewContainerHasRuleApplied:285 
Expected: <0>
     but: was <1000>
```

andrewazores:
> `RecordingWorkflowIT` failed at the part where it expects the in-memory jfr file to be larger than the saved jfr file. Why would it fail this time but pass most of the time? I'm re-running the tests to see what happens
> 
> ```
> Error:  Failures: 
> Error:    RecordingWorkflowIT.testWorkflow:222 
> Expected: a value greater than <9919>
> but: <8305> was less than <9919>
> ```

Hmm. It sounds like it could simply be that the in-memory ring buffers filled up and some older event data was dropped in between the two recordings being retrieved, assuming that the recording isn't somehow configured with a maximum event age.

It's not a good fix, but changing the template in that test from `ALL` to just `Continuous` should cut down the number of events recorded and hopefully work around this edge case. Could you file a bug for this failure with a link to the CI log and whatever other findings you may have?

andrewazores:
> Hmm, now `AutoRulesIT` is failing because I forgot to reset the default recording options. Let me fix that
> 
> ```
> Error:  Failures: 
> Error:    AutoRulesIT.testNewContainerHasRuleApplied:285 
> Expected: <0>
>      but: was <1000>
> ```

Aha - that would do it. `maxSize=1000` in your 4th test case would be likely to cause the behaviour I described in the last comment, or at least cause it to occur much earlier/more frequently.

src/test/java/itest/TargetRecordingOptionsIT.java
```diff
@@ -63,6 +64,35 @@
...
+
+        MatcherAssert.assertThat(getResponse.get(), Matchers.equalTo(expectedGetResponse));
```
andrewazores, line 93: I think we can skip the test assertion in this `AfterAll`, and the `BeforeAll` too if you add it.

jan-law: Sounds good

src/test/java/itest/TargetRecordingOptionsIT.java
```diff
@@ -63,6 +64,35 @@
             String.format("/api/v2/targets/%s/recordingOptionsList", SELF_REFERENCE_TARGET_ID);
     static final String RECORDING_NAME = "test_recording";
 
+    @AfterAll
+    static void resetDefaultRecordingOptions() throws Exception {
+        CompletableFuture<JsonObject> getResponse = new CompletableFuture<>();
+        MultiMap form = MultiMap.caseInsensitiveMultiMap();
+        form.add("maxAge", "0");
+        form.add("toDisk", "false");
+        form.add("maxSize", "0");
```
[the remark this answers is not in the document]
andrewazores: Actually - I guess the `toDisk` bug with 263 would bite us here, so at least that one would need to be overridden/hardcoded as `true` again. Nevermind then.

andrewazores (approved):

PR #638, opened 2021-08-10
feat(targetconnectionmanager): cache TTL/size configuration

### LINKED ISSUE #637: Connection cache size and TTL should be configurable
Currently, the TargetConnectionManager caches target JMX connections for 90 seconds, and will cache an unlimited number of simultaneous connections. The TTL and size should be configurable by environment variables.

andrewazores:
Fixes #637


andrewazores:
@hareetd or @jan-law , would either of you like to review this?

src/main/java/io/cryostat/net/NetworkModule.java
```diff
@@ -88,12 +93,27 @@ static NetworkResolver provideNetworkResolver() {
...
+    @Provides
+    @Named(MAX_TARGET_CONNECTIONS)
+    static int provideMaxTargetConnections(Environment env) {
+        return Integer.parseInt(env.getEnv(MAX_TARGET_CONNECTIONS, "-1"));
+    }
+
+    @Provides
+    @Named(MAX_TARGET_TTL)
+    static Duration provideMaxTargetTTL(Environment env) {
+        return Duration.ofSeconds(Integer.parseInt(env.getEnv(MAX_TARGET_TTL, "10")));
```
hareetd, line 105: As it is now, the TTL will either be 5 or 10 seconds depending on whether the environment variable exists or not. 

Is this because the previous TTL of 90 seconds was overkill?

pom.xml
```diff
@@ -354,6 +354,10 @@
               <argument>--mount</argument>
               <argument>type=tmpfs,target=/opt/cryostat.d/truststore.d</argument>
               <argument>--env</argument>
+              <argument>CRYOSTAT_MAX_TARGET_CONNECTIONS=1</argument>
```
hareetd, line 357: I'm guessing the previous default value of unlimited target connections was unnecessary since we typically stick to 1 target when testing etc. anyways?

src/test/java/io/cryostat/net/TargetConnectionManagerTest.java
```diff
@@ -210,4 +210,74 @@ public JFRConnection answer(InvocationOnMock invocation)
...
+                        new Answer<JFRConnection>() {
+                            @Override
+                            public JFRConnection answer(InvocationOnMock invocation)
+                                    throws Throwable {
+                                return Mockito.mock(JFRConnection.class);
+                            }
+                        });
+        ConnectionDescriptor desc = new ConnectionDescriptor("foo");
+        JFRConnection conn1 = mgr.executeConnectedTask(desc, a -> a);
+        Thread.sleep(10);
```
hareetd, line 244: I wasn't sure how a `Caffeine` cache behaves when the maximum size is zeroed out so I looked it up and the docs say the connection will be immediately evicted upon being loaded into the cache. In that case, I could be mistaken but I don't think it's necessary to sleep the thread for the TTL duration because a new connection should be created right away, no?

hareetd (commented):
Looks good to go besides the paragraph I mentioned, the rest of the comments are mainly just questions for my own understanding.

src/main/java/io/cryostat/net/TargetConnectionManager.java
```diff
@@ -68,19 +68,20 @@
     public static final Pattern HOST_PORT_PAIR_PATTERN =
             Pattern.compile("^([^:\\s]+)(?::(\\d{1,5}))?$");
 
-    static final Duration DEFAULT_TTL = Duration.ofSeconds(90);
-
```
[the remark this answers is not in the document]
andrewazores: Fixed, thanks.

src/main/java/io/cryostat/net/TargetConnectionManager.java
```diff
@@ -119,8 +120,11 @@ public void onRemoval(
                                             }
                                         }
                                     }
-                                })
-                        .build(this::connect);
+                                });
+        if (maxTargetConnections >= 0) {
+            cacheBuilder = cacheBuilder.maximumSize(maxTargetConnections);
+        }
```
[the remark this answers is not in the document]
andrewazores: Right, it may potentially grow unbounded if there is no maximum size or maximum total weight configured.

src/main/java/io/cryostat/net/NetworkModule.java
andrewazores: It's set to 5 in integration tests, and it defaults to 10 if the user/deployment doesn't set their own value. I set it to a non-default value in testing mostly just to exercise that a bit and make sure setting a non-default value doesn't blow anything up.

I feel 90 seconds was overkill in retrospect, although this is really just my feeling. It's best not to hold the connections open for too long/indefinitely since there may be arbitrarily many of them, but it's also preferable for performance reasons to not open a new connection for every single request or else the end user will likely experience bad latency and UI responsiveness, so there is some balance to be found.

Ideally this should probably be defaulted to some value that minimizes the occurrence of an interactive web-client user incurring JMX reconnections while they click through some typical workflows at a typical pace, but I don't really know what that value would be for certain.

pom.xml
andrewazores: Well, there is the `InterleavedExternalTargetRequestsIT`, which does have multiple targets and intentionally connects to them in a varying order. That test was written pretty much specifically to exercise the TargetConnectionManager and ensure that the caching behaviour works as expected and you don't end up with requests going to the wrong target etc. However, the number of concurrent connections open is really a performance optimization, and from an API perspective it doesn't matter, so the behaviour should be the same regardless of the cache size or number of targets.

I set it to 1 here again mostly just to ensure things don't blow up, but it might make sense to either set it to 0 for immediate eviction (highest test suite stress) or -1/unset for unbounded cache (also could cause stress if another test is written which has very many targets, but in general this would result in the fastest test suite execution time).

src/test/java/io/cryostat/net/TargetConnectionManagerTest.java
andrewazores: It will be "immediately" evicted, but this depends on the Caffeine/system scheduler, which does eviction and other tasks like this asynchronously on a worker thread. If I don't add a sleep here then the scheduler doesn't actually have a chance to evict the connection from the cache before I grab it again. That's what the [docs](https://javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Caffeine.html) mean by:

> As eviction is scheduled on the configured executor, tests may instead prefer to configure the cache to execute tasks directly on the same thread. 

I could, and maybe should, make it so that the tests use a same-thread scheduler, rather than the system scheduler, and that would make this test deterministic and remove the need for this sleep.

src/main/java/io/cryostat/net/NetworkModule.java
hareetd: Makes sense, thanks.

pom.xml
hareetd: Oh, that clears it up then, I didn't realize the connection concurrency was a performance optimization; thought it was a strict bound on the number of targets able to be connected to.

andrewazores: Yea, I realized the name was misleading, so I pushed a commit renaming the new variables in a way that (I hope) carries my intent better.

src/test/java/io/cryostat/net/TargetConnectionManagerTest.java
hareetd: I'll approve the current changes in the meantime then.

pom.xml
hareetd: Nice, the new variable names are definitely more clear.

hareetd (approved):

src/test/java/io/cryostat/net/TargetConnectionManagerTest.java
andrewazores: I just added a commit that configures the cache to use the current thread and no background scheduling for the unit tests where this makes sense. Otherwise the default executor (ForkJoinPool) and the system scheduler thread are still used.

hareetd (approved):

Issue #634, opened 2021-08-09
itest failures due to ordering, failed/lack of cleanup


andrewazores:
We're currently having some flaky itest results with sporadic failures that seem to change frequently, so I suspect instability due to (intentionally, for this kind of reason) randomized test ordering and probably some missing cleanup somewhere, particularly if test failures like #633 may cause the associated cleanup to be skipped or also fail, causing more cascading failures elsewhere.

`repeated-integration-tests.sh` will probably be helpful to track down these various failures.

- [ ] #632 
- [ ] #636

Issue #633, opened 2021-08-09
CI itest failures result in hanging run


andrewazores:

andrewazores:
I don't think this is true anymore. Failing tests seem to, well fail, just fine now.

PR #639, opened 2021-08-10
test(itest): fail itests fast to prevent cascades

andrewazores:
Cascading failures make it more difficult to troubleshoot failing tests, and there is not much reason currently to continue executing further tests past the first failure

https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#skipAfterFailureCount

Related to #633
Related to #634


andrewazores:
This is [unsupported in JUnit 5](https://maven.apache.org/surefire/maven-surefire-plugin/featurematrix.html) (unfortunately).

Issue #632, opened 2021-08-09
ReportIT fails due to empty recording


andrewazores:
https://github.com/cryostatio/cryostat/runs/3280669571?check_suite_focus=true#step:9:1096

andrewazores:
```
HTTP 500: Internal Server Error
Error:  Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.554 s <<< FAILURE! - in itest.ReportIT
Error:  testGetReportShouldSendFile  Time elapsed: 0.53 s  <<< ERROR!
java.util.concurrent.ExecutionException: io.vertx.ext.web.handler.impl.HttpStatusException: Internal Server Error
	at itest.ReportIT.testGetReportShouldSendFile(ReportIT.java:180)
Caused by: io.vertx.ext.web.handler.impl.HttpStatusException: Internal Server Error
```

and

```
Caused by: java.io.IOException: java.io.IOException: No recording data available
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.invokeOperation(JfrRecordingInputStreamV2.java:206)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.openStream(JfrRecordingInputStreamV2.java:181)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.openStream(JfrRecordingInputStreamV2.java:167)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.readStreamIdentifier(JfrRecordingInputStreamV2.java:120)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.fill(JfrRecordingInputStreamV2.java:104)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.read(JfrRecordingInputStreamV2.java:94)
	at java.base/java.io.InputStream.read(InputStream.java:271)
	at java.base/java.io.InputStream.transferTo(InputStream.java:704)
	at java.base/java.nio.file.Files.copy(Files.java:3078)
	at io.cryostat.core.sys.FileSystem.copy(FileSystem.java:91)
	at io.cryostat.recordings.RecordingArchiveHelper.writeRecordingToDestination(RecordingArchiveHelper.java:169)
[...]

PR #640, opened 2021-08-10
fix(recordingoptions): client may explicitly unset options

### LINKED ISSUE #634: itest failures due to ordering, failed/lack of cleanup

andrewazores:
Related to #632
Fixes #634

### DIFF
--- a/src/test/java/io/cryostat/net/web/http/api/v1/TargetRecordingOptionsPatchHandlerTest.java
+++ b/src/test/java/io/cryostat/net/web/http/api/v1/TargetRecordingOptionsPatchHandlerTest.java
@@ -115,16 +115,16 @@ class TargetRecordingOptionsPatchHandlerTest {
 
     @Test
     void shouldSetRecordingOptions() throws Exception {
-        Map<String, String> defaultValues =
+        Map<String, String> originalValues =
                 Map.of("toDisk", "true", "maxAge", "50", "maxSize", "32");
         Mockito.when(recordingOptionsBuilderFactory.create(Mockito.any())).thenReturn(builder);
         Mockito.when(builder.build()).thenReturn(recordingOptions);
-        Mockito.when(recordingOptions.get("toDisk")).thenReturn(defaultValues.get("toDisk"));
-        Mockito.when(recordingOptions.get("maxAge")).thenReturn(defaultValues.get("maxAge"));
-        Mockito.when(recordingOptions.get("maxSize")).thenReturn(defaultValues.get("maxSize"));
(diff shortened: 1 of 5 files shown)


hareetd:
Out of curiosity, how exactly was not being able to explicitly unset the recording options leading to the test failures from before?

andrewazores:
> Out of curiosity, how exactly was not being able to explicitly unset the recording options leading to the test failures from before?

This seems to have been the key: https://github.com/cryostatio/cryostat/pull/640/files#diff-5191e54db18231ddfc4a99eb2dd6289336bc68561507563c9c471d4f8986b0d7L74

When retrieving the options from the JVM, the default value being unset is reported as 0. But, explicitly setting a value and then "resetting" it to 0 seems to turn out to actually tell the target JVM to not record anything at all, as in the recording may have a maximum size of 0 bytes.

There's probably a better/correct way to get the `GET` recordingOptions data from the JMC API we're using to retrieve this so that we can return data to the client in the first place that distinguishes unset options from options set to 0/false, too.

hareetd (dismissed):
Looks good.

src/test/java/io/cryostat/net/web/http/api/v1/TargetRecordingOptionsPatchHandlerTest.java
```diff
@@ -159,6 +159,52 @@ public Map answer(InvocationOnMock args) throws Throwable {
         }
     }
 
+    @Test
+    void shouldUnsetRecordingOptions() throws Exception {
+        Map<String, String> defaultValues =
+                Map.of("toDisk", "true", "maxAge", "50", "maxSize", "32");
```
jan-law, line 165: Where do these default values come from (unless the values are arbitrary)?

jan-law (commented):
Besides this last question, looks good

src/test/java/io/cryostat/net/web/http/api/v1/TargetRecordingOptionsPatchHandlerTest.java
andrewazores: These are actually just arbitrary for this test. I should rename that variable.

hareetd (approved):

jan-law (approved):

PR #647, opened 2021-08-17
feat(rules): add cleanup query parameter

### LINKED ISSUE #642: Deleting a rule definition should optionally remove active recordings from targets
Currently, `DELETE`ing a rule definition simply removes the rule, preventing it from applying to any new targets in the future and cancelling any periodic archiving associated with the rule. However, the active recording created in matching targets will be left intact and running, and any associated archived recordings also remain in storage. This is because it is safer to leave that data in place to ensure it is not unintentionally lost, in case the user wishes to stop a rule from triggering on new targets in the future but still needs the recording data collected thus far.

There should be a way for the client to perform a `DELETE` on a rule definition and have this also perform cleanup on the matching targets. Archived recordings should most likely stay in place anyway, but active recordings can be stopped (and deleted?).

For example, `DELETE /api/v2/rules/someRule?clean=true` could specify that the "cleanup" of active recordings should be performed. `clean=false` or clean being unspecified should retain the current behaviour where recordings are left running.

An example use case is that the user has created a rule to start the Continuous template on all of their targets. Then, the user creates a customized template from the Continuous template, perhaps adjusting some sampling rates or event duration thresholds, or adding some application custom events. The user then wishes to replace the original rule with a new rule that applies their customized template.

In this scenario, the user would currently need to delete the original rule, then loop over all of their targets and issue requests to stop/delete the recordings created by that rule, and then create a new rule definition. If the user simply deletes the original rule and creates a new one with their new template then each of their targets will have two running recordings.

With this enhancement, the user would simply delete the original rule with the `clean=true` parameter and then create a new rule with their new template. The existing recordings would be stopped and new ones using the new template started instead.

andrewazores:
Fixes #642

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java
@@ -174,7 +174,7 @@ class RuleDeleteHandler extends AbstractV2RequestHandler<List<RuleDeleteHandler.
             }
         }
         if (failures.size() == 0) {
-            return new IntermediateResponse<List<RuleDeleteHandler.CleanupFailure>>().body(null);
+            return new IntermediateResponse<List<CleanupFailure>>().body(null);
         } else {
             return new IntermediateResponse<List<CleanupFailure>>().statusCode(500).body(failures);
         }
--- a/src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java
@@ -38,33 +38,55 @@
 package io.cryostat.net.web.http.api.v2;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Set;
 
 import javax.inject.Inject;
 
+import io.cryostat.configuration.CredentialsManager;
 import io.cryostat.core.log.Logger;
 import io.cryostat.net.AuthManager;
+import io.cryostat.net.ConnectionDescriptor;
+import io.cryostat.net.TargetConnectionManager;
 import io.cryostat.net.security.ResourceAction;
 import io.cryostat.net.web.http.HttpMimeType;
 import io.cryostat.net.web.http.api.ApiVersion;
+import io.cryostat.platform.PlatformClient;
+import io.cryostat.platform.ServiceRef;
 import io.cryostat.rules.Rule;
 import io.cryostat.rules.RuleRegistry;
 
 import com.google.gson.Gson;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.vertx.core.http.HttpMethod;
 
-class RuleDeleteHandler extends AbstractV2RequestHandler<Void> {
+class RuleDeleteHandler extends AbstractV2RequestHandler<List<RuleDeleteHandler.CleanupFailure>> {
 
     static final String PATH = RuleGetHandler.PATH;
+    static final String CLEAN_PARAM = "clean";
 
     private final RuleRegistry ruleRegistry;
+    private final TargetConnectionManager targetConnectionManager;
+    private final PlatformClient platformClient;
+    private final CredentialsManager credentialsManager;
     private final Logger logger;
 
     @Inject
-    RuleDeleteHandler(AuthManager auth, RuleRegistry ruleRegistry, Gson gson, Logger logger) {
+    RuleDeleteHandler(
+            AuthManager auth,
+            RuleRegistry ruleRegistry,
+            TargetConnectionManager targetConnectionManager,
+            PlatformClient platformClient,
+            CredentialsManager credentialsManager,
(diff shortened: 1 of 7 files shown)


src/main/java/io/cryostat/rules/RuleProcessor.java
```diff
@@ -238,6 +238,7 @@ private void startRuleRecording(ConnectionDescriptor connectionDescriptor, Rule
                             recordingTargetHelper.parseEventSpecifierToTemplate(
                                     rule.getEventSpecifier());
                     recordingTargetHelper.startRecording(
+                            true,
```
hareetd, line 241: I forgot the explanation from yesterday. Is this set to `true` so that every time a `Rule` is processed, if it has already been previously processed and a recording already exists, instead of throwing an `Exception` the old recording is 'quietly' replaced with a new one that is exactly the same?

andrewazores: Yea exactly. This is only done for recordings (re)started by rules, not ones created explicitly and manually by a user. Without this, if a user created rule `A` which in turn caused the creation and start of rule R, and then the user deleted `A` with cleanup, then `R` would be stopped and remain in the target. If the user then tries to (re)create a rule named `A'` that has the same definition as the original `A`, the rule would never be able to start a new recording `R'` because `R` is still there in a stopped state.

hareetd: Okay, that makes sense. Is there a reason for stopping the `Rule` recording instead of simply closing (i.e. deleting) it?

hareetd: Actually, I guess this way the user can still create a snapshot or dump the recording to archives.

andrewazores: It's more or less harmless to leave the recording in place but stopped - it'll remove any additional resource overhead that JFR was consuming while the recording ran, and the in-memory JFR ring buffers are pre-allocated regardless, so the only additional expense involved in stopping rather than deleting is that some disk space may be taken up if the recording was configured to also transfer chunks to disk. Since the cost is low and configurable, I figured it was best to leave the data there in case the user does still want it in the future, rather than delete it and have it lost forever if they haven't performed an archive on it in advance.

src/test/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandlerTest.java
```diff
@@ -169,5 +176,193 @@ void shouldRespondWith404ForNonexistentRule() throws Exception {
...
         }
+
+        @Test
+        void shouldRespondWith500ForCleanupFailures() throws Exception {
+            Mockito.when(params.getPathParams()).thenReturn(Map.of("name", testRuleName));
+            MultiMap queryParams = MultiMap.caseInsensitiveMultiMap();
+            queryParams.set("clean", "true");
+            Mockito.when(params.getQueryParams()).thenReturn(queryParams);
+
+            Rule rule =
```
jan-law, line 187: What is the advantage of creating `MultiMap queryParams`, `Rule rule` and `ServiceRef serviceRef` inside each test case vs declaring them at the top of the class and using `@BeforeEach void setup()` to instantiate them?

hareetd (dismissed):

src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java
```diff
@@ -99,16 +121,68 @@ public HttpMimeType mimeType() {
...
+                    failure.ref = ref;
+                    failure.message = e.getMessage();
+                    failures.add(failure);
+                }
+            }
+        }
+        if (failures.size() == 0) {
+            return new IntermediateResponse<List<RuleDeleteHandler.CleanupFailure>>().body(null);
+        } else {
+            return new IntermediateResponse<List<CleanupFailure>>().statusCode(500).body(failures);
```
jan-law, line 179: What is the difference between `List<RuleDeleteHandler.CleanupFailure>` and `List<CleanupFailure>` in this context?

andrewazores: Ah, nothing really. They refer to the same type, but I suppose my editor/code-completion did it one way and I hand-wrote it the other. I'll clean that up.

src/test/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandlerTest.java
andrewazores: It's just a bit of scope cleanliness. If not every test case uses `rule` and `serviceRef` then there is no need to have them as class members and in-scope for test cases where they are unused. But, *most* of the tests within this `@Nested` class scope do use them, so maybe it would be clearer to read if those were all pulled up to class members. I'm fine with either way, I'll perform that refactor if you'd prefer it.

jan-law: I'm good with it as is

jan-law (approved):

PR #648, opened 2021-08-18
feat(rules): implement "archive" rules

### LINKED ISSUE #643: POST JSON Rule definition can result in NPE
```
$ https -v --json POST :8181/api/v2/rules name:foo matchExpression="target.annotations.cryostat['JAVA_MAIN']=='demoApp'" eventSpecifier="template=Continuous,type=TARGET"
POST /api/v2/rules HTTP/1.1
Connection: keep-alive
Content-Length: 127
Host: localhost:8181
User-Agent: HTTPie/2.4.0
name: foo

{
    "eventSpecifier": "template=Continuous,type=TARGET",
    "matchExpression": "target.annotations.cryostat['JAVA_MAIN']=='demoApp'"
}


HTTP/1.1 500 Internal Server Error
content-encoding: gzip
content-length: 103

{
    "data": {
        "reason": null
    },
    "meta": {
        "status": "Internal Server Error",
        "type": "text/plain"
    }
}
```

I sent a `POST /api/v2/rules` and incorrectly formatted my request, so the `name` key was not included in the JSON body (instead, it was sent as an HTTP header). This resulted in an HTTP 500:

```
Aug 12, 2021 7:28:28 PM io.cryostat.core.log.Logger error
SEVERE: Exception thrown
	at io.cryostat.rules.Rule$Builder.from(Rule.java:234)
	at io.cryostat.util.RuleDeserializer.deserialize(RuleDeserializer.java:58)
	at io.cryostat.util.RuleDeserializer.deserialize(RuleDeserializer.java:51)
	at com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:69)
	at com.google.gson.Gson.fromJson(Gson.java:932)
	at com.google.gson.Gson.fromJson(Gson.java:897)
	at com.google.gson.Gson.fromJson(Gson.java:846)
	at com.google.gson.Gson.fromJson(Gson.java:817)
	at io.cryostat.net.web.http.api.v2.RulesPostHandler.handle(RulesPostHandler.java:144)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:114)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:68)

Aug 12, 2021 7:28:28 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:60442): POST /api/v2/rules 500 4ms
```

This should have instead been a `400`.

### LINKED ISSUE #645: Users should be able to create "one-shot"/"instant" Rule definitions
There should be a way to create Rule definition that is "one-shot", as in its effects are applied exactly once at creation time, but the rule is not persisted for future checks. This implies that one-shot rules will not support periodic archiving, so the `archivalPeriodSeconds` and `preservedArchives` rules parameters should either be ignored or should produce an error if provided for a one-shot rule definition.

One-shot rules would be used to start a recording using a given template (and maxAge in seconds or maxSize in bytes) exactly once, using the matchExpression to filter which targets will be affected.

Combined with #642 , this would allow users to start and stop recordings on sets of targets with single Cryostat API requests, allowing the automated rules backend to also handle #495 batch operations for the initially intended use cases. #495 could still implement new API endpoint handlers specific to starting/stopping/deleting recordings in batches, but the implementation would essentially work by piggybacking on top of the internal automated rules system.

andrewazores:
Closes #645

Fixes #643

This implements an alternative and simplified version of the "one-shot" rules.

If a Rule definition is given where `"eventSpecifier":"archive"`, the rule definition will not be persisted, does not require a unique name, may not have `archivalPeriodSeconds/preservedArchives/maxAge/maxSize`, and rather than starting a new recording will cause a snapshot to be taken, copied to archives, and the snapshot immediately removed.

The motivation is to allow a user to have a workflow such as the following:

```
POST /api/v2/rules
```
```json
{
    "name": "monitor",
    "description": "Monitor application, all the time",
    "eventSpecifier": "template=Continuous",
    "matchExpression": "target.labels.app=='MyApp'",
    "archivalPeriodSeconds": 3600,
    "maxAgeSeconds": 300,
    "preservedArchives": 6
}
```

Wait some time and observe application metrics (JFR or otherwise), either manually or by an external automated process. When some trigger event occurs, such as resource utilization exceeding a threshold for longer than some duration, the user wants to get copies of their application's JFR data at that moment and for the previous lookback window (5min in this case due to `maxAgeSeconds`).

Note that this rule configuration gives the user a 5min window of data in the recording at any time, and that that 5 minute sample will be copied to archives only once per hour (3600 seconds). 6 such copies will be retained in the archives at any given time. Such a configuration could be used to gather ongoing continuous monitoring data, leaving long gaps in between archival periods to prevent excessive use of disk space, for example.

Alternatively, the rule definition may only start a continuous recording in the targets, without any configuration for periodic archiving.

```
POST /api/v2/rules
```
```json
{
    "eventSpecifier": "archive",
    "matchExpression": "target.labels.app=='MyApp'"
}
```

After POSTing this second rule definition, Cryostat will connect to all of the same original set of targets and perform the snapshot/archive/remove snapshot procedure as explained earlier. The original recordings created by the first rule are still in place and running, so the user may re-POST the second rule definition as many times as they wish in the future as more external trigger events occur.

### DIFF
--- a/src/main/java/io/cryostat/rules/Rule.java
+++ b/src/main/java/io/cryostat/rules/Rule.java
@@ -64,10 +64,14 @@ public class Rule {
     private final int maxSizeBytes;
 
     Rule(Builder builder) throws MatchExpressionValidationException {
-        this.name = sanitizeRuleName(requireNonBlank(builder.name, Attribute.NAME));
+        this.eventSpecifier = builder.eventSpecifier;
+        if (isArchiver()) {
+            this.name = builder.name;
+        } else {
+            this.name = sanitizeRuleName(requireNonBlank(builder.name, Attribute.NAME));
+        }
         this.description = builder.description == null ? "" : builder.description;
         this.matchExpression = builder.matchExpression;
-        this.eventSpecifier = builder.eventSpecifier;
         this.archivalPeriodSeconds = builder.archivalPeriodSeconds;
         this.preservedArchives = builder.preservedArchives;
         this.maxAgeSeconds =
@@ -131,11 +135,8 @@ public class Rule {
     }
 
     public void validate() throws IllegalArgumentException, MatchExpressionValidationException {
-        requireNonBlank(this.name, Attribute.NAME);
         requireNonBlank(this.matchExpression, Attribute.MATCH_EXPRESSION);
         validateEventSpecifier(requireNonBlank(this.eventSpecifier, Attribute.EVENT_SPECIFIER));
-        requireNonNegative(this.archivalPeriodSeconds, Attribute.ARCHIVAL_PERIOD_SECONDS);
-        requireNonNegative(this.preservedArchives, Attribute.PRESERVED_ARCHIVES);
         validateMatchExpression(this);
 
         if (isArchiver()) {
@@ -143,6 +144,10 @@ public class Rule {
             requireNonPositive(this.preservedArchives, Attribute.PRESERVED_ARCHIVES);
             requireNonPositive(this.maxSizeBytes, Attribute.MAX_SIZE_BYTES);
             requireNonPositive(this.maxAgeSeconds, Attribute.MAX_AGE_SECONDS);
+        } else {
+            requireNonBlank(this.name, Attribute.NAME);
+            requireNonNegative(this.archivalPeriodSeconds, Attribute.ARCHIVAL_PERIOD_SECONDS);
+            requireNonNegative(this.preservedArchives, Attribute.PRESERVED_ARCHIVES);
         }
     }
 
@@ -188,10 +193,10 @@ public class Rule {
     }
 
     public static class Builder {
-        private String name;
-        private String description;
-        private String matchExpression;
-        private String eventSpecifier;
+        private String name = "";
+        private String description = "";
+        private String matchExpression = "";
+        private String eventSpecifier = "";
         private int archivalPeriodSeconds = 0;
         private int preservedArchives = 0;
         private int maxAgeSeconds = -1;
--- a/src/main/java/io/cryostat/rules/Rule.java
+++ b/src/main/java/io/cryostat/rules/Rule.java
@@ -39,6 +39,8 @@ package io.cryostat.rules;
 
 import java.util.function.Function;
 
+import io.cryostat.recordings.RecordingTargetHelper;
+
 import com.google.gson.JsonObject;
 import io.vertx.core.MultiMap;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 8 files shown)


src/main/java/io/cryostat/rules/RuleProcessor.java
```diff
@@ -255,13 +255,12 @@ private void startRuleRecording(ConnectionDescriptor connectionDescriptor, Rule
...
-                                    .name(rule.getRecordingName())
-                                    .toDisk(true);
+                                    .name(rule.getRecordingName());
                     if (rule.getMaxAgeSeconds() > 0) {
-                        builder = builder.maxAge(rule.getMaxAgeSeconds());
+                        builder = builder.maxAge(rule.getMaxAgeSeconds()).toDisk(true);
                     }
                     if (rule.getMaxSizeBytes() > 0) {
-                        builder = builder.maxSize(rule.getMaxSizeBytes());
+                        builder = builder.maxSize(rule.getMaxSizeBytes()).toDisk(true);
```
jan-law, line 263: Why are rules with `maxAge` or `maxSize`  always stored to disk?

andrewazores: https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/Recording.html#setMaxAge(java.time.Duration)

maxAge/maxSize control the size of the disk repository for JFR data that gets flushed to disk. If toDisk=false, maxAge/maxSize have no effect.

andrewazores: The in-memory JFR ring buffers have a relatively small fixed size, so it's almost certain that a Cryostat user's use cases would include enabling toDisk. If they're configuring maxAge/maxSize it seems reasonable to assume they intend toDisk=true as well.

src/main/java/io/cryostat/rules/Rule.java
```diff
@@ -60,10 +65,14 @@
     private final int maxSizeBytes;
 
     Rule(Builder builder) throws MatchExpressionValidationException {
-        this.name = sanitizeRuleName(requireNonBlank(builder.name, Attribute.NAME));
+        this.eventSpecifier = builder.eventSpecifier;
+        if (isArchiver()) {
+            this.name = builder.name;
```
jan-law, line 70: Why aren't archive names sanitized?
```
{
    "name": "unsanitized name",    
    "eventSpecifier": "archive",
    "matchExpression": "target.labels.app=='MyApp'"
}
```

jan-law (commented):
Looks good otherwise

hareetd (commented):
Looks good

src/main/java/io/cryostat/rules/Rule.java
andrewazores: They're actually just optional on archive rules, since they aren't used to create any recording name and aren't persisted. You can omit the `name` field entirely when creating an archive rule.

jan-law (approved):

PR #650, opened 2021-08-19
feat(health): add cryostatVersion

### LINKED ISSUE #649: Build should include git tag/hash as resource

andrewazores:
Fixes #649

Depends on https://github.com/cryostatio/cryostat-web/pull/241

```
$ https :8181/health
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 117

{
    "cryostatVersion": "v2.0.0-SNAPSHOT-188-ga975bef7",
    "dashboardAvailable": false,
    "datasourceAvailable": false
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -235,6 +235,30 @@
       </configuration>
     </plugin>
     <plugin>
+      <groupId>org.codehaus.mojo</groupId>
+      <artifactId>exec-maven-plugin</artifactId>
+      <version>${org.codehaus.mojo.exec.plugin.version}</version>
+      <executions>
+        <execution>
+          <id>generate-git-version</id>
+          <phase>generate-resources</phase>
+          <goals>
+            <goal>exec</goal>
+          </goals>
+          <configuration>
+            <executable>git</executable>
+            <arguments>
+              <argument>describe</argument>
+              <argument>--dirty</argument>
+              <argument>--long</argument>
+              <argument>--always</argument>
+            </arguments>
+            <outputFile>${project.build.directory}/assets/app/resources/io/cryostat/version</outputFile>
+          </configuration>
+        </execution>
+      </executions>
+    </plugin>
+    <plugin>
       <artifactId>maven-resources-plugin</artifactId>
       <version>${org.apache.maven.plugins.resources.version}</version>
       <executions>
@@ -273,7 +297,7 @@
           <id>create-pod</id>
           <phase>pre-integration-test</phase>
           <goals>
-              <goal>exec</goal>
+            <goal>exec</goal>
           </goals>
           <configuration>
             <executable>${imageBuilder}</executable>
--- a/pom.xml
+++ b/pom.xml
@@ -235,30 +235,6 @@
       </configuration>
     </plugin>
     <plugin>
-      <groupId>org.codehaus.mojo</groupId>
-      <artifactId>exec-maven-plugin</artifactId>
-      <version>${org.codehaus.mojo.exec.plugin.version}</version>
-      <executions>
-        <execution>
-          <id>generate-git-version</id>
-          <phase>generate-resources</phase>
-          <goals>
-            <goal>exec</goal>
-          </goals>
-          <configuration>
-            <executable>git</executable>
-            <arguments>
-              <argument>describe</argument>
-              <argument>--dirty</argument>
-              <argument>--long</argument>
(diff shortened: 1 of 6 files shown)


andrewazores:
Huh, the `git describe` run by maven-exec-plugin exits with a 128 status code in the CI runner. Any idea why that is?

ebaron:
That is odd. What about trying `git rev-parse --short HEAD`? (https://stackoverflow.com/a/949391)

andrewazores:
That could work, but it just gives the shortened hash of the current commit - I was hoping to get the name of the tag pointing to the current commit, if any, which `git describe` does do (ex. `v2.0.0-SNAPSHOT-189-gcad25f27` on this branch, but if you check out a tag then it just outputs the tag name).

andrewazores:
Ah, I think the problem with `describe` is probably because of the way that the `actions/checkout@v2` goes about cloning the repo and pulling the PR branch.

```
andrew@vengeance ~ $ d=`mktemp -d` && cd $d
andrew@vengeance /tmp/tmp.o2Q4MzzzXf $ git init foo
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
Initialized empty Git repository in /tmp/tmp.o2Q4MzzzXf/foo/.git/
candrew@vengeance /tmp/tmp.o2Q4MzzzXf $ cd foo
andrew@vengeance /tmp/tmp.o2Q4MzzzXf/foo (master) $ git describe
fatal: No names found, cannot describe anything.
andrew@vengeance /tmp/tmp.o2Q4MzzzXf/foo :( (master) $ echo $?
128
```

Looks like there are no tags present, so `describe` prints out that error and exits with the 128 status.

andrewazores:
Looks like that worked. I'm putting together a small -web PR as well that goes with this, so I'll update this PR with a commit to update the web-client submodule as well after that is ready and merged in the -web side.

pom.xml
```diff
@@ -234,6 +234,30 @@
...
+      <groupId>org.codehaus.mojo</groupId>
+      <artifactId>exec-maven-plugin</artifactId>
+      <version>${org.codehaus.mojo.exec.plugin.version}</version>
+      <executions>
+        <execution>
+          <id>generate-git-version</id>
+          <phase>generate-resources</phase>
+          <goals>
+            <goal>exec</goal>
+          </goals>
```
ebaron, line 247: I think this has to be combined with the existing `exec-maven-plugin` definition. I see this in the build logs:
```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for io.cryostat:cryostat:jar:2.0.0-SNAPSHOT
[WARNING] 'build.plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.codehaus.mojo:exec-maven-plugin @ line 285, column 13
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

andrewazores: Good catch, fixed.

ebaron (commented):
With your feature branch checked out, I get this:
```
$ git describe --dirty --long --always
v0.20.0-209-gc4ed02f8
```

Kinda weird, but maybe not an issue in practice.

andrewazores:
Hmm, there should be a more recent "ancestor" tag than v0.20.0. What if you do a `git fetch --tags upstream` (or whatever you have named the remote)? I see:

```
$ git describe --dirty --long --always
v2.0.0-SNAPSHOT-192-gc4ed02f8
```

andrewazores:
If I check out that `v2.0.0-SNAPSHOT` tag specifically rather than this feature branch, then the output is `v2.0.0-SNAPSHOT-0-g0a4045cf`, which I think looks pretty good. It's a very precise description that this is exactly the commit with that named tag.

ebaron:
That did the trick. I see the same now.

ebaron (approved):

PR #651, opened 2021-08-19
Match expression regex

jan-law:
Related #495 

Adds the ability to apply rules to targets matching regular expressions in `matchExpression`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "web-client"]
-	path = web-client
-	url = https://github.com/cryostatio/cryostat-web
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+    
+        {
+            "type": "java",
+            "name": "Debug (Attach)",
+            "request": "attach",
+            "hostName": "localhost",
+            "port": 5005
+            } 
+        
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "automatic"
+}
--- a/src/main/java/io/cryostat/rules/MatchExpressionTreeVisitor.java
+++ b/src/main/java/io/cryostat/rules/MatchExpressionTreeVisitor.java
@@ -171,11 +171,6 @@ class MatchExpressionTreeVisitor extends SimpleTreeVisitorES5_1<Void, String> {
     }
 
     @Override
-    public Void visitRegExpLiteral(RegExpLiteralTree node, String matchExpression) {
-        return fail(node, matchExpression);
-    }
-
-    @Override
     public Void visitReturn(ReturnTree node, String matchExpression) {
         return fail(node, matchExpression);
     }
--- a/src/test/.vscode/launch.json
+++ b/src/test/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+    
+        {
+            "type": "java",
+            "name": "Debug (Attach)",
+            "request": "attach",
+            "hostName": "localhost",
+            "port": 8000
+            } 
+        
+    ]
+}
--- a/src/test/java/itest/AutoRulesIT.java
+++ b/src/test/java/itest/AutoRulesIT.java
@@ -228,6 +228,55 @@ class AutoRulesIT extends ExternalTargetsTest {
 
     @Test
     @Order(3)
+    void testAddRuleCreatedWithRegex() throws Exception {
+        CompletableFuture<JsonObject> postResponse = new CompletableFuture<>();
+        JsonObject regexRule = new JsonObject();
+        regexRule.put("name", "Regex_Rule");
+        regexRule.put("description", "AutoRulesIT automated rule");
+        regexRule.put("eventSpecifier", "template=Continuous,type=TARGET");
+        regexRule.put("matchExpression", "/^[a-zA-Z0-9]+$/");
+
+        try {
+        webClient
+                .post("/api/v2/rules")
+                .sendJsonObject(
+                    regexRule,
+                        ar -> {
+                            if (assertRequestStatus(ar, postResponse)) {
+                                MatcherAssert.assertThat(
+                                        ar.result().statusCode(), Matchers.equalTo(201));
+                                postResponse.complete(ar.result().bodyAsJsonObject());
+                            }
+                        });
+
+            JsonObject expectedPostResponse =
+                new JsonObject(
+                        Map.of(
+                                "meta",
+                                        Map.of(
+                                                "type",
+                                                HttpMimeType.PLAINTEXT.mime(),
+                                                "status",
+                                                "Created"),
+                                "data", Map.of("result", "Regex_Rule")));
+        MatcherAssert.assertThat(postResponse.get(), Matchers.equalTo(expectedPostResponse));
+        } finally {
+             CompletableFuture<JsonObject> response = new CompletableFuture<>();
+            webClient
+                .delete(String.format("/api/v2/rules/%s", "Regex_Rule"))
+                .send(
+                        ar -> {
+                            if (assertRequestStatus(ar, response)) {
+                                MatcherAssert.assertThat(
+                                        ar.result().statusCode(), Matchers.equalTo(200));
+                                response.complete(ar.result().bodyAsJsonObject());
+                            }
+                        });
+        }
+    }
+
+    @Test
+    @Order(4)
     void testAddCredentials() throws Exception {
         CompletableFuture<JsonObject> response = new CompletableFuture<>();
         MultiMap form = MultiMap.caseInsensitiveMultiMap();
--- a/src/test/java/itest/AutoRulesIT.java
+++ b/src/test/java/itest/AutoRulesIT.java
@@ -227,14 +227,14 @@ class AutoRulesIT extends ExternalTargetsTest {
     }
 
     @Test
-    @Order(3)
+    @Order(4)
     void testAddRuleCreatedWithRegex() throws Exception {
         CompletableFuture<JsonObject> postResponse = new CompletableFuture<>();
         JsonObject regexRule = new JsonObject();
         regexRule.put("name", "Regex_Rule");
         regexRule.put("description", "AutoRulesIT automated rule");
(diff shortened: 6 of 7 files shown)


jan-law:
How does the script engine in ` public boolean applies(Rule rule, ServiceRef serviceRef)` compare `matchExpression` to targets? 

```
            Object result =
                    this.scriptEngine.eval(rule.getMatchExpression(), createBindings(serviceRef));
            if (result instanceof Boolean) {
                return (Boolean) result;
```

How would it interpret these two match expressions? 
`regexRule.put("matchExpression", "target.alias == /^[a-zA-Z0-9.]+$/");`

`regexRule.put("matchExpression", "/^[a-zA-Z0-9.]+$/");`

Based on my itest results, I don't think either of these rule definitions are applied to the cryostat target.

andrewazores:
> How does the script engine in ` public boolean applies(Rule rule, ServiceRef serviceRef)` compare `matchExpression` to targets?
> 
> ```
>             Object result =
>                     this.scriptEngine.eval(rule.getMatchExpression(), createBindings(serviceRef));
>             if (result instanceof Boolean) {
>                 return (Boolean) result;
> ```
> 
> How would it interpret these two match expressions?
> `regexRule.put("matchExpression", "target.alias == /^[a-zA-Z0-9.]+$/");`
> 
> `regexRule.put("matchExpression", "/^[a-zA-Z0-9.]+$/");`
> 
> Based on my itest results, I don't think either of these rule definitions are applied to the cryostat target.

It "runs" the matchExpression as a script, but the "script" is expected to really be an expression that evaluates to a boolean value. That result comes back as a Java Boolean, and the rule is considered to have matched the target depending on that `if (result instanceof Boolean) return (Boolean) result` piece.

> `regexRule.put("matchExpression", "/^[a-zA-Z0-9.]+$/");`

This one won't work because this means the expression just evaluates to a RegExp literal, not a boolean value.

> `regexRule.put("matchExpression", "target.alias == /^[a-zA-Z0-9.]+$/");`

This one does evaluate to a boolean value, but `target.alias` on the LHS is a `string`, and on the RHS you have another RegExp literal, so those two things do not `==` each other.

It seems like what we need is to be able to do ex. `target.alias.match(/[a-zA-Z0-9]+/)` (which returns a boolean), but this is tricky because it involves calling the `match` function. So we somehow want to allow calling this function, but not calling any others, or at least not arbitrary ones like `System.exit(1)`.

jan-law:
In my latest commit, I tried using `node.getFunctionSelect().toString()` to compare the function to "match", except `getFunctionSelect` returns "jdk.nashorn.api.tree.MemberSelectTreeImpl@62244557".

`Tree.Kind` returns the type of tree instead of the function name. ([functionCallTree docs](https://docs.oracle.com/en/java/javase/12/docs/api/jdk.scripting.nashorn/jdk/nashorn/api/tree/FunctionCallTree.html))
I suppose I could check if the arguments in `node.getArguments()` are of type `REGEXP_LITERAL` but it still doesn't tell me the name of the function?

I'd appreciate some direction as to how to compare the function node with the `match` function.

andrewazores:
> `getFunctionSelect` returns "jdk.nashorn.api.tree.MemberSelectTreeImpl@62244557".

This looks like calling `getFunctionSelect` returns you a tree node that represents a selected member of the string object, which in this case would be a member that is a function named `match`, hopefully. The string you have there looks a lot like the default `toString()` of a MemberSelectTreeImpl instance. The API type definition corresponding to that type should be this:

https://docs.oracle.com/en/java/javase/12/docs/api/jdk.scripting.nashorn/jdk/nashorn/api/tree/MemberSelectTree.html

`((MemberSelectTree) node.getFunctionSelect()).getIdentifier()` might give you `"match"`. Or if not, chaining a call to `((MemberSelectTree) node.getFunctionSelect()).getExpression()` might reveal more info you can use to get where we need to be.

(Check with `instanceof` before making the cast to `MemberSelectTree` in the snippets above)

jan-law:
Thanks, `((MemberSelectTree) node.getFunctionSelect()).getIdentifier()` returns `"match"`. 

When the `scriptEngine` evaluates the matchExpression, the `result` is `io.cryostat.Cryostat` instead of a boolean.
I tried putting parentheses around the target alias, `(target.alias).match(/[a-zA-Z0-9.]+/)` with the same result. I also can't step into `return super.visitFunctionCall(node, matchExpression);`. What is happening here?
image

andrewazores:
> Thanks, `((MemberSelectTree) node.getFunctionSelect()).getIdentifier()` returns `"match"`.
> 
> When the `scriptEngine` evaluates the matchExpression, the `result` is `io.cryostat.Cryostat` instead of a boolean.
> I tried putting parentheses around the target alias, `(target.alias).match(/[a-zA-Z0-9.]+/)` with the same result. I also can't step into `return super.visitFunctionCall(node, matchExpression);`. What is happening here?
> image

Agh. Looks like the JS string.match() function is a little funky with its return values:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

Maybe what we want instead for simplicity's sake is RegExp.test():

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

So the test case would instead be like:

```json
{
  "matchExpression": "/^[a-z]+$/.test(target.alias)"
}
```

jan-law:
My integration test returns an HTTP 500 somewhere inside `recordingTargetHelper.startRecording()` which was called by `startRuleRecording()`. The script engine returns `true` when matching any target aliases with `/[a-zA-Z0-9.]/.test(target.alias)`. 

If I step over `startRuleRecording()` or `serviceRef -> activate(event.getPayload(), serviceRef)`, I see the HTTP 500. If I step into `startRecording()`, the HTTP 500 error gets thrown even if I paused at a breakpoint. How can I find the cause of this error?

```
public synchronized void onEvent(Event<RuleEvent, Rule> event) {
        switch (event.getEventType()) {
            case ADDED:
                platformClient.listDiscoverableServices().stream()
                        .filter(serviceRef -> registry.applies(event.getPayload(), serviceRef))
                        .forEach(serviceRef -> activate(event.getPayload(), serviceRef));
```
```
recordingTargetHelper.startRecording(
                            true,
                            connectionDescriptor,
                            builder.build(),
                            template.getLeft(),
                            template.getRight())
```

andrewazores:
You might find more detail by doing `cat target/cryostat-itest-*.log` in the cryostat project directory, after running that integration test. This will let you see the Cryostat backend logs from the itest run, which should hopefully give you a stack trace and other details about the failure.

Attaching a debugger and stepping through may be difficult/confusing there because things may be executed on different threads.

jan-law:
It looks like cryostat successfully applies my new rule to cryostat:9091 but fails to connect to the fib demo on port 9093. The error disappears if I move the regex test after `testNewContainerHasRuleApplied()`. 

Why does adding another container from `testNewContainerHasRuleApplied()` resolve the HTTP 500 error? Or is it because the fib demo needed credentials before applying rules?
```
Aug 25, 2021 2:17:12 PM io.cryostat.core.log.Logger info
INFO: Creating connection for service:jmx:rmi:///jndi/rmi://cryostat-itests:9091/jmxrmi
Aug 25, 2021 2:17:12 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"RecordingCreated","type":{"type":"application","subType":"json"},"serverTime":1629901032},"message":{"recording":"auto_Regex_Rule","target":"service:jmx:rmi:///jndi/rmi://cryostat-itests:9091/jmxrmi"}}
Aug 25, 2021 2:17:12 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:60284): POST /api/v2/rules 201 794ms
Aug 25, 2021 2:17:12 PM io.cryostat.core.log.Logger info
INFO: Creating connection for service:jmx:rmi:///jndi/rmi://cryostat-itests:9093/jmxrmi
Aug 25, 2021 2:17:12 PM io.cryostat.core.log.Logger info
INFO: connection attempt failed.
Aug 25, 2021 2:17:12 PM io.cryostat.core.log.Logger error
SEVERE: Exception thrown
Caused by: org.openjdk.jmc.rjmx.ConnectionException caused by java.io.IOException: Failed to retrieve RMIServer stub: javax.naming.ServiceUnavailableException [Root exception is java.rmi.ConnectException: Connection refused to host: cryostat-itests; nested exception is: 
	java.net.ConnectException: Connection refused (Connection refused)]
	at org.openjdk.jmc.rjmx.internal.RJMXConnection.connect(RJMXConnection.java:441)
	at io.cryostat.core.net.JFRConnection.attemptConnect(JFRConnection.java:208)
	at io.cryostat.core.net.JFRConnection.connect(JFRConnection.java:165)
	at io.cryostat.core.net.JFRConnection.getService(JFRConnection.java:105)
	at io.cryostat.net.web.http.api.v1.TargetRecordingsGetHandler.lambda$handleAuthenticated$0(TargetRecordingsGetHandler.java:115)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:136)
	at io.cryostat.net.web.http.api.v1.TargetRecordingsGetHandler.handleAuthenticated(TargetRecordingsGetHandler.java:111)
...
```

jan-law:
I'm getting some inconsistent behaviour with the default recording options. Updating the snapshot itest with `continuous:false` makes sense to me. What about recording `maxAge` and snapshot `duration`?

line 368: `MatcherAssert.assertThat(recording.getInteger("maxAge"), Matchers.equalTo(0));`
```
Expected: <0>
     but: was <30000>
	at itest.AutoRulesIT.testAddRuleCreatedWithRegex(AutoRulesIT.java:368)
```
```
Error:    AutoRulesIT.testAddRuleCreatedWithRegex:368 
Expected: <30000>
     but: was <0>
```
```
Expected: ..."duration":0,"continuous":true,...
but: was ..."duration":87000,"continuous":false...
Expected: <{"meta":{"type":"text/plain","status":"Created"},"data":{"result":
	at itest.SnapshotIT.testPostV2ShouldCreateSnapshot(SnapshotIT.java:297)
```
```
Expected: ..."duration":0,"continuous":true,...
but: was ..."duration":34000,"continuous":false...
Expected: <{"meta":{"type":"text/plain","status":"Created"},"data":{"result":
	at itest.SnapshotIT.testPostV2ShouldCreateSnapshot(SnapshotIT.java:297)
```

andrewazores:
By "inconsistent", you mean it's changing between test runs? If you do `mvn -DrunOrder=filesystem clean verify` (or whatever - add the `-DrunOrder=filesystem` to what you're normally doing) so that the tests always run in a consistent order, rather than randomized, do the results become consistent?

jan-law:
After running `mvn -Dcryostat.minimal=true -DskipUTs=true -DrunOrder=filesystem clean verify` a couple times, the `maxAge` for rule recordings is always 0 - my last commit should fix that.

The snapshot recording `duration` still changes between test runs. I've seen
`duration:35000`, `duration:10500` and `duration:38000`.

andrewazores:
If the `maxAge` on the recording is inconsistent when the runOrder is randomized then this still makes me think we're suffering from some kind of inter-test dependencies or improper cleanup.

The duration of the snapshot probably makes sense that it changes. Snapshots are created at a specific instant in time, but they are still a Recording in the end, in the STOPPED state and with finite data, so the reported duration is probably something like the time difference between the timestamp of the first recorded event and the last one, or the first event and when the snapshot was taken.

PR #662, opened 2021-08-25
feat(openshift): configurable resource mapping from properties

### LINKED ISSUE #635: Fine-tune Kubernetes/OpenShift RBAC permissions
https://github.com/cryostatio/cryostat/pull/599#pullrequestreview-723720661

Related #659

andrewazores:
Fixes #635


andrewazores:
@ebaron I haven't exercised this in an actual OpenShift/k8s environment in a while, but I just finished rebasing this and getting it back into shape. Let me know what you think and if this is a good step toward the better Kubernetes RBAC stuff we need. The configuration is done in the file `src/main/java/io/cryostat/net/openshift/OpenShiftAuthManager.properties`. For now it just maps the Cryostat application-level resource `RECORDING` to k8s `operator.cryostat.io/recordings` (the Recordings CRD), and the `CERTIFICATE` to k8s `apps/deployments,pods`.

src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
```diff
@@ -0,0 +1,13 @@
...
+# DEPLOYMENTS="apps/deployments"
+# PODS="pods"
+# CRYOSTATS="operator.cryostat.io/cryostats"
+# FLIGHTRECORDERS="operator.cryostat.io/flightrecorders"
+# RECORDINGS="operator.cryostat.io/recordings"
+
+# ResourceType to Kubernetes resource kind mapping:
+RECORDING=operator.cryostat.io/recordings
+CERTIFICATE=apps/deployments,pods
+
```
ebaron, line 13: For the syntax, how about instead of `group/resource` we use `resource.group`? This matches what you would see with `kubectl auth can-i --list`. The resource will always be the substring up This has the advantage of allowing us to add support for subresources like `pod/exec` or `deployments.apps/scale`, which use the slash as a separator.

ebaron (commented):
Hey Andrew, seems good overall. One comment about the property file syntax.

src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
andrewazores: Makes sense, thanks. I've pushed one extra commit that changes the format to `resource.group/sub`

ebaron: Do we need to definitions for the other Cryostat resource types here (e.g. `TARGET`, `CREDENTIALS`, etc.)?

andrewazores: Sure, that's what this PR aims to make easier :-)

Do you have any suggestions for what k8s RBAC types we should map those application-level resources to? ie if a user wants to `CREATE TARGET`, aka create a custom target definition, what permissions should their OpenShift account have?

ebaron: Would `CREATE TARGET` have some effect on the target JVM? Like Cryostat establishing a remote JMX connection for example?

andrewazores: No, not until some client tries to do something like start a recording on that target.

andrewazores: > What if we just add the existing set of permissions to the properties file?

Sure - you mean to just transfer the pre-existing code-defined permissions map (around here: https://github.com/cryostatio/cryostat/blob/b44770868ca2ef66991404a03c8e161e60012e90/src/main/java/io/cryostat/net/openshift/OpenShiftAuthManager.java#L518) into the `.properties`?

ebaron: Right

andrewazores: Done in the latest commit

ebaron (approved):
Works well!

Issue #667, opened 2021-08-27
Revitalize upstream documentation


jiekang:
The upstream documentation (README.md, website, etc.) could use another pass over to:

- link more user-relevant information
- review and re-organize existing information
- standardize format among the cryostatio org repositories

- [ ] Update cryostat developer documentation https://github.com/cryostatio/cryostat/pull/669
- [ ] Update cryostat-web developer documentation https://github.com/cryostatio/cryostat-web/pull/269

PR #669, opened 2021-08-30
fix(docs): Update README.md

jiekang:
Related to https://github.com/cryostatio/cryostat/issues/667

I plan to do do multiple PRs in multiple repositories against that issue, but maybe I should create an issue per PR. Let me know what you think.

This PR mainly touches up the developer-related information. I'm still brainstorming how things will be organized. At the moment I'm leaning to having the user-guide documentation moved to the website, cryostat.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,9 +77,6 @@ exec:exec@destroy-pod`
 
 ## RUN
 
-### Development Run Outside of Containers
-* `MAVEN_OPTS="-Dcom.sun.management.jmxremote.port=9091 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.autodiscovery=true" mvn clean prepare-package exec:java`.
-
 ### Run on Kubernetes/Openshift
 * See the [cryostat-operator](https://github.com/cryostatio/cryostat-operator)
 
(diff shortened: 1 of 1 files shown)


andrewazores:
> I plan to do do multiple PRs in multiple repositories against that issue, but maybe I should create an issue per PR. Let me know what you think.

A single issue is fine. Maybe just put a list in it with action items for each PR opened.

README.md
```diff
@@ -25,141 +27,106 @@ a recommended tool for more full-featured analysis of JFR files beyond what
...
-
-Some of Cryostat's dependencies also use java.util.logging for their logging.
-Cryostat disables
-[some of these](https://github.com/cryostatio/cryostat-core/tree/main/src/main/resources/config/logging.properties)
-by default, because they generate unnecessary logs.
-However, they can be reenabled by overriding the default configuration file
-and setting the disabled loggers to the desired level.
+
+### Development Run Outside of Containers
+* `MAVEN_OPTS="-Dcom.sun.management.jmxremote.port=9091 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.autodiscovery=true" mvn clean prepare-package exec:java`.
```
andrewazores, line 81: I think we might just want to remove this subsection. It doesn't work as written and would probably take a fair bit of work to get it back to working, and I'm not sure the value of supporting it at this time.

README.md
```diff
@@ -25,141 +27,106 @@ a recommended tool for more full-featured analysis of JFR files beyond what
...
+
+## CONFIGURATION
+
+Cryostat can be configured via the following environment variables
+
+#### Configuration for cryostat-web
+
+* `CRYOSTAT_WEB_HOST`: the hostname used by cryostat-web
+* `CRYOSTAT_WEB_PORT`: the internal port used by cryostat-web
+* `CRYOSTAT_EXT_WEB_PORT`: the external port used by cryostat-web
```
andrewazores, line 100: The three env vars above aren't really just for -web, they are for the Cryostat backend internal HTTP webserver. `WEB_HOST` and `WEB_PORT` define the address that the webserver binds on, for example.

README.md
```diff
@@ -25,141 +27,106 @@ a recommended tool for more full-featured analysis of JFR files beyond what
...
+* `CRYOSTAT_CORS_ORIGIN`: the origin for CORS to load a different cryostat-web instance
+
+#### Configuration for cryostat
+
+* `CRYOSTAT_MAX_WS_CONNECTIONS`: the maximum number of webscoket client connections allowed (minimum 1, maximum 64, default 2)
+* `CRYOSTAT_AUTH_MANAGER`: the authentication/authorization manager used for validating user accesses. See the `USER AUTHENTICATION / AUTHORIZATION` section for more details. Set to the fully-qualified class name of the auth manager implementation to use, ex. `io.cryostat.net.BasicAuthManager`.
+* `CRYOSTAT_PLATFORM`: the platform client  used for performing platform-specific actions, such as listing available target JVMs. If `CRYOSTAT_AUTH_MANAGER` is not specified then a default auth manager will also be selected corresponding to the platform, whether that platform is specified by the user or automatically detected. Set to the fully-qualified name of the platform detection strategy implementation to use, ex. `io.cryostat.platform.internal.KubeEnvPlatformStrategy`.
```
andrewazores, line 107: There appear to be two space chars between `client` and `used`

README.md
```diff
@@ -25,141 +27,106 @@ a recommended tool for more full-featured analysis of JFR files beyond what
...
+* `CRYOSTAT_CONFIG_PATH`: the local filesystem path for the configuration directory (default `/opt/cryostat.d/conf.d`)
```
andrewazores, line 108: Should this explain that "local" means within the container, ie local to the Cryostat process?

README.md
```diff
@@ -25,141 +27,106 @@ a recommended tool for more full-featured analysis of JFR files beyond what
...
+
+* `CRYOSTAT_JUL_CONFIG` : the `java.util.logging.config.file` configuration file for logging via SLF4J Some of Cryostat's dependencies also use java.util.logging for their logging. Cryostat disables [some of these](https://github.com/cryostatio/cryostat-core/tree/main/src/main/resources/config/logging.properties) by default, because they generate unnecessary logs. However, they can be reenabled by overriding the default configuration file and setting the disabled loggers to the desired level.
+
+#### Configuration for Event Templates
+
+* `CRYOSTAT_TEMPLATE_PATH`: the local storage path for Cryostat event templates
+
+#### Configuration for Archiving
+
+* `CRYOSTAT_ARCHIVE_PATH`: the local storage path for archived recordings
```
andrewazores, line 129: same here and for TEMPLATE_PATH re: "local"

README.md
jiekang: I can remove this. Is it something to also remove elsewhere (maybe pom.xml) s.t. users also cannot run this anyore?

README.md
jiekang: Hm; I'm a little confused here. Should they be `CRYOSTAT_HOST` then? Or I suppose WEB refers to web server rather than cryostat-web?

README.md
jiekang: An interesting question. When I think of local in the context of containers, I do mean local to the container already heh :P But then local on non-containers is the host machine. I'll try to think of something that's generically less confusing.

README.md
andrewazores: Yea, it was meant to refer to the webserver. It was named this way originally as a contrast to the old `NET` variables which did the same thing but for the raw TCP socket connection.

README.md
andrewazores: I don't think anything in the pom.xml needs to be removed, no. That's a pretty standard mvn invocation, nothing is there specifically to support this.

README.md
jiekang: After further investigation, local in the context of containers is used to refer to the host machine filesystem. I guess that was just me. I've removed that confusing term for now.

PR #686, opened 2021-09-21
[WIP] Agent plugin Template management handlers

Josh-Matsuoka:
Expect this to change as more things are added. Like the accompanying core PR the goal here is to identify issues with style/design early.


run.sh
```diff
@@ -72,13 +72,16 @@ podman run \
...
+    --mount type=tmpfs,target=/opt/cryostat.d/probes.d \
     --mount type=bind,source="$(dirname $0)/conf",destination=/opt/cryostat.d/conf.d,relabel=shared,bind-propagation=shared \
     --mount type=bind,source="$(dirname $0)/truststore",destination=/truststore,relabel=shared,bind-propagation=shared \
     --mount type=bind,source="$(dirname $0)/certs",destination=/certs,relabel=shared,bind-propagation=shared \
     --mount type=bind,source="$(dirname $0)/clientlib",destination=/clientlib,relabel=shared,bind-propagation=shared \
     -e CRYOSTAT_PLATFORM=$CRYOSTAT_PLATFORM \
     -e CRYOSTAT_DISABLE_SSL=$CRYOSTAT_DISABLE_SSL \
     -e CRYOSTAT_DISABLE_JMX_AUTH=$CRYOSTAT_DISABLE_JMX_AUTH \
+    -e CRYOSTAT_DISABLE_SSL="true" \
+    -e CRYOSTAT_DISABLE_JMX_AUTH="true" \
```
andrewazores, line 84: It isn't required to disable these, is it? - just easier for testing temporarily?

src/main/java/io/cryostat/net/web/http/api/v1/ProbeTemplateDeleteHandler.java
```diff
@@ -0,0 +1,125 @@
...
+        super(auth);
+        this.notificationFactory = notificationFactory;
+        this.logger = logger;
+        this.probeTemplateService = probeTemplateService;
+        this.fs = fs;
+    }
+
+    @Override
+    public ApiVersion apiVersion() {
+        return ApiVersion.V1;
```
andrewazores, line 86: This should be a v2 handler

andrewazores: Actually, it should be a beta API handler, but the implementation should extend `AbstractV2RequestHandler`

src/main/java/io/cryostat/net/web/http/api/v1/ProbeTemplateUploadHandler.java
```diff
@@ -0,0 +1,136 @@
...
+        super(auth);
+        this.notificationFactory = notificationFactory;
+        this.logger = logger;
+        this.probeTemplateService = probeTemplateService;
+        this.fs = fs;
+    }
+
+    @Override
+    public ApiVersion apiVersion() {
+        return ApiVersion.V1;
```
andrewazores, line 88: Same here re. v2/beta

src/main/java/io/cryostat/templates/TemplatesModule.java
```diff
@@ -55,4 +56,11 @@ static LocalStorageTemplateService provideLocalStorageTemplateService(
             FileSystem fs, Environment env) {
         return new LocalStorageTemplateService(fs, env);
     }
+
+    @Provides
+    @Singleton
+    static LocalProbeTemplateService provideLocalProbeTemplateService(
```
andrewazores, line 62: We should probably create a new module for probes rather than piggybacking them into event templates, but for now this will do.

run.sh
Josh-Matsuoka: Yeah it's just easier for temporary testing, I'll revert that before officially opening this PR.

andrewazores:
@Josh-Matsuoka I can close this as it's been superceded by #731, right?

Josh-Matsuoka:
@andrewazores yes, this can be closed.

Issue #691, opened 2021-09-28
Authorization check failure results in Internal Server Error


ebaron:
If a user doesn't have all permissions needed to make a particular HTTP request, the handler returns a 500 - Internal Server Error response instead of the expected 401 - Unauthorized.

Related to: #687

PR #692, opened 2021-09-28
fix(permissions): respond with 401 when authz check fails

ebaron:
This updates the Exception cause check to account for Exceptions that may be nested within two ExecutionExceptions. I didn't use ExceptionUtils.getRootCause because KubernetesClientException may have a cause of its own.

Fixes: #691


andrewazores:
Needs backport label?

andrewazores (approved):
Looks good. I'd like to come back around and refactor this at some point so that there are no Kubernetes details leaked into this abstract handler, but that's a task for later.

PR #696, opened 2021-09-29
chore(web-client): update to latest

andrewazores:
Updates the web-client to latest from cryostatio/cryostat-web#308 , and updates pom.xml with newer node/yarn versions to match.


jan-law (approved):
Looks good!

PR #703, opened 2021-10-06
feat(auth): add hook and handler to retrieve user info on POST auth

### LINKED ISSUE #704: Implement user info retrieval on auth
A new handler should be implemented which returns some basic information about the authenticated user after a successful `POST /api/v2/auth`. The information returned should include at least the successfully authenticated user's username, but may include further information as well.

In the case of Basic auth, the username returned must be the same as the Basic auth username provided.

In the case of Noop auth, the username should be null or the empty string to indicate no user.

In the case of Bearer auth, such as in the OpenShiftAuthManager, the username should be the one that corresponds to the supplied Bearer token according to the delegate auth server.

https://github.com/cryostatio/cryostat-web/pull/261#issuecomment-936975117

andrewazores:
Fixes #704

Example:

```
$ https POST https://cryostat-sample-myproject.apps-crc.testing/api/v2/auth Authorization:"Bearer $(oc whoami -t)"
HTTP/1.1 200 OK
Set-Cookie: ac6fe4a79b8104f3ee178c2547bbced6=57139fb6f3a455474c40d75a234afe3c; path=/; HttpOnly; Secure; SameSite=None
content-encoding: gzip
content-length: 107
set-cookie: ac6fe4a79b8104f3ee178c2547bbced6=57139fb6f3a455474c40d75a234afe3c; path=/; HttpOnly; Secure; SameSite=None

{
    "data": {
        "result": {
            "username": "developer"
        }
    },
    "meta": {
        "status": "OK",
        "type": "application/json"
    }
}
```

This is more-or-less a drop-in replacement for the existing `POST /api/v1/auth` handler, but it returns user information in a successful response rather than only sending a 200 status code. The user information is currently just the username. For noop auth it's the empty string. For basic auth, it decodes and extracts the `Basic b64(user:pass)` header and returns the `user` portion. And for Bearer auth (in OpenShift) it uses a TokenReview to ask the backing auth server for the user information for the given Bearer token.

The handler itself is also guarded by the standard API handler auth check, so unauthorized/invalid users will hit a 401 response and not see the userinfo body.

### DIFF
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -106,6 +106,8 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
                         new AuthorizationErrorException("Authentication Failed"));
             }
             return CompletableFuture.completedFuture(new UserInfo(status.getUser().getUsername()));
+        } catch (ExecutionException ee) {
+            return CompletableFuture.failedFuture(ee.getCause());
         } catch (Exception e) {
             return CompletableFuture.failedFuture(e);
         }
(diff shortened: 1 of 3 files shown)


andrewazores:
That last commit is another just-in-case error handling that I don't think would normally happen in practice. Before we get to the point of performing the handler's TokenReview to look up the username, the handler has already actually done a TokenReview to ensure that the supplied token is valid (as is done for all other auth-required handlers). So the only way this case would be tripped is if the token is revoked or the owning user is deleted in between the initial TokenReview checking that the token is valid and the second TokenReview retrieving the owner's name - or some transient auth server error.

Issue #666, opened 2021-08-27
HTTP 500: NPE during report generation


andrewazores:
Depends on: https://github.com/cryostatio/cryostat-core/issues/95

Cause: https://github.com/cryostatio/cryostat/issues/666#issuecomment-989105967

```
INFO: (10.0.2.100:40302): GET /api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:9091%2Fjmxrmi/recordings/workflow_itest 200 150ms
Aug 27, 2021 7:07:20 PM io.cryostat.core.log.Logger info
INFO: io.cryostat.net.reports.SubprocessReportGenerator starting
Aug 27, 2021 7:07:20 PM io.cryostat.core.log.Logger info
INFO: io.cryostat.net.reports.SubprocessReportGenerator adjusting subprocess OOM score
Aug 27, 2021 7:07:20 PM io.cryostat.core.log.Logger info
INFO: io.cryostat.net.reports.SubprocessReportGenerator processing report
java.lang.NullPointerException
	at org.openjdk.jmc.flightrecorder.parser.synthetic.SyntheticAttributeExtension$ModuleExportSink.addEvent(SyntheticAttributeExtension.java:206)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager$EventTypeEntry.readEvent(TypeManager.java:318)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager.readEvent(TypeManager.java:406)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:91)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:47)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
[...]

PR #731, opened 2021-10-19
feat(jmcagent): add API handlers for JMC agent plugin

### LINKED ISSUE #577: [Task] Implement HTTP API support for Agent Plugin functionality
Subtask of https://github.com/cryostatio/cryostat/issues/576

Goal: Implement the set of HTTP Handlers that will be needed for communication between the user/UI and the agent

- TargetProbeDeleteHandler to facilitate removing probes from a target

- TargetProbePostHandler to facilitate adding probes to a target

- TargetProbesGetHandler to facilitate querying the currently active probes in a target

- ProbeTemplateUploadHandler to facilitate uploading an xml probe template to the probes directory

- ProbeTemplateDeleteHandler to facilitate deleting a probe template

Subtasks: https://github.com/cryostatio/cryostat/issues/578 , https://github.com/cryostatio/cryostat/issues/579

### LINKED ISSUE #578: [Task] Implement target probe handlers
Subtask of https://github.com/cryostatio/cryostat/issues/577

**Subtask:** Implement TargetProbeDeleteHandler

TargetProbeDeleteHandler will facilitate removal of probes from targets and have the following parameters and response types

DELETE /api/v2/targets/:targetId/probes/:probeTemplateName

targetId - The location of the target JVM to connect to, in the form of a service:rmi:jmx:// JMX Service URL. Should use percent-encoding.

probeTemplateName - Name of the template for the probes to remove from the target JVM

Responses

200 - No body

401 - User authentication failed. The body is an error message. There will be an X-WWW-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

404 - The target could not be found. The body is an error message.

427 - JMX authentication failed. The body is an error message. There will be an X-JMX-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

**Subtask**: Implement TargetProbePostHandler

TargetProbePostHandler will facilitate adding probes to a target and will have the following form and response types:

POST /api/v2/targets/:targetId/probes/

targetId - The location of the target JVM to connect to, in the form of a service:rmi:jmx:// JMX Service URL. Should use percent-encoding.

Parameters

probeTemplate - name of the probe template to use

Responses

200 - No body

401 - User authentication failed. The body is an error message. There will be an X-WWW-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

404 - The target could not be found. The body is an error message.

427 - JMX authentication failed. The body is an error message. There will be an X-JMX-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

**Subtask:** Implement TargetProbesGetHandler

TargetProbesGetHandler will return the currently active probe template if one is active. It will have the following form and responses:

GET /api/v1/targets/:targetId/probes

targetId - The location of the target JVM to connect to, in the form of a service:rmi:jmx:// JMX Service URL. Should use percent-encoding.

response

200 - The body is a JSON array of probe template objects.

{"downloadUrl":"$DOWNLOAD_URL", "id":"$ID", "name":"$NAME", "state":"$STATE"}

401 - User authentication failed. The body is an error message. There will be an X-WWW-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

404 - The target could not be found. The body is an error message.

### LINKED ISSUE #579: [Task] Implement probe template handlers
Subtask of https://github.com/cryostatio/cryostat/issues/577

**Subtask:** Implement ProbeTemplateUploadHandler

ProbeTemplateUploadHandler will be responsible for uploading a new probe template to the probes directory. It will have the following form and responses:

Templates should be submitted in a form with the name template, e.g.

template=@foo.xml

POST /api/v1/probes

responses

200 - No body.

400 - The template being uploaded is an invalid event template. The body is an error message.

401 - User authentication failed. The body is an error message. There will be an X-WWW-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

**Subtask:** Implement ProbeTemplateDeleteHandler

ProbeTemplateDeleteHandler will be responsible for deleting a specified probe template that was previously uploaded. It will have the following form/responses:

DELETE /api/v1/probes/:templatename

response

200 - No body.

400 - The template could not be found. The body is an error message.

401 - User authentication failed. The body is an error message. There will be an X-WWW-Authenticate: $SCHEME header that indicates the authentication scheme that is used.

404 - The target could not be found. The body is an error message.

Josh-Matsuoka:
Fixes #577
Fixes #578
Fixes #579

This PR adds jmc agent support to cryostat through a set of API handlers for managing probe templates (xml files detailing the desired transformations) as well as interacting with the agent (adding, removing, retrieving currently active event probes). Accompanying tests have been added. The CI failures should be resolved once the core support (https://github.com/cryostatio/cryostat-core/pull/103) is integrated in core.


src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
```diff
@@ -48,4 +48,29 @@
...
+    @IntoSet
+    abstract RequestHandler bindTargetProbePostHandler(TargetProbePostHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindTargetProbeDeleteHandler(TargetProbePostHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindTargetProbesGetHandler(TargetProbePostHandler handler);
```
andrewazores, line 75: This is re-binding the `TargetProbePostHandler` into the `RequestHandler` set. The corrected method definition should look like:

```java
    @Binds
    @IntoSet
    abstract RequestHandler bindTargetProbesGetHandler(TargetProbesGetHandler handler);
```

Note that the type of the argument has been updated to reflect the handler that was intended to be bound.

src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
```diff
@@ -48,4 +48,29 @@
...
+    @IntoSet
+    abstract RequestHandler bindProbeTemplateDeleteHandler(ProbeTemplateDeleteHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindTargetProbePostHandler(TargetProbePostHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindTargetProbeDeleteHandler(TargetProbePostHandler handler);
```
andrewazores, line 71: This one is also re-binding a handler instead of binding the intended handler.

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateDeleteHandler.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public boolean isOrdered() {
+        return true;
+    }
+
+    @Override
+    public boolean requiresAuthentication() {
+        return false;
```
andrewazores, line 114: I think authentication should be required for this handler.

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateUploadHandler.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return EnumSet.of(ResourceAction.CREATE_PROBE_TEMPLATE);
+    }
+
+    @Override
+    public boolean requiresAuthentication() {
+        return false;
```
andrewazores, line 123: This one should also require authentication

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateUploadHandler.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Override
+    public IntermediateResponse handle(RequestParameters requestParams) throws Exception {
+        try {
+            for (FileUpload u : requestParams.getFileUploads()) {
+                String templateName = requestParams.getPathParams().get("probetemplateName");
```
andrewazores, line 130: The `probetemplateName` path param is effectively a constant for the duration of this HTTP request, but it's used here within the loop for the name of each template added corresponding to each file upload. So, if I make a POST request to this handler that contains 2 or more file uploads, I will get an exception thrown by the template service:

https://github.com/cryostatio/cryostat-core/blob/af317dfbcd84def96e0a15d18c74015e19265fbf/src/main/java/io/cryostat/core/agent/LocalProbeTemplateService.java#L92

This seems a bit non-obvious. The way this is typically done in other API handlers is to require the `FileUpload` to have a specific identifying name, like `template`: https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/web/http/api/v1/TemplatesPostHandler.java#L120

This way, the client should be aware that only one single file upload is expected by this handler. If they send it with the wrong name, or send other files along with other names, then those will be ignored and deleted.

src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
Josh-Matsuoka: This was a change that got lost since I worked on these changes in two separate branches, fixed it now.

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateDeleteHandler.java
Josh-Matsuoka: Agreed, I had it marked as such for ease of testing, I've gone ahead and changed it now

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateUploadHandler.java
Josh-Matsuoka: Makes sense, I've added a check for the upload name along the same lines as the linked example, thanks!

andrewazores:
Please rebase.

This will also need an update to `pom.xml` to use `cryostat-core` version `2.4.0` once it is tagged and released, as this will be the minor version update that includes the corresponding JMC Agent plugin classes. https://github.com/cryostatio/cryostat-core/issues/104

andrewazores:
Not sure what that itest CI failure is exactly - some brief googling suggests that exit status 124 indicates that the command timed out. @Josh-Matsuoka could you also rebase this again?

Josh-Matsuoka:
@andrewazores I've fixed the issue that was causing the itest container to fail to run, the current build failure is 

ReportJwtDownloadIT.testDownloadRecordingUsingJwt:72

which fails on a clean version of cryostat for me as well and seems to be unrelated to this PR.

andrewazores:
@Josh-Matsuoka please file a new issue for this test failure if there isn't one already, and include logs from both the frontend Maven test runner as well as the backend logs (`target/cryostat-itest-sometimestamp.log`, or the `print itest logs` task in the CI run)

andrewazores:
@Josh-Matsuoka rebase please

andrewazores:
> @Josh-Matsuoka please file a new issue for this test failure if there isn't one already, and include logs from both the frontend Maven test runner as well as the backend logs (`target/cryostat-itest-sometimestamp.log`, or the `print itest logs` task in the CI run)

https://github.com/cryostatio/cryostat/runs/4461031558?check_suite_focus=true#step:9:1392

Looks like this is #666, so it is indeed unrelated to this PR and has already been filed.

andrewazores:
CI is currently failing due to spotless formatting violation

run.sh
```diff
@@ -97,9 +97,12 @@ podman run \
     --mount type=bind,source="$(dirname $0)/conf",destination=/opt/cryostat.d/conf.d,relabel=shared \
     --mount type=bind,source="$(dirname $0)/templates",destination=/opt/cryostat.d/templates.d,relabel=shared \
     --mount type=bind,source="$(dirname $0)/truststore",destination=/truststore,relabel=shared \
+    --mount type=tmpfs,target=/opt/cryostat.d/probes.d \
     -e CRYOSTAT_PLATFORM=$CRYOSTAT_PLATFORM \
     -e CRYOSTAT_DISABLE_SSL=$CRYOSTAT_DISABLE_SSL \
     -e CRYOSTAT_DISABLE_JMX_AUTH=$CRYOSTAT_DISABLE_JMX_AUTH \
+    -e CRYOSTAT_DISABLE_SSL="false" \
+    -e CRYOSTAT_DISABLE_JMX_AUTH="false" \
```
andrewazores, line 105: Can these two lines be removed?

src/main/java/io/cryostat/templates/TemplatesModule.java
```diff
@@ -55,4 +56,17 @@ static LocalStorageTemplateService provideLocalStorageTemplateService(
...
+    @Provides
+    @Singleton
+    static LocalProbeTemplateService provideLocalProbeTemplateService(
+            FileSystem fs, Environment env) {
+        try {
+            return new LocalProbeTemplateService(fs, env);
+        } catch (Exception e) {
+            // Dagger doesn't like constructors that can throw exceptions, the probeTemplateService
+            // throws an exception if the sanity checks fail so we need to deal with it here
+            return null;
```
andrewazores, line 69: Dagger doesn't like when `@Provides` methods throw checked exceptions, but it'll let you throw an unchecked exception. This could be:

```java
try {
  return new LocalProbeTemplateService(fs, env);
} catch (Exception e) {
  throw new RuntimeException(e);
}
```

and it would compile and work as expected. This pattern is used in a few places already, for example:

https://github.com/cryostatio/cryostat/blob/059e6de8ac450a162c130285dc34c3375ec6dd66/src/main/java/io/cryostat/net/web/WebModule.java#L79

This way, wherever you are trying to inject the `LocalProbeTemplateService` doesn't have to do a null check. If we're at the point of instantiating the dependent class then the sanity checks must have already passed, and the dependency is non-null.

andrewazores (commented):
Looks good other than the sanity check RuntimeException thing.

run.sh
Josh-Matsuoka: yes, these were for local testing, I've removed them now

src/main/java/io/cryostat/templates/TemplatesModule.java
Josh-Matsuoka: Makes sense, I've cleaned this up, thanks

andrewazores (approved):

PR #742, opened 2021-10-29
fix(archiveupload): improve performance of archive uploads and validation

### LINKED ISSUE #740: POST /api/v1/recordings of uploaded file consumes more memory than expected
Description of issue: https://github.com/cryostatio/cryostat/issues/667#issuecomment-952303304

~~Suspected cause: https://github.com/cryostatio/cryostat/blob/b5c547a7163d4b90b0869c53e8349f8160076716/src/main/java/io/cryostat/net/web/http/api/v1/RecordingsPostHandler.java#L226~~

~~In the `RecordingsPostHandler`, we invoke:~~

```java
                        JfrLoaderToolkit.loadEvents(
                                new File(recordingFile));
```

~~in an attempt to validate that the uploaded file we have received is really a JFR binary file.~~

~~The implementation of this class for JMC7 is here: https://github.com/JDKMissionControl/jmc/blob/ffcd895e728b168f5b8d313890088c4ae3190d6d/core/org.openjdk.jmc.flightrecorder/src/main/java/org/openjdk/jmc/flightrecorder/JfrLoaderToolkit.java#L113~~

~~This ends up creating an InputStream out of the file, then passing that along to a parser and loader. The loader attempts to allocate a bunch of memory for a thread pool to use while parsing and loading chunk data ( https://github.com/JDKMissionControl/jmc/blob/develop/core/org.openjdk.jmc.flightrecorder/src/main/java/org/openjdk/jmc/flightrecorder/internal/FlightRecordingLoader.java#L201 ). This memory allocation may also not work as intended when running inside a container with memory limits different than the host operating system.~~

~~This `JfrLoaderToolkit` call should be removed and replaced with a simpler and lighter file validation. At the least, the parallelized parsing and loading done should be replaced by a simple serial parse, and each parsed chunk can be discarded after it is successfully parsed as a JFR chunk.~~

Current Vert.x-web version is 3.9.7, which uses Netty 4.1.60.Final.

https://github.com/netty/netty/pull/11188

Vert.x-web should be updated to a version that uses Netty 4.1.64.Final or later:

https://github.com/vert-x3/wiki/wiki/3.9.8-Release-Notes
https://netty.io/news/2021/05/19/4-1-65-Final.html

andrewazores:
Fixes #740

The primary fix in this PR is in the V1 `RecordingPostHandler` - the implementation of the function for validating JFR binary data has been changed from one that parses the data and constructs a list of events - potentially using a threadpool -  to one that simply serially parses the JFR binary chunk-by-chunk, thereby ensuring that that binary is parseable as JFR data and contains at least 1 chunk of data. This is simpler, faster, and less resource intensive, and provides strong enough validation of the uploaded data for this use case.

Along the way I discovered and fixed some other minor deficiencies in uploaded file handling, for `RecordingPostHandler` and others using the Vert.x `BodyHandler` class - namely that body handlers in several instances were accepting file uploads (the default) where only a standard HTTP form upload was expected, or that invalid file uploads would result in a rejected API request but the uploaded file would remain in Vert.x's temporary file-upload storage.

Finally, I also determined that with large (50MB+) JFR files, `POST`ing to the V1 `RecordingPostHandler` is very slow - before the Cryostat handler implementation even gets to run. The Vertx `BodyHandler` itself is actually quite slow here and consumes significant memory while processing the multipart form-data file upload:

~~1. As an initial experimental workaround I have implemented the Beta API version of the `RecordingPostHandler`, which accepts the JFR data directly as the POST request body without multipart file boundaries etc., and which directly handles the streaming data instead of delegating to a Vert.x body handler. This has some caveats, including that the `Expect: 100-continue` header must be sent by the requesting client - this is respected by `cURL` by default but is not implemented by Firefox nor Chromium, so it is only reliably useful from the CLI and not the web-client. However, the upload performance using this handler is superior - there is no multipart decode occurring and the upload is handled using streaming IO to the filesystem, and then proceeds to the chunk-by-chunk JFR binary validation.~~ Dropped from the PR
2. Diving deeper into the cause of the poor Vertx Body Handler performance I discovered there was a Netty performance regression in the multipart decoder (https://github.com/netty/netty/issues/10508). Upgrading the Vert.x version from 3.9.7 to 3.9.9 also bumps the Netty version, and the original V1 API upload handler using multipart form-data uploads performs nearly as well as the Beta API binary file upload handler after the upgrade, while still remaining compatible with web browsers (not requiring the `Expect:100-continue` header).

This second point is responsible for the vast majority of the performance fix in the PR. Simply using a newer version of Netty (4.1.60.Final -> 4.1.67.Final) makes the `BodyHandler` run at a much more reasonable speed and I haven't triggered the OOM kill with it, either. Combined with the slimmed down JFR parsing, the overall resource overhead is greatly reduced.

I have a ~150MB JFR file that I can share privately for reviewers to test this PR with.


andrewazores:
@ebaron I'm not too attached to the new Beta API handler I added since the Vert.x 3.9.9 version upgrade brings the performance of the original handler back up to an acceptable level. In my testing using curl, uploading to either the V1 or the Beta endpoints and using the `Expect:100-continue` header in both cases (to narrow down the timing to file I/O differences and not incur additional roundtrip latency), the 150MB JFR file takes ~1.8s either way. With the Vert.x 3.9.7 V1 API handler this time is more like 10 minutes. So, I am open to simply dropping the Beta handler from the PR - the Vert.x version upgrade transitively pulling in a Netty version upgrade seems to fix the vast majority of the performance issue. What little improvement remains is easily applied to the V1 handler as well by the JFR parsing change I made.

andrewazores:
> I think if the Beta handler is simply a workaround for the Netty performance issue, then it makes sense to drop it. If it has a purpose going forward then we can keep it.
> 
> I did notice that the linked issue [netty/netty#10508](https://github.com/netty/netty/issues/10508) was supposedly fixed in 4.1.60.Final, which is used by Vert.x 3.9.7. I wonder if there's another bug.

There was another Netty bug: https://github.com/netty/netty/pull/11188

I updated the linked issue to mention this but forgot to update this PR's description of the problem too, sorry. That Netty fix is in 4.1.65.Final, or Vert.x 3.9.8+.

andrewazores:
From what I can tell, it's 10508 that's responsible for the slow apparent upload speed I observed. That's not actually an upload speed limitation but Netty being slow to process data out of TCP buffers, causing the transfer rate to suffer. 11188 causes additional memory pressure when processing the file upload, so that's responsible for (or heavily contributes to) the OOM side of the original issue.

ebaron (commented):
Code changes look good. I just need to test it with that large JFR file.

andrewazores:
> Code changes look good. I just need to test it with that large JFR file.

I sent you an email with a link to it.

HTTP_API.md
```diff
@@ -1696,3 +1697,31 @@ The handler-specific descriptions below describe how each handler populates the
...
+## Beta API
+
+### Quick Reference
+
+| What you want to do                                                       | Which handler you should use                                                    |
+| ------------------------------------------------------------------------- | --------------------------------------------------------------------------------|
+| **Miscellaneous**                                                         |                                                                                 |
+| View targets in overall deployment environment                            | [`DiscoveryGetHandler`](#DiscoveryGetHandler)                                   |
+| **Recordings in archive**                                                 |                                                                                 |
+| Upload a recording to archive                                             | [`RecordingsPostHandler`](#RecordingsPostHandler-1)                               |
```
ebaron, line 1710: Should this be removed?

andrewazores: Ah yes, thanks.

ebaron (approved):
Tested out in CRC with the provided JFR file. Uploaded quickly.

PR #743, opened 2021-11-02
fix(snapshot): Cleanly handle empty snapshots

### LINKED ISSUE #715: Cleanly handle Snapshots taken when there is no source data
It is currently possible to take a Snapshot recording even when the target VM has no running recordings, or has only empty recordings, etc. This will reliably create recordings with 0 length.

This can be easily reproduced:
```bash
$ sh smoketest.sh
$ https POST :8181/api/v1/targets/localhost:9093/snapshot
```

Or:
1. Open the web-client
2. Select a target application with no active recordings
3. Navigate to Recordings
4. Click Create
5. Switch to Snapshot tab
6. Click Create

These 0-length snapshots cause various problems for other features:

Report Generation:
image

View in Grafana / Download Recording / Download Report:
image

Attempting to archive such a recording appears to fail silently in the web-client UI, but the backend logs show this:

```
Oct 12, 2021 7:01:26 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:53814): POST /api/v1/auth 200 0ms
Oct 12, 2021 7:01:28 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:53814): POST /api/v1/auth 200 0ms
Oct 12, 2021 7:01:28 PM io.cryostat.core.log.Logger info
INFO: Creating connection for service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi
Oct 12, 2021 7:01:28 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:53814): PATCH /api/v1/targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcryostat%3A9093%2Fjmxrmi/recordings/snapshot-1 204 79ms
Oct 12, 2021 7:01:28 PM io.cryostat.core.log.Logger error
SEVERE: HTTP 500: io.cryostat.recordings.EmptyRecordingException: java.io.IOException: java.io.IOException: No recording data available
Caused by: java.util.concurrent.ExecutionException: io.cryostat.recordings.EmptyRecordingException: java.io.IOException: java.io.IOException: No recording data available
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at io.cryostat.net.web.http.api.v1.TargetRecordingPatchSave.handle(TargetRecordingPatchSave.java:67)
	at io.cryostat.net.web.http.api.v1.TargetRecordingPatchHandler.handleAuthenticated(TargetRecordingPatchHandler.java:108)
Caused by: io.cryostat.recordings.EmptyRecordingException: java.io.IOException: java.io.IOException: No recording data available
	at io.cryostat.recordings.RecordingArchiveHelper.writeRecordingToDestination(RecordingArchiveHelper.java:361)
	at io.cryostat.recordings.RecordingArchiveHelper.lambda$saveRecording$0(RecordingArchiveHelper.java:138)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:136)
	at io.cryostat.recordings.RecordingArchiveHelper.saveRecording(RecordingArchiveHelper.java:131)
	... 11 more
Caused by: java.io.IOException: java.io.IOException: No recording data available
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.invokeOperation(JfrRecordingInputStreamV2.java:206)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.openStream(JfrRecordingInputStreamV2.java:181)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.openStream(JfrRecordingInputStreamV2.java:167)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.readStreamIdentifier(JfrRecordingInputStreamV2.java:120)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.fill(JfrRecordingInputStreamV2.java:104)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.read(JfrRecordingInputStreamV2.java:94)
	at java.base/java.io.InputStream.read(InputStream.java:271)
	at java.base/java.io.BufferedInputStream.fill(BufferedInputStream.java:252)
	at java.base/java.io.BufferedInputStream.read(BufferedInputStream.java:271)
	at io.cryostat.recordings.RecordingArchiveHelper.writeRecordingToDestination(RecordingArchiveHelper.java:356)
	... 14 more
Caused by: java.io.IOException: java.io.IOException: No recording data available
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderCommunicationHelperV2.invokeOperation(FlightRecorderCommunicationHelperV2.java:84)
	at org.openjdk.jmc.rjmx.services.jfr.internal.JfrRecordingInputStreamV2.invokeOperation(JfrRecordingInputStreamV2.java:197)
	... 23 more
Caused by: javax.management.MBeanException: java.io.IOException: No recording data available
	at java.management/com.sun.jmx.mbeanserver.MBeanIntrospector.unwrapInvocationTargetException(MBeanIntrospector.java:313)
	at java.management/com.sun.jmx.mbeanserver.MBeanIntrospector.invokeM(MBeanIntrospector.java:239)
	at java.management/com.sun.jmx.mbeanserver.PerInterface.invoke(PerInterface.java:138)
	at java.management/com.sun.jmx.mbeanserver.MBeanSupport.invoke(MBeanSupport.java:252)
	at java.management/javax.management.StandardMBean.invoke(StandardMBean.java:405)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.invoke(DefaultMBeanServerInterceptor.java:809)
	at java.management/com.sun.jmx.mbeanserver.JmxMBeanServer.invoke(JmxMBeanServer.java:801)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl.doOperation(RMIConnectionImpl.java:1466)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl$PrivilegedOperation.run(RMIConnectionImpl.java:1307)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl.doPrivilegedOperation(RMIConnectionImpl.java:1399)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl.invoke(RMIConnectionImpl.java:827)
	at java.base/jdk.internal.reflect.GeneratedMethodAccessor18.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.rmi/sun.rmi.server.UnicastServerRef.dispatch(UnicastServerRef.java:359)
	at java.rmi/sun.rmi.transport.Transport$1.run(Transport.java:200)
	at java.rmi/sun.rmi.transport.Transport$1.run(Transport.java:197)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.rmi/sun.rmi.transport.Transport.serviceCall(Transport.java:196)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport.handleMessages(TCPTransport.java:562)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run0(TCPTransport.java:796)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.lambda$run$0(TCPTransport.java:677)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run(TCPTransport.java:676)
	at java.rmi/sun.rmi.transport.StreamRemoteCall.exceptionReceivedFromServer(StreamRemoteCall.java:303)
	at java.rmi/sun.rmi.transport.StreamRemoteCall.executeCall(StreamRemoteCall.java:279)
	at java.rmi/sun.rmi.server.UnicastRef.invoke(UnicastRef.java:164)
	at jdk.remoteref/jdk.jmx.remote.internal.rmi.PRef.invoke(Unknown Source)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl_Stub.invoke(Unknown Source)
	at java.management.rmi/javax.management.remote.rmi.RMIConnector$RemoteMBeanServerConnection.invoke(RMIConnector.java:1021)
	at org.openjdk.jmc.rjmx.internal.MCMBeanServerConnection.invoke(MCMBeanServerConnection.java:259)
	at org.openjdk.jmc.rjmx.ConnectionToolkit.invokeOperation(ConnectionToolkit.java:196)
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderCommunicationHelperV2.invokeJfrOperation(FlightRecorderCommunicationHelperV2.java:91)
	at org.openjdk.jmc.rjmx.services.jfr.internal.FlightRecorderCommunicationHelperV2.invokeOperation(FlightRecorderCommunicationHelperV2.java:82)
	... 24 more
Caused by: java.io.IOException: No recording data available
	at jdk.management.jfr/jdk.management.jfr.FlightRecorderMXBeanImpl.openStream(FlightRecorderMXBeanImpl.java:152)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at sun.reflect.misc.Trampoline.invoke(MethodUtil.java:71)
	at jdk.internal.reflect.GeneratedMethodAccessor10.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.base/sun.reflect.misc.MethodUtil.invoke(MethodUtil.java:260)
	at java.management/com.sun.jmx.mbeanserver.ConvertingMethod.invokeWithOpenReturn(ConvertingMethod.java:193)
	at java.management/com.sun.jmx.mbeanserver.ConvertingMethod.invokeWithOpenReturn(ConvertingMethod.java:175)
	at java.management/com.sun.jmx.mbeanserver.MXBeanIntrospector.invokeM2(MXBeanIntrospector.java:117)
	at java.management/com.sun.jmx.mbeanserver.MXBeanIntrospector.invokeM2(MXBeanIntrospector.java:54)
	at java.management/com.sun.jmx.mbeanserver.MBeanIntrospector.invokeM(MBeanIntrospector.java:237)
	at java.management/com.sun.jmx.mbeanserver.PerInterface.invoke(PerInterface.java:138)
	at java.management/com.sun.jmx.mbeanserver.MBeanSupport.invoke(MBeanSupport.java:252)
	at java.management/javax.management.StandardMBean.invoke(StandardMBean.java:405)
	at java.management/com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.invoke(DefaultMBeanServerInterceptor.java:809)
	at java.management/com.sun.jmx.mbeanserver.JmxMBeanServer.invoke(JmxMBeanServer.java:801)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl.doOperation(RMIConnectionImpl.java:1466)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl$PrivilegedOperation.run(RMIConnectionImpl.java:1307)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl.doPrivilegedOperation(RMIConnectionImpl.java:1399)
	at java.management.rmi/javax.management.remote.rmi.RMIConnectionImpl.invoke(RMIConnectionImpl.java:827)
	at java.base/jdk.internal.reflect.GeneratedMethodAccessor18.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at java.rmi/sun.rmi.server.UnicastServerRef.dispatch(UnicastServerRef.java:359)
	at java.rmi/sun.rmi.transport.Transport$1.run(Transport.java:200)
	at java.rmi/sun.rmi.transport.Transport$1.run(Transport.java:197)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.rmi/sun.rmi.transport.Transport.serviceCall(Transport.java:196)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport.handleMessages(TCPTransport.java:562)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run0(TCPTransport.java:796)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.lambda$run$0(TCPTransport.java:677)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run(TCPTransport.java:676)
```

When a Snapshot is created, Cryostat should check that the resultant recording is non-empty. If it is empty then the snapshot should be deleted and an appropriate HTTP response sent to the user, other than a 200, so that there is some indication that the request was received and processed but there is no result (HTTP 204 sounds appropriate, but this implies there is no response body. HTTP 202?)

hareetd:
Fixes #715 

Depends on https://github.com/cryostatio/cryostat-web/pull/346

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -120,7 +120,7 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
                 throw new HttpStatusException(
                     500,
                     String.format(
-                            "An error occured during the creation of snapshot %s",
+                            "An error occured during the creation of snapshot %s.",
                             snapshotName));
             }
             throw e;
@@ -130,7 +130,7 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
             ctx.response().setStatusCode(202);
             ctx.response()
                     .setStatusMessage(
-                            String.format("Snapshot %s failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from", snapshotName));
+                            String.format("Snapshot %s failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from.", snapshotName));
             ctx.response().end();
         }
 
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -160,5 +160,5 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
             return false;
         }
         return true;
-     }
+    }
 }
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -107,7 +107,7 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
 
     @Override
     public void handleAuthenticated(RoutingContext ctx) throws Exception {
-        final ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
+        ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
         String result =
                 targetConnectionManager.executeConnectedTask(
                         connectionDescriptor,
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -119,10 +119,10 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
                             String.format(
                                     "Snapshot %s failed to create: The resultant recording was unreadable for some reason, likely due to a lack of Active, non-Snapshot source recordings to take event data from.",
                                     snapshotName));
+        } else {
+            ctx.response().setStatusCode(200);
+            ctx.response().end(snapshotName);
         }
-
-        ctx.response().setStatusCode(200);
-        ctx.response().end(snapshotName);
     }
 
     private void handleExecutionException(ExecutionException e) throws ExecutionException {
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -37,17 +37,10 @@
  */
 package io.cryostat.net.web.http.api.v1;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.PushbackInputStream;
 import java.util.EnumSet;
-import java.util.Optional;
 import java.util.Set;
 
 import javax.inject.Inject;
-
-import org.openjdk.jmc.flightrecorder.configuration.recording.RecordingOptionsBuilder;
-import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
 
 import io.cryostat.net.AuthManager;
 import io.cryostat.net.ConnectionDescriptor;
@@ -55,10 +48,10 @@ import io.cryostat.net.TargetConnectionManager;
 import io.cryostat.net.security.ResourceAction;
 import io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler;
 import io.cryostat.net.web.http.api.ApiVersion;
-import io.cryostat.recordings.RecordingOptionsBuilderFactory;
 import io.cryostat.recordings.RecordingTargetHelper;
 import io.cryostat.recordings.RecordingTargetHelper.SnapshotCreationException;
 import io.cryostat.recordings.RecordingTargetHelper.SnapshotMinimalDescriptor;
+
 import io.vertx.core.http.HttpMethod;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
@@ -66,18 +59,15 @@ import io.vertx.ext.web.handler.impl.HttpStatusException;
 class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
 
     private final TargetConnectionManager targetConnectionManager;
-    private final RecordingOptionsBuilderFactory recordingOptionsBuilderFactory;
     private final RecordingTargetHelper recordingTargetHelper;
 
     @Inject
     TargetSnapshotPostHandler(
             AuthManager auth,
             TargetConnectionManager targetConnectionManager,
-            RecordingOptionsBuilderFactory recordingOptionsBuilderFactory,
             RecordingTargetHelper recordingTargetHelper) {
         super(auth);
         this.targetConnectionManager = targetConnectionManager;
-        this.recordingOptionsBuilderFactory = recordingOptionsBuilderFactory;
         this.recordingTargetHelper = recordingTargetHelper;
     }
 
@@ -113,25 +103,28 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
                 targetConnectionManager.executeConnectedTask(
                         connectionDescriptor,
                         connection -> {
-                            SnapshotMinimalDescriptor snapshot = recordingTargetHelper.createSnapshot(connection).get();
+                            SnapshotMinimalDescriptor snapshot =
+                                    recordingTargetHelper.createSnapshot(connection).get();
                             return snapshot.getName();
                         });
-    
+
         boolean verificationSuccessful = false;
         try {
-            verificationSuccessful = recordingTargetHelper.verifySnapshot(connectionDescriptor, snapshotName).get();
+            verificationSuccessful =
+                    recordingTargetHelper.verifySnapshot(connectionDescriptor, snapshotName).get();
         } catch (SnapshotCreationException e) {
             throw new HttpStatusException(
                     500,
                     String.format(
-                            "Successful creation verification of snapshot %s failed", snapshotName));
+                            "An error occured during the creation of snapshot %s",
+                            snapshotName));
         }
 
         if (!verificationSuccessful) {
             ctx.response().setStatusCode(202);
             ctx.response()
                     .setStatusMessage(
-                            "Snapshot failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from");
+                            String.format("Snapshot %s failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from", snapshotName));
             ctx.response().end();
         }
 
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -43,15 +43,14 @@ import java.util.concurrent.ExecutionException;
 
 import javax.inject.Inject;
 
+import io.cryostat.jmc.serialization.HyperlinkedSerializableRecordingDescriptor;
 import io.cryostat.net.AuthManager;
 import io.cryostat.net.ConnectionDescriptor;
-import io.cryostat.net.TargetConnectionManager;
 import io.cryostat.net.security.ResourceAction;
 import io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler;
 import io.cryostat.net.web.http.api.ApiVersion;
 import io.cryostat.recordings.RecordingTargetHelper;
 import io.cryostat.recordings.RecordingTargetHelper.SnapshotCreationException;
-import io.cryostat.recordings.RecordingTargetHelper.SnapshotMinimalDescriptor;
 
 import io.vertx.core.http.HttpMethod;
 import io.vertx.ext.web.RoutingContext;
@@ -60,16 +59,11 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 
 class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
 
-    private final TargetConnectionManager targetConnectionManager;
     private final RecordingTargetHelper recordingTargetHelper;
 
     @Inject
-    TargetSnapshotPostHandler(
-            AuthManager auth,
-            TargetConnectionManager targetConnectionManager,
-            RecordingTargetHelper recordingTargetHelper) {
+    TargetSnapshotPostHandler(AuthManager auth, RecordingTargetHelper recordingTargetHelper) {
         super(auth);
-        this.targetConnectionManager = targetConnectionManager;
         this.recordingTargetHelper = recordingTargetHelper;
     }
 
@@ -101,28 +95,21 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
     @Override
     public void handleAuthenticated(RoutingContext ctx) throws Exception {
         ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
-        String snapshotName =
-                targetConnectionManager.executeConnectedTask(
-                        connectionDescriptor,
-                        connection -> {
-                            SnapshotMinimalDescriptor snapshot =
-                                    recordingTargetHelper.createSnapshot(connection).get();
-                            return snapshot.getName();
-                        });
+
+        HyperlinkedSerializableRecordingDescriptor snapshotDescriptor = null;
+        try {
+            snapshotDescriptor = recordingTargetHelper.createSnapshot(connectionDescriptor).get();
+        } catch (ExecutionException e) {
+            handleExecutionException(e);
+        }
+        String snapshotName = snapshotDescriptor.getName();
 
         boolean verificationSuccessful = false;
         try {
             verificationSuccessful =
                     recordingTargetHelper.verifySnapshot(connectionDescriptor, snapshotName).get();
         } catch (ExecutionException e) {
-            if (ExceptionUtils.getRootCause(e) instanceof SnapshotCreationException) {
-                throw new HttpStatusException(
-                        500,
-                        String.format(
-                                "An error occured during the creation of snapshot %s.",
-                                snapshotName));
-            }
-            throw e;
+            handleExecutionException(e);
         }
 
         if (!verificationSuccessful) {
@@ -130,11 +117,19 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
             ctx.response()
                     .end(
                             String.format(
-                                    "Snapshot %s failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from.",
+                                    "Snapshot %s failed to create: The resultant recording was unreadable for some reason, likely due to a lack of Active, non-Snapshot source recordings to take event data from.",
                                     snapshotName));
         }
 
         ctx.response().setStatusCode(200);
         ctx.response().end(snapshotName);
+    }
+
+    private void handleExecutionException(ExecutionException e) throws ExecutionException {
+        Throwable cause = ExceptionUtils.getRootCause(e);
+        if (cause instanceof SnapshotCreationException) {
+            throw new HttpStatusException(500, cause.getMessage());
+        }
+        throw e;
     }
 }
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -139,7 +139,7 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
                     500,
                     String.format(
                             "Successful creation verification of snapshot %s failed", result));
-        } else if (snapshotIsEmpty(snapshotOptional.get())) {
+        } else if (!snapshotIsReadable(snapshotOptional.get())) {
             recordingTargetHelper.deleteRecording(connectionDescriptor, result).get();
             ctx.response().setStatusCode(202);
             ctx.response()
@@ -152,19 +152,14 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
         }
     }
 
-    private boolean snapshotIsEmpty(InputStream snapshot) throws IOException {
-        PushbackInputStream pushbackSnapshot = new PushbackInputStream(snapshot);
+    private boolean snapshotIsReadable(InputStream snapshot) throws IOException {
         try {
+            PushbackInputStream pushbackSnapshot = new PushbackInputStream(snapshot);
             int b = pushbackSnapshot.read();
-            // If this point is reached (i.e. no IOException was thrown) the stream must
-            // be non-empty so push back the last read byte
             pushbackSnapshot.unread(b);
-            return false;
+            return true;
         } catch (IOException e) {
-            if (e.getMessage().equals("java.io.IOException: No recording data available")) {
-                return true;
-            }
-            throw e;
+            return false;
         }
     }
 }
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
@@ -156,8 +156,7 @@ class TargetSnapshotPostHandler extends AbstractAuthenticatedRequestHandler {
         PushbackInputStream pushbackSnapshot = new PushbackInputStream(snapshot);
         try {
             int b = pushbackSnapshot.read();
-
-            // If this point is reached (i.e. no IOException is thrown) the stream must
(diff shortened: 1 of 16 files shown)


src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
```diff
@@ -120,7 +131,21 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
...
-        ctx.response().setStatusCode(200);
-        ctx.response().end(result);
+        
+        Optional<InputStream> snapshotOptional = recordingTargetHelper.getRecording(connectionDescriptor, result);
+        if (snapshotOptional.isEmpty()) {
+            throw new HttpStatusException(500, String.format("Successful upload verification of %s failed", result));
+        } else if (snapshotIsEmpty(snapshotOptional.get())) {
+            recordingTargetHelper.deleteRecording(connectionDescriptor, result);
+            ctx.response().setStatusCode(202);
+            ctx.response().setStatusMessage("Snapshot failed to create: Cryostat is not aware of any Active, non-Snapshot source recordings to take event data from");
```
andrewazores, line 141: Looks like a `ctx.response().end()` is missing here

andrewazores:
Nice work here so far. This is exactly how I would have done it.

hareetd:
Thanks, once tests are updated I'll open it up for a full review.

hareetd:
This PR should be ready to go with just a couple of tweaks.

The first issue is that when an empty Snapshot is created and subsequently deleted, the user shouldn't receive any recording creation or deletion notifications. I was able to separate the recording deletion notification process from the empty Snapshot handling but when it comes to the creation notification it looks like I'll have to change the API handling for Snapshot creation in `Api.service.tsx` on the frontend.

The second issue is that the -web PR I submitted to display a warning for empty Snapshots (https://github.com/cryostatio/cryostat-web/pull/329) is outdated now. The warning modal should instead say something along the lines of "Snapshot failed to create....". 

I'll open a -web issue to take care of those. Feel free to review this PR whenever you have time as the only remaining change will be to bump the web client version once the -web issue is resolved.

src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
```diff
@@ -120,7 +131,40 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
...
+                    .setStatusMessage(
+                            "Snapshot failed to create: Cryostat is not aware of any Active, non-Snapshot source recordings to take event data from");
+            ctx.response().end();
+        } else {
+            ctx.response().setStatusCode(200);
+            ctx.response().end(result);
+        }
+    }
+
+    private boolean snapshotIsEmpty(InputStream snapshot) throws IOException {
```
andrewazores, line 155: This looks handy enough that it's worth extracting out somewhere.

src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
```diff
@@ -120,7 +131,40 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
...
+        PushbackInputStream pushbackSnapshot = new PushbackInputStream(snapshot);
+        try {
+            int b = pushbackSnapshot.read();
+            // If this point is reached (i.e. no IOException was thrown) the stream must
+            // be non-empty so push back the last read byte
+            pushbackSnapshot.unread(b);
+            return false;
+        } catch (IOException e) {
+            if (e.getMessage().equals("java.io.IOException: No recording data available")) {
```
andrewazores, line 164: Relying on the string contents here looks fragile. Is there any other way to perform this check?

src/main/java/io/cryostat/net/web/http/api/v2/TargetSnapshotPostHandler.java
```diff
@@ -145,10 +156,45 @@ public boolean isAsync() {
...
+        Optional<InputStream> snapshotOptional =
+                recordingTargetHelper.getRecording(connectionDescriptor, snapshotName).get();
+        if (snapshotOptional.isEmpty()) {
+            throw new ApiException(
+                    500,
+                    String.format(
+                            "Successful creation verification of snapshot %s failed",
+                            snapshotName));
+        } else if (snapshotIsEmpty(snapshotOptional.get())) {
+            recordingTargetHelper.deleteRecording(connectionDescriptor, snapshotName).get();
```
andrewazores, line 170: Seeing this duplicated logic between the v1 and v2 handlers, I'm thinking there is now enough repeated code here that it's worth extracting and reusing. The `executeConnectedTask` block, `optional.isEmpty()` check, and `snapshotIsEmpty()` check together look like they belong together and could be extracted into some method like `RecordingTargetHelper#createSnapshot()`. Both API handlers (and the `RuleProcessor`) could then just call that method and always implement the same behaviour.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -156,6 +168,65 @@ public IRecordingDescriptor startRecording(
                 .findFirst();
     }
 
+    public Future<Optional<InputStream>> getRecording(
```
andrewazores, line 171: Please add a comment here with some link to `TargetConnectionManager#markConnectionInUse`. The returned `InputStream`, if any, is only readable while the remote connection remains open. Callers need to be aware of this so that they don't try to read from the stream after the connection has been closed due to inactivity, or so that the TCM doesn't close the stream while it's still being consumed without the TCM realizing it.

src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
hareetd: From my initial debugging, the message string was what alerted me to the fact that an `IOException` is thrown when attempting to read empty streams. I can take another look and see if there are any other identifying fields in the `IOException` object.

hareetd: Just realized, by "this check" do you mean checking if the `IOException` is due to an empty stream or if there exists another method for checking if a stream is empty?

andrewazores: It looks like this is roughly the code that throws that exception:

https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/jdk.management.jfr/share/classes/jdk/management/jfr/FlightRecorderMXBeanImpl.java#L150

This may change with different target JVM versions that we're communicating with, which is why I say that the check seems fragile. I don't see any immediately obvious other way to distinguish this kind of IOException from any others, unfortunately. Perhaps this method should be renamed from `snapshotIsEmpty` to `snapshotIsReadable` or something, and if any IOException is thrown when trying to read the first byte for any reason, then it's `false`. If the snapshot is unreadable immediately after creating it then perhaps the most likely reason is that it's simply empty, but it may be ambiguous. In any case I suppose we want to delete the unreadable snapshot.

hareetd: Ah, that makes sense, right. I agree with the `snapshotIsReadable` approach. I can change the HTTP API to say that the 202 response is due to "an unreadable snapshot, likely caused by a lack of Active source recordings..." (something along those lines). I'll do that for front-end notification as well.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -207,4 +333,48 @@ public IRecordingDescriptor startRecording(
...
+            ConnectionDescriptor connectionDescriptor, InputStream snapshot) throws IOException {
+        if (!targetConnectionManager.markConnectionInUse(connectionDescriptor)) {
+            throw new IOException(
+                    "Target connection unexpectedly closed while streaming recording");
+        }
+
+        try {
+            PushbackInputStream pushbackSnapshot = new PushbackInputStream(snapshot);
+            int b = pushbackSnapshot.read();
+            pushbackSnapshot.unread(b);
```
andrewazores, line 347: Now that this has been refactored and extracted in this way, I wonder if the `PushbackInputStream` is actually needed at all. This `snapshotIsReadable` method is only called in a branch of `verifySnapshot` that occurs after `verifySnapshot` has also called through to `this.getRecording` on its own - which means it's potentially opening a new target connection (probably just reusing a cached one), but it is definitely opening a new underlying stream of the snapshot recording data in the target JVM. The pushback/unread functionality would only seem to be needed if we were trying to verify that a provided InputStream was non-empty without opening a whole new stream to the same underlying remote recording. If we're opening a new stream to perform the verification then I think we can just use the InputStream as-is and try to read from it (https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read()) - if the return value is `-1` then it was empty, or if an exception was thrown then the connection was closed or the network failed etc.

Please check and test if replacing the PushbackInputStream with simply reading the first byte out of the newly-created stream works, and if so, let's go with that simpler version for this PR and leave the rest of this refactored structure the same. At this point it's just an internal detail of the `snapshotIsReadable` method - everything else seems to be put together well.

Where the PushbackInputStream *would* be extremely useful is in ex. `TargetRecordingGetHandler`. Then the handler could return `404` if there is no such requested recording, or ex. `204` if the recording (including a snapshot recording) could be located in the target but the stream contained 0 bytes. This would definitely be an enhancement for a future PR.

src/main/java/io/cryostat/net/web/http/api/v2/TargetSnapshotPostHandler.java
```diff
@@ -118,33 +122,46 @@ public boolean isAsync() {
...
+                                snapshotName));
+            }
+            throw e;
+        }
+
+        if (!verificationSuccessful) {
+            return new IntermediateResponse<HyperlinkedSerializableRecordingDescriptor>()
+                    .statusCode(202)
+                    .statusMessage(
+                            "Snapshot failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from.")
```
andrewazores, line 161: This is a really, really long status message. I'd rather that be part of the body instead.

src/main/java/io/cryostat/net/web/http/api/v1/TargetSnapshotPostHandler.java
```diff
@@ -97,30 +100,42 @@ public boolean isAsync() {
...
+            }
+            throw e;
+        }
+
+        if (!verificationSuccessful) {
+            ctx.response().setStatusCode(202);
+            ctx.response()
+                    .setStatusMessage(
+                            String.format(
+                                    "Snapshot %s failed to create: The resultant recording was unreadable for some reason, possibly due to a lack of Active, non-Snapshot source recordings to take event data from.",
```
andrewazores, line 133: Same here re. status message vs response body

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
hareetd: That all makes sense, I agree. I'll go ahead and try out `InputStream` instead; I'm fairly certain it'll work out. 

The only thing I'd like to add is that when I originally used `InputStream` instead of `PushbackInputStream`, I was checking to see if `read()` returns `-1` (since logically, an empty recording means the end of the stream will be reached immediately) but what actually happens is that an `IOException` is thrown. Therefore, the code will be the same except for the replacement of the `PushbackInputStream`.

andrewazores: Okay, makes sense. That might just be something specific to the `InputStream` implementation we're using under the covers (I think it's this: https://github.com/openjdk/jmc/blob/master/application/org.openjdk.jmc.rjmx.services.jfr/src/main/java/org/openjdk/jmc/rjmx/services/jfr/internal/JfrRecordingInputStreamV2.java). It seems like it "should" be returning `-1`, but there are cases where it throws `IOException: stream closed`, so it seems we fall through that for whatever reason.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -342,9 +341,7 @@ private boolean snapshotIsReadable(
         }
 
         try {
-            PushbackInputStream pushbackSnapshot = new PushbackInputStream(snapshot);
-            int b = pushbackSnapshot.read();
-            pushbackSnapshot.unread(b);
+            snapshot.read();
```
andrewazores, line 344: How about `return snapshot.read() != -1`? Just in case the stream doesn't throw - cover our bases.

andrewazores: Also even though this method is private, I think it's worth leaving a comment here (maybe a method-level javadoc comment) explaining that this will consume the first byte of the stream that it is verifying, so it should only be used when the stream passed to it is one that will not be used for any other later purpose.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -156,6 +172,115 @@ public IRecordingDescriptor startRecording(
...
+
+    public Future<Boolean> verifySnapshot(
+            ConnectionDescriptor connectionDescriptor, String snapshotName) {
+        CompletableFuture<Boolean> future = new CompletableFuture<>();
+        try {
+            Optional<InputStream> snapshotOptional =
+                    this.getRecording(connectionDescriptor, snapshotName).get();
+            if (snapshotOptional.isEmpty()) {
+                throw new SnapshotCreationException(snapshotName);
+            } else if (!snapshotIsReadable(connectionDescriptor, snapshotOptional.get())) {
```
andrewazores, line 272: The result of `snapshotOptional.get()` is an `InputStream` that we acquired within this method, so we should also be careful to close this stream within this method. It looks a little hairy, but maybe that looks something like:


```java
Optional<InputStream> opt = this.getRecording().get();
if (opt.isEmpty() {
  throw
} else {
  try (InputStream stream = opt.get()) {
    if (!snapshotIsReadable(connDesc, stream)) {
      this.deleteRecording(connDesc, snapshotName).get();
      future.complete(false);
    } else {
      future.complete(true);
    }
  }
}
```

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -156,6 +172,115 @@ public IRecordingDescriptor startRecording(
...
+                        return null;
+                    });
+            future.complete(null);
+        } catch (Exception e) {
+            future.completeExceptionally(e);
+        }
+        return future;
+    }
+
+    public Future<SnapshotMinimalDescriptor> createSnapshot(JFRConnection connection) {
```
andrewazores, line 241: For consistency with the other methods like `startRecording`, `getRecording`, `deleteRecording`, should this method take a `ConnectionDescriptor` parameter instead of a `JFRConnection`? Then the current method body would basically just get wrapped in a lambda and passed to `targetConnectionManager.executeConnectedTask` like the other methods do.

hareetd: The reason I passed in a `JFRConnection` instead is because that the `v1` and `v2` `TargetSnapshotPostHandler` implementations of the `targetConnectionManager.executeConnectedTask` both return different objects (`String` vs. `SnapshotDescriptor`, respectively) with the `v2` version also making using of the `connection` lambda parameter inside the returned `SnapshotDescriptor` object. That's why I decided to just extract the maximally mutual code between the two.

It's a bit hard to explain but if you look at both versions maybe you have a better idea of how I could go about this with a `ConnectionDescriptor` parameter instead?

src/main/java/io/cryostat/net/web/http/api/v2/TargetSnapshotPostHandler.java
hareetd: Since the body has to be a `HyperlinkedSerializableRecordingDescriptor`, what's the best way to go about returning that `String`? 

I could add a `String status` field to the object, which can be populated with `Healthy` for successful recordings and a message like the one above for failed recordings, with all other fields `null` in the latter case. Thoughts?

andrewazores: Actually, it will already be included as part of the V2 response body in the `meta` field. Now that I'm looking at that again (`AbstractV2RequestHandler#writeResponse`), there is no way to have the `.meta.status` of the body differ from the HTTP status message. So this is fine as-is.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
andrewazores: I think it would be a slightly heavier refactor, but the `SnapshotMinimalDescriptor` seems like it can be removed, too. If you opened a target connection in this method, created the snapshot, renamed it, and then queried again for recordings matching the new rename and returned that `IRecordingDescriptor`, would that fit the needs of both V1 and V2 handlers? V1 would just return the name from that descriptor since that is already the rename, and V2 would use that same name plus the rest of the whole descriptor (no more need for V2's own `SnapshotDescriptor` that overrides the name field).

andrewazores:
Please add a commit updating the web-client submodule as well now that the relevant frontend feature PR has been merged.

andrewazores (approved):
Great work here, thanks

PR #748, opened 2021-11-04
feat(auth): Redirect to OpenShift login on empty auth request

### LINKED ISSUE #717: OAuth Redirect Authentication flow
https://vertx.io/docs/vertx-auth-oauth2/java/

Vert.x supports OAuth redirect authentication flow. Cryostat could support this when using the OpenShiftAuthManager, or any other AuthManager implementation that delegates to an OAuth server. This way the user would not supply their Bearer token directly to Cryostat. Instead, the user would visit the Cryostat web-client and be redirected to the platform OAuth server, allowing the user to log in with whatever credentials that server is configured to require (OpenShift cluster username/password, LDAP, or some other SSO). After successful authentication the user would be redirected back to the web-client with their access token as a query parameter, which the web-client should be able to capture and store as usual.

jan-law:
Related #717 
Depends on https://github.com/cryostatio/cryostat-operator/pull/292
Depends on https://github.com/cryostatio/cryostat-web/pull/350

Edit: changed Fixes to Related as we will need a follow-up PR to add the logout feature

https://user-images.githubusercontent.com/84587295/144308041-ca14b8d9-9338-48d0-a2b2-269de3c23ea5.mp4

As of now, I've also set the `grantMethod` to `prompt`. If a user logs into Cryostat via the OCP login page for the first time, they will see this page before they are redirected to the dashboard. Alternatively, I can set `grantMethod: auto` to hide this page. ([OAuth Grant Options docs](https://docs.openshift.com/container-platform/4.8/authentication/configuring-internal-oauth.html#oauth-grant-options_configuring-internal-oauth))

image


src/main/java/io/cryostat/net/web/http/api/v2/AbstractV2RequestHandler.java
```diff
@@ -110,8 +111,17 @@ public final void handle(RoutingContext ctx) {
...
                             || cause instanceof KubernetesClientException) {
                         throw new ApiException(401, "HTTP Authorization Failure", ee);
+                    } else if (cause instanceof EmptyOpenShiftTokenException) {
+                        ctx.response()
+                                .putHeader(
+                                        "X-Location",
+                                        OpenShiftAuthManager.getAuthorizationEndpoint())
+                                .putHeader("access-control-expose-headers", "Location")
+                                .setStatusCode(302)
+                                .end();
```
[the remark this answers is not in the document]
andrewazores: I don't have a solid overall design suggestion for exactly how to do this, but here are some unorganized thoughts.

1. This is a common enough case that we should use normal control flow here, not relying on thrown/caught exceptions to control the flow
2. Expanding on that, maybe it makes sense to add a method to the `AuthManager` interface that looks something like `Optional<Headers> suppliedHeadersSufficient(Headers requestHeaders)` (this name sucks). Given the headers provided by the client on the request, it returns an empty Optional if the headers are "sufficient" from the AuthManager's perspective - ex. there is an Authorization header and, for the OpenShiftAuthManager in particular, the auth scheme is Bearer and the token is not blank. If the headers are not sufficient for any reason, then it returns an Optional containing Headers that should be sent back in a response to the client. In this case we would set the Location header. When the `AbstractV2RequestHandler` receives a request it should call `suppliedHeadersSufficient`. If the returned Optional is empty it proceeds as it usually does now. If it is non-empty then it inspects the returned response Headers and short-circuits into an early response to the client, determining the status code by the contents of the Headers (Location header implies a 302 status code, otherwise a 401, or something like that).
That's a really, really roughed out idea and needs some refinement. I already don't particularly like that the AuthManagers have to know anything about HTTP headers, and I'd like to avoid adding any more of a coupling there, so passing in and out the Headers objects directly isn't ideal and I would rather some abstraction be introduced here to represent `Your auth request is OK | Your auth request might be OK but there's another step required | Your auth request is invalid | Your auth request was rejected`. My rough idea above embeds this information into the Headers object and conditional logic about whether certain keys are present in the map or the form of the associated values.
3. The `authorizationUrl` shouldn't just be a "raw" `public static` field - it should only be accessible by calling a method on the `OpenShiftAuthManager`, which can have some kind of memoization so that it it isn't performing an HTTP request every time some caller needs to know the OAuth endpoint URL. You might just be able to reuse the `Future` returned by your existing `webClient` request for that - calling `.get()` on the response should just return back the value immediately on the 2nd, 3rd, nth invocation, I think. If not then it's fairly straightforward to memoize it with `CompletableFuture.complete`/`CompletableFuture.supplyAsync` and return that reference every time some caller needs it.
4. > except I'm not sure how to reach the AuthPostHandler if validateRequestAuthorization() throws a 401 before writeResponse(ctx, handle(requestParams)) is called
If the `AbstractV2RequestHandler`'s assumptions don't work or don't make sense for your concrete implementation class, then don't extend `AbstractV2RequestHandler` :-) . You can rename `AbstractV2RequestHandler` to `AbstractAuthenticatedV2RequestHandler` and create another one that doesn't do authorization checks on its own, and pull the authorization logic you do need to implement into your concrete `AuthPostHandler` instead.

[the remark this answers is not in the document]
jan-law: My latest refactor removes the headers inside `AuthPostHandler` and removed the custom exception. I have one last question before I add tests and open the PR:

The web browser will make an automatic GET request whenever it sees a response header named `Location` with a status code `302` and won't follow the redirect url. Some possible solutions, including setting the status code to a `401` or `403` instead of a `302` are outlined here: https://stackoverflow.com/a/52290824 

I don't think an error status code would be appropriate here, so I renamed the `Location` header to `X-Location`. What do you think?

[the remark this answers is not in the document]
andrewazores: Sure, that makes sense to me.

src/main/java/io/cryostat/net/AuthManager.java
```diff
@@ -37,18 +37,25 @@
...
 
 import io.cryostat.net.security.ResourceAction;
+import io.cryostat.net.web.http.api.v2.IntermediateResponse;
 
 public interface AuthManager {
     AuthenticationScheme getScheme();
 
     Future<UserInfo> getUserInfo(Supplier<String> httpHeaderProvider);
 
+    Optional<IntermediateResponse<UserInfo>> sendLoginRedirectIfRequired(
```
andrewazores, line 55: Hmm... do we need to include the `IntermediateResponse` type in this interface? I would prefer to keep that used solely within HTTP handlers (within the `.http.api` subpackage) if possible.

The `UserInfo` generic type parameter looks like it isn't really needed here since the actual `IntermediateResponse` returned by the concrete implementation doesn't actually set a `UserInfo` response body object, and I think the `302` status code can simply be inferred by the Handler implementation if this method returns a non-empty `Optional`. So, the only required information actually encoded here is some HTTP headers, which can be represented with either just the `String` for the redirect URL (and the `X-Location` inferred by the Handler), or a simple struct-like object if more than one Header value needs to be returned.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -113,6 +133,29 @@ public AuthenticationScheme getScheme() {
...
 
+    @Override
+    public Optional<IntermediateResponse<UserInfo>> sendLoginRedirectIfRequired(
+            Supplier<String> headerProvider, Set<ResourceAction> resourceActions)
+            throws ExecutionException, InterruptedException {
+        Boolean hasValidHeader = false;
+        try {
+            hasValidHeader = this.validateHttpHeader(headerProvider, resourceActions).get();
+
+            if (hasValidHeader) {
```
andrewazores, line 144: `hasValidHeader` is nullable since it is a `Boolean`, unfortunately, so this line could NPE if `validateHttpHeader` behaves badly. It would be safer to perform this check as `Boolean.FALSE.equals(hasValidHeader)`.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -287,6 +345,54 @@ private String getTokenFromHttpHeader(String rawHttpHeader) {
         }
     }
 
+    private Future<URI> computeAuthorizationEndpoint() {
```
andrewazores, line 348: Is this cacheable? It doesn't appear to use any end user-specific information, and the inputs like environment variables won't change during the Cryostat process' lifetime.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -287,6 +345,54 @@ private String getTokenFromHttpHeader(String rawHttpHeader) {
...
+                        String namespace = this.getNamespace();
+                        String clientId =
+                                String.format(
+                                        "system:serviceaccount:%s:%s",
+                                        namespace, env.getEnv("CRYOSTAT_OAUTH_CLIENT_ID"));
+
+                        String scope =
+                                String.format(
+                                        "user:check-access role:%s:%s",
+                                        env.getEnv("CRYOSTAT_OAUTH_ROLE"), namespace);
```
andrewazores, line 363: Can we add some guards in case the env vars are not set/blank? It looks to me like the failure mode here would be a little tough to figure out just looking at logs if one of these env vars wasn't set. Better to fail earlier and with a very clear error message in case the OpenShift platform is selected and the prerequisite environment variables appear to be missing.

src/main/java/io/cryostat/net/web/http/api/v2/AuthPostHandler.java
```diff
@@ -96,14 +99,32 @@ public boolean isAsync() {
...
-                                .get());
+
+        try {
+            Optional<IntermediateResponse<UserInfo>> redirectResponse =
+                    auth.sendLoginRedirectIfRequired(
+                            () -> requestParams.getHeaders().get(HttpHeaders.AUTHORIZATION),
+                            resourceActions());
+
+            if (!redirectResponse.isEmpty()) {
+                return redirectResponse.get();
```
andrewazores, line 110: Might be nice to avoid the `if not empty, return optional.get()` pattern and instead use `Optional#orElse` or something like it, ie:

```
Optional response = redirectIfRequired().orElse(new Response()/* normal non-redirect response*/).get();
return response;
```

andrewazores (commented):
Despite my nitpicking of fine internal details, this looks excellent and I'm really excited to have it in a future release. It will really smooth out the user experience and makes it feel much more polished and well-integrated into the deployment platform.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
jan-law: Yes this is cacheable. I tried using Guava at first but am still looking into using Caffeine instead to avoid adding another dependency.

andrewazores: Yes definitely avoid adding a dependency just for this.

Since it's only a single value that needs to be cached/memoized you don't even need a whole Caffeine cache, since that would give a lot more power and flexibility with cache eviction policies etc. that aren't needed here. All we basically need is lazy initialization and memoization of the result. [`ConcurrentHashMap#computeIfAbsent`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-) is simple, adds no dependency, and probably suits all the requirements here well enough - just initialize it with a capacity of 1 and use some `static final` key for accessing the one memoized value.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -330,83 +332,68 @@ private String getTokenFromHttpHeader(String rawHttpHeader) {
...
-                                                | IOException
-                                                | MissingEnvironmentVariableException e) {
-                                            throw new CompletionException(e);
-                                        }
-                                    });
-                        })
-                .get();
+    private CompletableFuture<String> computeAuthorizationEndpoint() {
+
+        authorizationUrl.computeIfAbsent(
```
andrewazores, line 337: I think you can just `return authorizationUrl.computeIfAbsent`

src/main/java/io/cryostat/net/AuthManager.java
```diff
@@ -49,6 +51,10 @@
 
     Future<UserInfo> getUserInfo(Supplier<String> httpHeaderProvider);
 
+    Optional<String> sendLoginRedirectIfRequired(
```
andrewazores, line 54: I think this name can be simplified a bit. `IfRequired` seems to already be implied by the `Optional` return type. Maybe just `getLoginRedirectUrl`?

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -69,25 +74,42 @@
...
+    private static final String OAUTH_WELL_KNOWN_HOST = "openshift.default.svc";
+    private static final String WELL_KNOWN_PATH = "/.well-known/oauth-authorization-server";
+    private static final String OAUTH_ENDPOINT_KEY = "authorization_endpoint";
 
+    private final Environment env;
     private final FileSystem fs;
     private final Function<String, OpenShiftClient> clientProvider;
+    private final WebClient webClient;
+
+    private ConcurrentHashMap<String, CompletableFuture<String>> authorizationUrl;
```
andrewazores, line 99: Can this be `final`?

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -69,25 +74,42 @@
...
+            Logger logger,
+            FileSystem fs,
+            Function<String, OpenShiftClient> clientProvider,
+            WebClient webClient) {
         super(logger);
+        this.env = env;
         this.fs = fs;
         this.clientProvider = clientProvider;
+        this.webClient = webClient;
+        authorizationUrl = new ConcurrentHashMap<>(1);
```
andrewazores, line 112: Just for consistent style, please use `this.authorizationUrl`

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -287,6 +332,70 @@ private String getTokenFromHttpHeader(String rawHttpHeader) {
...
+    private CompletableFuture<String> computeAuthorizationEndpoint() {
+
+        authorizationUrl.computeIfAbsent(
+                OAUTH_ENDPOINT_KEY,
+                key -> {
+                    CompletableFuture<JsonObject> oauthMetadata = new CompletableFuture<>();
+                    try {
+                        String namespace = this.getNamespace();
+                        Optional<String> clientId =
+                                Optional.ofNullable(env.getEnv("CRYOSTAT_OAUTH_CLIENT_ID"));
```
andrewazores, line 344: Please extract these repeated env var name Strings to constants

andrewazores:
Awesome, seems to be working great.

One question I have, which is definitely out of the scope of this specific PR, is if we can somehow get the OAuth redirected login page to remember our credentials so that they don't need to be provided every time. Is there some cookie we need to set and send along with the request when we redirect the user to OAuth perhaps?

andrewazores (approved):

jan-law:
Here's what happens if I login a second time. Do you mean to make a different request to the OAuth server if we already have a token to prevent another redirection?
I also believe you'll have to provide your credentials on the container platform login page again once the token expires after 24 hours.

https://user-images.githubusercontent.com/84587295/145612862-bde34105-43de-49d0-a034-0975f1118cbd.mp4
```
INFO: (10.129.0.1:42292): POST /api/v2.1/auth 302 74ms
Dec 10, 2021 4:57:51 PM io.cryostat.core.log.Logger info
INFO: (10.129.0.1:42294): GET /health 200 95ms
Dec 10, 2021 4:57:52 PM io.cryostat.core.log.Logger info
INFO: (10.129.0.1:42290): GET /api/v1/grafana_datasource_url 200 1ms
Dec 10, 2021 4:57:52 PM io.cryostat.core.log.Logger info
INFO: (10.129.0.1:42296): GET /api/v1/grafana_dashboard_url 200 2ms
Dec 10, 2021 4:57:52 PM io.cryostat.core.log.Logger info
INFO: (10.129.0.1:42292): GET /api/v1/notifications_url 200 2ms
Dec 10, 2021 4:57:53 PM io.cryostat.core.log.Logger info
INFO: (10.129.0.1:42296): POST /api/v2.1/auth 200 493ms
```

andrewazores:
Hmm. Let me try that again with Chrome. I'm in Firefox and when I tried closing and re-opening the Cryostat tab, or even just F5 refreshing it, I was redirected back to the platform OAuth login page and had to re-enter my credentials. Maybe it's a browser-specific thing, or maybe it's one of my installed add-ons.

andrewazores:
Nope, doesn't seem to be browser-specific. Same behaviour in Firefox, Chrome, and ungoogled-chromium. Maybe this is a CRC thing?

andrewazores:
https://user-images.githubusercontent.com/3787464/145617088-f5f91699-2c38-471b-b0c5-4874cd35df86.mp4

Here's what I see deployed in CRC and using Chrome in this video. The last OAuth login page at the end is from me pressing `Ctrl-R`.

jan-law:
> Maybe this is a CRC thing?
Possibly? Firefox and Chrome show the same behavior as what I posted above. Refreshing keeps you logged in.

PR #752, opened 2021-11-12
fix(runsh): run container as original uid

### LINKED ISSUE #751: `run.sh` mounted volumes permissions result in AccessDeniedExceptions
Labelling this as `build` since it only impacts the development setup.

Here's an example: https://github.com/cryostatio/cryostat-web/pull/319#pullrequestreview-805088834

The `run.sh` volume mounting of subdirectories of the working directory results in various permissions issues on different machines. This may depend on different default permissions settings (644/664), different podman versions, or other factors.

To simplify the setup and prevent these problems from occurring, a basic fix is to replace the host mounted volumes with simple tmpfs, like so:

```diff
diff --git a/run.sh b/run.sh
index 5642e29d..c0d6c79e 100755
--- a/run.sh
+++ b/run.sh
@@ -83,12 +83,12 @@ fi
 podman run \
     --pod cryostat \
     --memory 512M \
-    --mount type=bind,source="$(dirname $0)/archive",destination=/opt/cryostat.d/recordings.d,relabel=shared,bind-propagation=shared \
-    --mount type=bind,source="$(dirname $0)/certs",destination=/certs,relabel=shared,bind-propagation=shared \
-    --mount type=bind,source="$(dirname $0)/clientlib",destination=/clientlib,relabel=shared,bind-propagation=shared \
-    --mount type=bind,source="$(dirname $0)/conf",destination=/opt/cryostat.d/conf.d,relabel=shared,bind-propagation=shared \
-    --mount type=bind,source="$(dirname $0)/templates",destination=/opt/cryostat.d/templates.d,relabel=shared,bind-propagation=shared \
-    --mount type=bind,source="$(dirname $0)/truststore",destination=/truststore,relabel=shared,bind-propagation=shared \
+    --mount type=tmpfs,destination=/opt/cryostat.d/recordings.d \
+    --mount type=tmpfs,destination=/certs \
+    --mount type=tmpfs,destination=/clientlib \
+    --mount type=tmpfs,destination=/opt/cryostat.d/conf.d \
+    --mount type=tmpfs,destination=/opt/cryostat.d/templates.d \
+    --mount type=tmpfs,destination=/truststore \
     -e CRYOSTAT_PLATFORM=$CRYOSTAT_PLATFORM \
     -e CRYOSTAT_DISABLE_SSL=$CRYOSTAT_DISABLE_SSL \
     -e CRYOSTAT_DISABLE_JMX_AUTH=$CRYOSTAT_DISABLE_JMX_AUTH \
```

Related: #750

andrewazores:
Fixes #751

This runs the container using the uid 0. In rootless podman, this results in the container's processes running under the uid of the original user on the host system, though within the container the processes believe themselves to be root. For development purposes and binding subdirectories to the container for development testing this is easy and convenient.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -72,6 +72,10 @@ if [ ! -d "$(dirname $0)/clientlib" ]; then
     mkdir "$(dirname $0)/clientlib"
 fi
 
+if [ ! -d "$(dirname $0)/templates" ]; then
+    mkdir "$(dirname $0)/templates"
+fi
+
 if ! podman pod exists cryostat; then
     podman pod create \
         --hostname cryostat \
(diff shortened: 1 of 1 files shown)


hareetd (commented):
Tested and is working as expected but I'll leave final approval to @jan-law.

Edit: Since I had `AccessDeniedException` issues with `devserver.sh` but not with `smoketest.sh` like Janelle did.

run.sh
```diff
@@ -48,22 +48,6 @@ if [ -z "$KEYSTORE_PATH" ] && [ -f "$(dirname $0)/certs/cryostat-keystore.p12" ]
...
-if [ ! -d "$(dirname $0)/archive" ]; then
-    mkdir "$(dirname $0)/archive"
-fi
-
-if [ ! -d "$(dirname $0)/conf" ]; then
-    mkdir "$(dirname $0)/conf"
-fi
-
-if [ ! -d "$(dirname $0)/conf/credentials" ]; then
-    mkdir "$(dirname $0)/conf/credentials"
```
jan-law, line 60: Where should `cryostat-user.properties` be stored? If I put the file in `conf/` or `conf/credentials` I get a 401 when logging in with the `BasicAuthManager`:
```
WARNING: User properties file "/opt/cryostat.d/conf.d/cryostat-users.properties" does not exist
Nov 15, 2021 3:33:49 PM io.cryostat.core.log.Logger warn
WARNING: HTTP 401: Unauthorized
```

andrewazores: That's true, with this patch applied there is no longer anywhere to put a `cryostat-user.properties`. The `conf/` and `conf/credentials` in the working directory are no longer mounted into the container, nor is any other host directory mounted to the container at `/opt/cryostat.d/conf.d`.

I'll rethink this PR and see if there's a better and more compatible way to set up the mounts that doesn't result in access denied exceptions, while also giving us the flexibility of adding files like this.

andrewazores:
@hareetd @jan-law I've redone this PR with a new approach and updated the original PR text to suit. The existing `archives`, `templates`, `conf`, etc. subdirectories are mounted as usual here, so you can still add a `cryostat-users.properties` for example to test Basic auth.

jan-law (commented):
This is working well with both `deserver.sh` and `smoketest.sh` for me. I'll leave approval to @hareetd

hareetd (approved):

PR #758, opened 2021-12-01
fix(reports): ensure target connection remains open while copying active data

### LINKED ISSUE #759: Report generation may fail due to connection cache eviction
https://github.com/cryostatio/cryostat/blob/5394bd47bf9aa911f157e5c60467e32b537a47b0/src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java#L202

The remote target connection opened when generating a report for an active recording may be closed prematurely due to cache eviction (see also #756). The cache eviction overall is still a potential source of problems in general, but in this particular issue, it is possible to improve the scenario for active report generation in particular by implementing the stream I/O differently, similarly to:

https://github.com/cryostatio/cryostat/blob/5394bd47bf9aa911f157e5c60467e32b537a47b0/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingGetHandler.java#L138

In this case, the stream is processed in chunks by application code. Between each chunk the `TargetConnectionManager` is informed that the connection is in fact still in use. This does not completely prevent cache eviction but it does mark the connection as more recently used in the cache, lowering the connection's eviction score.

The current implementation opens the connection once and then performs a single long-running, blocking I/O operation to copy the remote recording data into a local temp file. No update to the connection cache is made to mark the connection as still in-use during this transfer. Perhaps worse, the connection can be evicted from cache simply due to expiring by TTL even though the connection is in fact still actively in-use.

andrewazores:
Fixes #759

### DIFF
--- a/src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java
+++ b/src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java
@@ -37,12 +37,14 @@
  */
 package io.cryostat.net.reports;
 
+import java.io.BufferedOutputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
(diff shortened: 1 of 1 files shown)


hareetd (approved):
Looks good, tested it out as well.

PR #764, opened 2021-12-01
fix(pom): remove unused "listen" port

### LINKED ISSUE #761: `pom.xml` still references `listen` env var

andrewazores:
Fixes #761

Remove unused listenPort property from pom, which also removes the declared port number 9090 from the OCI metadata


hareetd (dismissed):
Is this "listen" port a leftover from when the two-way command channel was a thing?

andrewazores:
> Is this "listen" port a leftover from when the two-way command channel was a thing?

Exactly, this was used for the separate TCP interactive command channel port, since that was separate from the webserver (at the time, the "exporter" - since the webserver was only used for serving JFR data to clients).

andrewazores:
Cleanly rebased.

hareetd (approved):

PR #774, opened 2021-12-03
fix(pom): use latest jfr-datasource and grafana-dashboard

### LINKED ISSUE #772: `pom.xml` uses outdated jfr-datasource and cryostat-grafana-dashboard images for itests

andrewazores:
Fixes #772

This updates the `pom.xml` versions for jfr-datasource and cryostat-grafana-dashboard used in integration testing from `1.0.0` to `latest`. Once the backport PR is opened I will change these versions from `latest` in `main` to `2.0.0` in the `v2` backport release version. Later, when `2.1.0` release is approaching and we create the upstream branch for it, then that branch should also have `latest` replaced by `2.1.0` for these two itest container versions as well.

Additionally, this PR uses `xpath` in `run.sh` and `smoketest.sh` to read `pom.xml` to determine things like the JMX and HTTP API port numbers to use by default. This is just a cleanup so that the values aren't hardcoded in multiple places, which is what leads to some places lagging behind as was the problem with the itest container versions above.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
   <cryostat.rjmxPort>9091</cryostat.rjmxPort>
   <cryostat.webPort>8181</cryostat.webPort>
   <cryostat.itest.webPort>8181</cryostat.itest.webPort>
-  <cryostat.itest.webHost>localhost</cryostat.itest.webHost>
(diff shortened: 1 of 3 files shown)


jan-law (commented):
This is working well for me, will leave approval to Hareet

hareetd (approved):

PR #779, opened 2021-12-07
feat(reports): sidecar container report generation

### LINKED ISSUE #324: Apply available memory-based heuristic when forking report subproces
See #8 #286 #287

andrewazores:
Related to https://github.com/cryostatio/cryostat/issues/8
Fixes #324 (no longer necessary)

This abstracts report generation to reduce the coupling of the report caches and report handlers to the SubprocessReportGenerator, and then introduces an alternate report generator implementation that performs HTTP calls out to a separate sidecar container (https://github.com/cryostatio/cryostat-reports). The user selects which report generation strategy is used by setting the environment variable `CRYOSTAT_REPORT_GENERATOR`. If this env var is unset or empty then Cryostat will default to forking a subprocess to generate the report as usual. If it is set then it is expected to be the URL to a `cryostat-reports` instance (or a loadbalanced service in front of some replicas), and Cryostat will delegate report generation off to the sidecar(s).

### DIFF
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -79,13 +79,14 @@ function runGrafana() {
 }
 
 function runReportGenerator() {
+    local RJMX_PORT=10000
     podman run \
         --name reports \
         --pod cryostat \
         --cpus 1 \
         --memory 512M \
         --restart on-failure \
-        --env JAVA_OPTIONS="-XX:ActiveProcessorCount=1 -XX:+UseSerialGC -Dorg.openjdk.jmc.flightrecorder.parser.singlethreaded=true" \
+        --env JAVA_OPTIONS="-XX:ActiveProcessorCount=1 -XX:+UseSerialGC -Dorg.openjdk.jmc.flightrecorder.parser.singlethreaded=true -Dcom.sun.management.jmxremote.autodiscovery=true -Dcom.sun.management.jmxremote.port=${RJMX_PORT} -Dcom.sun.management.jmxremote.rmi.port=${RJMX_PORT} -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false" \
         --env QUARKUS_HTTP_PORT=10001 \
         --rm -d quay.io/cryostat/cryostat-reports:latest
 }
@@ -112,6 +113,7 @@ function createPod() {
         --publish 8082:8082 \
         --publish 9990:9990 \
         --publish 9991:9991 \
+        --publish 10000:10000 \
         --publish 10001:10001
     # 8081: vertx-fib-demo
     # 9093: vertx-fib-demo-1 RJMX
@@ -121,8 +123,9 @@ function createPod() {
     # 9999: quarkus-test HTTP
     # 8082: Wildfly HTTP
     # 9990: Wildfly Admin Console
-    # 9990: Wildfly RJMX
-    # 10001: cryostat-reports
+    # 9991: Wildfly RJMX
+    # 10000: cryostat-reports RJMX
+    # 10001: cryostat-reports HTTP
 }
 
 function destroyPod() {
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -86,7 +86,7 @@ function runReportGenerator() {
         --memory 512M \
         --restart on-failure \
         --env QUARKUS_HTTP_PORT=10001 \
-        --rm -d quay.io/andrewazores/cryostat-reports:1.0.0-SNAPSHOT
+        --rm -d quay.io/cryostatio/cryostat-reports:1.0.0-SNAPSHOT
 }
 
 function createPod() {
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -85,8 +85,9 @@ function runReportGenerator() {
         --cpus 1 \
         --memory 512M \
         --restart on-failure \
+        --env JAVA_OPTIONS="-XX:ActiveProcessorCount=1 -XX:+UseSerialGC -Dorg.openjdk.jmc.flightrecorder.parser.singlethreaded=true" \
         --env QUARKUS_HTTP_PORT=10001 \
-        --rm -d quay.io/cryostat/cryostat-reports:1.0.0-SNAPSHOT
+        --rm -d quay.io/cryostat/cryostat-reports:latest
 }
 
 function createPod() {
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -87,6 +87,7 @@ function runReportGenerator() {
         --pod cryostat \
         --cpus 1 \
         --memory 512M \
+        --restart on-failure \
         --env QUARKUS_HTTP_PORT=10001 \
         --rm -d quay.io/andrewazores/cryostat-reports:1.0.0-SNAPSHOT
 }
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -85,6 +85,8 @@ function runReportGenerator() {
     podman run \
         --name reports \
         --pod cryostat \
+        --cpus 1 \
+        --memory 512M \
         --env QUARKUS_HTTP_PORT=10001 \
         --rm -d quay.io/andrewazores/cryostat-reports:1.0.0-SNAPSHOT
 }
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -16,7 +16,7 @@ function runCryostat() {
         CRYOSTAT_RJMX_USER=smoketest \
         CRYOSTAT_RJMX_PASS=smoketest \
         CRYOSTAT_ALLOW_UNTRUSTED_SSL=true \
-        CRYOSTAT_REPORT_GENERATOR=$CRYOSTAT_REPORT_GENERATOR \
+        CRYOSTAT_REPORT_GENERATOR="http://0.0.0.0:10001" \
         exec "$DIR/run.sh"
 }
 
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -86,7 +86,7 @@ function runReportGenerator() {
         --memory 512M \
         --restart on-failure \
         --env QUARKUS_HTTP_PORT=10001 \
-        --rm -d quay.io/cryostatio/cryostat-reports:1.0.0-SNAPSHOT
+        --rm -d quay.io/cryostat/cryostat-reports:1.0.0-SNAPSHOT
 }
 
 function createPod() {
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -4,6 +4,9 @@ set -x
 set -e
 
 function runCryostat() {
+    if [ -z "$CRYOSTAT_REPORT_GENERATOR" ]; then
+        CRYOSTAT_REPORT_GENERATOR=subprocess
+    fi
     local DIR="$(dirname "$(readlink -f "$0")")"
     local host="$(xpath -q -e 'project/properties/cryostat.itest.webHost/text()' pom.xml)"
     local datasourcePort="$(xpath -q -e 'project/properties/cryostat.itest.jfr-datasource.port/text()' pom.xml)"
@@ -13,7 +16,7 @@ function runCryostat() {
         CRYOSTAT_RJMX_USER=smoketest \
         CRYOSTAT_RJMX_PASS=smoketest \
         CRYOSTAT_ALLOW_UNTRUSTED_SSL=true \
-        CRYOSTAT_REPORT_GENERATOR="http://0.0.0.0:10001" \
+        CRYOSTAT_REPORT_GENERATOR=$CRYOSTAT_REPORT_GENERATOR \
         exec "$DIR/run.sh"
 }
 
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -12,6 +12,8 @@ function runCryostat() {
         GRAFANA_DASHBOARD_URL="http://${host}:${grafanaPort}" \
         CRYOSTAT_RJMX_USER=smoketest \
         CRYOSTAT_RJMX_PASS=smoketest \
+        CRYOSTAT_ALLOW_UNTRUSTED_SSL=true \
+        CRYOSTAT_REPORT_GENERATOR="http://0.0.0.0:10001" \
         exec "$DIR/run.sh"
 }
 
@@ -76,6 +78,14 @@ function runGrafana() {
         --rm -d "${stream}:${tag}"
 }
 
+function runReportGenerator() {
+    podman run \
+        --name reports \
+        --pod cryostat \
+        --env QUARKUS_HTTP_PORT=10001 \
+        --rm -d quay.io/andrewazores/cryostat-reports:1.0.0-SNAPSHOT
+}
+
 function createPod() {
     local jmxPort="$(xpath -q -e 'project/properties/cryostat.rjmxPort/text()' pom.xml)"
     local webPort="$(xpath -q -e 'project/properties/cryostat.webPort/text()' pom.xml)"
@@ -97,7 +107,8 @@ function createPod() {
         --publish 9999:9999 \
         --publish 8082:8082 \
         --publish 9990:9990 \
-        --publish 9991:9991
+        --publish 9991:9991 \
+        --publish 10001:10001
     # 8081: vertx-fib-demo
     # 9093: vertx-fib-demo-1 RJMX
     # 9094: vertx-fib-demo-2 RJMX
@@ -107,6 +118,7 @@ function createPod() {
     # 8082: Wildfly HTTP
     # 9990: Wildfly Admin Console
     # 9990: Wildfly RJMX
+    # 10001: cryostat-reports
 }
 
 function destroyPod() {
@@ -119,4 +131,5 @@ createPod
 runDemoApps
 runJfrDatasource
 runGrafana
+runReportGenerator
 runCryostat
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -4,9 +4,6 @@ set -x
 set -e
 
 function runCryostat() {
-    if [ -z "$CRYOSTAT_REPORT_GENERATOR" ]; then
-        CRYOSTAT_REPORT_GENERATOR=subprocess
-    fi
     local DIR="$(dirname "$(readlink -f "$0")")"
     local host="$(xpath -q -e 'project/properties/cryostat.itest.webHost/text()' pom.xml)"
     local datasourcePort="$(xpath -q -e 'project/properties/cryostat.itest.jfr-datasource.port/text()' pom.xml)"
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -13,7 +13,7 @@ function runCryostat() {
         CRYOSTAT_RJMX_USER=smoketest \
         CRYOSTAT_RJMX_PASS=smoketest \
         CRYOSTAT_ALLOW_UNTRUSTED_SSL=true \
-        CRYOSTAT_REPORT_GENERATOR="http://0.0.0.0:10001" \
+        CRYOSTAT_REPORT_GENERATOR="http://${host}:10001" \
         exec "$DIR/run.sh"
 }
 
--- a/src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
+++ b/src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
@@ -114,14 +114,17 @@ class ActiveRecordingReportCache {
             try {
                 saveFile = reportGeneratorServiceProvider.get().exec(recordingDescriptor).get();
                 return fs.readString(saveFile);
-            } catch (ExecutionException | CompletionException ee) {
-                logger.error(ee);
+            } catch (ExecutionException | CompletionException e) {
+                logger.error(e);
 
                 delete(recordingDescriptor.connectionDescriptor, recordingDescriptor.recordingName);
 
-                if (ee.getCause() instanceof SubprocessReportGenerator.ReportGenerationException) {
-                    SubprocessReportGenerator.ReportGenerationException generationException =
-                            (SubprocessReportGenerator.ReportGenerationException) ee.getCause();
+                if (e.getCause()
+                        instanceof SubprocessReportGenerator.SubprocessReportGenerationException) {
+                    SubprocessReportGenerator.SubprocessReportGenerationException
+                            generationException =
+                                    (SubprocessReportGenerator.SubprocessReportGenerationException)
+                                            e.getCause();
 
                     SubprocessReportGenerator.ExitStatus status = generationException.getStatus();
                     if (status == SubprocessReportGenerator.ExitStatus.OUT_OF_MEMORY) {
@@ -143,7 +146,7 @@ class ActiveRecordingReportCache {
                                 });
                     }
                 }
-                throw ee;
+                throw e;
             }
         } finally {
             if (saveFile != null) {
--- a/src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
+++ b/src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
@@ -62,20 +62,20 @@ import com.github.benmanes.caffeine.cache.Scheduler;
 
 class ActiveRecordingReportCache {
 
-    protected final Provider<SubprocessReportGenerator> subprocessReportGeneratorProvider;
+    protected final Provider<ReportGeneratorService> reportGeneratorServiceProvider;
     protected final FileSystem fs;
     protected final ReentrantLock generationLock;
-    protected final LoadingCache<SubprocessReportGenerator.RecordingDescriptor, String> cache;
+    protected final LoadingCache<RecordingDescriptor, String> cache;
     protected final TargetConnectionManager targetConnectionManager;
     protected final Logger logger;
 
     ActiveRecordingReportCache(
-            Provider<SubprocessReportGenerator> subprocessReportGeneratorProvider,
+            Provider<ReportGeneratorService> reportGeneratorServiceProvider,
             FileSystem fs,
             @Named(ReportsModule.REPORT_GENERATION_LOCK) ReentrantLock generationLock,
             TargetConnectionManager targetConnectionManager,
             Logger logger) {
-        this.subprocessReportGeneratorProvider = subprocessReportGeneratorProvider;
+        this.reportGeneratorServiceProvider = reportGeneratorServiceProvider;
         this.fs = fs;
         this.generationLock = generationLock;
         this.targetConnectionManager = targetConnectionManager;
@@ -93,10 +93,7 @@ class ActiveRecordingReportCache {
     Future<String> get(ConnectionDescriptor connectionDescriptor, String recordingName) {
         CompletableFuture<String> f = new CompletableFuture<>();
         try {
-            f.complete(
-                    cache.get(
-                            new SubprocessReportGenerator.RecordingDescriptor(
-                                    connectionDescriptor, recordingName)));
+            f.complete(cache.get(new RecordingDescriptor(connectionDescriptor, recordingName)));
         } catch (Exception e) {
             f.completeExceptionally(e);
         }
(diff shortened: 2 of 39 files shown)


andrewazores:
Test failure: https://github.com/cryostatio/cryostat/runs/4448618311?check_suite_focus=true#step:9:1165
#666

src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java
```diff
@@ -137,12 +116,15 @@
                         // See https://github.com/cryostatio/cryostat/issues/287
```
jan-law, line 116: Why is #324 no longer necessary? Can this comment be removed then?

smoketest.sh
```diff
@@ -12,6 +12,8 @@ function runCryostat() {
         GRAFANA_DASHBOARD_URL="http://${host}:${grafanaPort}" \
         CRYOSTAT_RJMX_USER=smoketest \
         CRYOSTAT_RJMX_PASS=smoketest \
+        CRYOSTAT_ALLOW_UNTRUSTED_SSL=true \
+        CRYOSTAT_REPORT_GENERATOR="http://0.0.0.0:10001" \
```
jan-law, line 16: What is the difference between `${host}` and `0.0.0.0` here?

smoketest.sh
```diff
@@ -76,6 +78,17 @@ function runGrafana() {
...
 
+function runReportGenerator() {
+    podman run \
+        --name reports \
+        --pod cryostat \
+        --cpus 1 \
+        --memory 512M \
+        --restart on-failure \
+        --env QUARKUS_HTTP_PORT=10001 \
+        --rm -d quay.io/cryostatio/cryostat-reports:1.0.0-SNAPSHOT
```
jan-law, line 89: When running smoketest.sh with `CRYOSTAT_AUTH_MANAGER=io.cryostat.net.BasicAuthManager CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true sh smoketest.sh`, I'm denied access to `quay.io/cryostatio/cryostat-reports`:
```
HTTP_PORT=10001 --rm -d quay.io/cryostatio/cryostat-reports:1.0.0-SNAPSHOT
Trying to pull quay.io/cryostatio/cryostat-reports:1.0.0-SNAPSHOT...
Error: initializing source docker://quay.io/cryostatio/cryostat-reports:1.0.0-SNAPSHOT: reading manifest 1.0.0-SNAPSHOT in quay.io/cryostatio/cryostat-reports: unauthorized: access to the requested resource is not authorized
```

src/main/java/io/cryostat/net/reports/ReportsModule.java
```diff
@@ -57,44 +56,75 @@
...
         })
 public abstract class ReportsModule {
 
-    static final String REPORT_GENERATION_LOCK = "REPORT_GENERATION_LOCK";
-
-    @Provides
-    @Singleton
-    @Named(REPORT_GENERATION_LOCK)
-    /** Used to ensure that only one report is generated at a time */
-    static ReentrantLock provideReportGenerationLock() {
```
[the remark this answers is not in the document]
andrewazores: There is still a lock involved, but it isn't so explicit:

https://github.com/andrewazores/cryostat/blob/sidecar-reports/src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java#L93

The `synchronized` keyword on this method will perform the same function of only allowing one thread to cause a forked subprocess to kick off and generate a report. In the case that a sidecar generator is configured then this code path isn't hit and so the report caches should be free to call out and make multiple report requests at the same time. No lock is applied in the case of the sidecar generators because it's a desirable feature that more than one generator replica could be available behind a loadbalancer - or it could even be a loadbalancer service running a native build of the report generator as "serverless".

src/main/java/io/cryostat/net/reports/SubprocessReportGenerator.java
andrewazores: #324 is about trying to fine-tune and eke out every last bit of optimized performance out of the subprocess generator, but that was always a compromised position to start out with. It's just a heavyweight operation by nature, and setting the subprocess heap size to take up as much space within the Cryostat container without causing Cryostat to OOM while also being able to process large reports requires an impossible balance between minimizing Cryostat's overall resource footprint and over-provisioning Cryostat enough that the subprocess generator can run and produce results.

Since this PR removes the resource footprint of report generation from the Cryostat container in the new preferred deployment mode, #324 can be dropped. And anyway, since the OOM score work that I did for the forked subprocess a while ago, the report generator should always be killed on out of memory and not the main Cryostat process. So the stability and robustness concern that was originally there in #324 has also been solved as well.

smoketest.sh
andrewazores: Good catch. I missed that substitution when I did my rebase. I've fixed it now to use `${host}`

smoketest.sh
andrewazores: Whoops, that should be `quay.io/cryostat/cryostat-reports` - I fixed the namespace.

That image repository is currently still empty on quay.io itself - I haven't pushed a manually-built image there and there is no CI auto-push set up yet, either. For now, clone `https://github.com/cryostatio/cryostat-reports` and build it locally yourself with `./mvnw package`. It should be loaded into your local podman image registry after that.

hareetd:
A general question before I start reviewing the actual code changes: if I'm understanding this PR correctly, by delegating report generation to a sidecar container used strictly for that purpose, there is more memory available for report generation (compared to the Cryostat instance container which must also handle every other function of Cryostat), meaning OOM errors should not occur?

andrewazores:
> A general question before I start reviewing the actual code changes: if I'm understanding this PR correctly, by delegating report generation to a sidecar container used strictly for that purpose, there is more memory available for report generation (compared to the Cryostat instance container which must also handle every other function of Cryostat), meaning OOM errors should not occur?

There isn't necessarily more memory available, since that depends on the container limits set by the container runtime platform.

I explained a bit of the resource requirements rationale in this comment: https://github.com/cryostatio/cryostat/pull/779#discussion_r766053568

The main problem is that report generation is going to be a relatively less-frequently used feature, but it requires by far the most resources. This leads to a classic resource provisioning problem where an end user would need to provision the Cryostat container with enough resources to be able to meet peak resource demands (during report generation), but the majority of the time those resources are not needed and end up being wasted. In a cloud environment, overprovisioned resources are a very direct waste of money at the end of the day.

By splitting the report generation into a separate container then it becomes much easier to provision resources appropriately. The main Cryostat container only needs enough resources to run its webserver and perform various relatively lightweight operations over HTTP and JMX, so its footprint can be small and therefore running it and keeping it deployed can be cheap. Users who only occasionally use automated reports can provision just a single container for it and only allocate it a small amount of resources, letting it fail if those resources are insufficient - or just continue using subprocess generation if they're really, low priority. The more you care about reports and the more often you want to see them, the more resources you can provision to the report container and/or the more replicas of the container you can spin up.

In the future I hope to have cryostat-reports built and runnable as a Quarkus native image as well, which would further reduce the resource footprint and also open the door to running it [serverless](https://docs.openshift.com/container-platform/4.6/serverless/serverless-getting-started.html), meaning that the cryostat-reports container would actually be started up on-demand when a reports generation request is made, and stopped after the response has been completed. This should make Cryostat as a whole just about the smallest and cheapest to run that it can be.

> 
> Edit: And this process of requesting report generation and receiving the generated report is done strictly through HTTP calls between the Cryostat container and the sidecar container?

Correct, the sidecar container has exactly one HTTP endpoint that expects a JFR binary to be POSTed to it and it responds with the automated analysis HTML document. So when you make a request to Cryostat for a report, then behind the scenes it actually delegates out to cryostat-reports with another HTTP request, and when it gets a response it wraps that up and sends you back another HTTP response.

smoketest.sh
jan-law: Thanks, I can see the reports in the recording dropdowns and in a new tab now

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -93,20 +88,15 @@
     Future<String> get(ConnectionDescriptor connectionDescriptor, String recordingName) {
         CompletableFuture<String> f = new CompletableFuture<>();
         try {
-            f.complete(
-                    cache.get(
-                            new SubprocessReportGenerator.RecordingDescriptor(
-                                    connectionDescriptor, recordingName)));
+            f.complete(cache.get(new RecordingDescriptor(connectionDescriptor, recordingName)));
```
hareetd, line 91: From this method, I'm assuming reports are saved as String representations? Or is this a String representation of the Path to the file? Also, if ActiveRecordingReportCache#get returns null then a call to ActiveRecordingReportCache#getReport would have to be made to generate the report, correct?

andrewazores: > From this method, I'm assuming reports are saved as String representations? Or is this a String representation of the Path to the file?

It is the report HTML itself as a String. The cache is configured with `softValues`, so if the available heap memory in the JVM is low then the garbage collector may reclaim space by removing cached reports. We could write them to disk and cache the paths and then use a cache eviction listener to delete the files on disk, too. That would make the `ActiveReportCache` more similar to the `ArchivedReportCache`, which does write to disk and has no time-based expiration policy (since archived recordings will not change and therefore the reports are always current for a given archived recording once generated), and which returns `Path`s instead of `String`s.

> Also, if ActiveRecordingReportCache#get returns null then a call to ActiveRecordingReportCache#getReport would have to be made to generate the report, correct?

`ActiveRecordingReportCache#get` should never return null - it calls `cache.get()` internally, and `cache.get()` will either return a cached value or will compute the value (by calling `getReport` as configured by line 85), cache it, and return that.

src/test/java/io/cryostat/net/reports/ActiveRecordingReportCacheTest.java
```diff
@@ -86,21 +84,21 @@
     class TestSubprocessReportGenerator extends SubprocessReportGenerator {
```
hareetd, line 84: Where is this nested class used? It seems the only place it's referenced is in the class and constructor definition.

andrewazores: Ah right, nice catch. I think it used to actually be used in tests but it was replaced by a more typical mock at some point. I'll take it out.

hareetd (dismissed):
Tested it out, looks good.

andrewazores:
It's a fairly large change, not just in line count/patch size but in functionality being moved around, so I'll wait until @jan-law gets another chance to take a look and @ebaron gets some eyes on this before merging.

src/main/java/io/cryostat/net/reports/RemoteReportGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+                MultipartForm.create()
+                        .binaryFileUpload(
+                                "file",
+                                recording.getFileName().toString(),
+                                recording.toAbsolutePath().toString(),
+                                HttpMimeType.OCTET_STREAM.mime());
+        var f = new CompletableFuture<Path>();
+        this.http
+                .postAbs(String.format("%s/report", reportGenerator))
+                .timeout(TimeUnit.MINUTES.toMillis(1))
```
jan-law, line 94: When I tried uploading a ~10MB recording to the cryostat pod archives (noop auth), the upload succeeds. Then when I click the dropdown to generate a report, the backend hangs for a little while before I see this exception:
```
Dec 10, 2021 7:45:48 PM io.cryostat.core.log.Logger info
INFO: POSTing /opt/cryostat.d/recordings.d/file-uploads/aa06f216-5dd5-4e64-a9aa-799994659a80 to http://localhost:10001
Dec 10, 2021 7:46:18 PM io.cryostat.core.log.Logger error
SEVERE: Exception thrown
java.util.concurrent.ExecutionException: io.vertx.core.VertxException: Connection was closed
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at io.cryostat.net.reports.ArchivedRecordingReportCache.get(ArchivedRecordingReportCache.java:82)
	at io.cryostat.net.reports.ReportService.get(ReportService.java:57)
	at io.cryostat.net.web.http.api.v1.ReportGetHandler.handleAuthenticated(ReportGetHandler.java:110)
Caused by: io.vertx.core.VertxException: Connection was closed

Dec 10, 2021 7:46:18 PM io.cryostat.core.log.Logger error
SEVERE: HTTP 500: java.util.concurrent.ExecutionException: io.vertx.core.VertxException: Connection was closed
Caused by: java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: io.vertx.core.VertxException: Connection was closed
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at io.cryostat.net.web.http.api.v1.ReportGetHandler.handleAuthenticated(ReportGetHandler.java:110)
Caused by: java.util.concurrent.ExecutionException: io.vertx.core.VertxException: Connection was closed
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at io.cryostat.net.reports.ArchivedRecordingReportCache.get(ArchivedRecordingReportCache.java:82)
	at io.cryostat.net.reports.ReportService.get(ReportService.java:57)
	... 10 more
Caused by: io.vertx.core.VertxException: Connection was closed
Dec 10, 2021 7:46:18 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:48070): GET /api/v1/reports/aa06f216-5dd5-4e64-a9aa-799994659a80 500 29582ms
```
If I click View Report, the new tab keeps loading and the backend hangs with the latest logs here:
```
Dec 10, 2021 7:56:02 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:48106): POST /api/beta/auth/token 200 9ms
Dec 10, 2021 7:56:02 PM io.cryostat.core.log.Logger info
INFO: POSTing /opt/cryostat.d/recordings.d/file-uploads/aa06f216-5dd5-4e64-a9aa-799994659a80 to http://localhost:10001
```
I don't see this behaviour with a smaller ~6MB recording. Does this happen to you?

Here's the 10MB recording I was using:
[allocation_before_gc_fix.zip](https://github.com/cryostatio/cryostat/files/7695124/allocation_before_gc_fix.zip)

andrewazores: Could you check the logs at `podman logs -f reports` to see what's happening in the cryostat-reports container?

andrewazores: Either way it sounds like I need to apply a fix so that it doesn't hang at least, but I'm curious what's happening in the report generator container to cause it to fail to respond.

jan-law: The pod looks like it terminated:
```
> podman logs -f reports
exec java -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/quarkus-run.jar
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/deployments/lib/main/org.jboss.slf4j.slf4j-jboss-logmanager-1.1.0.Final.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/deployments/lib/main/org.slf4j.slf4j-jdk14-1.7.30.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Slf4jLoggerFactory]
__  ____  __  _____   ___  __ ____  ______ 
 --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
 -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \   
--\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2021-12-10 20:56:39,161 INFO  [io.quarkus] (main) cryostat-reports 1.0.0-SNAPSHOT on JVM (powered by Quarkus 2.4.2.Final) started in 3.367s. Listening on: http://0.0.0.0:10001
2021-12-10 20:56:39,168 INFO  [io.quarkus] (main) Profile prod activated. 
2021-12-10 20:56:39,168 INFO  [io.quarkus] (main) Installed features: [cdi, resteasy-reactive, smallrye-context-propagation, vertx]
exec java -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/quarkus-run.jar
jalaw@localhost ~/cryostat-operator (main)> 
```

andrewazores: What happens if you increase the `--cpus` and `--memory` flag in `smoketest.sh` on the `reports` container? Or even just remove those flags altogether?

jan-law: >  Or even just remove those flags altogether?

If I upload the recording, then click View Report, it hangs.

If I upload the recoridng, then click the dropdown, the report shows up nicely in the dropdown after 5s, and a subsequent click on View Reports show the report in a new tab.

andrewazores: That's a little strange. Thanks for checking for me - I'll take a closer look into this soon.

andrewazores: Testing it now with your `allocation_before_gc_fix.jfr`. That should fail to upload to archives because the name pattern doesn't match, but it looks like it gets rejected with a `400` and yet the original UUID-named file corresponding to it is still present on the server. That's a bug in the V1 `RecordingsPostHandler`. I have renamed it to be similar to one of my other sample recordings that does have the correct name pattern, so now it is `janlaw-6f4775cdbf-82dvl_10mb_20211006T152006Z.jfr`. I'll see what's going on with this, but it definitely looks like the report generator container is just getting killed and restarted.

andrewazores: Reviewing the -web implementation,

> If I upload the recording, then click View Report, it hangs.

This goes through the newer `/api/beta` endpoint for report generation using a JWT token.

> If I upload the recording, then click the dropdown, the report shows up nicely in the dropdown after 5s, and a subsequent click on View Reports show the report in a new tab.

This still goes through the old `/api/v1` endpoint using the `reportUrl` given by `GET /api/v1/recordings`.

Could explain the difference in behaviour. They both end up calling through to the same report generator service which is behind the same archived report cache, which is why in the second case doing the dropdown first causes the subsequent "View Reports" to successfully open (and probably very quickly).

andrewazores: @jan-law the hanging problem should be fixed with the latest commit I just added. This applies a timeout to all of the `Future.get()`s involved in report generation, so it should take at most ~30s to get a response, and no more indefinite hang.

andrewazores: As for the problem of the sidecar dying when run with `--cpus 1 --memory 512M`, it appears it's getting OOM killed. That seems pretty weird - it's generating the reports the same way that the SubprocessReportGenerator does it, so I can't see why it's requiring so much memory. `run.sh` runs Cryostat itself with `--memory 512M` so it seems very suspicious that the subprocess generator can make do with whatever memory is available to it, but the sidecar version can't do it when it's the only thing running in a container with the same memory limit. I would hazard a guess that there is some kind of Quarkus/Vertx/Netty file upload handling or multipart decoding problem going on here again. In any case, that's a bug report for cryostatio/cryostat-reports.

jan-law:
After pulling your latest changes, I got a dependency error looking for the 2.4.0 cryostat-core jar. 
`Could not resolve dependencies for project io.cryostat:cryostat:jar:2.1.0-SNAPSHOT: Failure to find io.cryostat:cryostat-core:jar:2.4.0`

When I looked at cryostat-core, I see the versioning changed from 2.3.1 to 2.5.0-SNAPSHOT. Your branch builds fine when I change `<io.cryostat.core.version>2.4.0</io.cryostat.core.version>` to `2.5.0-SNAPSHOT`. Where is the 2.4.0 -core version from?

andrewazores:
> After pulling your latest changes, I got a dependency error looking for the 2.4.0 cryostat-core jar. `Could not resolve dependencies for project io.cryostat:cryostat:jar:2.1.0-SNAPSHOT: Failure to find io.cryostat:cryostat-core:jar:2.4.0`
> 
> When I looked at cryostat-core, I see the versioning changed from 2.3.1 to 2.5.0-SNAPSHOT. Your branch builds fine when I change `<io.cryostat.core.version>2.4.0</io.cryostat.core.version>` to `2.5.0-SNAPSHOT`. Where is the 2.4.0 -core version from?

You'll have to check out and build the upstream `v2` branch: https://github.com/cryostatio/cryostat-core/blob/v2/pom.xml#L8

src/main/java/io/cryostat/net/reports/RemoteReportGenerator.java
hareetd: Is the merging of this PR now dependent on this sidecar OOM bug being fixed first?

andrewazores: I think so. I haven't gotten around to checking out what's going on with the sidecar container, but there is no point to this PR until the sidecar is useful in practice.

ebaron:
I have a quick/dirty setup of this in OpenShift, but hit a reflection problem with the reports container:
```
2021-12-16 22:18:19,814 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-0) HTTP Request to /report failed, error id: 64f46a06-c1e9-4822-932e-c8cd734f5498-1: java.lang.RuntimeException: java.lang.InstantiationException: Type `org.openjdk.jmc.flightrecorder.internal.parser.v1.StructTypes.JfrOldObject` can not be instantiated reflectively as it does not have a no-parameter constructor or the no-parameter constructor has not been added explicitly to the native image.
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:526)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager$TypeEntry.readConstant(TypeManager.java:294)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager.readConstants(TypeManager.java:412)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.readConstantPoolEvent(ChunkLoaderV1.java:110)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:77)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:47)
	at java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.lang.Thread.run(Thread.java:829)
	at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:567)
	at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)
Caused by: java.lang.InstantiationException: Type `org.openjdk.jmc.flightrecorder.internal.parser.v1.StructTypes.JfrOldObject` can not be instantiated reflectively as it does not have a no-parameter constructor or the no-parameter constructor has not been added explicitly to the native image.
	at java.lang.Class.newInstance(DynamicHub.java:911)
	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:516)
	... 13 more
```

Should I be testing it in JVM mode for now?

andrewazores:
> I have a quick/dirty setup of this in OpenShift, but hit a reflection problem with the reports container:
> 
> ```
> 2021-12-16 22:18:19,814 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-0) HTTP Request to /report failed, error id: 64f46a06-c1e9-4822-932e-c8cd734f5498-1: java.lang.RuntimeException: java.lang.InstantiationException: Type `org.openjdk.jmc.flightrecorder.internal.parser.v1.StructTypes.JfrOldObject` can not be instantiated reflectively as it does not have a no-parameter constructor or the no-parameter constructor has not been added explicitly to the native image.
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:526)
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager$TypeEntry.readConstant(TypeManager.java:294)
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.TypeManager.readConstants(TypeManager.java:412)
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.readConstantPoolEvent(ChunkLoaderV1.java:110)
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:77)
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ChunkLoaderV1.call(ChunkLoaderV1.java:47)
> 	at java.util.concurrent.FutureTask.run(FutureTask.java:264)
> 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
> 	at java.util.concurrent.FutureTask.run(FutureTask.java:264)
> 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
> 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
> 	at java.lang.Thread.run(Thread.java:829)
> 	at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:567)
> 	at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)
> Caused by: java.lang.InstantiationException: Type `org.openjdk.jmc.flightrecorder.internal.parser.v1.StructTypes.JfrOldObject` can not be instantiated reflectively as it does not have a no-parameter constructor or the no-parameter constructor has not been added explicitly to the native image.
> 	at java.lang.Class.newInstance(DynamicHub.java:911)
> 	at org.openjdk.jmc.flightrecorder.internal.parser.v1.ValueReaders$ReflectiveReader.read(ValueReaders.java:516)
> 	... 13 more
> ```
> 
> Should I be testing it in JVM mode for now?

Yea, native mode isn't ready yet: https://github.com/cryostatio/cryostat-reports/issues/4

ebaron (dismissed):
Code changes look good to me. Tested in OpenShift with a slightly modified operator to set the `CRYOSTAT_REPORT_GENERATOR` variable to a manually deployed service. I was able to generate a report for a 3 min. recording.

src/main/java/io/cryostat/net/reports/RemoteReportGenerator.java
jan-law: > Is the merging of this PR now dependent on this sidecar OOM bug being fixed first?

Resolving this conversation because https://github.com/cryostatio/cryostat-reports/pull/7 is merged

andrewazores:
Rebased, just one very minor conflict in MessagingModule since I moved the MAX_CONNECTIONS env var out of there, but Janelle added the PRUNER constant.

jan-law (approved):
Changes look good to me after the rebase. 

The todo bot should detect and create issues for any todo comments, right?

andrewazores:
The todo bot is not active anymore, actually. I can't remember anymore why.

jan-law:
No worries. It would be nice, although not necessary, if the todos were tracked as issues somehow. None of the todos are high priority anyway.

hareetd (approved):
Looks good!

Issue #709, opened 2021-10-08
Improved Handling of Active and Archived Recordings


andrewazores:
Brainstorming ideas for V2+ API handlers and how we want to deal with active and archived recordings, in ways that will support more flexible client use-cases and therefore better web-client UX. These may also include changes to related classes like `SerializableRecordingDescriptor`.

#### Merged active/archive queries with state filtering by query param, supporting negation

- `GET /api/v2/targets/:targetId/recordings`
response:
```json
{
  "meta": {
    "status": "OK",
    "type": "application/json"
  },
  "data": [
    {
        "id": 1,
        "name": "myRecording",
        "state": "RUNNING",
        "startTime": 12345,
        "duration": 0,
        "continuous": true,
        "toDisk": true,
        "maxSize": 0,
        "maxAge": 0
    },
    {
        "name": "myRecording.1",
        "state": "ARCHIVED",
        "startTime": 12345,
        "duration": 30000,
        "continuous": true,
        "toDisk": true,
        "maxSize": 0,
        "maxAge": 0
    }
  ]
}
```

Here, active recordings for the `:targetId` are included in the response, just like `GET /api/v1/targets/:targetId/recordings` - however, a `state: ARCHIVED` recording is also included in the response. This represents an archived copy of that recording that was taken 30000ms after the source recording was first started. It does not have a unique `id` of its own, and it is named in a way that also indicates it is archived and which source recording it came from. Other recording options, like `continuous` and `maxSize`, from the source recording are also preserved and included in the data about this archived recording.

Implementation details: we already have archives that are subdivided by target, so it should be relatively easy to create an API handler that queries the JVM for its active recordings, queries the filesystem for archived recordings from that target, and joins the results. The recording options can be stored in the filesystem alongside the archived recording - here, the archived recording would be at `myRecording.1.jfr` in the archive, and the continuous/toDisk/maxSize/maxAge/duration data could be stored in `myRecording.1.[properties|json]` in the same directory.

- `GET /api/v2/targets/:targetId/recordings?state=ARCHIVED`
response:
```json
{
  "meta": {
    "status": "OK",
    "type": "application/json"
  },
  "data": [
    {
        "name": "myRecording.1",
        "state": "ARCHIVED",
        "startTime": 12345,
        "duration": 30000,
        "continuous": true,
        "toDisk": true,
        "maxSize": 0,
        "maxAge": 0
    }
  ]
}
```

[...]

PR #783, opened 2021-12-10
feat(graphql): prototyping

andrewazores:
Related to #495


andrewazores:
@hareetd @jan-law @ebaron any questions/thoughts/concerns about GraphQL and the capabilities we have with it so far in this draft?

(The code is a mess - it definitely should not just all be stuffed into a GraphModule, so before merging any of this I would refactor it out and split it up in some way generally similar to how the HttpModule and RequestHandlers are.)

andrewazores:
Also, this looks like a reasonable approach to the TODO I have left in `queries.graphqls`:

https://www.yld.io/blog/query-by-2-or-more-fields-on-graphql/

I'm still not sure if this should be an `AND` or `OR` or if we want to make that a selectable option. GraphQL doesn't like nested structures, at least not ones to arbitrary depth, so we can't really do something like building an expression out of the label/annotation selector (key/value match pair) with that. More feasibly we could have the query support a list of operands and allow the user to select whether they are `AND`ed or `OR`ed - ie *any* of the conditions matches or *all* must match for a target to be included in the result set.

graphql/discovery-query.graphql
```diff
@@ -0,0 +1,66 @@
...
+            annotations {
+                cryostat
+                platform
+            }
+        }
+    }
+}
+
+# Arbitrarily query up to 8 levels deep. GraphQL doesn't support arbitrary recursive
+# querying for nested/recurisve structures - the query has to have an explicit depth
```
hareetd, line 18: I think I remember you mentioning this being an issue, could you explain why again?

andrewazores: I can define a structure in GraphQL that looks something like (pseudo"code"):

```
A {
  B[] b
  String name
}

B {
  C[] c
  String name
}

C {
  String name
}
```

We could use this to do something like:

```
var c = C("leaf 1")
var c = C("leaf 1")
var c = C("leaf 1")

Comment (hareetd) [reply graphql/discovery-query.graphql:18]: Okay, that makes sense. Would you say this is the biggest potential issue (as far as we know) if we do decide to use GraphQL for batch operations? From first impressions, I like the JSON-esque queries and it looks like mutations have us covered when it comes to performing actions on Cryostat. The `DataFetchers` themselves don't seem overly complex either.

Comment (andrewazores) [reply graphql/discovery-query.graphql:18]: Yes, this seems to be the biggest limitation IMO. Luckily it only really matters for the querying side of things - mutations can operate on arbitrarily nested structures just fine, the only problem is that we can't return results with arbitrarily nested structures. For responses to mutations that probably doesn't matter, but for querying what exists it's definitely a limitation.

In practise we could probably just allow depths up to say 16 or 32 or some arbitrary "large" depth and it should cover us in the practical use cases we support, and we can document that limitation.

File: src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java [hunk 2]
```diff
@@ -0,0 +1,721 @@
...
+    }
+
+    @Provides
+    @Singleton
+    @Named("recordings")
+    @SuppressFBWarnings(
+            value = "URF_UNREAD_FIELD",
+            justification =
+                    "The Recordings fields are serialized and returned to the client by the GraphQL engine")
+    static DataFetcher<Recordings> provideRecordingsFetcher(
```
jan-law, line 531: Let’s say I have several JVMs, some of which contain archived recordings. If I started a recording in all of my JVMs with a fixed duration, then fetched them after they’ve been automatically stopped and archived, how could I search for the recordings that I started in that specific batch operation?

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
```diff
@@ -0,0 +1,721 @@
...
+            TargetConnectionManager tcm,
+            RecordingTargetHelper helper,
+            RecordingOptionsBuilderFactory recordingOptionsBuilderFactory,
+            CredentialsManager credentialsManager,
+            Logger logger) {
+        return env -> {
+            Map<String, Object> settings = env.getArgument("recording");
+            List<Map<String, String>> labelSelectors = env.getArgument("labels");
+
+            EnvironmentNode root = client.getDiscoveryTree();
```
jan-law, line 441: > For performance reasons we might want to keep an in-memory model of the whole Kubernetes deployment cached for some time and periodically refresh it

Revisiting this, is the list of active and archived recordings for each target considered part of this deployment model? If so, would we need to refresh this cache every time a mutation operates on your targets? For example, if you accidentally started several recordings and wanted to stop them all, the `provideStopRecordingByLabelsFetcher` would need to know what recordings were started on those targets.

src/main/resources/mutations.graphqls
```diff
@@ -0,0 +1,5 @@
+type Mutation {
+    startRecordingByNodes(recording: RecordingSettings!, nodes: [NodeSelector!]!): [TargetNode!]!
+    startRecordingByAnnotations(recording: RecordingSettings!, annotations: [AnnotationSelector!]!): [TargetNode!]!
+    startRecordingByLabels(recording: RecordingSettings!, labels: [LabelSelector!]!): [TargetNode!]!
```
jan-law, line 4: >  TODO - can these all be done in one query? With an implied "AND" or "OR"
    of the various selector parameters? Or would the AND/OR be another argument?
    targetsWithLabel(key: String!, value: String): [TargetNode!]!
    targetsWithAnnotation(key: String!, value: String): [TargetNode!]!
    targetsWithAnnotations(annotations: [String!]!): [TargetNode!]!

Similar to the Queries, if a user can query for targets with annotations `AND`/`OR` labels, would it make sense for users to be able to start/stop recordings with the same amount of flexibility to select by annotation and/or label? How would the mutation definitions need to change?

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
andrewazores: There would be nothing linking a recording (active or archived) to a specific batch operation that was used to create that recording. If we implemented something that allows us to store some more metadata about recordings (see https://github.com/cryostatio/cryostat/issues/709) then perhaps we could include some piece of information about the batch operation execution in that metadata, but I don't know what that would be - the exact timestamp of the operation's execution? Does the user need to provide an operation ID that we record in the metadata, or perhaps we generate such an ID to include in the metadata and we include the ID in the response we send to the user?

I would think that the useful kinds of query here would not be "show me the recordings that were created by this specific batch operation I performed" but "show me all recordings with start time greater than XYZ" or "show me all recordings from JVMs with annotation `com.app/suspiciousService`". The operation itself doesn't seem like it's the thing of interest - it's the recordings that are interesting, and the properties they have, or the properties of the JVMs that they come from.

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
andrewazores: The active and archived recordings aren't things that Kubernetes knows about, so mutations to those resources doesn't require a re-query from Kubernetes for the deployment model. We would be able to just update our in-memory model of the active and archived recordings that we know about. We may still want to periodically refresh those models to ensure that our model isn't out of sync with reality, particularly for active recordings since it's possible that some other JMX client has mutated the recordings on some targets, or it's possible that a target has been restarted and lost its active recordings (if they are not started by JVM flags on the target, instead started by Cryostat either manually or via automated rules).

Anyway, your example of wanting to stop specific recordings within a set of targets is an interesting one, and I think very worthy of consideration. This also links back to the other comment chain and #709. Maybe we want to add metadata to allow users to attach annotations and labels to the recordings themselves. Then a user can make a single query like "start a recording with annotation `com.app/jfrProfiling` on all targets with label `app=checkoutService`", and later they can say "stop recordings with annotation `com.app/jfrProfiling` on all targets with label `app=checkoutService`".

As it is right now, there is an implicit underlying assumption that users only really use one recording at a time per target. It is supported to have multiple because JFR supports this very cleanly, but it's true that Cryostat's API tends to operate on things at the target level and all recordings belonging to a target.

src/main/resources/mutations.graphqls
andrewazores: IMO what makes the most sense is that the mutation definitions should generally mirror the query definitions, except they also require this `RecordingSettings` parameter. Given a Selector, or some series of Selectors and some Operation, we can expect the Query to tell us about all of the targets or recordings matching the Selector(s)/Operation, or we can expect the Mutation to start a recording using the given settings on all targets matching the Selector(s)/Operation.

The nice thing GraphQL gives us is that the Query side can always be defined as telling the client about the Targets, but that the Targets also contain reference to the Recordings, and GraphQL knows how to resolve the Recordings. If the user only cares about the Targets then they just make their query for that level of detail and that's what they get back, and the server does no unnecessary work. If they do need to know about the recordings then they add that request into their query and the GraphQL engine figures out how to call the resolver and prepare the result, only causing the server to do the extra work of connecting to the targets (or consulting the in-memory cache as described earlier) to populate the active recordings list for each target if that's actually requested and required.

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
jan-law: > it's the recordings that are interesting, and the properties they have, or the properties of the JVMs that they come from.

Ah I see, as long as we can select the JVMs we're interested in, we can get all of the recordings we need.

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
jan-law: > We would be able to just update our in-memory model of the active and archived recordings that we know about.

This makes sense, thanks.

> As it is right now, there is an implicit underlying assumption that users only really use one recording at a time per target.

Do you know of any use cases that would involve running more than one recording at once? For example, starting a Profiling recording for a short duration while an older Continuous recording runs in the background?
It seems like adding metadata would be worthwhile if you needed to differentiate between multiple running recordings. If there's only one running recording, you could just ask Cryostat to stop all recordings on the JVM.

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
andrewazores: Yea, exactly. There definitely appears to be some opportunity to give the client a way to filter out which recordings they are interested in, but that seems less necessary than filtering by target.

src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphModule.java
andrewazores: Yes, your example is a very valid one and one that we should probably expect to happen quite a lot. A Continuous template recording should be running at all times on most services, and a Profiling template recording would be used to help diagnose and troubleshoot when a specific problem is detected using the information provided by the Continuous recording. We shouldn't ask the user to stop the Continuous recording before starting the Profiling recording, and we don't want to do any of that manual juggling ourselves (stop the Continuous, start the Profiling, start the Continuous again when that's done) because it's unnecessary overhead and bookkeeping.

In other words, the implicit underlying assumption is not a good one. But it's the nature of how the API has been built so far, where recordings are only really differentiated by the target they belong to and their human-friendly name. I don't think we want to start enforcing specific names for recordings, so this metadata idea is pretty powerful and important.

andrewazores:
Just had a neat idea to really use the power of the GraphQL engine to our advantage here, after I've gone through and done some refactoring and cleanup to make this more presentable. Check it out:

```
$ https -v :8181/api/beta/graphql query=@graphql/start-recording-by-target.graphql
```

```graphql
query {
    targetsDescendedFrom(nodes: [{ name: "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi", nodeType: "JVM" }]) {
        name
        nodeType
        target {
            alias
            serviceUri
            labels
            annotations {
                platform
                cryostat
            }
        }
        recordings {
            active {
                name
                downloadUrl
                reportUrl
                state
                startTime
                duration
                continuous
                toDisk
                maxSize
                maxAge
            }
            archived {
                name
                downloadUrl
                reportUrl
            }
        }
        startRecording(recording: {
            name: "foo",
            template: "Profiling",
            templateType: "TARGET",
            duration: 30
            }) {
                name
                downloadUrl
                reportUrl
                state
                startTime
                duration
                continuous
                toDisk
                maxSize
                maxAge
        }
    }
}
```

This happens to be specifying a specific target node (the 9093 vertx-fib-demo) as the node whose descendants should be queried, so in this case the list of target descendants is trivially just the singleton list of the specified node itself. The query will be carried out against all of the nodes in that list and it will return a whack of information about each of those nodes, like its name and type and labels, etc. The GraphQL engine will also call another data fetcher to resolve the `recordings` field of the query, which causes Cryostat to connect to the target and check what its active recordings are, so the query includes information about those recordings' name, downloadUrl, reportUrl, etc., and it will also include the equivalent information for archived recordings on disk belonging to each target.

Most notably and new in the latest commit is the `startRecording` "field" of the query, which is obviously actually a mutating function call with each target node in the descendants list as the source object. This means that a client can make a read-only query or a mutating query using the same "shape" of query at the top level, starting with the `targetsDescendedFrom` function, but by adding the `startRecording` field it can become a mutating query that GraphQL will actually call out to another data fetcher, causing Cryostat to connect again to each of the target nodes and start a recording on them with the given settings. The result of this mutation is a `HyperlinkedSerializableRecordingDescriptor`, so the client can ask again for whichever fields of that they would like in the response.

Speaking of response, here's what the response to the big query above looks like:

```json
{
    "data": {
        "targetsDescendedFrom": [
            {
                "nodeType": "JVM",
                "recordings": {
                    "active": [],
                    "archived": []
                },
                "startRecording": {
                    "continuous": false,
                    "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/foo",
                    "duration": 30000,
                    "maxAge": 0,
                    "maxSize": 0,
                    "name": "foo",
                    "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/foo",
                    "startTime": 1643340857661,
                    "state": "RUNNING",
                    "toDisk": true
                },
                "target": {
                    "annotations": {
                        "cryostat": {
                            "HOST": "cryostat",
                            "JAVA_MAIN": "es.andrewazor.demo.Main",
                            "PORT": "9093"
                        },
                        "platform": {}
                    },
                    "labels": {},
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi"
                }
            }
        ]
    }
}
```

We can see that the active recordings list is empty, so this means (and this is also supported by the logs) that the read part of the query was executed before the mutation part. But if we simply reorder the components of the query then we can cause the mutation to happen first so that the query includes that result as well.

```graphql
query {
    targetsDescendedFrom(nodes: [{ name: "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi", nodeType: "JVM" }]) {
        startRecording(recording: {
            name: "bar",
            template: "Profiling",
            templateType: "TARGET",
            duration: 30
            }) {
                name
                downloadUrl
                reportUrl
                state
                startTime
                duration
                continuous
                toDisk
                maxSize
                maxAge
        }
        name
        nodeType
        target {
            alias
            serviceUri
            labels
            annotations {
                platform
                cryostat
            }
        }
        recordings {
            active {
                name
                downloadUrl
                reportUrl
                state
                startTime
                duration
                continuous
                toDisk
                maxSize
                maxAge
            }
            archived {
                name
                downloadUrl
                reportUrl
            }
        }
    }
}
```

```json
{
    "data": {
        "targetsDescendedFrom": [
            {
                "nodeType": "JVM",
                "recordings": {
                    "active": [
                        {
                            "continuous": false,
                            "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/foo",
                            "duration": 30000,
                            "maxAge": 0,
                            "maxSize": 0,
                            "name": "foo",
                            "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/foo",
                            "startTime": 1643340857661,
                            "state": "STOPPED",
                            "toDisk": true
                        },
                        {
                            "continuous": false,
                            "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/bar",
                            "duration": 30000,
                            "maxAge": 0,
                            "maxSize": 0,
                            "name": "bar",
                            "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/bar",
                            "startTime": 1643341706359,
                            "state": "RUNNING",
                            "toDisk": true
                        }
                    ],
                    "archived": []
                },
                "startRecording": {
                    "continuous": false,
                    "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/bar",
                    "duration": 30000,
                    "maxAge": 0,
                    "maxSize": 0,
                    "name": "bar",
                    "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/bar",
                    "startTime": 1643341706359,
                    "state": "RUNNING",
                    "toDisk": true
                },
                "target": {
                    "annotations": {
                        "cryostat": {
                            "HOST": "cryostat",
                            "JAVA_MAIN": "es.andrewazor.demo.Main",
                            "PORT": "9093"
                        },
                        "platform": {}
                    },
                    "labels": {},
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi"
                }
            }
        ]
    }
}
```

andrewazores:
Here are some resources following up from yesterday's discussion:

https://stackoverflow.com/questions/46160002/graphql-cascade-deletion

https://graphql-api.com/guides/schema/using-nested-mutations/

https://blog.logrocket.com/supporting-opt-in-nested-mutations-in-graphql/

https://fedidat.com/280-graphql-java-nesting/

graphql/environment-nodes-query-1.graphql
```diff
@@ -0,0 +1,15 @@
+query {
+    environmentNodes(filter: { nodeName: "JDP" }) {
+        descendantTargets {
+            startRecording(recording: {
+                name: "foo",
```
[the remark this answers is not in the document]
andrewazores: Ah, that's definitely somewhere to add better error handling. I think you are missing the JMX credentials required on some of the smoketest targets. The GraphQL API request generally completed and it was able to create some of the recordings you asked for, but there were some errors along the way for some targets. When you try to re-issue the request then all of the attempts fail either due to lacking credentials or if you do have credentials then because there is already a recording with the same name.

Just like for automated rules, currently you need to manually add credentials for such targets in advance using https://github.com/cryostatio/cryostat/blob/main/HTTP_API.md#TargetCredentialsPostHandler .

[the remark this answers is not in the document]
andrewazores: You could test that that's the problem by changing the `environmentNodes` filter - instead of selecting the JDP node, try adding a custom target pointing either to Cryostat itself or to the `vertx-fib-demo:9093` target which doesn't require credentials (or both), and then select the node named `Custom Targets` instead of `JDP`.

graphql/environment-nodes-query-4.graphql
```diff
@@ -0,0 +1,17 @@
+query {
+    environmentNodes(filter: { nodeName: "JDP" }) {
+        descendantTargets {
+            snapshot {
+                name
+                state
+            }
+            recordings {
+                active {
+                    archive {
```
jan-law, line 10: I was looking at this example and mistakenly confused the keywords `archived` (query) and `archive` ("mutation") with each other. What do you think about making the mutation keywords easier to identify, or at least change `archive` to `archiveRecording` to match `startRecording`?

andrewazores: Yea, I agree. I made the same mistake and thought it was a bug when I demoed this to you. I do think it would be better to have some kind of naming convention that marks mutations. My first thought was to just start them with `do`, in this case `doArchive`, but that gets pretty weird for `doStart` or `doStartRecording`. I think it would be nice to have some kind of uniform prefix/postfix for all mutations still, so they can be identified very quickly at a glance. Maybe `doStartRecording` is okay if it's also accompanied by `doArchiveRecording` and `doDeleteRecording`.

ebaron (approved):
Changes seem good to me. Lots of FIXMEs/TODOs throughout, but this is mostly self-contained, so I'm not worried about getting this in and fixing it up later.

hareetd:
> Just had a neat idea to really use the power of the GraphQL engine to our advantage here, after I've gone through and done some refactoring and cleanup to make this more presentable. Check it out:
> 
> ```
> $ https -v :8181/api/beta/graphql query=@graphql/start-recording-by-target.graphql
> ```
> 
> ```graphql
> query {
>     targetsDescendedFrom(nodes: [{ name: "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi", nodeType: "JVM" }]) {
>         name
>         nodeType
>         target {
>             alias
>             serviceUri
>             labels
>             annotations {
>                 platform
>                 cryostat
>             }
>         }
>         recordings {
>             active {
>                 name
>                 downloadUrl
>                 reportUrl
>                 state
>                 startTime
>                 duration
>                 continuous
>                 toDisk
>                 maxSize
>                 maxAge
>             }
>             archived {
>                 name
>                 downloadUrl
>                 reportUrl
>             }
>         }
>         startRecording(recording: {
>             name: "foo",
>             template: "Profiling",
>             templateType: "TARGET",
>             duration: 30
>             }) {
>                 name
>                 downloadUrl
>                 reportUrl
>                 state
>                 startTime
>                 duration
>                 continuous
>                 toDisk
>                 maxSize
>                 maxAge
>         }
>     }
> }
> ```
> 
> This happens to be specifying a specific target node (the 9093 vertx-fib-demo) as the node whose descendants should be queried, so in this case the list of target descendants is trivially just the singleton list of the specified node itself. The query will be carried out against all of the nodes in that list and it will return a whack of information about each of those nodes, like its name and type and labels, etc. The GraphQL engine will also call another data fetcher to resolve the `recordings` field of the query, which causes Cryostat to connect to the target and check what its active recordings are, so the query includes information about those recordings' name, downloadUrl, reportUrl, etc., and it will also include the equivalent information for archived recordings on disk belonging to each target.
> 
> Most notably and new in the latest commit is the `startRecording` "field" of the query, which is obviously actually a mutating function call with each target node in the descendants list as the source object. This means that a client can make a read-only query or a mutating query using the same "shape" of query at the top level, starting with the `targetsDescendedFrom` function, but by adding the `startRecording` field it can become a mutating query that GraphQL will actually call out to another data fetcher, causing Cryostat to connect again to each of the target nodes and start a recording on them with the given settings. The result of this mutation is a `HyperlinkedSerializableRecordingDescriptor`, so the client can ask again for whichever fields of that they would like in the response.
> 
> Speaking of response, here's what the response to the big query above looks like:
> 
> ```json
> {
>     "data": {
>         "targetsDescendedFrom": [
>             {
>                 "name": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi",
>                 "nodeType": "JVM",
>                 "recordings": {
>                     "active": [],
>                     "archived": []
>                 },
>                 "startRecording": {
>                     "continuous": false,
>                     "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/foo",
>                     "duration": 30000,
>                     "maxAge": 0,
>                     "maxSize": 0,
>                     "name": "foo",
>                     "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/foo",
>                     "startTime": 1643340857661,
>                     "state": "RUNNING",
>                     "toDisk": true
>                 },
>                 "target": {
>                     "alias": "es.andrewazor.demo.Main",
>                     "annotations": {
>                         "cryostat": {
>                             "HOST": "cryostat",
>                             "JAVA_MAIN": "es.andrewazor.demo.Main",
>                             "PORT": "9093"
>                         },
>                         "platform": {}
>                     },
>                     "labels": {},
>                     "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi"
>                 }
>             }
>         ]
>     }
> }
> ```
> 
> We can see that the active recordings list is empty, so this means (and this is also supported by the logs) that the read part of the query was executed before the mutation part. But if we simply reorder the components of the query then we can cause the mutation to happen first so that the query includes that result as well.
> 
> ```graphql
> query {
>     targetsDescendedFrom(nodes: [{ name: "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi", nodeType: "JVM" }]) {
>         startRecording(recording: {
>             name: "bar",
>             template: "Profiling",
>             templateType: "TARGET",
>             duration: 30
>             }) {
>                 name
>                 downloadUrl
>                 reportUrl
>                 state
>                 startTime
>                 duration
>                 continuous
>                 toDisk
>                 maxSize
>                 maxAge
>         }
>         name
>         nodeType
>         target {
>             alias
>             serviceUri
>             labels
>             annotations {
>                 platform
>                 cryostat
>             }
>         }
>         recordings {
>             active {
>                 name
>                 downloadUrl
>                 reportUrl
>                 state
>                 startTime
>                 duration
>                 continuous
>                 toDisk
>                 maxSize
>                 maxAge
>             }
>             archived {
>                 name
>                 downloadUrl
>                 reportUrl
>             }
>         }
>     }
> }
> ```
> 
> ```json
> {
>     "data": {
>         "targetsDescendedFrom": [
>             {
>                 "name": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi",
>                 "nodeType": "JVM",
>                 "recordings": {
>                     "active": [
>                         {
>                             "continuous": false,
>                             "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/foo",
>                             "duration": 30000,
>                             "maxAge": 0,
>                             "maxSize": 0,
>                             "name": "foo",
>                             "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/foo",
>                             "startTime": 1643340857661,
>                             "state": "STOPPED",
>                             "toDisk": true
>                         },
>                         {
>                             "continuous": false,
>                             "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/bar",
>                             "duration": 30000,
>                             "maxAge": 0,
>                             "maxSize": 0,
>                             "name": "bar",
>                             "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/bar",
>                             "startTime": 1643341706359,
>                             "state": "RUNNING",
>                             "toDisk": true
>                         }
>                     ],
>                     "archived": []
>                 },
>                 "startRecording": {
>                     "continuous": false,
>                     "downloadUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/recordings/bar",
>                     "duration": 30000,
>                     "maxAge": 0,
>                     "maxSize": 0,
>                     "name": "bar",
>                     "reportUrl": "https://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat:9093%2Fjmxrmi/reports/bar",
>                     "startTime": 1643341706359,
>                     "state": "RUNNING",
>                     "toDisk": true
>                 },
>                 "target": {
>                     "alias": "es.andrewazor.demo.Main",
>                     "annotations": {
>                         "cryostat": {
>                             "HOST": "cryostat",
>                             "JAVA_MAIN": "es.andrewazor.demo.Main",
>                             "PORT": "9093"
>                         },
>                         "platform": {}
>                     },
>                     "labels": {},
>                     "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi"
>                 }
>             }
>         ]
>     }
> }
> ```

I've been looking over this code and playing around with GraphQL on the main branch to refresh my memory before giving input on #825. I decided to try this query out but got an error saying `Validation error of type FieldUndefined: Field 'targetsDescendedFrom' in type 'Query' is undefined @ 'targetsDescendedFrom'`. I added `TypeRuntimeWiring` connecting the `targetsDescendedFrom` query with the `TargetDescendentsFetcher` as well as the necessary `Query/Type` definitions and now it's working. I might be doing something wrong though but if I'm correct, would you like me to open an issue with a PR of my changes?

andrewazores:
You didn't do anything wrong there really, but that old comment is out of date and doesn't reflect what got merged anymore. It does look like I left the `Fetcher` implementation for it in place erroneously. Generally though, that query was first written for making a top-level query that would select some EnvironmentNode in the tree and return a list of all of its TargetNode children. That got replaced by the top-level `environmentNodes` query that takes the whole tree, flattens it into a list, and removes the TargetNodes. This can then be combined with a nested `descendantTargets` query to list or perform actions on targets that descend from the EnvironmentNodes returned by the outer query. With the right query filter provided to the outer query you get a singleton list, and the combination outer/inner query is equivalent to the `targetsDescendedFrom`.

`graphql/environment-nodes-query-1.graphql`
```graphql
query {
        descendantTargets {
            doStartRecording(recording: {
                name: "foo",
                template: "Profiling",
                templateType: "TARGET",
                duration: 30
                }) {
                    name
                    state
            }
        }
    }
}
```

hareetd:
Ah makes sense, thanks.

PR #786, opened 2021-12-16
feat(auth): Add OAuth logout

### LINKED ISSUE #717: OAuth Redirect Authentication flow

jan-law:
Fixes #717 
Depends on https://github.com/cryostatio/cryostat-web/pull/355
Depends on https://github.com/cryostatio/cryostat-operator/pull/319
Depends on https://github.com/cryostatio/cryostat-operator/issues/320

Please see the -web PR for more information about the OAuth server logout behaviour.

This PR also updates the web client and caches the logout redirect url.

### DIFF
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -73,7 +73,6 @@ import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReviewB
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.openshift.api.model.OAuthAccessToken;
-import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.client.WebClient;
@@ -315,16 +314,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
     private Future<Boolean> deleteToken(String token) {
         try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
             String serviceAccountAsOAuthClient = this.getServiceAccountName();
-
-            // FIXME reuse performTokenReview instead of copying it here
-            TokenReview review =
-                    new TokenReviewBuilder().withNewSpec().withToken(token).endSpec().build();
-            review = client.tokenReviews().create(review);
-            TokenReviewStatus status = review.getStatus();
-            if (StringUtils.isNotBlank(status.getError())) {
-                return CompletableFuture.failedFuture(
-                        new AuthorizationErrorException(status.getError()));
-            }
+            TokenReviewStatus status = performTokenReview(token).get();
             String uid = status.getUser().getUid();
 
             List<OAuthAccessToken> userOauthAccessTokens =
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -321,7 +321,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
                     Optional.ofNullable(client.oAuthAccessTokens().delete(userOauthAccessTokens))
                             .orElseThrow(TokenNotFoundException::new);
 
-            if (!deleted) {
+            if (Boolean.FALSE.equals(deleted)) {
                 throw new TokenNotFoundException();
             }
 
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -165,7 +165,8 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
     }
 
     @Override
-    public Optional<String> logout(Supplier<String> httpHeaderProvider) throws ExecutionException, InterruptedException {
+    public Optional<String> logout(Supplier<String> httpHeaderProvider)
+            throws ExecutionException, InterruptedException {
 
         String token = getTokenFromHttpHeader(httpHeaderProvider.get());
         deleteToken(token).get();
@@ -319,10 +320,12 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
 
             List<OAuthAccessToken> userOauthAccessTokens =
                     client.oAuthAccessTokens().list().getItems().stream()
-                            .filter(t -> {
-                                return (t.getClientName().equals(serviceAccountAsOAuthClient)
-                                    && t.getUserUID().equals(uid));
-                            })
+                            .filter(
+                                    t -> {
+                                        return (t.getClientName()
+                                                        .equals(serviceAccountAsOAuthClient)
+                                                && t.getUserUID().equals(uid));
+                                    })
                             .collect(Collectors.toList());
 
             Boolean deleted =
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -166,10 +166,10 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
 
     @Override
     public Optional<String> logout(Supplier<String> httpHeaderProvider)
-            throws ExecutionException, InterruptedException {
+            throws ExecutionException, InterruptedException, IOException, TokenNotFoundException {
 
         String token = getTokenFromHttpHeader(httpHeaderProvider.get());
-        deleteToken(token).get();
+        deleteToken(token);
 
         return Optional.of(this.computeLogoutRedirectEndpoint().get());
     }
@@ -311,7 +311,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
         }
     }
 
-    private Future<Boolean> deleteToken(String token) {
+    private Boolean deleteToken(String token) throws IOException, TokenNotFoundException {
         try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
             Boolean deleted =
                     Optional.ofNullable(
@@ -324,10 +324,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
                 throw new TokenNotFoundException();
             }
 
-            return CompletableFuture.completedFuture(deleted);
-        } catch (Exception e) {
-            logger.error(e);
-            return CompletableFuture.failedFuture(e);
+            return deleted;
         }
     }
 
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -311,7 +311,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
         }
     }
 
-    private Boolean deleteToken(String token) throws IOException, TokenNotFoundException {
+    private boolean deleteToken(String token) throws IOException, TokenNotFoundException {
         try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
             Boolean deleted =
                     Optional.ofNullable(
--- a/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
+++ b/src/main/java/io/cryostat/net/OpenShiftAuthManager.java
@@ -72,7 +72,6 @@ import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReview;
 import io.fabric8.kubernetes.api.model.authorization.v1.SelfSubjectAccessReviewBuilder;
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.KubernetesClientException;
-import io.fabric8.openshift.api.model.OAuthAccessToken;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.client.WebClient;
@@ -80,6 +79,7 @@ import jdk.jfr.Category;
 import jdk.jfr.Event;
 import jdk.jfr.Label;
 import jdk.jfr.Name;
+import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.utils.URIBuilder;
 
@@ -313,22 +313,11 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
 
     private Future<Boolean> deleteToken(String token) {
         try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
-            String serviceAccountAsOAuthClient = this.getServiceAccountName();
-            TokenReviewStatus status = performTokenReview(token).get();
-            String uid = status.getUser().getUid();
-
-            List<OAuthAccessToken> userOauthAccessTokens =
-                    client.oAuthAccessTokens().list().getItems().stream()
-                            .filter(
-                                    t -> {
-                                        return (t.getClientName()
-                                                        .equals(serviceAccountAsOAuthClient)
-                                                && t.getUserUID().equals(uid));
-                                    })
-                            .collect(Collectors.toList());
-
             Boolean deleted =
-                    Optional.ofNullable(client.oAuthAccessTokens().delete(userOauthAccessTokens))
+                    Optional.ofNullable(
+                                    client.oAuthAccessTokens()
+                                            .withName(this.getOauthAccessTokenName(token))
+                                            .delete())
                             .orElseThrow(TokenNotFoundException::new);
 
             if (Boolean.FALSE.equals(deleted)) {
@@ -380,6 +369,7 @@ public class OpenShiftAuthManager extends AbstractAuthManager {
     }
 
     private CompletableFuture<String> computeAuthorizationEndpoint() {
+
         return oauthUrls.computeIfAbsent(
                 AUTHORIZATION_URL_KEY,
                 key -> {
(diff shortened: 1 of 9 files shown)


src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -296,6 +309,29 @@ public AuthenticationScheme getScheme() {
...
+            List<OAuthAccessToken> userOauthAccessTokens =
+                    client.oAuthAccessTokens().list().getItems().stream()
+                            .filter(t -> t.getClientName().equals(serviceAccountAsOAuthClient))
+                            .collect(Collectors.toList());
+
+            Boolean deleted =
+                    Optional.ofNullable(client.oAuthAccessTokens().delete(userOauthAccessTokens))
+                            .orElseThrow(TokenNotFoundException::new);
+
+            if (!deleted) {
```
andrewazores, line 324: I think, since `deleted` is nullable, this would be safer as `Boolean.FALSE.equals(deleted)`

andrewazores (commented):
Code looks pretty reasonable. I'll take a look at the OpenShift auth docs next.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -296,6 +309,29 @@ public AuthenticationScheme getScheme() {
...
+        try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
+            String serviceAccountAsOAuthClient = this.getServiceAccountName();
+            List<OAuthAccessToken> userOauthAccessTokens =
+                    client.oAuthAccessTokens().list().getItems().stream()
+                            .filter(t -> t.getClientName().equals(serviceAccountAsOAuthClient))
+                            .collect(Collectors.toList());
+
+            Boolean deleted =
+                    Optional.ofNullable(client.oAuthAccessTokens().delete(userOauthAccessTokens))
+                            .orElseThrow(TokenNotFoundException::new);
```
ebaron, line 322: Will this delete tokens for every user? Say Alice and Bob are both logged into Cryostat. One of them logs out, would the other be logged out as well?

jan-law: Oops, I think I mistakenly assumed that all tokens in a single namespace correspond to a single user. I'll add a fix to compare usernames too. Do you know of another user besides `kube:admin` that I can use to test this?

andrewazores: There's also the `developer` user in a default CRC install. But the username could be basically any arbitrary string, really.

jan-law: > There's also the developer user in a default CRC install

Is there an equivalent user for Quicklab?

> the username could be basically any arbitrary string

Good point, I should compare user UIDs instead.

ebaron: I found this article helpful before: https://medium.com/kubelancer-private-limited/create-users-on-openshift-4-dc5cfdf85661.

You'll need to authorize the new user by adding them to roles with:
`oc adm policy add-role-to-user`
The `edit` cluster role will give you most built-in permissions you need, then there are a variety of cluster roles for our operator that will help for the operator's CRDs: `cryostat-editor-role`, `flightrecorder-editor-role`, and `recording-editor-role`.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -296,6 +312,46 @@ public AuthenticationScheme getScheme() {
...
 
+    private Future<Boolean> deleteToken(String token) {
+        try (OpenShiftClient client = clientProvider.apply(getServiceAccountToken())) {
+            String serviceAccountAsOAuthClient = this.getServiceAccountName();
+
+            // FIXME reuse performTokenReview instead of copying it here
+            TokenReview review =
+                    new TokenReviewBuilder().withNewSpec().withToken(token).endSpec().build();
+            review = client.tokenReviews().create(review);
+            TokenReviewStatus status = review.getStatus();
```
jan-law, line 323: How does the `clientProvider` instantiate the `OpenShiftClient` and apply the service account token to the client? 
I know I can't call `performTokenReview` inside this try block because I get a `token already set` error when the `clientProvider` tries to apply the service account token again in `performTokenReview`. I also tried calling `performTokenReview` outside of this try block  with the same outcome. 

Alternatively, is there a simpler way to determine the userUID given a token from the authorization header?

andrewazores: https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/NetworkModule.java#L207

It's just a simple abstraction over the `new DefaultOpenShiftClient()` call so that the unit tests can construct the AuthManager with a `clientProvider` that provides a mock `OpenShiftClient`.

Is `token already set` happening in actual use or just in tests? https://github.com/cryostatio/cryostat/blob/main/src/test/java/io/cryostat/net/OpenShiftAuthManagerTest.java#L570

IIRC this was just a quick hack in the test implementation so that a single `client` instance could be reused, because at the time it was written there was never a case where nested clients were needed. So that `token already set` in the mock wrapper was just a guard against accidentally misusing the mock, which I guess bit me at some point and led to weird test results.

If this is happening in actual practice with the real `clientProvider` that calls `new DefaultOpenShiftClient()` then something is really funky.

> Alternatively, is there a simpler way to determine the userUID given a token from the authorization header?

This part I'm not sure about. Cryostat definitely doesn't know anything about the tokens and the OpenShiftAuthManager doesn't know anything about UIDs, and I don't think we want to do anything like trying to intercept and cache user auth token requests. The best bet should be doing essentially what you're trying to do here, taking the user-provided token and asking the platform auth server who that belongs to, using only our own service account credentials to do so.

jan-law: > Is token already set happening in actual use or just in tests?

Thankfully I only see the error in the unit tests, which explains why I didn't see the error on OpenShift. 

> IIRC this was just a quick hack in the test implementation so that a single client instance could be reused, because at the time it was written there was never a case where nested clients were needed. 

How would you recommend modifying the test implementation such that nested clients are allowed and I can still guard against these other weird test results?

andrewazores: @jan-law I was in the middle of editing my comment when this one came in - see my `Edit:` on the first paragraph.

tl;dr it seems there's leftover technical cruft in the unit tests from when I switched it from using a Mockito mock client to using the fabric8 mock-server and a real client that talks to the mock server. 

The `OpenShiftClient client;` field in the test is injected by the `OpenShiftMockServerExtension.class`/`@EnabledOpenShiftMockClient` test extensions, so it looks right now like the test class can only have one instance of the client at a time. I don't recall for sure but there is probably a way to get an instance of the client without relying on it being injected by the test extension. This is just useful because it's pre-configured to point to the mock-server and generally set up for unit testing (maybe this also means not caching results or reusing HTTP connections, things like that).

If there is no easy way to get additional client instances and it really does just need to be one instance injected by the extension, then perhaps a `@Nested` class could give you a second instance?

andrewazores: @jan-law

https://github.com/fabric8io/kubernetes-client/blob/master/openshift-server-mock/src/main/java/io/fabric8/openshift/client/server/mock/OpenShiftMockServer.java#L52

Looks like you can get a second, third, *n*th client pretty easily for testing. So I suppose the injected `OpenShiftClient client` and the `@EnableOpenShiftMockClient` can probably be removed, and the test `TokenProvider` modified so that each `apply` call invokes that `createOpenShiftClient` method.

The next difficulty is getting those returned clients to use the test-supplied tokens passed into the `provider#apply()` method.

Maybe the easiest thing to do here is create a subclass extending `OpenShiftMockServer` and adding a method that accepts the token to be used and merges that into the generated mock config, and then uses the new merged config to construct the client instance?

andrewazores: Here's what the MockServer class looks like in the actual version we're currently using:

https://github.com/fabric8io/kubernetes-client/blob/v5.4.1/openshift-server-mock/src/main/java/io/fabric8/openshift/client/server/mock/OpenShiftMockServer.java

The client creation method here is extra straightforward, so extending the class and adding a new method that does the exact same thing but adds in a token to be used seems easy enough.

It might need to be fixed when we eventually upgrade the fabric8 client version, but I don't think it'll be too bad to maintain.

jan-law: While refactoring, I found a cleaner way to lookup tokens using `oauthAccessTokens().withName(tokenName)`. This way, I don't need to use token reviews or nested clients. The token name is the base64 encoded checksum of the raw token: https://github.com/openshift/console/blob/9f352ba49f82ad693a72d0d35709961428b43b93/pkg/server/server.go#L609-L613

andrewazores: That's interesting. Is there anything in the OAuth server documentation that at least suggests this? I worry that it may simply be an internal implementation detail that is subject to change in the future, or worse, may be inconsistent between OpenShift versions or distributions with different OAuth versions/implementations. Or is it perhaps possible to get the OAuth server to tell us the token name some other way? Just as an example, when the OAuth server provides us with the token by URI fragment, can we request the token name from the server at the same time?

jan-law: It looks like the OAuth server will only return an access token (and/or a optional refresh token) https://docs.openshift.com/container-platform/4.8/authentication/configuring-internal-oauth.html#oauth-token-request-flows_configuring-internal-oauth

Here's everything I found related to token names. I couldn't find anything in the official docs:
1) [customer portal OAuth Logout API thread](https://access.redhat.com/discussions/5374431)
2) [a github issue that resulted in the same fix as 1)](https://github.com/kiali/kiali/issues/3595)

Also, when deleting tokens by user UID, I noticed that the kube:admin tokens didn't get deleted, which is why I tried using OAuthAccessToken names instead. If I create another user, that user's tokens are deleted successfully. Not sure if it was a problem with my implementation or something else.

andrewazores: Okay, thanks - nice investigation. I suppose that's the best we can do, and if it's the same as what the OpenShift Console does then I hope we can rely on this being relatively stable.

> Also, when deleting tokens by user UID, I noticed that the kube:admin tokens didn't get deleted, which is why I tried using OAuthAccessToken names instead. If I create another user, that user's tokens are deleted successfully. Not sure if it was a problem with my implementation or something else.

This is the same as what we're discussing over in the corresponding -web PR, right?

jan-law: > when deleting tokens by user UID, I noticed that the kube:admin tokens didn't get deleted,

I believe this was a different issue, because the -web PR has an issue with making a successful POST to the oauth-issuer/logout endpoint, whereas deleting the OAuthAccessToken uses the OpenShiftClient delete() endpoint. If I try to delete the kube:admin token by user UID, the delete call returns true, but I still see the token when I run `oc get oauthaccesstokens`. The original token is still present after making a subsequent successful /logout call using a good commit from the -web PR. I think the /logout call only invalidates the kube:admin token, which is why I can see a login page, but doesn't delete the token. 
```
jalaw@localhost ~/cryostat ((b208d427…))> oc get oauthaccesstokens
NAME                                                 USER NAME    CLIENT NAME                                     CREATED                EXPIRES                         REDIRECT URI                                                                           SCOPES
sha256~<insensitive token name>   kube:admin   system:serviceaccount:default:cryostat-sample   2021-12-21T19:28:26Z   2021-12-22 19:28:26 +0000 UTC   https://cryostat-sample-default.apps.jalaw0.lab.upshift.rdu2.redhat.com                user:check-access,role:cryostat-operator-oauth-client:default
```

andrewazores: Ah, I see. I wonder if `kube:admin` is being treated specially by the OAuth server to prevent people from locking themselves out of the admin account or something.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -334,67 +369,104 @@ private String getTokenFromHttpHeader(String rawHttpHeader) {
...
                             | URISyntaxException
                             | IOException
                             | MissingEnvironmentVariableException e) {
-                        throw new CompletionException(e);
+                        return CompletableFuture.failedFuture(e);
                     }
                 });
     }
 
+    private CompletableFuture<String> computeLogoutRedirectEndpoint() {
```
jan-law, line 402: In order for the frontend to access the OAuth server logout endpoint, the APIServer needs the cryostat web host added to the spec `additionalCORSAllowedOrigins`. The [docs](https://docs.openshift.com/container-platform/4.8/rest_api/config_apis/apiserver-config-openshift-io-v1.html) mention this "may be needed to access the API and the integrated OAuth server from JavaScript applications". If I don't include this spec, the /logout request fails with a CORS error.
Here's a hardcoded example:
```
apiVersion: config.openshift.io/v1
kind: APIServer
metadata:
  ...
spec:
  additionalCORSAllowedOrigins:
  - cryostat-sample-default\.apps\.jalaw0\.lab\.upshift\.rdu2\.redhat\.com
``` 

The request would look something like this:
```
client.config()
                    .apiServers()
                    .edit(
                            APIServerSpec.class,
                            spec -> spec.setAdditionalCORSAllowedOrigins(cryostatOrigin));
```


Where's the best place to put this configuration request in the backend?

andrewazores: IMO that fits best in -operator, actually. That configuration could be added when a Cryostat CR is created, and removed when the Cryostat CR is deleted.

jan-law: Sounds good, I created an issue for it

andrewazores:
This and https://github.com/cryostatio/cryostat-web/pull/355 are looking pretty good. I'll wait until https://github.com/cryostatio/cryostat-operator/issues/320 is resolved so I can also test that out - right now I'm seeing the (currently expected) CORS failure when trying to log out from the admin account.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
ebaron: Do you think calling this `/logout` endpoint will end up logging the `kube:admin` user out of the API server entirely?

jan-law: > Do you think calling this /logout endpoint will end up logging the kube:admin user out of the API server entirely?

I think the /logout endpoint only invalidates a single session. If I open multiple cryostat sessions, login with kube:admin on all of them, then logout of one session, the other sessions stay logged in. My terminal also remains logged in with kube:admin.

ebaron: Sounds good, thanks for checking!

jan-law:
Note: The "Token Authentication" section in https://cryostat.io/get-started/ should also be updated when this OAuth feature becomes available in the next Cryostat release.

jan-law:
Rebased just now. The only conflict was fixed by rebasing the web-client with the latest web-client commits on the main branch.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
jan-law: > then there are a variety of cluster roles for our operator that will help for the operator's CRDs

For future reference, I can add my new user to the same role used to request am OAuth access_token with all of the necessary permissions: 
```
cd ~/cryostat-operator/config/rbac
oc adm policy add-cluster-role-to-user cluster-admin ocadmin --rolebinding-name=oauth_client
```
If I have two cryostat tabs open, with one logged in as `kube:admin` and one as `ocadmin`, various combinations of logging in, refreshing, and logging out do not affect the other tab as expected.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -296,6 +311,26 @@ public AuthenticationScheme getScheme() {
...
+                            .orElseThrow(TokenNotFoundException::new);
+
+            if (Boolean.FALSE.equals(deleted)) {
+                throw new TokenNotFoundException();
+            }
+
+            return CompletableFuture.completedFuture(deleted);
+        } catch (Exception e) {
+            logger.error(e);
+            return CompletableFuture.failedFuture(e);
```
andrewazores, line 330: Since the `Future` returned here is always one that is created and populated with its value/exception immediately, and this is only an internal helper method, I think it can just be simplified to return a `boolean` rather than `Future<Boolean>`. It's all synchronous/single-threaded as it is, anyway, so it would simplify it down a little and at least lead to less `ExecutionException` wrapping/unwrapping in failure cases.

andrewazores (commented):
LGTM otherwise. I would be okay with deferring my one suggestion above to a follow-up PR as well if you'd prefer, since that sort of cleanup also might change exception handling and the exceptions that are reported to the web-client.

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
jan-law: I'll refactor it in this PR

src/main/java/io/cryostat/net/OpenShiftAuthManager.java
```diff
@@ -296,6 +311,23 @@ public AuthenticationScheme getScheme() {
         }
     }
 
+    private Boolean deleteToken(String token) throws IOException, TokenNotFoundException {
```
andrewazores, line 314: I think the return type here can be an unboxed `boolean`, since the `Boolean.FALSE` check lower down takes care of the nullability case - no need to potentially propagate a `null` any further.

andrewazores (approved):

PR #790, opened 2022-01-11
docs(readme): Add cgroups v2 configuration instructions to README 

### LINKED ISSUE #789: Update README with proper cgroups configuration for local podman setup
From https://github.com/cryostatio/cryostat-reports/pull/7#discussion_r775297624:

> > It took me a while but I realized that I was running `podman` using `cgroups v1` which doesn't allow configuring resource limitation for rootless containers. I enabled `cgroups v2`, which also required updating my `Docker` setup (`podman` was good as is) and now the `--memory "${MEMORY}"` flag is working.
> 
> Good find. If you have any links to documentation you found discussing this then maybe it would be worth it for us to link those in the main Cryostat README file in the "run in local podman" section or something.

hareetd:
Fixes #789

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,6 @@ Build Requirements:
 Run Requirements:
 - Kubernetes/OpenShift/Minishift, Podman/Docker, or other container platform
 
-- If running on Podman/Docker, [cgroups v2](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html) must be enabled.
-This allows resource configuration for any rootless containers running on Podman. First, update your Docker installation; instructions to do so, as well as how to enable cgroups v2 can be found [here](https://docs.docker.com/config/containers/runmetrics/#changing-cgroup-version). Finally, to ensure Podman works with cgroups v2, follow these [instructions](https://podman.io/blogs/2019/10/29/podman-crun-f31.html)*.
-
-*If your Podman runtime is set to runc v1.0.0-rc91 or later it is not necessary to change it crun as recommended in these instructions. This version of 
-runc supports cgroups v2, as stated in the [Docker documentation on the subject](https://docs.docker.com/config/containers/runmetrics/#running-docker-on-cgroup-v2). 
-
 ## BUILD
 
 ### Setup Dependencies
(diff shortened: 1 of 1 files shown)


andrewazores (commented):
I find this text a bit confusing. Podman and Docker are separate container runtimes, although Podman (and all other OCI tooling) is heavily inspired by Docker. I don't think it should be necessary to have Docker installed alongside Podman, and the `Note:` at the bottom is talking about configuring Podman's `runc` version but links to Docker documentation?

hareetd:
When I attempted to build/run https://github.com/cryostatio/cryostat-reports/pull/7 my Docker installation complained after enabling `cgroups v2` so I had to update it accordingly. This made me think that Podman and Docker were linked when in reality it's likely because I had both installed on the system. The `Note:` regarding the `Podman` runtime is because of this confusion. I'll remove any mention of Docker from my changes, thanks.

hareetd:
Makes sense, I only installed it since the OperatorHub.io test suite was refusing to work with Podman.

README.md
```diff
@@ -88,12 +88,17 @@ exec:exec@destroy-pod`
...
 * `run.sh`
 
-### Run on local podman with Grafana, jfr-datasource and demo application
+### Run on local podman with Grafana, jfr-datasource and demo application*
 * `smoketest.sh`
 
+*To run on local podman, [cgroups v2](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html) must be enabled.
+This allows resource configuration for any rootless containers running on podman. To ensure podman works with cgroups v2, follow these [instructions](https://podman.io/blogs/2019/10/29/podman-crun-f31.html).
+
+Note: If your podman runtime is set to runc v1.0.0-rc91 or later it is not necessary to change it crun as recommended in the instructions, since this version of runc supports cgroups v2. The article refers to an older version of runc.
```
andrewazores, line 100: "to change it crun" - "to change it *to* crun"?

README.md
```diff
@@ -88,12 +88,17 @@ exec:exec@destroy-pod`
...
 
-### Run on local podman
+### Run on local podman*
 * `run.sh`
 
-### Run on local podman with Grafana, jfr-datasource and demo application
+### Run on local podman with Grafana, jfr-datasource and demo application*
 * `smoketest.sh`
 
+*To run on local podman, [cgroups v2](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html) must be enabled.
```
andrewazores, line 97: Let's replace "must" with simply "should". It should still be possible to run with cgroups v1 if some user really needs that and doesn't necessarily care that they don't get rootless resource constraints with it.

andrewazores (approved):

PR #792, opened 2022-01-12
feat(websocket): Add keepalive server ping

### LINKED ISSUE #757: WebSocket keepalive/pings
@jan-law Try adding these annotations to each Ingress:
```
nginx.ingress.kubernetes.io/proxy-read-timeout: 3600
nginx.ingress.kubernetes.io/proxy-send-timeout: 3600
```
This seems to be the recommended solution from the NGINX Ingress documentation: https://kubernetes.github.io/ingress-nginx/user-guide/miscellaneous/#websockets

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/279#issuecomment-983039470_

jan-law:
Related #757 

The server will ping accepted WebSocket clients every 5 seconds to prevent an idle timeout disconnection. The client will automatically respond with a pong as per [WebSocketBase docs](https://vertx.io/docs/apidocs/io/vertx/core/http/WebSocketBase.html#writePing-io.vertx.core.buffer.Buffer-io.vertx.core.Handler-)

> Once we have an upstream release that includes application-level keepalive then the downstream configuration for increasing the timeout can be removed.

After this PR is merged, I'll add another PR to remove the extended timeout annotations in [config.md](https://github.com/cryostatio/cryostat-operator/blob/main/docs/config.md#network-options) which will close #757.

### DIFF
--- a/src/main/java/io/cryostat/messaging/MessagingModule.java
+++ b/src/main/java/io/cryostat/messaging/MessagingModule.java
@@ -63,6 +63,7 @@ public abstract class MessagingModule {
 
     static final String WS_MAX_CONNECTIONS = "WS_MAX_CONNECTIONS";
     static final String LIMBO_PRUNER = "LIMBO_PRUNER";
+    static final String KEEPALIVE_PINGER = "KEEPALIVE_PINGER";
 
     static final String MAX_CONNECTIONS_ENV_VAR = "CRYOSTAT_MAX_WS_CONNECTIONS";
     static final int MIN_CONNECTIONS = 1;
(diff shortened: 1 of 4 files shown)


andrewazores:
> The client will automatically respond with a pong as per WebSocketBase docs

Nice, the Vert.x doc links to the WebSocket RFE (https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2) and this says that any compliant implementation should behave by responding with a pong frame. So this should mean that a user's web browser will behave as expected when Vert.x sends a ping.

PR #794, opened 2022-01-13
chore(workerthreads): use daemon threads

### LINKED ISSUE #793: Worker/background thread pools should use daemon threads

andrewazores:
Fixes #793

There are also some other small fixes and improvements included, all relating to thread pools/scheduled executor services. Details in each commit message.


jan-law (approved):
Looks good! Thanks for the extra details, the refactoring makes sense.

PR #804, opened 2022-01-26
feat(jdp): add env vars to enable/configure JDP

### LINKED ISSUE #803: JDP should be optional
JDP is not useful in all environments and it may be preferred to disable it when not needed, for example to prevent any unnecessary multicast advertisting packets.

andrewazores:
Fixes #803

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,18 +103,19 @@ Note: If your podman runtime is set to runc v1.0.0-rc91 or later it is not neces
 
 ## CONFIGURATION
 
-Cryostat can be configured via the following environment variables
+Cryostat can be configured via the following environment variables:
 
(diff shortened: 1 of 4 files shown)


ebaron (approved):
Looks good, thanks Andrew! Do you think there's value in the operator explicitly setting this new environment variable to false, even though it's false by default?

ebaron:
I had another thought about this for backporting. Should we keep the default as true in the backport for backwards compatibility?

andrewazores:
That's a good idea. I'll set the default to `true`, and I think then the operator should explicitly set it to `false`.

ebaron (approved):

PR #809, opened 2022-02-02
fix(devserver): include assets in classpath

### LINKED ISSUE #749: Devserver does not have complete classpath/assets
See https://github.com/cryostatio/cryostat/pull/745

The hot-reloading devserver setup does not have the complete classpath that is available when fully built into an OCI image. As outlined in the original devserver PR above, the following are some examples of broken features:

1. `target/assets/app/resources/io/cryostat/version` file is not a loadable resource, so `GET /health` reports `unknown` version
2. Report generation is broken because the forked process cannot load its main class, and if it could, it would also be missing its classpath
3. `web-client` assets are not served by the devserver, even if they are built (`-Dcryostat.minimal=false` or unset in the `devserver.sh`)

andrewazores:
Fixes #749

### DIFF
--- a/devserver.sh
+++ b/devserver.sh
@@ -33,7 +33,6 @@ fi
 cp -r target/assets/app/resources/* target/classes
 
 flags=(
-    "-cp 'target/classes:target/assets/app/resources:@target/dependency-plugin-classpath-file'"
     "-Dcom.sun.management.jmxremote.autodiscovery=true"
     "-Dcom.sun.management.jmxremote.port=9091"
     "-Dcom.sun.management.jmxremote.rmi.port=9091"
--- a/devserver.sh
+++ b/devserver.sh
@@ -1,23 +1,39 @@
 #!/bin/sh
 
-work_dir=$(mktemp -d)
+work_dir="$(mktemp -d)"
 
-if [ -z $MVN ]; then
-    MVN=$(which mvn)
+function cleanup() {
+    podman stop cryostat-devserver-reports
+    rm -rf "${work_dir}"
+}
+trap cleanup EXIT
+
+if [ -z "${MVN}" ]; then
+    MVN="$(which mvn)"
 fi
 
 for i in archive clientlib conf templates probes; do
-    if [ -e $work_dir/$i ]; then
-        if [ ! -d $work_dir/$i ]; then
-            echo "$work_dir/$i already exists but is not a directory"
+    if [ -e "${work_dir}/${i}" ]; then
+        if [ ! -d "${work_dir}/${i}" ]; then
+            echo "${work_dir}/${i} already exists but is not a directory"
             exit 1
         fi
     else
-        mkdir $work_dir/$i
+        mkdir "${work_dir}/${i}"
     fi
 done
 
+if [ -z "$CRYOSTAT_CORS_ORIGIN" ]; then
+    "${MVN}" -DskipTests=true prepare-package
+else
+    "${MVN}" -DskipTests=true -Dcryostat.minimal=true prepare-package
+fi
+
+# HACK. The vertx-maven-plugin doesn't include target/assets on the classpath, so copy its contents into target/classes which is included
+cp -r target/assets/app/resources/* target/classes
+
 flags=(
+    "-cp 'target/classes:target/assets/app/resources:@target/dependency-plugin-classpath-file'"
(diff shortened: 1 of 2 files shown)


devserver.sh
```diff
@@ -1,23 +1,39 @@
...
-        mkdir $work_dir/$i
+        mkdir "${work_dir}/${i}"
     fi
 done
 
+if [ -z "$CRYOSTAT_CORS_ORIGIN" ]; then
+    "${MVN}" -DskipTests=true prepare-package
+else
+    "${MVN}" -DskipTests=true -Dcryostat.minimal=true prepare-package
+fi
```
hareetd, line 30: What exactly is this `if-else` block doing?

devserver.sh
```diff
@@ -1,23 +1,39 @@
...
+
+# HACK. The vertx-maven-plugin doesn't include target/assets on the classpath, so copy its contents into target/classes which is included
+cp -r target/assets/app/resources/* target/classes
+
 flags=(
+    "-cp 'target/classes:target/assets/app/resources:@target/dependency-plugin-classpath-file'"
```
hareetd, line 36: Since we already copied the contents of `target/assets/app/resources/` to `target/classes` on line 33, why do we need to specify `target/assets/app/resources/` in the JVM classpath? Is it because the `vertx-maven-plugin` classpath configuration is only concerned with the proper building of our application, while the JVM classpath has to do with running the built application (i.e. their respective classpaths fulfill different purposes)?

devserver.sh
andrewazores: > However, in the original dev-server PR we were able to launch a hot-reloading instance of the web-client without involving the CRYOSTAT_CORS_ORIGIN flag

https://github.com/cryostatio/cryostat/pull/809/files#diff-ad8ed4ad8729e9fd13d92c8775318283cf8f2b58c1a49edd9eb3a0c195257cd1L35

Previously, the devserver script just always had the CORS origin environment variable set to the "standard" value.

> What exactly is this if-else block doing?

If you're using a separate web-client devserver then you must set the CORS origin to allow it to make requests to the backend. If that's the case then there is no need for the backend devserver to also build the web-client frontend assets. If you are not using a frontend devserver and have no CORS setup, then it does build the web-client frontend assets so that it can serve them to you.

devserver.sh
andrewazores: Whoops. I think this line is actually no longer needed - I was experimenting with different ways to set the classpath for the main JVM and for the forked Vert.x JVM before, but eventually just settled on the "cp -r" hack instead. I'll remove this.

devserver.sh
andrewazores: `prepare-package` does everything required for `package`, except it doesn't do the final step of actually creating the OCI image. Classes are generated, classes are compiled, resources are built/generated and copied into the correct locations, etc. This basically populates the `target/classes` and `target/assets` directories. With `-Dcryostat.minimal=true` the front-end build portion is skipped, so `target/assets` won't contain the HTML/JS/CSS of -web.

andrewazores:
Hmm. Looks like I broke this at some point - the devserver starts up successfully, but if you edit a source file then the recompile fails. Not sure what I did to break it.

andrewazores:
> Hmm. Looks like I broke this at some point - the devserver starts up successfully, but if you edit a source file then the recompile fails. Not sure what I did to break it.

Nope. Just a silly mistake. I tried to run it with my usual `mvnd` instead of `mvn`, and that didn't work. If I properly do `MVN=/usr/bin/mvn sh devserver.sh` then I think it's working.

jan-law (commented):
Report generation and the minimal build setup is working well for me. Did you intend to leave the version string on the About page blank?

andrewazores:
> Report generation and the minimal build setup is working well for me. Did you intend to leave the version string on the About page blank?

Looks like that's actually a bug on the frontend: https://github.com/cryostatio/cryostat-web/blob/16410a78be2a31b07e71739f8252b967504eb5f6/src/app/Shared/Services/Api.service.tsx#L112

Since there is no Grafana configuration, the `Api.service` fails the request, causing the "Grafana configuration not found" error popup. But this also prevents the version string from being correctly populated.

devserver.sh
hareetd: Makes sense, thanks.

jan-law (approved):

PR #810, opened 2022-02-02
feat(notifications): Add "Recording Stopped" notification on implied stop

### LINKED ISSUE #781: "Recording Stopped" notification on implied stop
Depends on #780

When Cryostat creates a new recording, it knows if that recording has a fixed duration. If it does, then Cryostat can easily infer the timestamp of when that recording is expected to finish by adding the recording duration to the current time. Cryostat does not receive a notification from the remote target JVM when recordings stop but it could reconnect to the target JVM after the previously calculated duration has elapsed (or shortly after, give a couple of extra seconds perhaps to minimize the chance of any timestamp drift between Cryostat and target influencing the result) and check whether the expected recording is still present and is in the `STOPPED` state. If it is, then Cryostat should emit a notification informing clients that this has occurred.

jan-law:
Fixes #781

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
@@ -77,7 +77,7 @@ class TargetRecordingPatchStop {
                         connection.getService().stop(descriptor.get());
                         recordingTargetHelper.cancelScheduledNotificationIfExists(recordingName);
                         recordingTargetHelper.notifyRecordingStopped(
-                        recordingName, connectionDescriptor.getTargetId());
+                                recordingName, connectionDescriptor.getTargetId());
                         return null;
                     } else {
                         throw new HttpStatusException(
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
@@ -75,6 +75,9 @@ class TargetRecordingPatchStop {
                                     .findFirst();
                     if (descriptor.isPresent()) {
                         connection.getService().stop(descriptor.get());
+                        recordingTargetHelper.cancelScheduledNotificationIfExists(recordingName);
+                        recordingTargetHelper.notifyRecordingStopped(
+                        recordingName, connectionDescriptor.getTargetId());
                         return null;
                     } else {
                         throw new HttpStatusException(
@@ -85,9 +88,5 @@ class TargetRecordingPatchStop {
                 });
         ctx.response().setStatusCode(200);
         ctx.response().end();
-
-        recordingTargetHelper.cancelScheduledNotificationIfExists(recordingName);
-        recordingTargetHelper.notifyRecordingStopped(
-                recordingName, connectionDescriptor.getTargetId());
     }
 }
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
@@ -37,14 +37,10 @@
  */
 package io.cryostat.net.web.http.api.v1;
 
-import java.util.Optional;
-
 import javax.inject.Inject;
 
-import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
-
 import io.cryostat.net.ConnectionDescriptor;
-import io.cryostat.net.TargetConnectionManager;
+import io.cryostat.recordings.RecordingNotFoundException;
 import io.cryostat.recordings.RecordingTargetHelper;
 
 import io.vertx.ext.web.RoutingContext;
@@ -52,40 +48,21 @@ import io.vertx.ext.web.handler.impl.HttpStatusException;
 
 class TargetRecordingPatchStop {
 
-    private final TargetConnectionManager targetConnectionManager;
     private final RecordingTargetHelper recordingTargetHelper;
 
     @Inject
-    TargetRecordingPatchStop(
-            TargetConnectionManager targetConnectionManager,
-            RecordingTargetHelper recordingTargetHelper) {
-        this.targetConnectionManager = targetConnectionManager;
+    TargetRecordingPatchStop(RecordingTargetHelper recordingTargetHelper) {
         this.recordingTargetHelper = recordingTargetHelper;
     }
 
     void handle(RoutingContext ctx, ConnectionDescriptor connectionDescriptor) throws Exception {
         String recordingName = ctx.pathParam("recordingName");
 
-        targetConnectionManager.executeConnectedTask(
-                connectionDescriptor,
-                connection -> {
-                    Optional<IRecordingDescriptor> descriptor =
-                            connection.getService().getAvailableRecordings().stream()
-                                    .filter(recording -> recording.getName().equals(recordingName))
-                                    .findFirst();
-                    if (descriptor.isPresent()) {
-                        connection.getService().stop(descriptor.get());
-                        recordingTargetHelper.cancelScheduledNotificationIfExists(recordingName);
-                        recordingTargetHelper.notifyRecordingStopped(
-                                recordingName, connectionDescriptor.getTargetId());
-                        return null;
-                    } else {
-                        throw new HttpStatusException(
-                                404,
-                                String.format(
-                                        "Recording with name \"%s\" not found", recordingName));
-                    }
-                });
+        try {
+            recordingTargetHelper.stopRecording(connectionDescriptor, recordingName).get();
+        } catch (RecordingNotFoundException e) {
+            throw new HttpStatusException(404, e);
+        }
         ctx.response().setStatusCode(200);
         ctx.response().end();
     }
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
@@ -37,7 +37,6 @@
  */
 package io.cryostat.net.web.http.api.v1;
 
-import java.util.Map;
 import java.util.Optional;
 
 import javax.inject.Inject;
@@ -47,7 +46,7 @@ import org.openjdk.jmc.rjmx.services.jfr.IRecordingDescriptor;
 import io.cryostat.messaging.notifications.NotificationFactory;
 import io.cryostat.net.ConnectionDescriptor;
 import io.cryostat.net.TargetConnectionManager;
-import io.cryostat.net.web.http.HttpMimeType;
+import io.cryostat.recordings.RecordingTargetHelper;
 
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.impl.HttpStatusException;
@@ -56,15 +55,16 @@ class TargetRecordingPatchStop {
 
     private final TargetConnectionManager targetConnectionManager;
     private final NotificationFactory notificationFactory;
-
-    private static final String NOTIFICATION_CATEGORY = "RecordingStopped";
+    private final RecordingTargetHelper recordingTargetHelper;
 
     @Inject
     TargetRecordingPatchStop(
             TargetConnectionManager targetConnectionManager,
-            NotificationFactory notificationFactory) {
+            NotificationFactory notificationFactory,
+            RecordingTargetHelper recordingTargetHelper) {
         this.targetConnectionManager = targetConnectionManager;
         this.notificationFactory = notificationFactory;
+        this.recordingTargetHelper = recordingTargetHelper;
     }
 
     void handle(RoutingContext ctx, ConnectionDescriptor connectionDescriptor) throws Exception {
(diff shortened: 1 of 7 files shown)


src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -365,6 +408,43 @@ public IRecordingDescriptor startRecording(
...
+                                                getDescriptorByName(connection, recordingName);
+
+                                        boolean recordingStopped =
+                                                desc.isPresent()
+                                                        && !desc.stream()
+                                                                .filter(
+                                                                        recording ->
+                                                                                recording.getState()
+                                                                                        == RecordingState
+                                                                                                .STOPPED)
```
andrewazores, line 430: I think this could be tidied up just slightly:

```java
desc.stream()
  .map(IRecordingDescriptor::getState)
  .filter(s -> s.equals(RecordingState.STOPPED))
  .count()
```

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -314,6 +340,23 @@ public IRecordingDescriptor startRecording(
...
+                .metaType(HttpMimeType.JSON)
+                .message(Map.of("recording", recordingName, "target", targetId))
+                .build()
+                .send();
+    }
+
+    public void cancelScheduledNotificationIfExists(String stoppedRecordingName) {
+        if (scheduledStopNotifications.containsKey(stoppedRecordingName)) {
+            scheduledStopNotifications.get(stoppedRecordingName).cancel(true);
+            scheduledStopNotifications.remove(stoppedRecordingName);
```
andrewazores, line 356: Not important, but this could be written as:

```java
Future f = notifications.remove(name);
if (f != null) {
  f.cancel(true);
}
```

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -314,6 +340,23 @@ public IRecordingDescriptor startRecording(
...
+                .createBuilder()
+                .metaCategory(STOP_NOTIFICATION_CATEGORY)
+                .metaType(HttpMimeType.JSON)
+                .message(Map.of("recording", recordingName, "target", targetId))
+                .build()
+                .send();
+    }
+
+    public void cancelScheduledNotificationIfExists(String stoppedRecordingName) {
+        if (scheduledStopNotifications.containsKey(stoppedRecordingName)) {
```
andrewazores, line 354: Also, I think we might want to add some thread synchronization when we're accessing this map. I believe it's possible that a client, or multiple clients, can send HTTP API requests in very tight timing to start and immediately stop a recording, which could lead to the Vert.x worker thread that handles scheduling this notification to still be underway doing it at this point, then another worker thread stops/deletes the recording, and then the first thread continues to schedule the notification for that recording after it has already been stopped or deleted. It would be very hard to replicate the behaviour and it would be a minor bug, but I think it's still possible and worth handling.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -314,6 +339,23 @@ public IRecordingDescriptor startRecording(
...
+        notificationFactory
+                .createBuilder()
+                .metaCategory(STOP_NOTIFICATION_CATEGORY)
+                .metaType(HttpMimeType.JSON)
+                .message(Map.of("recording", recordingName, "target", targetId))
+                .build()
+                .send();
+    }
+
+    public synchronized void cancelScheduledNotificationIfExists(String stoppedRecordingName) {
```
andrewazores, line 352: Looks better. I would be cautious of using the simple `synchronized` method-level keyword here. This uses the `RecordingTargetHelper` instance itself as the lock instance. For now, since the only two places where that lock instance is used are the two notifications-related methods, that's fine. But if in the future we decide we need to synchronize accesses to some other resource in here - say the `reportService` for some reason - then we don't want those two separate resources causing threads to contend for the same lock when they really are likely to be able to both proceed safely.

I think it might be better ("safer" as well as more explicit) to use the map instance as the lock.

```java
void cancelNotification(String name) {
  synchronized (notifications) {
    var f = notifications.remove(name);
    if (f != null) f.cancel(true);
  }
}

void scheduleNotification(String name) {
  synchronized (notifications) {
    notifications.put(name, scheduler.schedule());
  }
}

void stopRecording(String name) {
  synchronized (notifications) {
    tcm.executeConnectedTask(() -> cancelTheRecording());
    cancelNotification(name);
  }
}

void startRecording(String name) {
  synchronized (notifications) {
    tcm.executeConnectedTask(() -> startRecording());
    scheduleNotification(name);
  }
}
```

jan-law: > But if in the future we decide we need to synchronize accesses to some other resource in here - say the reportService for some reason - then we don't want those two separate resources causing threads to contend for the same lock when they really are likely to be able to both proceed safely.

Thanks, this makes sense.

In your example, why does the call to `scheduleNotification` in `startRecording()` need to be wrapped in a synchronized block as well as inside the `scheduleNotification()` method itself? Is it because `startRecording` or `stopRecording` need to acquire the lock before entering `scheduleNotification`?

andrewazores: I was thinking it makes sense that no other thread should be modifying the notifications map while there is a thread actively working on connecting to a target and starting/stopping a recording. If we only lock within `scheduleNotification` then two threads can both be working on connecting to a target with one starting a recording and the other trying to stop the same recording erroneously, but depending on thread scheduling we can end up with the recording starting, then being stopped, then the stop notification being scheduled. If we lock earlier, before even connecting to the target, then this can't happen - the stop can't interleave between the start and the scheduling.

Perhaps this should actually be a map of locks, with a lock for each targetId, so we don't end up reducing request throughput here when making parallel requests to separate targets. Acquiring such a target-specific lock before opening the connection to the target, then connecting, starting/stopping the recording, and emitting/scheduling a stop notification preserves the nice atomicity we want here without also creating a chokepoint where threads are blocked waiting for each other to do things on different targets.

jan-law:
The latest CI build failure looks like #666 again:
```
Feb 04, 2022 6:19:57 PM io.cryostat.core.log.Logger info
INFO: io.cryostat.net.reports.SubprocessReportGenerator processing report
java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
	at io.cryostat.net.reports.SubprocessReportGenerator.generateReportFromFile(SubprocessReportGenerator.java:336)
	at io.cryostat.net.reports.SubprocessReportGenerator.main(SubprocessReportGenerator.java:292)
	at org.openjdk.jmc.flightrecorder.parser.synthetic.SyntheticAttributeExtension$ModuleExportSink.addEvent(SyntheticAttributeExtension.java:206)
...
Feb 04, 2022 6:19:57 PM io.cryostat.core.log.Logger error
SEVERE: HTTP 500: Internal Server Error
Caused by: java.util.concurrent.ExecutionException: io.cryostat.net.reports.SubprocessReportGenerator$SubprocessReportGenerationException: [6] An unspecified unexpected exception occurred.
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2022)
...
Feb 04, 2022 6:19:57 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:53658): GET /api/beta/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat-itests:9091%2Fjmxrmi/reports/ReportJwtDownloadIT 500 981ms
```

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
jan-law: It looks like starting/stopping a recording and scheduling a stop notification always calls `executeConnectedTask()`, so I decided to put the map of locks in TargetConnectionManager. `executeConnectedTask()` will aqcuire a target lock before calling `task.execute()`.

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchStop.java
```diff
@@ -79,6 +75,9 @@ void handle(RoutingContext ctx, ConnectionDescriptor connectionDescriptor) throw
                                     .findFirst();
                     if (descriptor.isPresent()) {
                         connection.getService().stop(descriptor.get());
+                        recordingTargetHelper.cancelScheduledNotificationIfExists(recordingName);
+                        recordingTargetHelper.notifyRecordingStopped(
+                                recordingName, connectionDescriptor.getTargetId());
```
andrewazores, line 80: What if the `connection.getService().stop(desc)` call was extracted out and refactored into a method like `RecordingTargetHelper#stopRecording()`? We already have start/delete/get in there, so it would be a nice cleanup. I think it would also be nice to avoid having `notifyRecordingStopped` and `cancelScheduledNotificationIfExists` as public methods - those should be private and called by the helper whenever it makes sense to do so, not called by other callsites like HTTP handlers.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -314,6 +370,23 @@ public IRecordingDescriptor startRecording(
...
+        notificationFactory
+                .createBuilder()
+                .metaCategory(STOP_NOTIFICATION_CATEGORY)
+                .metaType(HttpMimeType.JSON)
+                .message(Map.of("recording", recordingName, "target", targetId))
+                .build()
+                .send();
+    }
+
+    public void cancelScheduledNotificationIfExists(String stoppedRecordingName) {
```
andrewazores, line 383: This can be private now too, right?

jan-law: Whoops, fixed now.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -314,6 +370,23 @@ public IRecordingDescriptor startRecording(
...
+        notificationFactory
+                .createBuilder()
+                .metaCategory(STOP_NOTIFICATION_CATEGORY)
+                .metaType(HttpMimeType.JSON)
+                .message(Map.of("recording", recordingName, "target", targetId))
+                .build()
+                .send();
+    }
+
+    private void cancelScheduledNotificationIfExists(String stoppedRecordingName) {
```
andrewazores, line 383: I think this needs to be called in `deleteRecording` too, right?

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -365,6 +438,42 @@ public IRecordingDescriptor startRecording(
...
+                                                    connectionDescriptor.getTargetId());
+                                        }
+
+                                        return desc;
+                                    });
+                        },
+                        delay + TIMESTAMP_DRIFT_SAFEGUARD,
+                        TimeUnit.MILLISECONDS);
+
+        scheduledStopNotifications.put(recordingName, scheduledFuture);
```
andrewazores, line 474: I think the notifications need to be mapped not only by the recording name but also the name of the associated target, since names are only unique within a target. So either this should be some computed key that can be unambiguously taken back apart into a targetId part and a recordingName part, or it should be a `Map<TargetId<Map<RecordingName, ScheduledFuture>>>`

andrewazores:
The expected happy-path workflow seems to be working well, just those last two kinda corner cases above left. This will go nicely with https://github.com/cryostatio/cryostat-web/issues/368 too - the active recordings table will update running recordings to the stopped state automatically, even without view refresh enabled.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
jan-law: Are TargetIds unique?

andrewazores: Yes - if two targetIds are the same then they are the same JMX service URL, or other future identifier/URL for a target.

It is possible that one target can be referred to by two or more different targetIds, however. For example, when you use our standard development setup with JDP, Cryostat can discover itself there with the JMX service URL `cryostat:9091`. You can then add a custom target for `localhost:0`, which tells the JVM to connect to itself locally. There are various other ways such a scenario can arise, like having multiple DNS records for different hostnames pointing to the same IP where a JVM is listening.

I think in those cases, at least for now, that'll be a bit of a caveat emptor. To properly fix that we would need to actually perform DNS queries whenever a targetId is discovered or added and ensure that it's really referring to a distinct JVM from any others we already know about, and in the future we might also need to know how to map JMX service URLs to Jolokia URLs etc.

I think this would only ever present a "problem" if a client started a recording `r1` using `targetId1` that refers to `jvmA`, and then deleted `r1` using `targetId2` that also refers to `jvmA`, or other strange scenarios like that. In this case, we would end up deleting the recording but not the scheduled notification, so we would waste a bit of time/energy in the future connecting back to `targetId1` and checking if `r1` is still there, even though we "should" know that we've already deleted it. The notification still wouldn't be emitted in the end, so I don't think it's really a real problem.

jan-law: Sounds good, thanks

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -314,6 +371,23 @@ public IRecordingDescriptor startRecording(
...
+                .createBuilder()
+                .metaCategory(STOP_NOTIFICATION_CATEGORY)
+                .metaType(HttpMimeType.JSON)
+                .message(Map.of("recording", recordingName, "target", targetId))
+                .build()
+                .send();
+    }
+
+    private void cancelScheduledNotificationIfExists(String stoppedRecordingName) {
+        var f = scheduledStopNotifications.remove(stoppedRecordingName);
```
andrewazores, line 385: This needs the targetId included, too

andrewazores (approved):

PR #832, opened 2022-02-11
fix(notifications): unlimited max simultaneous websocket connections

### LINKED ISSUE #830: WebSocket notification channel maximum and default connections should be increased
The notification channel has a very limited maximum of 64 connections and defaults to only 2 if no value is set by environment variable. The default is certainly too small and the maximum may also be too small given the extra utility the notification channel has gained over time.

Better to default to no limit (or Integer.MAX_VALUE) and allow the end user to set a maximum if they desire.

andrewazores:
Fixes #830


src/main/java/io/cryostat/messaging/MessagingModule.java
```diff
@@ -99,23 +96,20 @@ static MessagingServer provideWebSocketMessagingServer(
...
                             env.getEnv(
                                     Variables.MAX_CONNECTIONS_ENV_VAR,
-                                    String.valueOf(DEFAULT_MAX_CONNECTIONS)));
-            if (maxConn > MAX_CONNECTIONS) {
-                logger.info("Requested maximum WebSocket connections {} is too large.", maxConn);
-                return MAX_CONNECTIONS;
+                                    String.valueOf(Integer.MAX_VALUE)));
+            if (count <= 0) {
+                logger.warn(
+                        "{} was set to {} - ignoring", Variables.MAX_CONNECTIONS_ENV_VAR, count);
```
jan-law, line 106: Where does this log line get printed? I tried running `CRYOSTAT_MAX_WS_CONNECTIONS=0 sh run.sh
`. The max connections is properly set to `Integer.MAX_VALUE` and I don't see the log message in my logs ([logs](https://gist.github.com/jan-law/75fefd0fde4db733d2eb9b479bdec4af)).

jan-law (commented):
Works as expected.

It'd be a good idea to update the operator docs before the 2.1 release too:
https://github.com/cryostatio/cryostat-operator/blob/main/docs/config.md#cryostat-client-options
https://github.com/cryostatio/cryostat-operator/blob/main/api/v1beta1/cryostat_types.go#L83

src/main/java/io/cryostat/messaging/MessagingModule.java
andrewazores: `run.sh` doesn't handle that env var - I'll add it in.

hareetd (approved):
Looks good!

PR #835, opened 2022-02-17
feat(recordings): Add recording metadata labels

jan-law:
Related #709
Depends on https://github.com/cryostatio/cryostat-web/pull/383

Adds a patch handler and helper functions to optionally add labels when starting a recording on a target. The patch handler is used for editing labels for existing active or archived recordings.


src/main/java/io/cryostat/recordings/RecordingMetadataManager.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+
+public class RecordingMetadataManager {
+
+    private final Path recordingMetadataDir;
+    private final FileSystem fs;
+    private final Gson gson;
+    private final Logger logger;
+
+    private final Map<Pair<String, String>, String> recordingLabelsMap;
```
jan-law, line 62: This `recordingLabelsMap` stores the labels for a recording as a stringified JSON object. All of my unit tests are passing, however if I try to manually start a recording with labels, I can see this `recordingLabelsMap` was updated with the labels from the `201` response. A subsequent GET recordings/ shows the labels for that recording are empty (`labels: ""`). I also tried storing the labels as a `Map<String, String>` with the same curl output.

This is my latest attempt. Do you have any ideas as to why the map value becomes empty?
```
jalaw@localhost ~/cryostat-web (recording-labels)> curl -v --insecure -F recordingName=test -F events=template=ALL -F duration=10 -F labels="{\"KEY\":\"VALUE\",\"key.2\":\"some.value\",\"key3\":\"1234\"}" http://localhost:8181/api/v1/targets/localhost:0/recordings

Trying ::1:8181...
Connected to localhost (::1) port 8181 (#0)
POST /api/v1/targets/localhost:0/recordings HTTP/1.1
Host: localhost:8181
User-Agent: curl/7.76.1
Accept: /
Content-Length: 503
Content-Type: multipart/form-data; boundary=------------------------07df1e96a15e09f6

We are completely uploaded and fine
Mark bundle as not supporting multiuse
< HTTP/1.1 201 Created
< location: /test
< content-length: 472
<
Connection #0 to host localhost left intact
{"downloadUrl":"[http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/recordings/test",
"reportUrl":"http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/reports/test",
"labels":"{\"KEY\":\"VALUE\",\"key.2\":\"some.value\",\"key3\":\"1234\"}",
"id":1,"name":"test","state":"RUNNING","startTime":1645125006896,"duration":10000,"continuous":false,"toDisk":true,"maxSize":0,"maxAge":0}⏎](http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/recordings/test%22,%22reportUrl%22:%22http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/reports/test%22,%22labels%22:%22%7B/%22KEY/%22:/%22VALUE/%22,/%22key.2/%22:/%22some.value/%22,/%22key3/%22:/%221234/%22%7D%22,%22id%22:1,%22name%22:%22test%22,%22state%22:%22RUNNING%22,%22startTime%22:1645125006896,%22duration%22:10000,%22continuous%22:false,%22toDisk%22:true,%22maxSize%22:0,%22maxAge%22:0%7D%E2%8F%8E) jalaw@localhost ~/cryostat-web (recording-labels)> curl -v --insecure http://localhost:8181/api/v1/targets/localhost:0/recordings
Trying ::1:8181...
Connected to localhost (::1) port 8181 (#0)
GET /api/v1/targets/localhost:0/recordings HTTP/1.1
Host: localhost:8181
User-Agent: curl/7.76.1
Accept: /

Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 412
<
Connection #0 to host localhost left intact
[{"downloadUrl":"[http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/recordings/test",
"reportUrl":"http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/reports/test",
"labels":"","id":1,"name":"test","state":"STOPPED","startTime":1645125006896,"duration":10000,"continuous":false,"toDisk":true,"maxSize":0,"maxAge":0}]](http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/recordings/test%22,%22reportUrl%22:%22http://localhost:8181/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/reports/test%22,%22labels%22:%22%22,%22id%22:1,%22name%22:%22test%22,%22state%22:%22STOPPED%22,%22startTime%22:1645125006896,%22duration%22:10000,%22continuous%22:false,%22toDisk%22:true,%22maxSize%22:0,%22maxAge%22:0%7D]%E2%8F%8E)
```

src/main/java/io/cryostat/recordings/RecordingsModule.java
```diff
@@ -148,4 +157,29 @@ static ScheduledExecutorService provideNotificationScheduler() {
         Runtime.getRuntime().addShutdownHook(new Thread(ses::shutdown));
         return ses;
     }
+
+    @Provides
+    // FIXME which path to use?
```
andrewazores, line 162: Add a `@Singleton` here. Without that, Dagger will inject a new manager instance anywhere you've requested it. This means each of your API handlers has its own manager instance, with its own in-memory labels map - which is why when you POST to create a recording the response indicates the labels were captured, but when you GET after they appear to be missing.

jan-law: Thanks! I'll let you know when I'm ready for another review.

src/test/java/io/cryostat/recordings/RecordingMetadataManagerTest.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @Test
+    void shouldParseAndStoreLabelsInRecordingLabelsMap() throws Exception {
+        String targetId = "someTarget";
+        String recordingName = "someRecording";
+        String labels = "{\"KEY\":\"VALUE\",\"key.2\":\"some.value\",\"key3\":\"1234\"}";
+
+        Path mockPath = Mockito.mock(Path.class);
+        Mockito.when(recordingMetadataDir.resolve(Mockito.anyString())).thenReturn(mockPath);
```
jan-law, line 85: I'm getting a null pointer exception at this line when attempting to mock the `fs.writeString()` call. I tried following the rule registry test [here](https://github.com/cryostatio/cryostat/blob/main/src/test/java/io/cryostat/rules/RuleRegistryTest.java#L132) and am not sure what to try next. Any ideas?

```
 at io.cryostat.recordings.RecordingMetadataManagerTest.shouldOverwriteLabelsForExistingLabelEntries([RecordingMetadataManagerTest.java:179]())
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke([NativeMethodAccessorImpl.java:62]())
 at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke([DelegatingMethodAccessorImpl.java:43]())
 at java.base/java.lang.reflect.Method.invoke([Method.java:566]())
 at org.junit.platform.commons.util.ReflectionUtils.invokeMethod([ReflectionUtils.java:688]())
```

src/test/java/itest/RecordingMetadataIT.java
```diff
@@ -0,0 +1,131 @@
...
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import itest.bases.StandardSelfTest;
+import itest.util.ITestCleanupFailedException;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Test;
+import org.junit.jupiter.api.BeforeAll;
+
+public class RecordingMetadataIT extends StandardSelfTest {
```
jan-law, line 55: What am I missing in order to run this itest? 
`mvn -Dcryostat.minimal -DskipUTs -Dit.test=RecordingMetadataIT.java clean verify` doesn't run this new itest:
```
[INFO] --- maven-failsafe-plugin:2.22.2:integration-test (failsafe-tests) @ cryostat ---
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```

src/test/java/io/cryostat/recordings/RecordingMetadataManagerTest.java
andrewazores: I think you're just missing the `@ExtendWith` annotation, like here: https://github.com/cryostatio/cryostat/blob/main/src/test/java/io/cryostat/rules/RuleRegistryTest.java#L71

Without that, the `@Mock` on your fields doesn't do anything, so `recordingMetadataDir` is actually an uninitialized null when you get to this line.

src/test/java/itest/RecordingMetadataIT.java
andrewazores: I think you have the wrong `@Test` imported - `import org.junit.Test;` vs `import org.junit.jupiter.api.Test;`.

Wonder if there's an extra test dependency we can remove from the test classpath to avoid running into this. IIRC this is some difference between JUnit major versions, but I'm not sure why both are available and if we can get around that.

src/test/java/io/cryostat/recordings/RecordingMetadataManagerTest.java
jan-law: Thanks!

src/test/java/itest/RecordingMetadataIT.java
jan-law: Ah thanks, that fixed it.

src/main/java/io/cryostat/net/web/http/api/v2/HttpApiV2Module.java
```diff
@@ -184,4 +184,24 @@ abstract RequestHandler bindTargetCredentialsDeleteHandler(
...
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindRecordingMetadataPatchBodyHandler(
+            RecordingMetadataPatchBodyHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindTargetRecordingMetadataPatchBodyHandler(
+            TargetRecordingMetadataPatchBodyHandler handler);
```
andrewazores, line 206: Could you move these new handlers from the V2 module and API v2.1 namespace into the Beta module? I think we might want to hold off on including this in 2.1.0 since that's coming up very soon and this feature isn't in the current release planning roadmap. If we keep it in Beta then we're free to make more changes to it before the next major release where we promote it to the main API.

src/main/java/io/cryostat/net/web/http/api/v2/RecordingMetadataPatchHandler.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public String path() {
+        return basePath() + PATH;
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return Set.of(ResourceAction.READ_TARGET, ResourceAction.READ_RECORDING);
```
andrewazores, line 97: Since this handler operates on archived recordings, I don't think it needs the `READ_TARGET` permission. I think `UPDATE_RECORDING` might make sense, if we consider the metadata/labels of a recording to be a part of the recording resource itself.

src/main/java/io/cryostat/net/web/http/api/v2/TargetRecordingMetadataPatchHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    public String path() {
+        return basePath() + PATH;
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return Set.of(ResourceAction.READ_TARGET, ResourceAction.READ_RECORDING);
```
andrewazores, line 104: As above, I think this should have the `UPDATE_RECORDING` permission as well.

src/main/java/io/cryostat/recordings/RecordingsModule.java
```diff
@@ -148,4 +160,29 @@ static ScheduledExecutorService provideNotificationScheduler() {
         Runtime.getRuntime().addShutdownHook(new Thread(ses::shutdown));
         return ses;
     }
+
+    @Provides
+    @Singleton
+    static RecordingMetadataManager provideRecordingMetadataManager(
+            @Named(ConfigurationModule.CONFIGURATION_PATH) Path confDir,
```
andrewazores, line 167: I'm not sure if the `CONFIGURATION_PATH` is the best parent for this purpose, but since we eventually want to move this to be backed by a proper database of some form rather than just manually managing JSON files on disk, I think it's okay for now. Please just leave a comment here that this is a `FIXME` and the manager should either be updated to use a database connection, or a new filesystem path should be created and used for this kind of purpose instead.

andrewazores (commented):
Looks great overall, thanks for all this! Other than the small details I've left comments about, I have one higher-level internal implementation detail concern here. It seems that the API is defined to accept the labels as a string, but that string must be a JSON object/map. But, internally, the labels are passed around as the raw String, and when included in an API response (as a serialized recording descriptor), the labels field is actually again simply a String that has contents that appear to be JSON - but it will be wrapped in double quotes in the response as a String, not an actual JSON object.

I think it would be best to parse the client's request body as JSON right away in the handlers/manager, expecting it to be mappable to a Java `Map<String, String>`. If it isn't then that is a `400` response. If it does, then the labels object that is passed around internally should be the Map, and this is also what should be kept as a field reference in the serializable recording descriptor. This can still be serialized to JSON for storage on disk, and when included as an API response as a serialized recording descriptor, it will be nicely embedded as a JSON object/map rather than a String that looks like further encoded JSON.

src/test/java/io/cryostat/net/web/http/api/beta/TargetRecordingMetadataPatchHandlerTest.java
```diff
@@ -174,100 +179,98 @@ void shouldUpdateLabels() throws Exception {
...
+                    .message(
+                            Map.of(
+                                    "target",
+                                    targetId,
+                                    "recording",
+                                    recordingName,
+                                    "labels",
+                                    labels));
             Mockito.verify(notificationBuilder).build();
             Mockito.verify(notification).send();
```
jan-law, line 226: Before I replace the internal `labels` strings with maps, I wanted to update the unit tests to make sure everything else works properly. I'm getting an error with the `notificationBuilder` here. The notification stubs in `setup()` are the same as other tests. What am I missing?
```
Expected [notificationBuilder.message(

    {"recording" = "someRecording", "target" = "fooTarget", "labels" = "{key=value}"}

);] but was [notificationBuilder.metaCategory(

    "RecordingMetadataUpdated"

);

notificationBuilder.metaType(

    JSON

);

notificationBuilder.message(

    {"recordingName" = "someRecording", "target" = "fooTarget", "labels" = "{key=value}"}

);

notificationBuilder.build(

    

);]
```

andrewazores: That looks a little funny. I wonder if it's something silly like the `labels` in one is actually a string literal, but in the other it's some other expected object reference value that gets formatted the same way by its `toString()` to print out that error message. A Map's `toString()` might look like that for example, but the Map is not actually equal to the string.

jan-law: Ah whoops, I changed the message "recording" to "recordingName" - thanks for looking at this anyway

src/main/java/io/cryostat/recordings/RecordingMetadataManager.java
```diff
@@ -0,0 +1,202 @@
...
+                .filter(Objects::nonNull)
+                .map(reader -> gson.fromJson(reader, StoredRecordingMetadata.class))
+                .forEach(
+                        srm ->
+                                recordingLabelsMap.put(
+                                        Pair.of(srm.getTargetId(), srm.getRecordingName()),
+                                        srm.getLabels()));
+    }
+
+    public Future<Map<String, String>> addRecordingLabels(
```
andrewazores, line 106: Could you add an `addRecordingLabels(String recordingName, Map<String, String> labels)` method definition that just calls through to this one with that magic `ARCHIVES` constant? I think this will clean up the interface a bit, so that calling code doesn't need to know about the "targetId=='archives'" special case so explicitly.

andrewazores: Also, I think this method's name is a little misleading - should be it named `setRecordingLabels` instead of `addRecordingLabels`, since calling this will over-write the previous value, rather than joining the old map with the new map?

src/main/java/io/cryostat/net/web/http/api/beta/RecordingMetadataPatchBodyHandler.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public int getPriority() {
+        return DEFAULT_PRIORITY - 1;
+    }
+
+    @Override
+    public HttpMethod httpMethod() {
+        return HttpMethod.PATCH;
```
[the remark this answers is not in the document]
andrewazores: Along with this I would also propose that the `HyperlinkedSerializableRecordingDescriptor` and any other response objects that contain these new labels should contain a more generalized "metadata" field, which currently contains just the one nested "labels" field. This should be more extensible for the future.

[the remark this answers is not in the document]
jan-law: Yeah, `POST` makes sense. I was originally considering the labels as part of a recording state, but since the existing `PATCH` handlers can create new labels, it makes sense to rename them to `POST`.

For your `GET` handler examples, did you mean to add handlers specifically for metadata? As of now, the labels are returned with the recording `GET` handlers.

src/main/java/io/cryostat/recordings/RecordingMetadataManager.java
jan-law: Good point, should be fixed now.

src/main/java/io/cryostat/net/web/http/api/beta/RecordingMetadataPatchBodyHandler.java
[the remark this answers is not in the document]
andrewazores: I'm still tossing around the idea of `GET`s for `/metadata` and `/metadata/labels`. If we don't have these then the only way to get metadata for a recording is to `GET /recordings`/`GET /targets/:targetId/recordings` and then filter that result by the desired name on the client side. On the other hand, I don't like that this creates an implication or expectation that the API URLs are hierarchical in a way that totally mirrors the JSON response structures.

Really, the root problem here is that I initially designed the `GET /api/v1//targets/:targetId/recordings/:recordingName` to respond with the JFR binary file directly, and not the recording descriptor object. The binary should have been sent either by making a request on a different endpoint, or by the client setting the request header `Content-Type`, or something along those lines. This fix is way out of scope of this current PR though.

I suppose since we have no direct use case of our own for the `GET /metadata` endpoints, let's skip implementing them for now. Maybe they make sense to do later, or maybe we never do them and instead make API v2.2+ endpoints that re-implement the v1 APIs in a smarter way, which solves the recording metadata problem at the same time.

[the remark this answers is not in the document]
andrewazores: To clarify, even without a `GET /metadata` being implemented, I would still like the serialized recording descriptor to nest the `labels` field within a `metadata` field.

```json
{
  "name": "myRecording",
  "state": "RUNNING",
  "metadata": {
    "labels": {
      "myApp": "cryostat"
    }
  }
}
```

On the Java side this probably just means add a simple inner static class within the descriptor class that just holds a reference to the labels map, creating this nested structure.

src/main/java/io/cryostat/jmc/serialization/HyperlinkedSerializableRecordingDescriptor.java
```diff
@@ -85,6 +103,38 @@ public int hashCode() {
...
+        HyperlinkedSerializableRecordingDescriptor descriptor =
+                (HyperlinkedSerializableRecordingDescriptor) o;
+        return new EqualsBuilder()
+                .append(downloadUrl, descriptor.downloadUrl)
+                .append(reportUrl, descriptor.reportUrl)
+                .append(metadata.labels, descriptor.metadata.labels)
+                .build();
+    }
+
+    public static class Metadata {
```
andrewazores, line 125: This should probably also override `hashCode()`/`equals()`

src/main/java/io/cryostat/jmc/serialization/HyperlinkedSerializableRecordingDescriptor.java
```diff
@@ -85,6 +103,38 @@ public int hashCode() {
 
     @Override
     public boolean equals(Object o) {
```
andrewazores, line 105: `equals` was updated here for the metadata field, but `hashCode` was not. `hashCode` and `equals` should always be consistent with each other - is that still preserved with this change?

jan-law: Oops, should be fixed now

andrewazores:
Code looks good. Could you update HTTP_API.md?

andrewazores:
rebase please

andrewazores:
web-client should be updated again to the latest squashed/merged frontend commit

jan-law:
Latest build failure caused by `No recording data available` in `ArchivedReportJwtDownloadIT`
```
testDownloadRecordingUsingJwt  Time elapsed: 11.602 s  <<< ERROR!
java.util.concurrent.ExecutionException: java.lang.Exception: HTTP 500
	at itest.ArchivedReportJwtDownloadIT.testDownloadRecordingUsingJwt(ArchivedReportJwtDownloadIT.java:78)
Caused by: java.lang.Exception: HTTP 500
```
```
INFO: (10.0.2.100:58684): PATCH /api/v1/targets//service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Fcryostat-itests:9091%2Fjmxrmi/recordings/someRecording 204 241ms
Mar 09, 2022 9:09:52 PM io.cryostat.core.log.Logger error
SEVERE: HTTP 500: io.cryostat.recordings.EmptyRecordingException: java.io.IOException: java.io.IOException: No recording data available
Caused by: java.util.concurrent.ExecutionException: io.cryostat.recordings.EmptyRecordingException: java.io.IOException: java.io.IOException: No recording data available
```

andrewazores (approved):

PR #839, opened 2022-02-28
fix(graphQL): Restrict GraphiQL availability to dev mode

### LINKED ISSUE #819: GraphiQL handler should only be available in an explicit dev mode
https://github.com/cryostatio/cryostat/blob/a00358d379bb23e8833f028edbab51aca86f120b/src/main/java/io/cryostat/net/web/http/api/beta/graph/GraphiQLGetHandler.java#L83

We don't currently have an exact "dev mode" environment variable for Cryostat. The closest thing would be the `CRYOSTAT_CORS_ORIGIN`, but that is not quite right.

jan-law:
Fixes #819 

The GraphiQL web page is only available at `https://localhost:8181/api/beta/graphiql/` when running cryostat with `CRYOSTAT_DEV_MODE=true`.

### DIFF
--- a/devserver.sh
+++ b/devserver.sh
@@ -109,4 +109,5 @@ MAVEN_OPTS="${flags[@]}" \
     CRYOSTAT_CONFIG_PATH="${work_dir}/conf" \
     CRYOSTAT_TEMPLATE_PATH="${work_dir}/templates" \
     CRYOSTAT_PROBE_TEMPLATE_PATH="${work_dir}/probes" \
+    CRYOSTAT_WEB_ENV_DEV_MODE=true \
     "${MVN}" -Dcryostat.minimal=true -DskipTests=true vertx:run
--- a/devserver.sh
+++ b/devserver.sh
@@ -109,5 +109,5 @@ MAVEN_OPTS="${flags[@]}" \
     CRYOSTAT_CONFIG_PATH="${work_dir}/conf" \
     CRYOSTAT_TEMPLATE_PATH="${work_dir}/templates" \
     CRYOSTAT_PROBE_TEMPLATE_PATH="${work_dir}/probes" \
-    CRYOSTAT_WEB_ENV_DEV_MODE=true \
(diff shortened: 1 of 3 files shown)


andrewazores (approved):

andrewazores (approved):
Actually - could you also update the `devserver.sh` to set the dev mode env var to true?

src/main/java/io/cryostat/configuration/Variables.java
```diff
@@ -69,6 +69,7 @@ private Variables() {}
     public static final String MAX_CONNECTIONS_ENV_VAR = "CRYOSTAT_MAX_WS_CONNECTIONS";
     public static final String ENABLE_CORS_ENV = "CRYOSTAT_CORS_ORIGIN";
     public static final String HTTP_REQUEST_TIMEOUT = "CRYOSTAT_HTTP_REQUEST_TIMEOUT";
+    public static final String WEB_ENV_DEV_MODE = "CRYOSTAT_WEB_ENV_DEV_MODE";
```
andrewazores, line 72: WDYT about just calling this `CRYOSTAT_DEV_MODE`?

jan-law: Sure, sounds good

andrewazores (approved):

devserver.sh
```diff
@@ -109,4 +109,5 @@ MAVEN_OPTS="${flags[@]}" \
     CRYOSTAT_CONFIG_PATH="${work_dir}/conf" \
     CRYOSTAT_TEMPLATE_PATH="${work_dir}/templates" \
     CRYOSTAT_PROBE_TEMPLATE_PATH="${work_dir}/probes" \
+    CRYOSTAT_WEB_ENV_DEV_MODE=true \
```
andrewazores, line 112: Rename here too

andrewazores (changes requested):

andrewazores (approved):

Issue #846, opened 2022-03-04
GraphQL environmentNodes query filters


andrewazores:
https://github.com/cryostatio/cryostat/blob/a00358d379bb23e8833f028edbab51aca86f120b/src/main/resources/queries.graphqls#L3

environmentNodes query should have filters for node name, node type, as well as labels or annotations applied on the node.

andrewazores:
Partially implemented: environment nodes can be filtered by name or type in this query. Node labels/annotations are not yet handled because we would like to reuse Kubernetes label selectors for this and how exactly to best do that hasn't been sorted out.

Issue #820, opened 2022-02-11
[Story] GraphQL query filters


andrewazores:
Depends on #825

- [x] #846 
- [x] #821
- [x] #822 
- [x] #823

PR #847, opened 2022-03-04
feat(graphql): apply conjunction of filters

andrewazores:
Related to #820
Related to #846

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/beta/graph/EnvironmentNodesFetcher.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/graph/EnvironmentNodesFetcher.java
@@ -39,8 +39,6 @@ package io.cryostat.net.web.http.api.beta.graph;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.function.Function;
 
@@ -68,7 +66,10 @@ class EnvironmentNodesFetcher implements DataFetcher<List<EnvironmentNode>> {
 
     @Override
     public List<EnvironmentNode> get(DataFetchingEnvironment environment) throws Exception {
-        Map<String, String> filter = environment.getArgument("filter");
+        FilterInput filter = FilterInput.from(environment);
+        if (filter.acceptsNone()) {
+            return List.of();
+        }
         EnvironmentNode root = rootNodeFetcher.get(environment);
         List<EnvironmentNode> nodes =
                 recurseFetcher.get(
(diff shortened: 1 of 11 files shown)


src/main/java/io/cryostat/net/web/http/api/beta/graph/EnvironmentNodesFetcher.java
```diff
@@ -68,56 +66,24 @@
...
-            String nodeType = filter.get("nodeType");
-            List<EnvironmentNode> typedNodes =
-                    recurse(root, n -> n.getNodeType().getKind().equalsIgnoreCase(nodeType));
-            if (typedNodes.isEmpty()) {
-                throw new NoSuchElementException(String.format("Node of type %s", nodeType));
-            }
-            return typedNodes;
+        if (filter.contains(FilterInput.Key.NODE_TYPE)) {
+            String nodeType = filter.get(FilterInput.Key.NODE_TYPE);
+            nodes = recurse(root, n -> n.getNodeType().getKind().equalsIgnoreCase(nodeType));
```
jan-law, line 84: > The initial implementations of input objects will only support simple string-equality matches, so ex. { name: "myNode", kind: "JVM" } is an input object that logically means something like node.name.equals("myNode") && node.kind.equals("JVM")

Could you explain how this filters nodes with both `nodeName` `AND` `nodeType`? Wouldn't the filtered `nodes` from line 79 get overwritten here?

andrewazores: Ah you're correct, good catch. The idea was to perform the query for all nodes first and then sequentially apply each filter, which has the same result as logically ANDing the filter criteria and applying that, but that's not what is actually implemented here. I'll fix it.

src/main/java/io/cryostat/net/web/http/api/beta/graph/EnvironmentNodesFetcher.java
```diff
@@ -38,68 +38,75 @@
...
         }
-        return nodes;
+        return new ArrayList<>(nodes);
     }
 
-    List<EnvironmentNode> recurse(
-            EnvironmentNode node, Function<EnvironmentNode, Boolean> matcher) {
-        List<EnvironmentNode> result = new ArrayList<>();
+    Set<EnvironmentNode> flattenEnvNodes(EnvironmentNode root) {
+        return new HashSet<>(recurse(root, e -> e instanceof EnvironmentNode));
```
hareetd, line 84: Is the switch to `HashSet` from `ArrayList` due to performance reasons? I guess in this case, the order doesn't matter and nodes are unique so `HashSet` makes more sense?

andrewazores: Right, order doesn't matter/is undefined for this query. It isn't so much a performance reason as a cheap way to de-duplicate the collection. There shouldn't really be duplicates anyway if the tree is truly a tree and not a general graph (with potential cycles), but this is cheap insurance against that possibility.

hareetd:
While running `smoketest`, when I do the following query:

`query {
        name
        nodeType
        descendantTargets {
            name
            nodeType
        }
    }
}
`

I get back this result:

`HTTP/1.1 200 OK
content-encoding: gzip
content-length: 178

{
    "data": {
        "environmentNodes": [
            {
                "descendantTargets": [
                    {
                        "nodeType": "JVM"
                    },
                    {
                        "nodeType": "JVM"
                    },
                    {
                        "nodeType": "JVM"
                    },
                    {
                        "nodeType": "JVM"
                    },
                    {
                        "nodeType": "JVM"
                    },
                    {
                        "nodeType": "JVM"
                    }
                ],
                "name": "JDP",
                "nodeType": "REALM"
            }
        ]
    }
}
`
However, if I then do the same query but this time filtering for the `nodeType`:

`query {
    environmentNodes(filter: { name: "JDP", nodeType: "REALM" }) {
        name
        nodeType
        descendantTargets {
            name
            nodeType
        }
    }
}
`

I get back an empty list (i.e. no matches) when I expect to get the same result as the first query:

`HTTP/1.1 200 OK
content-encoding: gzip
content-length: 58

{
    "data": {
        "environmentNodes": []
    }
}
`

However, the query works as expected if your filter for `nodeType: "Realm"` instead of `nodeType: "REALM"`, so it looks like a casing mismatch issue on line 78 of the `EnvironmentNodesFetcher`.

andrewazores:
Nice catch @hareetd . I had an `equalsIgnoreCase()` on that line previously but it didn't seem right, so I removed it. But that's broken because in the discovery tree result, what the user sees is the `toString()` of the NodeType enum - this is where the uppercase `REALM` comes from. The `getKind()` of that `BaseNodeType.REALM` enum member is `Realm`, though. I updated that condition to accept either form, but I think the proper fix is that the GraphQL queries and the Discovery API should report the nodes' type/kind as the `Realm` form, ie the value specified by the enum members, and not just the enum members' names. I'll see about fixing that separately.

andrewazores:
^ https://github.com/cryostatio/cryostat/issues/853

hareetd (approved):
Looks good.

jan-law (approved):
lgtm

PR #856, opened 2022-03-11
fix(graphql): override nodetype tostrings for display

### LINKED ISSUE #853: GraphQL and Discovery API disagree on nodes' names
The Discovery API resports nodes' names as their implemented `Kind`, ex. `Realm` or `Pod`. The GraphQL API just serializes the enum member's name directly, ie `REALM` and `POD`. In all of our current use cases this only amounts to a casing difference, but it would still be best if these were consistent between the two API endpoints.

andrewazores:
Fixes #853

I found this simple workaround for the way that GraphQL treats enum values. Rather than creating some duplication within our internal GraphQL module that re-encodes all of the NodeType enums to their Kind strings, I simply add an override on the `toString()` method for all of the NodeType implementations that simply returns the `Kind`. GraphQL uses `toString()` to serialize the `nodeType` enum field, which displays the enum member's name by default, so this override is simple and effective and doesn't duplicate definitions.

To test:
Query `http :8181/api/beta/graphql query=@graphql/root-node-query.graphql`
Query `http :8181/api/v2.1/discovery`

Compare the results of the `kind`/`nodeType` fields of nodes, before and after this patch. Before you should see ex. `REALM`/`Realm`, whereas after they will both agree on `Realm`.

### DIFF
--- a/src/main/java/io/cryostat/platform/discovery/AbstractNode.java
+++ b/src/main/java/io/cryostat/platform/discovery/AbstractNode.java
@@ -40,7 +40,6 @@ package io.cryostat.platform.discovery;
 import java.util.Collections;
 import java.util.Map;
 
-import com.google.gson.annotations.SerializedName;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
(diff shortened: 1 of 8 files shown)


hareetd (approved):
Looks good!

PR #858, opened 2022-03-11
feat(credentials): Always persist to disk

### LINKED ISSUE #565: CredentialsManager should be able to persist credentials to disk
https://github.com/cryostatio/cryostat/blob/74328d3268156f001b8977ab769601da7a3e2fd5/src/main/java/io/cryostat/configuration/CredentialsManager.java#L93

jan-law:
Fixes #565 

Note: Most of the additions were caused by injecting the `CredentialsManager` into the `AbstractAuthenticatedRequestHandler`

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/AbstractAuthenticatedRequestHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/AbstractAuthenticatedRequestHandler.java
@@ -37,10 +37,12 @@
  */
 package io.cryostat.net.web.http;
 
+import java.io.IOException;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
 import java.rmi.ConnectIOException;
 import java.util.Base64;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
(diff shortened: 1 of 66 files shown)


src/main/java/io/cryostat/net/web/http/AbstractAuthenticatedRequestHandler.java
```diff
@@ -164,16 +169,35 @@ private boolean isAuthFailure(ExecutionException e) {
...
+    private void removeCredentialsIfPresent(RoutingContext ctx) {
+        Optional<String> targetId = Optional.ofNullable(ctx.pathParam("targetId"));
+
+        targetId.ifPresent(
+                id -> {
+                    if (credentialsManager.getCredentials(id) != null) {
+                        try {
+                            credentialsManager.removeCredentials(id);
+                        } catch (IOException unused) {
+                            // handleConnectionException already throws an HTTPStatusException
```
andrewazores, line 198: I think it'd be good to log this exception at least, even if it doesn't get propagated any further back the call stack.

andrewazores (dismissed):
Awesome, thanks

andrewazores (approved):

PR #864, opened 2022-03-16
feat(recordings): set recording template in label

### LINKED ISSUE #860: Recordings should have labels for recording settings
Recording settings like the template used should be stored in the labels attached to a recording when it is started (manually or by rule)

andrewazores:
Fixes #860

The main feature work is a tiny change in the `feat(recordings): set recording template in label` commit. The commit before that is some simple refactoring (affecting a lot of call sites, but a minor change in itself) to generalize the `RecordingMetadataManager` to deal with the `Metadata` type rather than just the labels map, so that we can more easily extend the `Metadata` to contain other fields in the future. And the first commit in the series fixes an oversight where GraphQL responses did not include the metadata/labels properties, so there was a discrepancy there vs the standard HTTP API.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/beta/RecordingMetadataLabelsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/RecordingMetadataLabelsPostHandler.java
@@ -54,12 +54,13 @@ import io.cryostat.net.web.http.api.v2.IntermediateResponse;
 import io.cryostat.net.web.http.api.v2.RequestParameters;
 import io.cryostat.recordings.RecordingArchiveHelper;
 import io.cryostat.recordings.RecordingMetadataManager;
+import io.cryostat.recordings.RecordingMetadataManager.Metadata;
 import io.cryostat.recordings.RecordingNotFoundException;
 
 import com.google.gson.Gson;
 import io.vertx.core.http.HttpMethod;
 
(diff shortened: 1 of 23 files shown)


src/main/java/io/cryostat/net/web/http/api/beta/RecordingMetadataLabelsPostHandler.java
```diff
@@ -116,28 +117,27 @@ public boolean requiresAuthentication() {
...
-                    recordingMetadataManager.setRecordingLabels(recordingName, labels).get();
+            Metadata updatedMetadata =
+                    recordingMetadataManager.setRecordingMetadata(recordingName, metadata).get();
 
             notificationFactory
                     .createBuilder()
                     .metaCategory(RecordingMetadataManager.NOTIFICATION_CATEGORY)
                     .metaType(HttpMimeType.JSON)
-                    .message(Map.of("recordingName", recordingName, "labels", labels))
+                    .message(Map.of("recordingName", recordingName, "metadata", updatedMetadata))
```
jan-law, line 136: Could you also update the -web side with the new notification message format?

andrewazores: Oh right, thanks.

jan-law (approved):

PR #867, opened 2022-03-17
fix(recordings): disallow labels with whitespace

### LINKED ISSUE #865: Recording labels should disallow whitespace, allow other characters
> I realized that I configured the frontend label validation to exclude commas and equal signs. We should probably make the validations more lenient to accommodate `template=Profiling,type=TARGET`: https://github.com/cryostatio/cryostat-web/blob/main/src/app/CreateRecording/EditRecordingLabels.tsx#L65

I'm thinking maybe we just want to enforce no whitespace. Using a regex like this is also forcing users into the Latin alphabet, and right now they also can't enter labels with a `/` or other plausibly desirable characters.

That validation should also be enforced on the backend side too, not just in the frontend form. I'll file a separate issue.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/391#issuecomment-1071151604_

jan-law:
Fixes #865

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingMetadataManager.java
+++ b/src/main/java/io/cryostat/recordings/RecordingMetadataManager.java
@@ -48,6 +48,8 @@ import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import io.cryostat.core.log.Logger;
 import io.cryostat.core.sys.FileSystem;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/recordings/RecordingMetadataManager.java
```diff
@@ -153,12 +155,19 @@ public Metadata deleteRecordingMetadataIfExists(String targetId, String recordin
     public Map<String, String> parseRecordingLabels(String labels) throws IllegalArgumentException {
         Objects.requireNonNull(labels, "Labels must not be null");
 
+        Pattern noWhitespace = Pattern.compile("^(\\S)+$");
+        Matcher matcher = noWhitespace.matcher(labels);
```
andrewazores, line 159: What is the value of `labels` at this point, or what is the source of the value? I think this is going to come in from ex. an API request, and it's the raw JSON representation of the labels map, right?

In that case it's probably better to apply the no-whitespace validation after the gson parser has already parsed it. We can be lenient about the JSON input being pretty-printed with whitespaces, it's just the actual values within that we want to enforce on.

andrewazores (approved):

PR #868, opened 2022-03-17
feat(graphql): filter active recordings by various properties

### LINKED ISSUE #823: GraphQL active recordings query filters
https://github.com/cryostatio/cryostat/blob/a00358d379bb23e8833f028edbab51aca86f120b/src/main/resources/types.graphqls#L56

Active recordings should be filterable by name, state, duration/continuous, startTime (before or after a timestamp).

andrewazores:
Fixes #823

Allows querying active recordings with filters on name, state, startTime (before or after timestamp inclusive), duration (greater/equal or lesser/equal to duration), continuous, toDisk, and the value of the "template" metadata label if any.

Example query for recordings using the Profiling template which have already stopped, running on Cryostat's own JVM:

```graphql
query {
    targetNodes(filter: { name: "service:jmx:rmi:///jndi/rmi://localhost:9091/jmxrmi" }) {
        recordings {
            active(filter: {
                state: "STOPPED",
                templateLabel: "template=Profiling,type=TARGET"
                }) {
                name
                state
                startTime
                duration
                continuous
                toDisk
                maxSize
                maxAge
                metadata {
                    labels
                }
                reportUrl
                downloadUrl
            }
        }
    }
}
```


src/main/resources/types.graphqls
```diff
@@ -67,7 +79,7 @@ interface Node {
 }
 
 type Recordings {
-    active(names: [String!]): [ActiveRecording!]! # TODO better filtering, not just on name
+    active(filter: ActiveRecordingFilterInput): [ActiveRecording!]!
```
jan-law, line 82: I created some recordings with the web UI, then tested some queries with curl. The backend keeps terminating after the second or third graphql query. Are there any other logs I can check to find out what happened?
```
POST /api/beta/graphql HTTP/1.1
Connection: keep-alive
Content-Length: 621
Host: localhost:8181
User-Agent: HTTPie/3.0.2

{
    "query": "query {\n    environmentNodes(filter: { name: \"JDP\" }) {\n        name\n        nodeType\n        descendantTargets {\n            recordings {\n                active(filter: { continuous: true, templateLabel: \"template=Cryostat,type=TARGET\" }) {\n                    name\n                    state\n                    metadata {\n                        labels\n                    }\n                }\n                archived {\n                    name\n                    metadata {\n                        labels\n                    }\n                }\n            }\n        }\n    }\n}\n"
}



http: error: ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')) while doing a POST request to URL: http://localhost:8181/api/beta/graphql
```
```
INFO: (10.0.2.100:51816): GET /api/v1/targets/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Fcryostat%3A9091%2Fjmxrmi/recordings 200 21ms
Mar 18, 2022 5:47:30 PM io.cryostat.core.log.Logger info
INFO: (10.0.2.100:51810): GET /api/v1/recordings 200 28ms
Mar 18, 2022 5:47:40 PM io.cryostat.core.log.Logger info
INFO: Removing cached connection for service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi: EXPIRED
Mar 18, 2022 5:47:40 PM io.cryostat.core.log.Logger info
INFO: Connection for service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi closed
Mar 18, 2022 5:48:45 PM io.cryostat.core.log.Logger info
INFO: Creating connection for service:jmx:rmi:///jndi/rmi://cryostat:10000/jmxrmi
+ cleanup
+ podman pod kill cryostat
cab957e3c614d9e0bafd13451c105f400dc7ffc19cf78aec08744ce478ec993a
+ podman pod rm cryostat
cab957e3c614d9e0bafd13451c105f400dc7ffc19cf78aec08744ce478ec993a
```

andrewazores: Hmm. My hunch is maybe your container is getting OOM killed... what does `podman stats` say?

jan-law: There's a `jovial_dirac` using 99% memory. What is this?
```

ID            NAME                CPU %       MEM USAGE / LIMIT  MEM %       NET IO      BLOCK IO      PIDS        CPU TIME    AVG CPU %
0a795e4f1959  vertx-fib-demo-2    20.07%      221.5MB / 16.54GB  1.34%       -- / --     -- / --       42          9.496745s   2.52%
20cd24843200  jovial_dirac        72.19%      536.5MB / 536.9MB  99.93%      -- / --     16.38kB / 0B  75          32.686578s  11.13%
2b64c72a3b9d  reports             7.29%       167MB / 536.9MB    31.10%      -- / --     -- / --       32          7.235555s   2.57%
47246e431a54  vertx-fib-demo-1    28.62%      217.3MB / 16.54GB  1.31%       -- / --     -- / --       42          8.471838s   3.26%
5e09bd44bc27  grafana             0.16%       82.77MB / 16.54GB  0.50%       -- / --     -- / --       13          858.297ms   0.12%
5e0a30d883cb  ecf5b216b396-infra  --          843.8kB / 16.54GB  0.01%       -- / --     -- / --       1           2.743ms     1.19%
741d4e554ea0  quarkus-test        22.30%      258.7MB / 16.54GB  1.56%       -- / --     -- / --       61          9.879247s   2.68%
a8bc4ff7889e  wildfly             0.24%       511.1MB / 16.54GB  3.09%       -- / --     -- / --       74          20.3384s    0.39%
eef9bc705eed  jfr-datasource      0.00%       38.41MB / 16.54GB  0.23%       -- / --     -- / --       21          31.854ms    0.01%
```

andrewazores:
```
ID            NAME        CPU %       MEM USAGE / LIMIT  MEM %       NET IO      BLOCK IO           PIDS        CPU TIME    AVG CPU %
0980633ef8c0  kind_allen  1.60%       515.2MB / 536.9MB  95.97%      -- / --     65.54kB / 16.38kB  84          32.133373s  23.10%
```

Whoops. Yep, I think you're probably OOMing - my test instance is right on the edge, too.

andrewazores: > What is this?

That's cryostat - the pod gets named `cryostat`, and the actual Cryostat container inside just gets a generated name assigned by podman.

andrewazores: @jan-law can you try a temporary workaround to see if this solves the problem? Just modify your `run.sh` and either remove the `--memory` flag or bump it up to say 1GB. If your crash problem still occurs then that's almost definitely not an OOM. However, just seeing the podman shutdown in the logs like that is usually an OOM from my experience.

jan-law (commented):
All the active recording filters with `run.sh` are working well

src/main/resources/types.graphqls
```diff
@@ -22,6 +22,18 @@ input DescendantTargetsFilterInput {
     name: String
 }
 
+input ActiveRecordingFilterInput {
+    name: String
+    state: String
+    continuous: Boolean
+    toDisk: Boolean
+    durationGreaterThanEqual: Long
+    durationLessThanEqual: Long
```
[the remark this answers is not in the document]
andrewazores: Everything here is just in whatever units are used by the backing representation, so yes these are ms. Because of the way that GraphQL works it's not trivial to have it format the durations in different units in responses, and I think it would be the [least surprising](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) for a user if the filter input and the response output use consistent units.

It is probably a good idea to rename these filter keys to include unit information where applicable, though.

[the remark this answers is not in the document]
jan-law: > It is probably a good idea to rename these filter keys to include unit information where applicable, though.

good idea, makes sense.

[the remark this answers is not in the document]
andrewazores: (this is also a blunder on the `POST /api/v1/targets/:targetId/recordings` API - it interprets the `duration` form attribute as seconds and converts that to milliseconds. It really should have been in milliseconds, or it should have accepted a formatted duration string like `30s`/`500ms`/`1h` etc)

src/main/resources/types.graphqls
jan-law: > bump it up to say 1GB

no more crashes :-)
`podman stats` is showing ~600M / 1G for the cryostat container, so it's very likely an OOM:
```
ID            NAME                CPU %       MEM USAGE / LIMIT  MEM %       NET IO      BLOCK IO      PIDS        CPU TIME    AVG CPU %
aaa76bf317f0  loving_ellis        15.22%      596MB / 1.074GB    55.50%      -- / --     270.3kB / 0B  87          35.785672s  10.38%
```

src/main/java/io/cryostat/net/web/http/api/beta/graph/ActiveRecordingsFetcher.java
```diff
@@ -0,0 +1,133 @@
...
+            result =
+                    result.stream()
+                            .filter(r -> r.getDuration() <= duration)
+                            .collect(Collectors.toList());
+        }
+        if (filter.contains(FilterInput.Key.START_TIME_BEFORE)) {
+            long startTime = filter.get(FilterInput.Key.START_TIME_BEFORE);
+            result =
+                    result.stream()
+                            .filter(r -> r.getStartTime() <= startTime)
```
hareetd, line 109: The filter `startTimeBeforeEqual` matches if the `startTime` value given is before or equal to a recording's start time, right? Shouldn't the equality be `r.getStartTime() >= startTime` then?

src/main/java/io/cryostat/net/web/http/api/beta/graph/ActiveRecordingsFetcher.java
```diff
@@ -0,0 +1,133 @@
...
+                            .collect(Collectors.toList());
+        }
+        if (filter.contains(FilterInput.Key.START_TIME_AFTER)) {
+            long startTime = filter.get(FilterInput.Key.START_TIME_AFTER);
+            result =
+                    result.stream()
+                            .filter(r -> r.getStartTime() >= startTime)
```
hareetd, line 116: The equality here should be switched to `<=` following the same logic as above.

src/main/java/io/cryostat/net/web/http/api/beta/graph/ActiveRecordingsFetcher.java
andrewazores: It's meant to be "if the recording was started at or before the given timestamp", but I guess the filter name isn't really clear about that. Any suggestion for a better name?

hareetd: Nevermind, I realize it matches if a recording's start time is before or equal to the filter `startTime` value.

hareetd: I think it's fine as is actually. More of a misunderstanding on my part than any confusion in the filter name.

hareetd (approved):
Works well, I didn't run into any OOM issues with `smoketest`.

jan-law (approved):

jan-law (approved):

hareetd (approved):

PR #872, opened 2022-03-25
feat(graphql): implement k8s-style label selectors

### LINKED ISSUE #820: [Story] GraphQL query filters

### LINKED ISSUE #821: GraphQL targetNodes query filters
https://github.com/cryostatio/cryostat/blob/a00358d379bb23e8833f028edbab51aca86f120b/src/main/resources/queries.graphqls#L4

targetNodes query should have filters for node name, as well as labels or annotations applied on the node.

### LINKED ISSUE #822: GraphQL descendantTargets query filters
https://github.com/cryostatio/cryostat/blob/a00358d379bb23e8833f028edbab51aca86f120b/src/main/resources/types.graphqls#L39

descendantTargets query should have filters for node name, as well as labels or annotations applied on the node.

### LINKED ISSUE #825: GraphQL filtering strategy
Using input objects as a sort of filter configuration seems to make sense. This allows the query's type signature to be short and sweet, and also makes it easy to extend the configuration input type with more filter kinds in the future. However, we need to decide what happens if a user specifies multiple of the optional fields within the configuration object. If a user is querying active recordings and applies a filter with both a "continuous==true" and "labels['myLabel']=='something'`, does this filter match recordings with both properties or with either property? It seems unclear. I suggest that it should be an "AND", since this allows the user to perform very narrow requests on pairs of properties that are computed server-side, alleviating a lot of requests and filtering work on the client side. If the user needs to do an equivalent "OR" then they can do so by decomposing their request into multiple requests where each filter property is supplied once on its own.

Another option is as @ebaron suggests, to re-use [Kubernetes label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) via the fabric8 client. If we do this then it should not only be limited to labels, and we should come up with a way to also apply this to other metadata about resources, so that the selector syntax can be the entire filter syntax rather than just a portion of the filter specifically for the labels field. The first, perhaps naive, solution that comes to mind is to copy these other fields into the labels map with some well-known key prefix, and perhaps to also prepend any actual labels with another prefix for consistency.

Consider following node resource (JSON-ish):

```
TargetNode {
  name: "myNode",
  kind: "JVM",
  annotations: {
    "myAnnotation": "anno1"
  },
  labels: {
    "com.example.MyApp/label": "labelled"
  }
}
```

This could be converted into a simple Map (made up syntax) to be processed using the label selector syntax:

```
  cryostat.node.name -> myNode
  cryostat.node.kind -> JVM
  cryostat.annotations.myAnnotation -> anno1
  cryostat.labels.com.example.MyApp/label -> labelled
```

### LINKED ISSUE #846: GraphQL environmentNodes query filters

andrewazores:
Closes #846
Closes #825
Closes #822
Closes #821
Closes #820

This adds `labels` to GraphQL query filters for nodes in the deployment tree and for recordings, as well as `annotations` to query filters on target nodes. For `annotations` the Cryostat-generated annotations and the platform-provided annotations are merged together into one map. Both the `labels` and `annotations` query fields are optional, and can be given a Kubernetes label selector expression to match against the labels or annotations maps on the objects being queried.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcher.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcher.java
@@ -63,8 +63,7 @@ public class LabelSelectorMatcher implements Predicate<Map<String, String>> {
     // parens, not a single matched pair. Regexes are hard. This works but is too tolerant of broken
     // expressions.
     static final Pattern SET_MEMBER_PATTERN =
-            Pattern.compile(
-                    "^(?<key>[a-zA-Z0-9-_./]+)[\\s]+(?<op>in|notin)[\\s]+(?<values>[\\(\\)a-zA-Z0-9-_./=, ]*)$",
+            Pattern.compile("(?<key>[a-zA-Z0-9-_./]+)[\\s]+(?<op>in|notin)[\\s]+\\((?<values>(?:[a-zA-Z0-9-_./=, ]+[,\\s]*)+)\\)",
                     Pattern.CASE_INSENSITIVE);
 
     // ex. "mykey" or "!mykey". Tests whether the given key name exists in the test label set as a
@@ -125,7 +124,7 @@ public class LabelSelectorMatcher implements Predicate<Map<String, String>> {
             String op = m.group("op");
             SetMatcher.Operator operator = SetMatcher.Operator.fromString(op);
             Objects.requireNonNull(operator, "Unknown set operator " + op);
-            String value = m.group("values").replaceAll("[\\(\\)]+", "");
+            String value = m.group("values");
             List<String> values =
                     Arrays.asList(value.split(",")).stream()
                             .map(String::trim)
--- a/src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcher.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcher.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package io.cryostat.net.web.http.api.beta.graph.labels;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class LabelSelectorMatcher implements Predicate<Map<String, String>> {
+
+    // ex. "my.prefix/label = something". Whitespaces around the operator are ignored. Left side
+    // must loosely look like a k8s label (not strictly enforced here), right side must loosely look
+    // like a k8s label value, which may be empty. Allowed operators are "=", "==", "!=".
+    static final Pattern EQUALITY_PATTERN =
+        Pattern.compile("^(?<key>[a-zA-Z0-9-_./]+)[\\s]*(?<op>=|==|!=)[\\s]*(?<value>[a-zA-Z0-9-_./]*)$");
+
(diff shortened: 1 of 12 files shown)


src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/SetMatcher.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public boolean test(String s) {
+        return operator.with(values).test(s);
+    }
+
+    public enum Operator {
+        IN("In", args -> v -> contains(args, v)),
+        NOT_IN("NotIn", args -> v -> !contains(args, v)),
```
jan-law, line 74: How do I test an expression with more than one value in the set? When I tried querying with `descendantTargets (filter: { annotations: "PORT in (9091, 9096)" })`, the results only show the `cryostat:9091` container:
```
query {
        name
        nodeType
        descendantTargets (filter: { annotations: "PORT in (9091, 9096)" }){
            target {
                alias
                annotations {
                cryostat
                }
            }
        }
    }
}
```
```
{
    "data": {
        "environmentNodes": [
            {
                "descendantTargets": [
                    {
                        "target": {
                            "alias": "io.cryostat.Cryostat",
                            "annotations": {
                                "cryostat": {
                                    "HOST": "cryostat",
                                    "JAVA_MAIN": "io.cryostat.Cryostat",
                                    "PORT": "9091"
                                }
                            }
                        }
                    }
                ],
                "name": "JDP",
                "nodeType": "Realm"
            }
        ]
    }
}
```

jan-law (commented):
The label equality matchers are working well with both labels and annotation filters. Just had a question about the set matchers

src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcher.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class LabelSelectorMatcher implements Predicate<Map<String, String>> {
+
+    // ex. "my.prefix/label = something". Whitespaces around the operator are ignored. Left side
+    // must loosely look like a k8s label (not strictly enforced here), right side must loosely look
+    // like a k8s label value, which may be empty. Allowed operators are "=", "==", "!=".
```
jan-law, line 55: Does this mean that custom recording labels would also need to 'loosely' match the k8s label format if users want to filter queries by those custom labels?

src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/SetMatcher.java
andrewazores: Ah, darn. Thanks for catching that. This is actually a bug due to the way that label selectors can be `AND`ed together by joining them with a comma, but the set membership expression also uses the comma inside the parentheses to denote the set members.

So it's supposed to be a supported syntax to do say `PORT in (9091, 9096), env != production`, but my naive approach to "parsing" breaks down here because of the comma and splits the expression inappropriately.

I see two solutions. The first is perhaps "most correct" but also the most annoying, and that's to implement a more proper parser rather than just leaning on `string.split(",")` and regexes, so that we can parse the expression and separate out the "clauses" joined by commas, and then parse out the different parts of the expressions as well.

The second is to deviate slightly from the k8s label selector syntax and just not do clause conjunction using a comma, but instead split that out into a list of clauses. The GraphQL query would change from `filter: { annotations: "PORT in (9091, 9096), env != production" }` to `filter: { annotations: ["PORT in (9091, 9096)", "env != production"] }. This is quick and easy, and IMO even somewhat nicer because we lean on the GraphQL syntax that provides some notion of a type system, rather than just encoding everything into one big string.

src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcher.java
andrewazores: That is true. It does violate something I asked you to do a while ago for label validations which is to be sure we can allow general unicode characters, not just the ASCII subset. I should modify this to be more tolerant and just match any non-whitespace characters, I think.

andrewazores: That is true. It does violate something I asked you to do a while ago for label validations which is to be sure we can allow general unicode characters, not just the ASCII subset. I should modify this to be more tolerant and just match any non-whitespace characters, I think.

src/main/java/io/cryostat/net/web/http/api/beta/graph/labels/SetMatcher.java
andrewazores: Unresolving this - we can do a query like `filter: { annotations: "PORT in (9091, 9096)" }` now, but we're now missing the ability to `AND` multiple annotation conditions like that together within the query, which is what ex. `filter: { annotations: "PORT in (9091, 9096), env != production" }` was supposed to give us. I can either implement support for that syntax myself, or I can change the syntax a little and let GraphQL's parser do the work by making it into a list type in the query filter object.

jan-law: >  let GraphQL's parser do the work by making it into a list type in the query filter object.

I think it makes sense to use a list here. Seems like a good middle ground between the GraphQL syntax and the k8s syntax

andrewazores: @ebaron Any opposition to using a GraphQL list type?

andrewazores: That's pretty much my opinion on it, too. It might be a minor tripping point for a Kubernetes user who sees the familiar label selector syntax and tries to just copy-paste a selector conjunction they're already using elsewhere, but that's the only potential downside I see and it's an easy enough fix once they realize it's not working and check what our expected syntax actually is.

jan-law (approved):
Looks good!

Looking ahead, will there be other documentation (eg on HTTP_API or cryostat.io) for these 2.1 GraphQL features besides the blog posts?

andrewazores:
I think a general description of the GraphQL endpoint in `HTTP_API.md` makes sense, but since that endpoint has so many possibilities in operation depending on the query sent, it's a little too hefty to describe it all there. There is still the `GRAPHQL.md` and the `graphql/` examples hanging around which I had intended as temporary and to be cleaned up later, but perhaps it makes sense to clean them up and keep them around as in-repo docs. There should also be some form of documentation for this feature that also makes it onto cryostat.io for upstream docs, as well as something similar to whatever that ends up being for downstream product docs.

src/test/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcherTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    final Map<String, String> labels =
+            Map.of(
+                    "foo", "bar",
+                    "something", "else",
+                    "my.prefixed/label", "expectedValue",
+                    "env", "prod",
+                    "present", "irrelevant");
+
+    @Disabled("Condition conjuction using comma separation is not currently supported")
```
jan-law, line 61: Can this test be enabled now?

jan-law (changes requested):
Whoops, just saw this now

src/test/java/io/cryostat/net/web/http/api/beta/graph/labels/LabelSelectorMatcherTest.java
andrewazores: Ah right. No, that would still need a real parser implemented to support, since this is testing the individual "clauses" of a label selector. The input to the `LabelSelectorMatcher` is down the line from the output of what GraphQL parses, so at this level the comma-separated conjunction is still not supported. I'll remove this test.

jan-law (approved):

PR #874, opened 2022-03-25
fix(labels): split event template name/type into separate labels

### LINKED ISSUE #873: Label for recording template/type is incompatible with GraphQL filtering/k8s label selectors
> @ebaron I just realized we have a grammar incompatibility between k8s label selectors and our current form of template specifier. We use ex. `template=Profiling,type=TARGET` to specify the event template to use to start a recording, and we capture that string verbatim and store it in the metadata labels attached to a recording (active or archived), with the map key `template`. For example:

```json
{
    "data": {
        "environmentNodes": [
            {
                "descendantTargets": [
                    {
                        "recordings": {
                            "active": [
                                {
                                    "metadata": {
                                        "labels": {
                                            "template": "template=Profiling,type=TARGET"
                                        }
                                    },
                                    "name": "foo",
                                    "state": "RUNNING"
                                }
                            ],
                            "archived": []
                        }
                    }
                ],
                "labels": {},
                "name": "JDP",
                "nodeType": "Realm"
            }
        ]
    }
}
```

> This is incompatible with both the equality-based and set-based k8s label selector syntaxes. The equality style expression would look like `template = template=Profiling,type=TARGET` and the set style expression would look like `template in (template=Profiling,type=TARGET)`. Both of these are ambiguous and invalid because our `template=Foo[,type=TYPE]` expression is not a valid k8s label value.

> Possible solutions:

> 1. Maintain the existing `templateLabel` special case label selector style, where the filter input simply takes the expected value of the label, and not a full label selector expression. All other kinds of label can be selected normally with k8s selector style.
> 2. Change our `template=Foo[,type=TYPE]` syntax. Doesn't seem appropriate for 2.1.0.
> 2b. We could alternately maintain the same input syntax, but replace the delimiters to something k8s label selector compatible when persisting them into recording metadata.
> 3. Implement some more specialized parser for the label selectors that can handle this ambiguous case for our specific scenario, but this is pretty hacky and can still end up being ambiguous with our template specifier syntax anyway.

> Thoughts? For 2.1.0, option 1 seems to make the most sense to me, or maybe 2b. For 3.0 we can change the event specifier syntax and avoid this issue entirely.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/825#issuecomment-1079222248_

andrewazores:
Fixes #873

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
@@ -168,11 +168,8 @@ public class RecordingTargetHelper {
                     Metadata metadata =
                             recordingMetadataManager.getMetadata(targetId, recordingName);
                     Map<String, String> labels = metadata.getLabels();
-                    labels.put(
-                            "template",
-                            String.format(
-                                    "template=%s,type=%s",
-                                    templateName, preferredTemplateType.name()));
+                    labels.put("template.name", templateName);
+                    labels.put("template.type", preferredTemplateType.name());
                     metadata = new Metadata(labels);
                     metadata =
                             recordingMetadataManager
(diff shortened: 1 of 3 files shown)


jan-law (approved):
LGTM

> We could alternately maintain the same input syntax, but replace the delimiters to something k8s label selector compatible when persisting them into recording metadata.

I also agree with keeping the `events: template=Continuous,type=TARGET` input form the same for 2.1

Issue #881, opened 2022-04-07
OpenShift OAuth server is not trusted


ebaron:
I think I see the problem now. It has nothing to do with Grafana or this PR. It's another side effect of removing `CRYOSTAT_ALLOW_UNTRUSTED_SSL`. Cryostat sends a GET request to OpenShift's OAuth server to discover its endpoints. Cryostat is not configured to trust the OAuth server's certificate. The fix should be something like adding `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` to its truststore.

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/360#issuecomment-1089170952_

ebaron:
Example stacktrace:
```
Apr 01, 2022 8:19:49 PM io.cryostat.core.log.Logger info
INFO: (10.217.0.1:33976): GET /health 200 41ms
Apr 01, 2022 8:19:52 PM io.cryostat.core.log.Logger info
INFO: (10.217.0.1:33754): GET /api/v1/notifications_url 200 0ms
Apr 01, 2022 8:19:52 PM io.cryostat.core.log.Logger error
SEVERE: HTTP 500: Internal Server Error
io.cryostat.net.web.http.api.v2.ApiException: Internal Server Error
Caused by: java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: javax.net.ssl.SSLHandshakeException: Failed to create SSL connection
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at io.cryostat.net.OpenShiftAuthManager.getLoginRedirectUrl(OpenShiftAuthManager.java:155)
	at io.cryostat.net.web.http.api.v2.AuthPostHandler.handle(AuthPostHandler.java:104)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:117)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:69)
	at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:313)
[...]

PR #882, opened 2022-04-07
fix(oauth): use Kubernetes client to query OAuth server

ebaron:
This PR uses the underlying HTTP client of the Fabric8 Kubernetes Client to make the GET request to query OAuth server metadata from the API server. The HTTP client is already configured to trust the cluster's root CA in order to talk to the API server as we do elsewhere in Cryostat [1].

We could instead add the CA cert bundle to our trust store at startup, but I'm concerned there may be edge cases where the CA bundle isn't mounted in the expected place. Leaving this to the Fabric8 client should give us less to think about. There's also some documentation that sounds like we should avoid trusting the CA bundle for anything other than talking to Kubernetes endpoints [2].

OkHttp and Jackson Databind are already dependencies of the Fabric8 Kubernetes Client, so we're not adding anything new.

In the tests, I had to remove the check that the TokenProvider is only called once. Some code paths now result in calling this twice. Maybe there's a better way to do this.

Fixes: #881 

[1] https://github.com/fabric8io/kubernetes-client#configuring-the-client
[2] https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#trusting-tls-in-a-cluster

### DIFF
--- a/src/main/java/io/cryostat/net/NetworkModule.java
+++ b/src/main/java/io/cryostat/net/NetworkModule.java
@@ -196,15 +196,14 @@ public abstract class NetworkModule {
     @Provides
     @Singleton
     static OpenShiftAuthManager provideOpenShiftAuthManager(
-            Environment env, Logger logger, FileSystem fs, WebClient webClient) {
+            Environment env, Logger logger, FileSystem fs) {
         return new OpenShiftAuthManager(
                 env,
                 logger,
                 fs,
                 token ->
                         new DefaultOpenShiftClient(
-                                new OpenShiftConfigBuilder().withOauthToken(token).build()),
-                webClient);
+                                new OpenShiftConfigBuilder().withOauthToken(token).build()));
     }
 
     @Binds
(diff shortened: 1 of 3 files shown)


andrewazores (commented):
Code looks good. Haven't had a chance to build and test this just yet.

Reviewing this reminded me that I had a branch in progress some time ago where I was working on creating a single long-lived `OpenShiftClient` instance, since that only depends on the service account token which shouldn't be changing within the lifecycle of a Cryostat container instance. There should be no need to create a new client instance for every request, and this would also apply to the new `OkHttpClient` instance, too. WDYT? This can/should be done as a separate refactoring, I think.

andrewazores (approved):
Built, deployed, and tested - everything looks like it works now.

jan-law (commented):
Code looks good to me. My quicklab cluster failed to install yesterday - trying again today.

ebaron:
> Reviewing this reminded me that I had a branch in progress some time ago where I was working on creating a single long-lived `OpenShiftClient` instance, since that only depends on the service account token which shouldn't be changing within the lifecycle of a Cryostat container instance. There should be no need to create a new client instance for every request, and this would also apply to the new `OkHttpClient` instance, too. WDYT? This can/should be done as a separate refactoring, I think.

I think that definitely makes sense for the requests using Cryostat's service account. We'd still need separate clients for the requests using the user's token, but I think this would still be a nice savings.

jan-law (approved):
Tested, looks good

PR #884, opened 2022-04-08
chore(openshift): use cached, lazy init client instances

### LINKED ISSUE #883: Use cached client instances when communicating with OpenShift API server
Reviewing this reminded me that I had a branch in progress some time ago where I was working on creating a single long-lived `OpenShiftClient` instance, since that only depends on the service account token which shouldn't be changing within the lifecycle of a Cryostat container instance. There should be no need to create a new client instance for every request, and this would also apply to the new `OkHttpClient` instance, too. WDYT? This can/should be done as a separate refactoring, I think.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/pull/882#pullrequestreview-935724938_

andrewazores:
Fixes #883 

Refactors the `OpenShiftAuthManager`, moving it into its own subpackage with its own DI module for encapsulation. The mounted namespace and account token files are only read from filesystem once, and the `OpenShiftClient` instance using the service account token is created once with lazy initialization and then that instance reused for the lifetime of the Cryostat container. `OpenShiftClient` instances using a user's token are stored in a cache, expiring (closing the client instance) after 5 minutes of inactivity on the client instance, so that we can reuse the client and its HTTP connection(s) while a user is actively making API requests through Cryostat. This reduces the number of memory allocations we're doing and should allow for some HTTP connection reuse, too. The time-based cache eviction ensures that the cache does not grow unbounded for long-lived Cryostat deployments which may have many different users over time.

### DIFF
--- a/src/main/java/io/cryostat/net/NetworkModule.java
+++ b/src/main/java/io/cryostat/net/NetworkModule.java
@@ -53,6 +53,7 @@ import io.cryostat.core.net.JFRConnectionToolkit;
 import io.cryostat.core.sys.Environment;
 import io.cryostat.core.sys.FileSystem;
 import io.cryostat.core.tui.ClientWriter;
+import io.cryostat.net.openshift.OpenShiftNetworkModule;
 import io.cryostat.net.reports.ReportsModule;
 import io.cryostat.net.security.SecurityModule;
 import io.cryostat.net.web.WebModule;
@@ -64,8 +65,6 @@ import dagger.Lazy;
 import dagger.Module;
 import dagger.Provides;
 import dagger.multibindings.IntoSet;
-import io.fabric8.openshift.client.DefaultOpenShiftClient;
-import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 import io.vertx.core.Vertx;
 import io.vertx.ext.web.client.WebClient;
 import io.vertx.ext.web.client.WebClientOptions;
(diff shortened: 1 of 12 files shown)


ebaron (commented):
Code changes look good to me. I'll give this a try in OpenShift.

Do you know if we're using any Caffeine 3.x API here? We still downgrade this to 2.x for our downstream builds to sync with Quarkus. Perhaps we should just switch upstream to 2.x as well since it's still being supported alongside 3.x.

andrewazores:
> Do you know if we're using any Caffeine 3.x API here? We still downgrade this to 2.x for our downstream builds to sync with Quarkus

I think it's all 2.x here still, the stuff being used here is the same as or even a smaller feature subset than what is already being used elsewhere ex. `TargetConnectionManager`

> Perhaps we should just switch upstream to 2.x as well since it's still being supported alongside 3.x.

Sure, I think that's fine too.

ebaron:
@andrewazores Is there a limit on the number of entries in the cache?

andrewazores:
No, the cache has unlimited size. The only eviction policy is the 5 minute inactivity timeout.

ebaron:
That should be fine. I doubt there would be many concurrent users in a single namespace in 5 minutes.

ebaron (approved):
Seems to work fine in OpenShift

Issue #885, opened 2022-04-11
CI configs should be updated for 2.1 release branch


ebaron:
Our Mergify config should be updated to backport PRs to the `cryostat-v2.1` branch. Additionally, the branch pattern matching in `ci.yaml` needs to be updated to support branches of the form `cryostat-vX.Y`.

PR #886, opened 2022-04-11
ci(release): update CI for cryostat-v2.1 branch

ebaron:
Fixes; #885

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,11 +6,13 @@ on:
       - main
       - v[0-9]+
       - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
   pull_request:
     branches:
       - main
       - v[0-9]+
       - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
 
 jobs:
   get-pom-properties:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -6,11 +6,13 @@ on:
       - main
       - v[0-9]+
       - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
   pull_request:
     branches:
       - main
       - v[0-9]+
       - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
```
andrewazores, line 15: Just noticing now but does the `.` here need to be escaped so that it matches a literal `.` character and not an "any-character"? I think this might match ex. `cryostat-v2a2`. I doubt we'll ever have branch names of that form, but still.

ebaron: It doesn't seem like `.` has special meaning in GitHub's syntax: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

andrewazores (approved):

andrewazores (approved):

PR #912, opened 2022-04-26
fix(snapshot): do not issue a Recording Deleted notification when empty snapshots are deleted

### LINKED ISSUE #911: Recording Deleted notification issued during empty snapshot handling
The empty snapshot handling workflow consists of creating a snapshot and if it turns out to be empty, deleting it before issuing a 202 response with a message explaining that the snapshot failed to create. When the empty snapshot is deleted there should be no Recording Deleted notification issued to avoid user confusion.

hareetd:
Fixes #911

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
@@ -252,7 +252,7 @@ public class RecordingTargetHelper {
     }
 
     public Future<Void> deleteRecording(
-            ConnectionDescriptor connectionDescriptor, String recordingName) {
+            ConnectionDescriptor connectionDescriptor, String recordingName, boolean isAnUnreadableSnapshot) {
         CompletableFuture<Void> future = new CompletableFuture<>();
         try {
             String targetId = connectionDescriptor.getTargetId();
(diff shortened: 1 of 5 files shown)


src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -252,7 +252,7 @@ public Future<Optional<InputStream>> getRecording(
     }
 
     public Future<Void> deleteRecording(
-            ConnectionDescriptor connectionDescriptor, String recordingName) {
+            ConnectionDescriptor connectionDescriptor, String recordingName, boolean isAnUnreadableSnapshot) {
```
andrewazores, line 255: Can/would callers from outside of this class be expected to know whether what they're deleting is an unreadable snapshot? The two examples here in the DeleteActiveRecordingMutator and the TargetRecordingDeleteHandler both pass `false` here, and I think that's likely to always be the case for other mutators/handlers.

I think it might be "nicer" to have methods like this:

```java
public Future<Void> deleteRecording(ConnectionDescriptor connectionDescriptor, String recordingName);
Future<Void> deleteRecording(ConnectionDescriptor connectionDescriptor, String recordingName, boolean sendNotification);
```

The mutators/handlers would call the `public` one, which in turn just wraps around the package-private one (which might also be better suited as a `protected` or `private`) and calls it with `true` for the final parameter, which is an inverted condition compared to how it is now.

andrewazores:
Change looks good. I think there's another half to this problem though. If I make an API request to create a snapshot recording, and that snapshot ends up being empty, then this prevents a notification from being emitted when that snapshot is automatically deleted due to being detected as empty. But I'm still going to receive a notification that the recording was created, right? However, when I go looking in the active recordings list for that recording, I won't find it there. The web-client that uses notifications to update state might also create a UI entry representing the snapshot based on that notification and then not automatically remove it since there is no corresponding deletion notification.

hareetd:
We're good on that end, the backend doesn't actually issue any notifications when creating snapshot recordings. Instead, the frontend is doing the notification handling in `ApiService#createSnapshot`, where a `Recording Creation` snapshot is only issued if the backend response is a 200, meaning the snapshot was created and passed verification.

It's been a while so I don't remember why I decided to do it this way, I could move the notification handling to the backend to make it more consistent with how notifications are handled for other recording-related actions?

andrewazores:
Oh, interesting. Yes I think that would be a good enhancement as well to bring that behaviour in line with everything else. I'll approve and merge this one on its own for now, but please file an issue for that.

andrewazores (approved):

PR #917, opened 2022-04-29
fix(snapshot): Cleanup notification handling for snapshots

### LINKED ISSUE #916: Snapshot creation notification handling should be delegated to the backend.
Currently, notification handling upon successful snapshot creation is done by the frontend whereas all other recording-specific notification handling is done by the backend. The `Recording Created` notification for snapshots should be issued by the backend for consistency reasons.

hareetd:
Fixes #916 
Depends on https://github.com/cryostatio/cryostat-web/pull/427

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
@@ -40,8 +40,6 @@ package io.cryostat.net.web.http.api.v1;
 import java.util.EnumSet;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.inject.Inject;
 
@@ -105,12 +103,7 @@ class TargetRecordingDeleteHandler extends AbstractAuthenticatedRequestHandler {
         String recordingName = ctx.pathParam("recordingName");
         ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
         try {
-            Matcher m = SNAPSHOT_NAME_PATTERN.matcher(recordingName);
-            if (m.matches()) {
-                recordingTargetHelper.deleteSnapshot(connectionDescriptor, recordingName).get();
-            } else {
-                recordingTargetHelper.deleteRecording(connectionDescriptor, recordingName).get();
-            }
+            recordingTargetHelper.deleteRecording(connectionDescriptor, recordingName).get();
             ctx.response().setStatusCode(200);
             ctx.response().end();
         } catch (ExecutionException e) {
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
@@ -59,7 +59,6 @@ import io.vertx.ext.web.handler.impl.HttpStatusException;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 
 class TargetRecordingDeleteHandler extends AbstractAuthenticatedRequestHandler {
-    private static final Pattern SNAPSHOT_NAME_PATTERN = Pattern.compile("^(snapshot\\-)([0-9]+)$");
 
     private final RecordingTargetHelper recordingTargetHelper;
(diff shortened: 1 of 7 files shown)


hareetd:
Thoughts before I finish updating tests?

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
```diff
@@ -102,7 +105,12 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
         String recordingName = ctx.pathParam("recordingName");
         ConnectionDescriptor connectionDescriptor = getConnectionDescriptorFromContext(ctx);
         try {
-            recordingTargetHelper.deleteRecording(connectionDescriptor, recordingName).get();
+            Matcher m = SNAPSHOT_NAME_PATTERN.matcher(recordingName);
+            if (m.matches()) {
+                recordingTargetHelper.deleteSnapshot(connectionDescriptor, recordingName).get();
+            } else {
+                recordingTargetHelper.deleteRecording(connectionDescriptor, recordingName).get();
+            }
```
andrewazores, line 113: This looks useful enough to extract out into a utility method within the `RecordingTargetHelper`, and that also helps encapsulate and consolidate the logic there so that this API handler is as simple as possible. ie the handler doesn't know anything about snapshot or not, it just knows how to delete active recordings by name, and the Helper is what figures out how to handle that exactly.

andrewazores (commented):
Looking good so far, just that one refactoring idea I commented

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingDeleteHandler.java
hareetd: I extracted determining whether a recording is a snapshot or not into the `RecordingTargetHelper` but I also got rid of the separate `deleteSnapshot` and `deleteRecording` methods. I feel it's less cluttered this way with the only recording deletion methods being `deleteRecording(ConnectionDescriptor connectionDescriptor, String recordingName` and `deleteRecording(ConnectionDescriptor connectionDescriptor, String recordingName, boolean isSnapshot, boolean issueNotification)`.

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -251,9 +250,13 @@ public Future<Optional<InputStream>> getRecording(
...
-    public Future<Void> deleteRecording(
-            ConnectionDescriptor connectionDescriptor, String recordingName) {
-        return this.deleteRecording(connectionDescriptor, recordingName, true);
+    public Future<Void> deleteRecording(ConnectionDescriptor connectionDescriptor, String recordingName) {
+        Matcher m = SNAPSHOT_NAME_PATTERN.matcher(recordingName);
+        if (m.matches()) {
+            return this.deleteRecording(connectionDescriptor, recordingName, false, true);
+        } else {
+            return this.deleteRecording(connectionDescriptor, recordingName, true, true);
+        }
```
andrewazores, line 259: Since the only difference here is the snapshot regex check and the boolean value passed into the call to the private method, does it make sense to just perform that check within the private method and remove the parameter from the method signature?

src/main/java/io/cryostat/recordings/RecordingTargetHelper.java
```diff
@@ -428,18 +447,17 @@ private Future<Void> deleteRecording(
...
+                                        if (isSnapshot) {
+                                            this.issueNotification(
+                                                    connectionDescriptor.getTargetId(),
+                                                    linkedDesc,
+                                                    SNAPSHOT_DELETION_NOTIFICATION_CATEGORY);
+                                        } else {
+                                            this.issueNotification(
+                                                    connectionDescriptor.getTargetId(),
+                                                    linkedDesc,
+                                                    DELETION_NOTIFICATION_CATEGORY);
```
andrewazores, line 459: I think it would be a little cleaner to just set the value of a `String notificationCategory` local variable based on `isSnapshot`, and then call `this.issueNotification(connectionDescriptor.getTargetId(), linkedDesc, notificationCategory)` identically either way.

andrewazores (commented):
Implementation looks good.

andrewazores (approved):

andrewazores (approved):

PR #924, opened 2022-05-11
feat(graphql): Add query for all archived recordings

### LINKED ISSUE #923: [Task] Enable revamp/redesign of the -web "All Archives" view
Backend equivalent of https://github.com/cryostatio/cryostat-web/issues/400. 

In order to populate the table containing uploaded recordings and not target-associated recordings, we need a way to query and filter all archived recordings. ~~New notifications for when targets are added/deleted are also required in order to keep the targets list up-to-date at all times.~~

hareetd:
Fixes #923 

Add a query for all archived recordings with the ability to filter on target source, name, and labels. Besides differentiating between JVM targets, the target source can also be used to retrieve only those recordings uploaded directly to archives (i.e. recordings not associated with a target).

Change the subdirectory name for uploaded recordings from "unlabelled" to "uploads" for clarity. 

Add a new field for archived recordings in the GraphQL Schema containing aggregate information such as the count of recordings. 

Depends on https://github.com/cryostatio/cryostat-web/pull/431

### DIFF
--- a/src/main/resources/types.graphqls
+++ b/src/main/resources/types.graphqls
@@ -116,7 +116,7 @@ type ArchivedRecording implements Recording {
 }
 
 type Archived {
-    recordings: [ArchivedRecording!]!
+    data: [ArchivedRecording!]!
     aggregate: AggregateInfo!
 }
 
--- a/src/main/resources/types.graphqls
+++ b/src/main/resources/types.graphqls
@@ -83,7 +83,7 @@ interface Node {
 
 type Recordings {
     active(filter: ActiveRecordingFilterInput): [ActiveRecording!]!
-    archived(filter: ArchivedRecordingFilterInput): [ArchivedRecording!]!
+    archived(filter: ArchivedRecordingFilterInput): Archived!
 }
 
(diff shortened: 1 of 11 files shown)


andrewazores:
`types.graphqls` needs to be updated as well to reflect the new addition to `ArchivedRecordingFilterInput`. This is just used for request validation to ensure that it matches the expected schema.

src/main/resources/types.graphqls
```diff
@@ -38,6 +38,7 @@ input ActiveRecordingFilterInput {
 input ArchivedRecordingFilterInput {
     name: String
     labels: [String]
+    sourceTarget: String
```
hareetd, line 41: @andrewazores I think I'm good to go on that front, unless I need to add something else here?

andrewazores: Oh, my bad. I could've sworn this change didn't show up in my diff viewer when I first looked at this.

andrewazores (approved):
Rebase please, I can't do it automatically as GitHub says there are conflicts.

andrewazores (approved):

PR #930, opened 2022-05-12
fix(vertx): deploy verticles with async decorated blocking handlers

### LINKED ISSUE #929: Concurrent API Requests can become blocked
In Cryostat 2.1.0, API requests are not properly handled concurrently. Issuing a slow API request and then issuing a fast API request before the first request finishes results in the fast request not getting a response until after the first request also completes.

2.1.0 bugged behaviour:
```
A
|    B
|    .    C
|    .    .
|    .    .
v    .    .
     |    .
     v    .
          |
          v
```

Expected behaviour:
```
A
|    B
|    |    C
|    v    |
|         v
v
```

Where the A, B, C start points represent the point at which the client issued the request. `|` represents time where Cryostat is processing the request - A is a slow request, B and C are fast. `.` represents time where the request is "stuck" and not being processed. `v` represents Cryostat finishing processing and sending a response to the client.

andrewazores:
Fixes #929

- tmp: add intentionally blocking API handler for testing
- fix(vertx): deploy verticles with async decorated blocking handlers

The first commit in this series is just for testing purposes. It adds an API handler at "/api/beta/block" that sleeps the current thread for 5 seconds, then returns a "hello" string response.

This can be used to test by doing:

```bash
https :8181/api/beta/block
```

in one terminal. In a second terminal, before that first request completes,

```bash
https :8181/health
```

In current `main`, the `/health` request will not return until after the `/block` request returns. After the fix in this PR is applied these requests are properly handled concurrently again.

### DIFF
--- a/src/main/java/io/cryostat/Cryostat.java
+++ b/src/main/java/io/cryostat/Cryostat.java
@@ -56,6 +56,8 @@ import io.cryostat.rules.RuleRegistry;
 
 import com.nimbusds.jose.crypto.bc.BouncyCastleProviderSingleton;
 import dagger.Component;
+import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Vertx;
 
 class Cryostat {
 
@@ -79,8 +81,20 @@ class Cryostat {
         client.credentialsManager().load();
         client.ruleRegistry().loadRules();
         client.ruleProcessor().enable();
-        client.httpServer().start();
-        client.webServer().start();
+        client.vertx()
+                .deployVerticle(
+                        client.httpServer(),
+                        new DeploymentOptions().setWorker(false),
+                        res -> {
+                            System.out.println("HTTP Server Verticle Started");
+                        });
+        client.vertx()
+                .deployVerticle(
+                        client.webServer(),
+                        new DeploymentOptions().setWorker(true),
+                        res -> {
+                            System.out.println("WebServer Verticle Started");
+                        });
         client.messagingServer().start();
         client.platformClient().start();
         client.recordingMetadataManager().load();
@@ -96,6 +110,8 @@ class Cryostat {
         RuleRegistry ruleRegistry();
 
         RuleProcessor ruleProcessor();
+
+        Vertx vertx();
 
         HttpServer httpServer();
 
--- a/src/main/java/io/cryostat/Cryostat.java
+++ b/src/main/java/io/cryostat/Cryostat.java
@@ -85,16 +85,12 @@ class Cryostat {
                 .deployVerticle(
                         client.httpServer(),
                         new DeploymentOptions().setWorker(false),
-                        res -> {
-                            System.out.println("HTTP Server Verticle Started");
-                        });
+                        res -> logger.info("HTTP Server Verticle Started"));
         client.vertx()
                 .deployVerticle(
                         client.webServer(),
                         new DeploymentOptions().setWorker(true),
-                        res -> {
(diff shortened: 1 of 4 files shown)


src/main/java/io/cryostat/Cryostat.java
```diff
@@ -79,8 +81,20 @@ public static void main(String[] args) throws Exception {
...
+                .deployVerticle(
+                        client.httpServer(),
+                        new DeploymentOptions().setWorker(false),
+                        res -> {
+                            System.out.println("HTTP Server Verticle Started");
+                        });
+        client.vertx()
+                .deployVerticle(
+                        client.webServer(),
+                        new DeploymentOptions().setWorker(true),
```
hareetd, line 94: The `WebServer` is deployed as a `Worker Verticle` here because we don't want the HTTP request handlers to be doing their work on an event loop thread, correct?

src/main/java/io/cryostat/net/web/WebModule.java
```diff
@@ -61,7 +61,6 @@ public abstract class WebModule {
     public static final String WEBSERVER_TEMP_DIR_PATH = "WEBSERVER_TEMP_DIR_PATH";
 
     @Provides
-    @Singleton
```
hareetd, line 64: From the Vert.x Core docs on Worker Verticles: 

"These run using a thread from the worker pool. An instance is never executed concurrently by more than one thread."

Is this why we no longer need to ensure the `WebServer` is a singleton (i.e. the WebServer Worker Verticle instance will only ever be executed on one thread)?

src/main/java/io/cryostat/Cryostat.java
andrewazores: Yep correct, we want the request handlers on worker threads.

src/main/java/io/cryostat/net/web/WebModule.java
andrewazores: In practice there is still only one WebServer instance since `Cryostat.java` only asks for one instance. But, it might be useful in the future to also be able to deploy mutliple WebServer verticles, since that's also a recommended way to structure concurrency with Vertx.

src/main/java/io/cryostat/Cryostat.java
andrewazores: The `BlockingHandlerDecorator` is a key piece of what's actually putting work onto worker threads too, btw. But from my experimentation here that only seemed to work correctly when the WebServer was deployed as a worker verticle.

andrewazores: I also just realized the naming is a little unfortunate. The `BlockingHandler` I added is the `/api/beta/block` API endpoint that blocks 5 seconds and then responds "hello", named because it blocks. `BlockingHandlerDecorator` is a Vertx class that wraps around a handler definition and perform its work on a worker thread then enqueues the next handler in series, named because it "decorates" (wraps) around handlers that block and lets them behave nice and async.

ebaron (approved):
Works as described for me. With just the block handler, the health check is blocked. With the fix, the health check doesn't block.

jan-law (dismissed):
Works well for me

andrewazores:
Just replaced `S.o.println` with `logger.info` and removed the `BlockingHandler` manual testing API endpoint. No other changes on that force push.

hareetd (approved):
Web-client is working well too with the changes.

ebaron (approved):

PR #956, opened 2022-05-25
chore(api): Move Probe handlers from beta to v2 package

### LINKED ISSUE #928: [Task] JMC Agent Probe handlers are in V2 API version but 'beta' package
Various Agent Probe handlers are marked as part of the V2 API version, but the files are located within the 'beta' subpackage. They should be moved into the correct `v2` package.

tthvo:
Fixes #928

Move Probe handlers from beta to v2 sub-package since they are available in v2 API. This includes moving source and test files.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
@@ -39,6 +39,12 @@ package io.cryostat.net.web.http.api.beta;
 
 import io.cryostat.net.web.http.RequestHandler;
 import io.cryostat.net.web.http.api.beta.graph.GraphModule;
+import io.cryostat.net.web.http.api.v2.ProbeTemplateDeleteHandler;
+import io.cryostat.net.web.http.api.v2.TargetProbePostHandler;
+import io.cryostat.net.web.http.api.v2.TargetProbeDeleteHandler;
+import io.cryostat.net.web.http.api.v2.TargetProbesGetHandler;
+import io.cryostat.net.web.http.api.v2.ProbeTemplateUploadBodyHandler;
+import io.cryostat.net.web.http.api.v2.ProbeTemplateUploadHandler;
 
 import dagger.Binds;
 import dagger.Module;
--- a/src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
@@ -39,12 +39,6 @@ package io.cryostat.net.web.http.api.beta;
 
 import io.cryostat.net.web.http.RequestHandler;
 import io.cryostat.net.web.http.api.beta.graph.GraphModule;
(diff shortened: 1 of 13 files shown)


src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
```diff
@@ -39,6 +39,12 @@
 
 import io.cryostat.net.web.http.RequestHandler;
 import io.cryostat.net.web.http.api.beta.graph.GraphModule;
+import io.cryostat.net.web.http.api.v2.ProbeTemplateDeleteHandler;
+import io.cryostat.net.web.http.api.v2.TargetProbePostHandler;
+import io.cryostat.net.web.http.api.v2.TargetProbeDeleteHandler;
+import io.cryostat.net.web.http.api.v2.TargetProbesGetHandler;
+import io.cryostat.net.web.http.api.v2.ProbeTemplateUploadBodyHandler;
+import io.cryostat.net.web.http.api.v2.ProbeTemplateUploadHandler;
```
andrewazores, line 47: Since these handlers have been moved out from the `/api/beta` version space and into `/api/v2`, I think it would also make sense for their `@Provides`/`@Binds` methods to be moved out of the `HttpApiBetaModule` and into `HttpApiV2Module`. What do you think? This way the handlers do not need to be `public` and do not need to be imported by any other class outside of their API version package.

tthvo: Yeh I think this will make more sense to that! I will move them now.

tthvo:
CI build failure looks like it runs into #666.

```
1371-2022-05-25T17:48:33.9875229Z [INFO] Running itest.ArchivedReportJwtDownloadIT
1372-2022-05-25T17:48:45.7598010Z [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 11.77 s <<< FAILURE! - in itest.ArchivedReportJwtDownloadIT
1373-2022-05-25T17:48:45.7599320Z [ERROR] testDownloadRecordingUsingJwt  Time elapsed: 11.769 s  <<< ERROR!
1374-2022-05-25T17:48:45.7599780Z java.util.concurrent.ExecutionException: java.lang.Exception: HTTP 500
1375:2022-05-25T17:48:45.7600374Z 	at itest.ArchivedReportJwtDownloadIT.testDownloadRecordingUsingJwt(ArchivedReportJwtDownloadIT.java:78)
1376-2022-05-25T17:48:45.7601033Z Caused by: java.lang.Exception: HTTP 500

```

src/main/java/io/cryostat/net/web/http/api/beta/HttpApiBetaModule.java
andrewazores: If the `public` isn't needed then I would remove it. Most of the time there is no need for a handler's implementation to be accessed directly by any other class, it only needs to be available to the relevant API module so that it can be `@Binds`/`@Provides`'d into the DI graph and registered with the webserver router generically. So the no-keyword, package-private visibility works well.

PR #957, opened 2022-05-26
fix(recordings): empty archive directories are now removed

### LINKED ISSUE #951: [Task] Empty archive directories are not removed
```
tree archive

archive
├── file-uploads
├── ONSXE5TJMNSTU2TNPA5HE3LJHIXS6L3KNZSGSL3SNVUTULZPMNZHS33TORQXIORZGA4TCL3KNV4HE3LJ
├── ONSXE5TJMNSTU2TNPA5HE3LJHIXS6L3KNZSGSL3SNVUTULZPMNZHS33TORQXIORZGA4TGL3KNV4HE3LJ
│   └── es-andrewazor-demo-Main_foo_20220401T212052Z.jfr
└── unlabelled
    ├── es-andrewazor-demo-Main_50mb_20211224T172734Z.2.jfr
    ├── es-andrewazor-demo-Main_50mb_20211224T172734Z.jfr
    └── es-andrewazor-demo-Main_foo_20220401T212052Z.jfr

4 directories, 4 files
```

Notice the first `ONSXE5...` subdirectory listed without any contents.

When the last archived recording for a given target is deleted, the archive subdirectory for that target is not removed. We should not leave these dangling since each one consumes some space within the file system. It may be a small amount, but over time these can accumulate to a large amount of space for no good reason.

turkeychikken:
Fixes #951 

There is a bug that is tracked with the spotbugs bug tracker 

`[ERROR] Medium: Nullcheck of entries at line 242 of value previously dereferenced in io.cryostat.recordings.RecordingArchiveHelper.pathIsEmpty(Path) [io.cryostat.recordings.RecordingArchiveHelper, io.cryostat.recordings.RecordingArchiveHelper] At RecordingArchiveHelper.java:[line 242]Redundant null check at RecordingArchiveHelper.java:[line 243] RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE`

I'm not sure why because I don't see a clear null checking in the code I added.

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
@@ -45,7 +45,6 @@ import java.net.URISyntaxException;
 import java.net.URLEncoder;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.temporal.ChronoUnit;
@@ -84,7 +83,6 @@ import io.cryostat.util.URIUtil;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.codec.binary.Base32;
-import org.apache.commons.io.FileUtils;
 
 public class RecordingArchiveHelper {
 
@@ -222,8 +220,8 @@ public class RecordingArchiveHelper {
                     .message(Map.of("recording", archivedRecordingInfo))
                     .build()
                     .send();
-            if (pathIsEmpty(parentPath)) {
-                FileUtils.deleteDirectory(parentPath.toFile());
+            if (fs.size(parentPath) == 0) {
+                fs.deleteIfExists(parentPath);
             }
             future.complete(archivedRecordingInfo);
         } catch (IOException | InterruptedException | ExecutionException | URISyntaxException e) {
@@ -233,15 +231,6 @@ public class RecordingArchiveHelper {
         }
 
         return future;
-    }
-
-    private Boolean pathIsEmpty(Path path) throws IOException {
-        if (Files.isDirectory(path)) {
-            try (DirectoryStream<Path> entries = Files.newDirectoryStream(path)) {
-                return !entries.iterator().hasNext();
-            }
-        }
-        return false;
     }
 
     private void validateSavePath(String recordingName, Path path) throws IOException {
--- a/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
@@ -220,7 +220,7 @@ public class RecordingArchiveHelper {
                     .message(Map.of("recording", archivedRecordingInfo))
                     .build()
                     .send();
-            if (fs.size(parentPath) == 0) {
+            if (fs.listDirectoryChildren(parentPath).isEmpty()) {
                 fs.deleteIfExists(parentPath);
             }
             future.complete(archivedRecordingInfo);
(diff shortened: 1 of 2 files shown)


andrewazores:
> There is a bug that is tracked with the spotbugs bug tracker

> [ERROR] Medium: Nullcheck of entries at line 242 of value previously dereferenced in io.cryostat.recordings.RecordingArchiveHelper.pathIsEmpty(Path) [io.cryostat.recordings.RecordingArchiveHelper, io.cryostat.recordings.RecordingArchiveHelper] At RecordingArchiveHelper.java:[line 242]Redundant null check at RecordingArchiveHelper.java:[line 243] RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE

> I'm not sure why because I don't see a clear null checking in the code I added.

Ah, yea. I've seen this before. IIRC SpotBugs is doing its analysis on the compiled bytecode, not directly on the Java source code you wrote. The try-with-resources construct gets compiled down to bytecode that performs a null check and so SpotBugs sometimes flags that as redundant. I thought we had got around that by updating our SpotBugs version at some point, but maybe not.

https://github.com/spotbugs/spotbugs/issues/756

You can just add a `@SuppressFBWarnings` annotation to ignore the false positive. Example:

https://github.com/cryostatio/cryostat/blob/5ebf150355095c071e265a04baec8f7256fb37d4/src/main/java/io/cryostat/net/web/WebServer.java#L258

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -220,6 +222,9 @@ public Future<ArchivedRecordingInfo> deleteRecording(String recordingName) {
                     .message(Map.of("recording", archivedRecordingInfo))
                     .build()
                     .send();
+            if (pathIsEmpty(parentPath)) {
+                FileUtils.deleteDirectory(parentPath.toFile());
```
andrewazores, line 226: Better to use our `FileSystem` abstraction:

https://github.com/cryostatio/cryostat-core/blob/main/src/main/java/io/cryostat/core/sys/FileSystem.java

There's already an instance of this within this class for you to use.

This is preferred because it's easily mockable in unit tests, basically. I think this class provides methods that will let you delete the directory as well as a simple way to check if it is empty.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -220,6 +220,9 @@ public Future<ArchivedRecordingInfo> deleteRecording(String recordingName) {
                     .message(Map.of("recording", archivedRecordingInfo))
                     .build()
                     .send();
+            if (fs.size(parentPath) == 0) {
```
andrewazores, line 223: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#size(java.nio.file.Path)

> The size of files that are not [regular](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#isRegularFile(java.nio.file.Path,%20java.nio.file.LinkOption...)) files is implementation specific and therefore unspecified.

I don't think `.size()` is the best check here due to the above description, since a directory is not a "regular file".

Perhaps `fs.listDirectoryChildren(parentPath).isEmpty()` would do the job?

turkeychikken: Ah, yes that makes sense thanks I'll try that.

turkeychikken: There is a test failure because I use fs.deleteIfExists(path) 3 times instead of 2 due to the addition of the method in the code. Is this a problem?

andrewazores: Not a problem, sounds like an expected change in behaviour. You'll have to update the unit tests as well.

turkeychikken:
running mvn clean verify on my laptop builds and passes all tests fine, i've run git status and git log and nothing seems wrong there...

andrewazores:
Looks like just a flakey test run. Those reports-related integration tests are usually the most problematic ones.

andrewazores (approved):

Issue #950, opened 2022-05-19
[Task] Add local-only healthcheck endpoint


andrewazores:
The current `GET /health` endpoint is used for liveness and readiness probes in OpenShfit/Kubernetes. This endpoint reports not only Cryostat's own status (by responding with a `2xx` response in a timely fashion) but also the status of its related services. That information about related services is useful when diagnosing deployment setups, but for the purposes of a platform liveness/readiness probe, we only need to check that Cryostat itself is up and responding to requests. If a related service like `jfr-datasource` is misconfigured this should have no impact on the probe status for Cryostat itself. For that reason we should have an endpoint at some other path that simply returns a `200` or `204` response unconditionally and immediately, and this should be used for automated platform health check probes. This will also need to be documented in `HTTP_API.md`

andrewazores:
Closed by #958

PR #958, opened 2022-05-27
feat(endpoint): added a local health endpoint

turkeychikken:
Solves #950 

haven't updated docs or added tests, thoughts?

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/generic/LocalHealthGetHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/generic/LocalHealthGetHandler.java
@@ -39,6 +39,8 @@ package io.cryostat.net.web.http.generic;
 
 import java.util.Set;
 
+import javax.inject.Inject;
+
 import io.cryostat.net.security.ResourceAction;
 import io.cryostat.net.web.http.RequestHandler;
 import io.cryostat.net.web.http.api.ApiVersion;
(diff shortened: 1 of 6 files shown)


andrewazores:
This pretty much looks like what I expected. I would say use the `204` status code and an empty response body, just for simplicity's sake. Rather than the path `/local`, which may not be immediately clear to an API client what that means, maybe use `/health/liveness`?

andrewazores:
Looks good, just update `HTTP_API.md` and add some basic unit tests please.

src/main/java/io/cryostat/net/web/http/generic/LocalHealthGetHandler.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.inject.Inject;
+
+import io.cryostat.net.security.ResourceAction;
+import io.cryostat.net.web.http.RequestHandler;
+import io.cryostat.net.web.http.api.ApiVersion;
+
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.RoutingContext;
+
+class LocalHealthGetHandler implements RequestHandler {
```
andrewazores, line 51: Rename this class please. Generally the class name follows the endpoint path and the HTTP verb, so in this case I would suggest "HealthLivenessGetHandler".

turkeychikken: Okay done.

andrewazores: Nope, not a problem at all. This is normal and expected :-)

andrewazores (approved):

Issue #973, opened 2022-06-02
Image tags must be lowercase


ebaron:
From a recent CI run:
```
Run redhat-actions/push-to-registry@v2
Creating temporary Podman image storage for pulling from Docker daemon
Overriding storage mount_program with "fuse-overlayfs" in environment
Warning: Reference to image and/or tag must be lowercase. Reference has been converted to be compliant with standard.
Combining image name "cryostat" and registry "quay.io/cryostat" to form registry path "quay.io/cryostat/cryostat"
🔍 Checking if the given image is manifest or not.
/usr/bin/podman version
/usr/bin/podman manifest exists cryostat:2.2.0-snapshot
/usr/bin/podman manifest exists cryostat:latest
🔍 Checking if "cryostat:2.2.0-snapshot, cryostat:latest" present in the local Podman image storage
/usr/bin/podman image exists cryostat:2.2.0-snapshot
/usr/bin/podman image exists cryostat:latest
Tag "cryostat:latest" found in Podman image storage
Warning: Tag "cryostat:2.2.0-snapshot" not found in Podman image storage
🔍 Checking if "cryostat:2.2.0-snapshot, cryostat:latest" present in the local Docker image storage
/usr/bin/podman --root /tmp/podman-from-docker-v3AXOl --storage-opt overlay.mount_program=/usr/bin/fuse-overlayfs pull docker-daemon:cryostat:2.2.0-snapshot
/usr/bin/podman --root /tmp/podman-from-docker-v3AXOl --storage-opt overlay.mount_program=/usr/bin/fuse-overlayfs pull docker-daemon:cryostat:latest
Removing temporary Podman image storage for pulling from Docker daemon
/usr/bin/podman --root /tmp/podman-from-docker-v3AXOl --storage-opt overlay.mount_program=/usr/bin/fuse-overlayfs rmi -a -f
Error: ❌ All tags were not found in either Podman image storage, or Docker image storage. Tag "cryostat:2.2.0-snapshot" not found in Podman image storage, and tags "cryostat:2.2.0-snapshot, cryostat:latest" not found in Docker image storage.
```

This is caused by a recent change in the push-to-registry action: https://github.com/redhat-actions/push-to-registry/issues/54

andrewazores:
This applies to all of our containers - [`-reports`](https://github.com/cryostatio/cryostat-reports/runs/6715676693?check_suite_focus=true) is failing too for the same reason.

ebaron:
[...]

PR #974, opened 2022-06-02
build(oci): transform image tag to lower case

ebaron:
Related to #973


andrewazores (commented):
Looks fine. Why use the plugin to transform it rather than just changing the `<version>` to lowercase directly?

ebaron:
I'm not sure if changing the version directly would impact how Maven does version comparisons. According to this `SNAPSHOT` is treated specially, but I couldn't find any official Maven docs about it: https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN401

andrewazores:
Ahh right, makes sense. Thanks.

Issue #946, opened 2022-05-17
[Story] Automated Analysis Scores in Dashboard


andrewazores:
Parent https://github.com/cryostatio/cryostat/issues/945

There should be an at-a-glance view for Automated Analysis reports scores embeddable in the cryostat-web Dashboard view. Rather than using the JMC-generated full-fledged HTML rules report, this should display as some more Patternfly-compliant components like Chips or Labels. N/A or green scores should not be included. This would provide users with a quick way to determine if there are any automatically detected performance issues on their application, without needing to select a specific recording and expand the table row to view the report document.

To facilitate this, Cryostat should simply take a snapshot recording of the selected target and then perform a rules report analysis on that snapshot. If there are no source recordings then no analysis is performed and some empty state content is displayed instead of scores indicators.

Tasks:
- [x] https://github.com/cryostatio/cryostat/issues/880
- [x] https://github.com/cryostatio/cryostat-reports/issues/23
- [x] https://github.com/cryostatio/cryostat/issues/918
- [x] https://github.com/cryostatio/cryostat-reports/issues/28
- [x] https://github.com/cryostatio/cryostat-core/issues/136
- [x] https://github.com/cryostatio/cryostat-web/issues/608

maxcao13:
Anything else for this story?

andrewazores:
Looks done.

PR #977, opened 2022-06-02
feat(reports): allow API requests to filter rules to evaluate

### LINKED ISSUE #918: [Task] API requests should be able to specify which rules to evaluate
Currently, all possible rules will be evaluated against the JFR file provided in the uploaded recording, and a requesting client can only try to limit the rules evaluations performed by disabling events before collecting the recording data. Once they do have a JFR file however there is no way to remove events that they are not interested in. The report generation endpoint should allow some way for clients to include a specification for names or categories of rules to evaluate against the accompanying JFR file, for example as a separate field in the POST form body, or as HTTP header. If this information is not provided then the report generation should default to evaluating all rules.

maxcao13:
Resolves #918 

Right now, I am wondering how I should have the optional `String filter` be used in the API, because there are a lot of methods and steps to go through receiving the request from the client, to going through execs, to finally sending a post request. 

I have two options right now as to how I should structure the optional `filter` string.

1. Should we overload all the methods such that we are able to specify an optional `filter` String as a parameter? The benefits of this I see are that it's easy to see if we are including "filters" in a request since a argument wouldn't be supplied, but the drawback is that it's doubling the methods that are needed to go through the process.

2. The second option is what I am doing right now in the code, where I am supplying a `filter` String parameter for all the methods that need it and I only check at the very end, (in this case: `exec` in `RemoteReportGenerator`) if the String is empty or not. If the string was empty, then we just send a POST request normally without the query string, if it wasn't etc. etc.... 

3. An unrelated question, but in `SubprocessReportGenerator` the class directly uses [cryostat-reports]'s InterruptibleReportGenerator method in `ReportResource.java` which I updated in the last pr. And in that class, I use I made my own predicate combiner thing, and I also use it in `SubprocessReportGenerator` by making a new util function. Is this the right approach so that I will eventually refactor the [cryostat-reports] class to use the util function as well?

Thanks!


andrewazores:
Approach 2 sounds good to me. No need to have overloaded versions for with- and without-filters - just let the filter String be nullable, or `Optional`.

> An unrelated question, but in SubprocessReportGenerator the class directly uses [cryostat-reports]'s InterruptibleReportGenerator method in ReportResource.java which I updated in the last pr. And in that class, I use I made my own predicate combiner thing, and I also use it in SubprocessReportGenerator by making a new util function. Is this the right approach so that I will eventually refactor the [cryostat-reports] class to use the util function as well?

Ah right, makes sense. I should have thought ahead to that point when reviewing your `-core` and `-reports` changes.

It's OK to add the util function and have it as essentially duplicate code across the repos for now. Later on we can extract that logic out into a util in `-core` so that both `-reports` and this project can both reuse it.

maxcao13:
Feedback please :)

src/main/java/io/cryostat/net/reports/RecordingDescriptor.java
```diff
@@ -47,10 +47,13 @@
 class RecordingDescriptor {
     final ConnectionDescriptor connectionDescriptor;
     final String recordingName;
+    final String filter;
```
andrewazores, line 50: `RecordingDescriptor` is meant to carry information about where the source recording file (which we're generating a report for) is located - the filter specification string doesn't seem like it fits well within this class with that context. Does that make sense? I think the `filter` field should be removed from here and passed around between method calls some other way.

src/main/java/io/cryostat/net/reports/ArchivedRecordingReportCache.java
```diff
@@ -71,22 +71,24 @@ class ArchivedRecordingReportCache {
...
         if (fs.isReadable(dest) && fs.isRegularFile(dest)) {
+            System.out.println("IM HERE"); // cached
             f.complete(dest);
             return f;
         }
 
         try {
             logger.trace("Archived report cache miss for {}", recordingName);
-
+            System.out.println("ACTUALLY HERE"); // non-cached
```
andrewazores, line 86: All these `printlns` will have to be removed before merge, but they're OK to keep for now while you're working through the implementation.

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -86,21 +86,26 @@ class ActiveRecordingReportCache {
...
     }
 
-    Future<String> get(ConnectionDescriptor connectionDescriptor, String recordingName) {
+    Future<String> get(
+            ConnectionDescriptor connectionDescriptor, String recordingName, String filter) {
         CompletableFuture<String> f = new CompletableFuture<>();
+        System.out.println("I AM IN ACTIVERECORDINGREPORTCACHE");
         try {
-            f.complete(cache.get(new RecordingDescriptor(connectionDescriptor, recordingName)));
+            f.complete(
```
andrewazores, line 97: How feasible is it, as a temporary solution, to ignore caching when the filter specification string is null/empty? In other words, only cache the report if it was generated using all available automated report rules. If the client specified some subset then we can generate the result and send it back, but not cache it.

src/main/java/io/cryostat/net/reports/RemoteReportGenerator.java
```diff
@@ -80,16 +81,27 @@ class RemoteReportGenerator extends AbstractReportGeneratorService {
...
+        System.out.println(String.format("IN REMOTE: THIS IS THE FILTER {%s}", filter));
         var form =
-                MultipartForm.create()
-                        .binaryFileUpload(
-                                "file",
-                                recording.getFileName().toString(),
-                                recording.toAbsolutePath().toString(),
-                                HttpMimeType.OCTET_STREAM.mime());
+                StringUtils.isNotBlank(filter)
+                        ? MultipartForm.create()
```
andrewazores, line 91: This seems like an awkward use of the ternary operator. Can't the form be created, the `.binaryFileUpload` applied regardless, and then just the `.attribute()` performed conditionally?

src/main/java/io/cryostat/net/web/http/api/v1/ReportGetHandler.java
```diff
@@ -119,10 +120,14 @@ public boolean isOrdered() {
     @Override
     public void handleAuthenticated(RoutingContext ctx) throws Exception {
         String recordingName = ctx.pathParam("recordingName");
+        List<String> queriedFilter = ctx.queryParam("filter");
+        String rawFilter = queriedFilter.isEmpty() ? null : queriedFilter.get(0);
```
andrewazores, line 124: If `null` here was `""`, then the null-checks performed everywhere else within the actual handling logic wouldn't need to be done, right? It would be nice to handle that argument as always being non-nullable after this point.

andrewazores: Even better, what if you used `RuleFilterParser` within each `Handler` implementation, and rather than passing the `String rawFilter` all the way deep into the call stack, you passed down something like an `Optional<Predicate<IRule>>`?

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
maxcao13: Seems like it would work as a temporary solution, I'll give it a shot for ActiveRecordingReportCache. Do you want the same functionality for the ArchivedRecordingReportCache as well? I think for ArchivedRecordingReportCache it's just an extra if statement to implement.

andrewazores: Yes, please. Ideally we can implement some smart caching so that we don't have to blow up the cache size by keeping a cached version for every possible permutation of the rules/topics, but that's a longer term feature and far out of scope for this particular issue.

In the end this is currently just mean to support https://github.com/cryostatio/cryostat/issues/946 , so as a proof-of-concept for that feature it's okay to compute the results on demand for each request. Once we nail down what we want that main feature to actually look like (which rules/topics to evaluate, and how we display that to the user on the dashboard) then we can worry about doing smarter caching.

src/main/java/io/cryostat/net/web/http/api/v1/ReportGetHandler.java
maxcao13: I looked into this and it feels like it's somewhat messy to be passing Optional<Predicate<IRule>> because of things like the `SubprocessReportGenerator` needing String arguments to pass through main(String[] args), as well as needing to import those 3 dependencies in every file. To me it feels like it would be easier to parse the String at the end since a String is easy to send and receive and such, what are your thoughts?

andrewazores: That's fair. At least doing the null-check at the point where you're receiving the client input seems like a good enough cleanup.

maxcao13: Also, the [cryostat-reports] endpoint that I implemented expects a rawString like "rule1,rule2,..." and `RemoteReportGenerator` just sends a request to that endpoint with a filter form attribute so I am not sure about this. The part I use the `RuleFilterParser` is with the `SubprocessReportGenerator` where (I think) [cryostat-reports] doesn't exist so the functionality is emulated within that process, so then I have to directly pass in the `Predicate<IRule>` filter into the `generateReportInterruptibly` method.

andrewazores: That's also a fair point. So, `String` it is.

maxcao13:
I will remove prints once OK'd.

andrewazores (commented):
Looking good, nice work.

PR #985, opened 2022-06-07
ci(submodule): submodule tracks branch

### LINKED ISSUE #984: [Meta] `web-client` submodule should track branch, not just commit hash

andrewazores:
Fixes #984

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "web-client"]
 	path = web-client
 	url = https://github.com/cryostatio/cryostat-web
+	branch = main
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -76,7 +76,7 @@ jobs:
         name: cryostat-core
         path: /home/runner/.m2/repository/io/cryostat/cryostat-core/
     - run: git submodule init
-    - run: git submodule update
+    - run: git submodule update --remote
     - run: mvn -B -U clean verify
     - name: Print itest logs
       if: ${{ failure() }}


ebaron (commented):
This never actually updates the submodule ref, right? It looks like our downstream builds aren't able to use `--remote` for `git submodule update`. What do you think about adding something to CI that pushes a commit to update the submodule whenever a push happens to the associated -web branch? Something like this maybe: https://stackoverflow.com/a/68213855

andrewazores:
Right, with this change it's still on the user or CI system to do the `git submodule update --remote` call to actually pull in the latest of whichever branch the submodule is tracking. And when that is done the main repo will still show the `web-client` submodule as having been changed in the index, so it'd still need a commit to clean that up.

> What do you think about adding something to CI that pushes a commit to update the submodule whenever a push happens to the associated -web branch? Something like this maybe: https://stackoverflow.com/a/68213855

Sure, that sounds good too. This way developers would just need to do a `git pull origin main` or equivalent to get the updated submodule automatically, right? It would also stick out as an "out of date with `main` branch" on any PRs that haven't done that.

ebaron:
> Sure, that sounds good too. This way developers would just need to do a `git pull origin main` or equivalent to get the updated submodule automatically, right? It would also stick out as an "out of date with `main` branch" on any PRs that haven't done that.

I think they would still need a `git pull origin main --recurse-submodules`, but that and `git submodule update` would both check out the most up-to-date web-client commit then.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "web-client"]
 	path = web-client
 	url = https://github.com/cryostatio/cryostat-web
+	branch = main
```
ebaron, line 4: This could really bite us when branching for a release. If we forget to update this, then CI will be including commits from main and we might not notice. There is the `.` option:
`
A special value of .  is used to indicate that the name of the branch in the submodule should be the same name as the current branch in the current repository.
`
But this breaks if you have a feature branch checked out...

andrewazores:
> This never actually updates the submodule ref, right? It looks like our downstream builds aren't able to use `--remote` for `git submodule update`. What do you think about adding something to CI that pushes a commit to update the submodule whenever a push happens to the associated -web branch? Something like this maybe: https://stackoverflow.com/a/68213855

I like this approach, and the more time I spend thinking about this the more it seems like the most reasonable and effective solution. Do you want to open a PR for it?

ebaron:
> > This never actually updates the submodule ref, right? It looks like our downstream builds aren't able to use `--remote` for `git submodule update`. What do you think about adding something to CI that pushes a commit to update the submodule whenever a push happens to the associated -web branch? Something like this maybe: https://stackoverflow.com/a/68213855
> 
> I like this approach, and the more time I spend thinking about this the more it seems like the most reasonable and effective solution. Do you want to open a PR for it?

Sure, I'll work on it.



> This will still require developers to do `git submodule update` locally to ensure that they're building images with the correct web-client version. I wonder if we can add a local git hook to do that automatically after `git pull`, or something? But then maybe that gets annoying when you're intentionally checking out a feature branch.

I think this should do it [1]: 
`git config --global submodule.recurse true`

[1] https://stackoverflow.com/a/49427199

ebaron (approved):
I think what I have planned can build off this PR.

PR #992, opened 2022-06-09
fix(notifications): added a target field to archivedRecording deletion notifications

### LINKED ISSUE #986: [Task] Allow ArchivedRecordings to include target field in archivedRecording notifications
Backend implementation of https://github.com/cryostatio/cryostat-web/issues/430

maxcao13:
Fixes #986 

Thoughts?

### DIFF
--- a/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
@@ -101,6 +101,7 @@ public class RecordingArchiveHelper {
     private static final String SAVE_NOTIFICATION_CATEGORY = "ActiveRecordingSaved";
     private static final String DELETE_NOTIFICATION_CATEGORY = "ArchivedRecordingDeleted";
 
+    public static final String UNLABELLED = "unlabelled";
     public static final String ARCHIVES = "archives";
 
     RecordingArchiveHelper(
@@ -218,7 +219,7 @@ public class RecordingArchiveHelper {
                             recordingMetadataManager.deleteRecordingMetadataIfExists(
                                     ARCHIVES, recordingName));
             String targetId = parentPath.getFileName().toString();
-            if (!targetId.equals("unlabelled")) {
+            if (!targetId.equals(UNLABELLED)) {
                 targetId = new String(base32.decode(targetId), StandardCharsets.UTF_8);
             }
             notificationFactory
--- a/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
@@ -218,10 +218,11 @@ public class RecordingArchiveHelper {
                             webServerProvider.get().getArchivedReportURL(filename),
                             recordingMetadataManager.deleteRecordingMetadataIfExists(
                                     ARCHIVES, recordingName));
-            String targetId = parentPath.getFileName().toString();
-            if (!targetId.equals(UNLABELLED)) {
-                targetId = new String(base32.decode(targetId), StandardCharsets.UTF_8);
-            }
+            String subdirectoryName = parentPath.getFileName().toString();
+            String targetId =
+                    (subdirectoryName.equals(UNLABELLED))
+                            ? ""
+                            : new String(base32.decode(subdirectoryName), StandardCharsets.UTF_8);
             notificationFactory
                     .createBuilder()
                     .metaCategory(DELETE_NOTIFICATION_CATEGORY)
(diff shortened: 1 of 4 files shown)


hareetd:
Looks good, nice work. However, there is one other (known) case affected by a lack of a target field, see [here](https://github.com/cryostatio/cryostat-web/pull/431#issuecomment-1149067098). Please update the necessary notifications to allow us to differentiate between uploaded vs. saved archived recordings on the frontend. Sorry, I should have documented both cases in the original issue; I'll do that now.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -213,11 +213,13 @@ public Future<ArchivedRecordingInfo> deleteRecording(String recordingName) {
                             webServerProvider.get().getArchivedReportURL(filename),
                             recordingMetadataManager.deleteRecordingMetadataIfExists(
                                     ARCHIVES, recordingName));
+            byte[] decodedParentPath = base32.decode(parentPath.getFileName().toString());
+            String decodedTargetId = new String(decodedParentPath, StandardCharsets.UTF_8);
```
hareetd, line 217: I just realized this doesn't account for the case where `parentPath.getFileName().toString() == "unlabelled"` (i.e. the recording to be deleted is in the uploads subdirectory). We should skip decoding in that case. 

Looks good otherwise.

andrewazores:
I think the `web-client` submodule changes in this PR are probably accidental and should be removed. They appear to wind the submodule back several commits to an older version.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -213,11 +214,15 @@ public Future<ArchivedRecordingInfo> deleteRecording(String recordingName) {
...
+            String targetId = parentPath.getFileName().toString();
+            if (!targetId.equals(UNLABELLED)) {
+                targetId = new String(base32.decode(targetId), StandardCharsets.UTF_8);
+            }
             notificationFactory
                     .createBuilder()
                     .metaCategory(DELETE_NOTIFICATION_CATEGORY)
                     .metaType(HttpMimeType.JSON)
-                    .message(Map.of("recording", archivedRecordingInfo))
+                    .message(Map.of("recording", archivedRecordingInfo, "target", targetId))
```
andrewazores, line 225: Kinda wondering about what to do in the case where the subdirectory/targetId is the unlabelled directory. I'm not sure we actually want to expose that `"unlabelled"` string value as part of the API surface here. IMO that's just an internal representation/implementation detail, but conceptually those recordings really don't have a targetId. So perhaps we leave that key out of the map still, or perhaps we set it to the empty string. @hareetd WDYT?

andrewazores: Yea, that's fair. It's nice to not have to check for null/undefined all the time, though on the frontend side the "falsey-ness" of `undefined`/`null` makes it a little less ugly (`if (!!notification.targetId)`, `notification?.targetId`) than null handling on the backend.

hareetd: It actually has more to do with how I've changed the `ArchivedRecordingsTable` than any null-checking, really. I've set it up such that it can be used both as a target-specific table (inside the `Recordings` and `AllTargetsArchivedRecordingsTable` components) and non-target-specific (inside the `Uploads` tab). In order to do this, I've included `target?: Observable<Target>` inside the props for the `ArchivedRecordingsTable`. The target field is optional since the `Uploads` doesn't have any real target to provide.

In order to account for this, I've had to add a number of conditionals throughout the component, such as in the notification handling. However, if the archived recording notifications were to always have a `target` field, even for uploaded recordings, I would be able to make the target prop required and pass in the following fake target for the Uploads table:

```
const target: Target = {
     connectUrl: ' ',
     alias: ' ',
}
```

The `connectUrl` would then be matched to the empty string provided by the notifications' target field for uploaded recordings, reducing the need for some of the conditionals used throughout the component. Thoughts?

andrewazores: That's fair. I think from the API perspective the empty string makes enough sense (though we should find somewhere to document this), so if that also simplifies the frontend then that's a nice net positive.

hareetd: I should add by "null-checking", I'm referring to that of the notifications' target field. The aforementioned conditionals do involve null-checking the (currently) optional target prop and acting accordingly.

andrewazores: The `*public* static final` on the constant isn't what I meant by "API surface area". What I meant by that is exposed values to clients communicating with Cryostat over the network, like clients listening to these WebSocket notifications. The constant can still be left as `public` so that it's visible at build time across different packages.

But yes, if the `"targetId"` (archive subdirectory) is the unlabelled one, then the notification's `target` field should be the empty string rather than the actual `"unlabelled"` value.

andrewazores (approved):
LGTM. @hareetd any other feedback?

hareetd (approved):

andrewazores (approved):

Issue #895, opened 2022-04-14
[Story] Stored credentials should be able to match multiple targets


andrewazores:
Currently, stored credentials are essentially a conceptual mapping like `<targetId, <jmx_user, jmx_pass>>`. This works well enough, but it doesn't work well for larger-scale application deployments, or in particular for deployments with varying numbers of application container replicas that run the same actual code and have the same configuration for JMX credentials. For example, defining a JMX credential for a specific target application replica enables Automated Rules actions on that replica, but scaling the deployment to add a second replica will result in the existing Automated Rule failing to connect to the new replica because the existing credentials will not be associated with the new replica's different targetId (different Endpoint IP, assuming this is OpenShift for simplicity).

We already have two kinds of similar mechanism that could be repurposed and reused here: the Automated Rules matchExpression, or the GraphQL query filter. The matchExpression seems more natural and more powerful and would be easy to repurpose. Stored credentials would then simply be a conceptual mapping like `<matchExpression, <jmx_user, jmx_pass>>`. The next question that arises is which stored credential to use if there are multiple stored credentials where the expression evaluates to `true` for a given target application.

Tasks:
- [x] ~~Determine which stored credential to use if there are multiple stored credentials where the expression evaluates to `true` for a given target application. Pick the first? Pick a random one? Try each? Do we record the result of what happened? Where?~~The first matching credential is picked. It is up to the user to ensure they do not create overlapping credentials.
- [ ] Implement better storage, ie in a document database. Related to cryostatio/cryostat#937, probably the same database but a different table/collection
- [x] cryostatio/cryostat-web#465

andrewazores:
The database chosen for pluggable discovery ended up being a relational/SQL db, not a document store, so storing the credentials within the database is not as much of a gain anymore. Closing this for now, but we may want to revisit this idea at a later date, perhaps for 3.0.

PR #1000, opened 2022-06-16
feat(credentials): implement matchExpression-based credentials

andrewazores:
Related to #895
Depends on https://github.com/cryostatio/cryostat-web/pull/475

This PR enhances the existing CredentialsManager by replacing the `targetId` field of stored credentials with a `matchExpression`. This uses the same expression evaluation internal infrastructure already in place for Automated Rules, so the syntax and behaviour is exactly the same. There is a migration function that checks all stored credentials already on disk and checks if they have the `targetId` field. If they do not then they are skipped over, since they are either invalid files or conform to the new format already.

```json
{
  "targetId": "foo",
  "credentials": {
    "username": "user",
    "password": "pass"
  }
}
```

is migrated by applying a simple transformation to:

```json
{
  "matchExpression": "target.connectUrl == \"foo\"",
  "credentials": {
    "username": "user",
    "password": "pass"
  }
}
```

This allows the new expression-based system to migrate existing credentials while maintaining the same semantics, and also allows the old credentials API endpoints to continue working with the same behaviour.

Still TODO in follow-up PRs:
1. Add new API endpoints to allow clients to define and delete credentials with generalized `matchExpression`, not only the backward-compatible `target.connectUrl == "foo"` style
2. ~~Use the new `CredentialsGetHandler` on the frontend to enhance the Security view to better reflect these expression-based credentials~~ https://github.com/cryostatio/cryostat-web/issues/465

Since it is now possible for one set of credentials to match multiple targets, the `AbstractAuthenticatedRequestHandler` no longer deletes credentials if they are used for a JMX connection and the connection fails - since the credentials may apply to multiple targets, they may be valid for other targets than the one that was just checked.

It is also possible that one target may be matched by multiple sets of credentials. There is no protection against this case, so it would simply be considered a user configuration error. If this occurs then the `credentialsManager` will simply provide the "first" matching set of credentials for a given target when requested. The ordering of credentials is intentionally undefined at this time. Multiple credentials for a given target is not an expected valid configuration and so no scheme of trying each of them in order is attempted, or any other more complex behaviours.

### DIFF
--- a/src/test/java/io/cryostat/configuration/CredentialsManagerTest.java
+++ b/src/test/java/io/cryostat/configuration/CredentialsManagerTest.java
@@ -39,6 +39,7 @@ package io.cryostat.configuration;
 
 import java.io.BufferedReader;
 import java.io.StringReader;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
@@ -48,10 +49,13 @@ import java.util.Map;
 import java.util.Set;
 
 import io.cryostat.MainModule;
+import io.cryostat.configuration.CredentialsManager.MatchedCredentials;
 import io.cryostat.core.log.Logger;
+import io.cryostat.core.net.Credentials;
 import io.cryostat.core.sys.FileSystem;
 import io.cryostat.messaging.notifications.NotificationFactory;
 import io.cryostat.platform.PlatformClient;
+import io.cryostat.platform.ServiceRef;
 import io.cryostat.rules.MatchExpressionEvaluator;
 
 import com.google.gson.Gson;
@@ -59,6 +63,7 @@ import com.google.gson.JsonObject;
 import org.apache.commons.codec.binary.Base32;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
@@ -67,7 +72,9 @@ import org.mockito.ArgumentCaptor;
 import org.mockito.InOrder;
 import org.mockito.Mock;
 import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.stubbing.Answer;
 
 @ExtendWith(MockitoExtension.class)
 class CredentialsManagerTest {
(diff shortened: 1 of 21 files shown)


jan-law (approved):
Works well

andrewazores:
Actually, hold on. I'll add unit tests for the CredentialsManager. It probably should have had them before and it certainly should now.

andrewazores:
Not done adding tests yet, just added basic tests for the new migration function so far. I'll add some more tomorrow.

hareetd (commented):
I haven't reviewed the tests yet but the implementation changes look good. However, since the credentials-related notifications have been changed to accommodate the match expression change, the `StoreJmxCredentials` component no longer updates the table state properly when credentials are added or deleted.

andrewazores:
Yea, I'll fix that with TODO 2 listed in the PR body.

andrewazores:
Filing the `-web` side to do later, since I'm on PTO next week and won't have any time to get started on it. I'll leave it open for anyone else to take a stab at if there's interest, otherwise I'll start on it when I'm back to work.

src/test/java/io/cryostat/configuration/CredentialsManagerTest.java
```diff
@@ -0,0 +1,540 @@
...
+            JsonObject credentials = json.getAsJsonObject("credentials");
+            String foundUsername = credentials.getAsJsonPrimitive("username").getAsString();
+            String foundPassword = credentials.getAsJsonPrimitive("password").getAsString();
+            MatcherAssert.assertThat(foundUsername, Matchers.equalTo(username));
+            MatcherAssert.assertThat(foundPassword, Matchers.equalTo(password));
+        }
+    }
+
+    @Nested
+    class Loading {
```
hareetd, line 452: Could you explain the difference between these two tests? I noticed that the second one doesn't require `Mockito.when(platformClient.listDiscoverableServices()).thenReturn(List.of(serviceRef));` even though they seem functionally identical.

andrewazores: Between the two `Loading` tests? One calls `credentialsManager.getCredentials(ServiceRef serviceRef)` and the other calls `credentialsManager.getCredentialsByTargetId(String targetId)`. The `byTargetId` one is there to preserve backward-compatibility with the older API and the `AbstractAuthenticatedRequestHandler`, so that the new style can be used for API paths like `/api/v1/targets/:targetId/recordings`, where we don't have a full `ServiceRef` object to evaluate the `matchExpression` against. In that case, the credentials manager searches for known targets that match the given `targetId`, maps that to a `ServiceRef` if it can, and then checks for matching credentials for that. The `getCredentials(ServiceRef)` just directly looks for credentials that match the supplied `ServiceRef`, without having to call out to the platform client to query for targets that match (since this is already a reference to a target).

hareetd: Makes sense, thanks. I should've just diffed the two files instead of looking to see what they differences were; missed the `getCredentials(...)` vs. `getCredentialsByTargetId(...)` calls :P

hareetd (approved):

jan-law:
Backend looks good, I'm assuming you'll merge this once the web client gets updated

andrewazores:
`-web` side merged and updated in this feature branch. Nothing else has changed since previous review approvals.

hareetd (approved):

hareetd (commented):
Looks like the web-client submodule hasn't been updated with the frontend changes.

andrewazores:
It isn't listed as a commit in this PR since I rebased on top, but the latest frontend change in the submodule is already in `main` thanks to our new CI-bot: https://github.com/cryostatio/cryostat/commit/d4b448ccc598cefc6d02e186b45b82aae5360db6

hareetd:
Oh that makes sense then, cool.

andrewazores:
No more manually updating the submodule :-)

PR #1018, opened 2022-07-07
feat(tests): repeat test script enhancements

### LINKED ISSUE #1017: [Task] `repeated-*-tests.bash` scripts fail without manual build first
It would also be nice for these scripts to allow selecting which tests to run, rather than always running the full suite.

https://github.com/cryostatio/cryostat/pull/1009#issuecomment-1176272931

andrewazores:
Fixes #1017

- feat(tests): allow MVN env var to override maven binary location
- fix(tests): ensure log file directories and files are created and truncated
- ensure tests are compiled before run
- implement second arg as test selector

### DIFF
--- a/repeated-integration-tests.bash
+++ b/repeated-integration-tests.bash
@@ -57,6 +57,10 @@ while [ "${runcount}" -lt "${runs}" ]; do
     timestamp="$(date -Iminutes)"
     client_logfile="$DIR/target/${POD_NAME}-${timestamp}.client.log"
     server_logfile="$DIR/target/${POD_NAME}-${timestamp}.server.log"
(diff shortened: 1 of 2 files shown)


andrewazores:
With this change you should be able to do ex.

```bash
$ mvn clean # simulate a freshly cloned repo - or actually freshly clone one
$ bash repeated-integration-tests.bash 2 ReportIT # run the ReportIT test only, and run it twice
```

similarly

```bash
$ bash repeated-unit-tests.bash 10 RecordingTargetHelperTest
```

maxcao13 (approved):
Works well!

hareetd (approved):

PR #1019, opened 2022-07-11
feat(jmxcredentials): add v2.2 API endpoints for JMX credentials

andrewazores:
Related to #895
Related to cryostatio/cryostat-web#465
Depends on https://github.com/cryostatio/cryostat-web/pull/481

- chore(credentials): refactor to store on-disk only using numeric ID
- add new methods for retrieving and referencing credentials by numeric ID
- feat(credentials): add v2.2 API endpoints

### DIFF
--- a/src/main/java/io/cryostat/configuration/CredentialsManager.java
+++ b/src/main/java/io/cryostat/configuration/CredentialsManager.java
@@ -282,49 +282,6 @@ public class CredentialsManager {
         return result;
     }
 
-    @Deprecated(since = "2.2")
-    /**
-     * @deprecated use {@link getAll} instead
-     */
-    Collection<String> getMatchExpressions()
-            throws JsonSyntaxException, JsonIOException, IOException {
-        Set<String> expressions = new HashSet<>();
-        for (String pathString : this.fs.listDirectoryChildren(credentialsDir)) {
-            Path path = credentialsDir.resolve(pathString);
-            try (BufferedReader br = fs.readFile(path)) {
-                StoredCredentials sc = gson.fromJson(br, StoredCredentials.class);
-                expressions.add(sc.getMatchExpression());
-            }
-        }
-        return expressions;
-    }
-
-    @Deprecated(since = "2.2")
-    /**
-     * @deprecated use {@link getAll} instead
-     */
-    public List<MatchedCredentials> getMatchExpressionsWithMatchedTargets()
-            throws JsonSyntaxException, JsonIOException, IOException {
-        List<MatchedCredentials> result = new ArrayList<>();
-        List<ServiceRef> targets = platformClient.listDiscoverableServices();
-        for (String expr : getMatchExpressions()) {
-            Set<ServiceRef> matchedTargets = new HashSet<>();
-            for (ServiceRef target : targets) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/net/web/http/api/v2/CredentialGetHandler.java
```diff
@@ -0,0 +1,127 @@
...
+            throws ApiException {
+        int id = Integer.parseInt(params.getPathParams().get("id"));
+        try {
+            String matchExpression = credentialsManager.get(id);
+            Set<ServiceRef> targets = credentialsManager.resolveMatchingTargets(id);
+            MatchedCredentials match = new MatchedCredentials(matchExpression, targets);
+            return new IntermediateResponse<MatchedCredentials>().body(match);
+        } catch (IOException ioe) {
+            throw new ApiException(500, ioe);
+        }
```
jan-law, line 125: If I try to get credentials with a nonexistent id, I get back an HTTP 500 instead of a 404. Should the response be a 404 Not Found instead?

```
curl http://0.0.0.0:8181/api/v2.2/credentials/2
{"meta":{"type":"text/plain","status":"Internal Server Error"},"data":{"reason":null}}
```

src/main/java/io/cryostat/configuration/CredentialsManager.java
```diff
@@ -204,11 +226,86 @@ public Collection<ServiceRef> getServiceRefsWithCredentials() {
...
+        }
+        return expressions;
     }
 
-    public List<MatchedCredentials> getMatchExpressionsWithMatchedTargets() {
+    @Deprecated(since = "2.2")
+    /**
+     * @deprecated use {@link getAll} instead
+     */
+    public List<MatchedCredentials> getMatchExpressionsWithMatchedTargets()
```
jan-law, line 307: VSCode > "find all references" says this method is only used for unit tests. If this method isn't used in the source code anymore, why is it advantageous to mark a method as deprecated instead of deleting it and updating the unit tests?

hareetd (commented):
Code changes look good and the API endpoints are working well. Just had the one comment above.

hareetd (approved):

src/main/java/io/cryostat/net/web/http/api/v2/CredentialGetHandler.java
andrewazores: Thanks, fixed. I also added unit tests which I realized were missing for these handlers.

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: Good catch. I'll tackle removing that and updating tests tomorrow.

andrewazores: Fixed in latest commits.

jan-law: Did you also mean to remove `getMatchExpressions()` and `getMatchExpressionsWithMatchedTargets()` or leave them annotated as deprecated?

PR #1020, opened 2022-07-12
feat(notifications): implement vertx event bus to listen for notifications emitted

### LINKED ISSUE #921: [Task] Report continues to show incomplete results after recording stops
It seems like the report for a running recording is not updated after the recording stops.

Example of `quarkus-test` sample app, 120s recording, ALL template. Recording stopped, but the report was initially generated while recording was running:
Screenshot 2022-05-04 at 15-46-40 Recordings

After archiving the recording, this also matches the report generated by downloading the JFR file and opening in JMC:
Screenshot 2022-05-04 at 15-50-44 Recordings

The "GC Caused by Heap Inspection" score increases, with the description going from 2 GCs to 4 GCs.

maxcao13:
Fixes #921 

So far I've implemented the feature very shallowly, thoughts on what needs to be done on a larger scale? I'm not quite sure what you meant by the NotificationListener thin wrapper feature. Right now, every consumer handlers trigger on any DeleteActiveRecording Notification, and that might not be good so I was thinking maybe passing the RecordingDescriptor key when publishing the notification to the address and then each consumer just if else checking the message? Is that why we need a wrapper around a Notification to better facilitate this?

Also, would you recommend resources for understanding the module, provides, component, injection design patterns of Dagger, Vertx and other inversion of control frameworks? This part of back-end development is definitely something I would like to get more comfortable with along my time on the project.


andrewazores:
> I'm not quite sure what you meant by the NotificationListener thin wrapper feature.

I just meant to define some interface like `NotificationListener`, so that you can invert control and rather than each class needing to have the Vertx injected and subscribe to the event bus, you could have such classes `implement NotificationListener` and have a Dagger module `@Binds @IntoSet` for these listeners (very much like `RequestHandler`s). This would mostly just be to reduce how much surface area Vertx has in our code base, so that if/when we migrate to another web framework it's already abstracted away and less of a chore to change implementations, and also to simplify unit testing in the meantime. I'll write some snippets of example code illustrating the idea here.

> Right now, every consumer handlers trigger on any DeleteActiveRecording Notification, and that might not be good so I was thinking maybe passing the RecordingDescriptor key when publishing the notification to the address and then each consumer just if else checking the message? Is that why we need a wrapper around a Notification to better facilitate this?

It's the nature of a pub/sub system or an event bus to send out messages broadly, and on the other end try to listen only for messages you're likely to care about. In this case the event bus "address", which for us is the notification category, fills that role. Beyond that, the message consumer may need to do some further check to determine if the message is something it does care about, which you've handled here.

> Also, would you recommend resources for understanding the module, provides, component, injection design patterns of Dagger, Vertx and other inversion of control frameworks? This part of back-end development is definitely something I would like to get more comfortable with along my time on the project.

Well, Vertx isn't an inversion-of-control framework. Dagger gives us dependency injection, which is generally a kind of inversion of control. For that, [Dagger's own docs](https://dagger.dev/dev-guide/) are pretty good. DI isn't limited to backend development either, in fact Dagger is probably much more commonly used in mobile development than it is in backend development.

andrewazores:
```java
interface NotificationListener<T> {
    Set<String> categories();
    void onMessage(String category, T message);
}

class NotificationPublisher {
    Vertx vertx;
    Set<NotificationListener> listeners;

    NotificationPublisher() {
        listener.forEach(listener -> {
            listener.categories().forEach(category -> {
                vertx.eventBus().consumer(category).handler(message -> listener.onMessage(category, message));
            });
        });
    }

    <T> void send(String category, T message) {
        vertx.eventBus().publish(category, message);
    }
}

class NotificationModule {
    @Binds @IntoSet abstract NotificationListener bindTestListener(TestListener listener);

    // in Cryostat we already have the NotificationFactory that is something like this
    @Provides static NotificationPublisher provideNotificationPublisher(Vertx vertx, Set<NotificationListener> listeners) {
        return new NotificationPublisher(vertx, listeners);
    }
}

class TestListener implements NotificationListener<Foo> {
    @Override
    public Set<String> categories() {
        return Set.of("cat", "dog");
    }

    @Override
    public void onMessage(String category, Foo message) {
        System.out.println(String.format("[%s]: %s", category, message.asJson()));
    }
}

class TestSender {
    NotificationPublisher publisher;

    void hello() {
        publisher.send("dog", new Foo(1234));
    }
}
```

That's a little sloppy, but I hope it gets the point across. This way only the publisher class actually knows anything about Vertx, and it uses it in a very isolated fashion. The listeners don't need to add a Vertx dependency injection, and the only notification-related tests needed on the `TestListener` would be to simply call `onMessage` and verify that the expected result occurred (ex. `Mockito.verify(dep).doSomething()`).

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -177,4 +183,27 @@ protected String getReport(RecordingDescriptor recordingDescriptor, String filte
             }
         }
     }
+
+    private void registerConsumer(RecordingDescriptor key) {
+        MessageConsumer<String> consumer = this.vertx.eventBus().consumer(STOP_NOTIFICATION_CATEGORY);
```
andrewazores, line 188: You can just subscribe to that `STOP_NOTIFICATION_CATEGORY` when this cache class is initialized. In the message handler callback you check if the cache contains the key from the message, and invalidate the cache entry if so. Otherwise you just ignore the message. The message itself should contain some information beyond just the category, so ex. you should be able to get the target and recording name out of it to check if that's in your cache. With the way this is implemented now I think you can run into some effects where the message handler you have registered can end up invalidating a cache entry for a different recording that was stopped, since the `message` contents are not actually checked - you just register the invalidation handler whenever `getReport` is called and then essentially invalidate the corresponding report cache entry the next time any recording is stopped.

maxcao13: So if I am understanding it correctly, the `ActiveRecordingReportCache` should implement `NotificationListener` since the class is trying to listen for the `ActiveRecordingStopped` notification, and implement its methods. The categories would be the Set of just that 1 notification category and in the overridden`onMessage`, it would check the cache and maybe the key and recording name for the entry and invalidate it if its there. 

Few questions.

1. What is wrong with publishing in the Notification class like I was doing? Is it because we only care about 1 set of notification categories right now so publishing every other notification would be a waste of resources? 
2. How do I see where the cache class is initialized? I thought the NotificationPublisher would be able to allow each listener implementation class to subscribe to its categories instead of doing that.
3. When I do 
```
public class NotificationPublisher {
    Vertx vertx;
    Set<NotificationListener> listeners;
    
    NotificationPublisher(Vertx vertx, Set<NotificationListener> listeners) {
        listeners.forEach(listener -> {
            listener.categories().forEach(category -> {
                vertx.eventBus().consumer(category).handler(message -> listener.onMessage(category, message));
            });
        });
    }

    <T> void send(String category, T message) {
        vertx.eventBus().publish(category, message);
    }
}
```
Java gives warnings and doesn't like the lack of parameterizations, but I'm not sure how to solve that other than to just put <?> in there. Each listener I assume would use their own generic `T` as a message but the actual 
```
.handler(message -> listener.onMessage(category, message));
            });
```
message callback object is of type Message<Object> so that doesn't help.

maxcao13:
Not sure how to bind the `ActiveRecordingReportCache` into the set of listeners in the `NotificationPublisher` without making it public.

Also may be better ways to parse the notification message but I don't think using POJOs here is smart nor sending them preparsed for generality of the onMessage method. 

Ignoring tests, thoughts on the structure?

src/main/java/io/cryostat/messaging/notifications/NotificationListener.java
```diff
@@ -0,0 +1,47 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package io.cryostat.messaging.notifications;
+
+import java.util.Set;
+
+import io.vertx.core.eventbus.Message;
+
+public interface NotificationListener {
```
andrewazores, line 44: What if this was like:

```java
public interface NotificationListener<T> {
  void onMessage(String category, Message<T> message);
}
```

andrewazores: This has the problem that it will enforce that all of the messages have the same type across all of the categories in the `Set`, though.

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -177,4 +189,32 @@ protected String getReport(RecordingDescriptor recordingDescriptor, String filte
...
+    @Override
+    public Set<String> categories() {
+        return Set.of(STOP_NOTIFICATION_CATEGORY);
+    }
+
+    @Override
+    public void onMessage(String category, Message<?> message) {
+        Map<String, Object> notification = gson.fromJson(message.body().toString(), Map.class);
+        Map<String, Object> msg = objectMapper.convertValue(notification.get("message"), Map.class);
+        Map<String, Object> recording = objectMapper.convertValue(msg.get("recording"), Map.class);
```
andrewazores, line 202: This is unexpected. Why does it need both the Jackson ObjectMapper and Gson? Both of those fill the same role of (de)serializing POJO to/from JSON.

andrewazores:
> Not sure how to bind the ActiveRecordingReportCache into the set of listeners in the NotificationPublisher without making it public.

Not sure exactly what you mean here.

https://github.com/cryostatio/cryostat/blob/4c45fbc451e2c41deadd09679e8bbc98d5f4849f/src/main/java/io/cryostat/net/web/http/api/v2/HttpApiV2Module.java#L94

https://github.com/cryostatio/cryostat/blob/4c45fbc451e2c41deadd09679e8bbc98d5f4849f/src/main/java/io/cryostat/net/web/http/api/v2/ApiGetHandler.java#L62

Does this example help answer the question at all?

> Also may be better ways to parse the Notification message but I don't think using POJOs here is smart

Why?

src/main/java/io/cryostat/messaging/notifications/NotificationListener.java
maxcao13: Yeah, and then I get problems trying to cast the `Message<Object>` message callback object into the `Message<T>` `onMessage` arg. I think maybe this problem could be solved if we used [MessageCodec](https://vertx.io/docs/apidocs/io/vertx/core/eventbus/MessageCodec.html)? I think then we could just use it to register codecs for each listener `onMessage` `T` type and it won't complain.

andrewazores: Sure, that looks like the right solution if we need to send generalized data types as the message.

> Usually the event bus only allows a certain set of message types to be sent across the event bus, including primitive types, boxed primitive types, byte[], [JsonObject](https://vertx.io/docs/apidocs/io/vertx/core/json/JsonObject.html), [JsonArray](https://vertx.io/docs/apidocs/io/vertx/core/json/JsonArray.html), [Buffer](https://vertx.io/docs/apidocs/io/vertx/core/buffer/Buffer.html).

That sounds like it might already cover our needs without having to implement a MessageCodec, though. Our notification messages generally just contain a simple serialized `Map`, which is more or less the same thing as a `JsonObject`, so perhaps we should just orient the interface toward that instead. ie rather than being generic, require the message type to be `JsonObject`.

andrewazores: Also, I would prefer to not expose the `Message` class from Vertx through the `NotificationListener` interface. Part of the design goal of this interface is to abstract the Vertx EventBus away, so listeners and publishers shouldn't know anything about Vertx, the EventBus, or the Message type.

The EventBus and Message have a lot more capability after all, and in this use case we're just trying to use one narrow facet.

```java
public interface NotificationListener<T> {
  void onMessage(String category, T message);
}
```

andrewazores: In the publisher, it would look something like:

```java
        listeners.forEach(listener -> {
            listener.categories().forEach(category -> {
                vertx.eventBus().consumer(category, message -> listener.onMessage(category, message.body()));
            });
        });
```

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
maxcao13: > > Not sure how to bind the ActiveRecordingReportCache into the set of listeners in the NotificationPublisher without making it public.
> 
> Not sure exactly what you mean here.
> 
> https://github.com/cryostatio/cryostat/blob/4c45fbc451e2c41deadd09679e8bbc98d5f4849f/src/main/java/io/cryostat/net/web/http/api/v2/HttpApiV2Module.java#L94
> 
> https://github.com/cryostatio/cryostat/blob/4c45fbc451e2c41deadd09679e8bbc98d5f4849f/src/main/java/io/cryostat/net/web/http/api/v2/ApiGetHandler.java#L62
> 
> Does this example help answer the question at all?
> 

Sorry, I meant that the `NotificationsModule` is in `java.io.cryostat.messaging.notifications.NotificationsModule` and the report cache is in `java..io.cryostat.net.reports.ActiveRecordingReportCache` so I think the `NotificationsModule` can't see the cache as a class since it's in a different package and that's what I meant. 

> > Also may be better ways to parse the Notification message but I don't think using POJOs here is smart
> 
> Why?

Uh, I think I was thinking like generally, since Notifications have different structures, it may be difficult to extend the parsing to all kinds, but I realize that Notifications are all generally the same structure of `Meta` and `Message` so it probably isn't that hard to parse.

maxcao13:
> > Not sure how to bind the ActiveRecordingReportCache into the set of listeners in the NotificationPublisher without making it public.
> 
> Not sure exactly what you mean here.
> 
> https://github.com/cryostatio/cryostat/blob/4c45fbc451e2c41deadd09679e8bbc98d5f4849f/src/main/java/io/cryostat/net/web/http/api/v2/HttpApiV2Module.java#L94
> 
> https://github.com/cryostatio/cryostat/blob/4c45fbc451e2c41deadd09679e8bbc98d5f4849f/src/main/java/io/cryostat/net/web/http/api/v2/ApiGetHandler.java#L62
> 
> Does this example help answer the question at all?
> 

Sorry, I meant that the `NotificationsModule` is in `java.io.cryostat.messaging.notifications.NotificationsModule` and the report cache is in `java..io.cryostat.net.reports.ActiveRecordingReportCache` so I think the `NotificationsModule` can't see the cache as a class since it's in a different package and that's what I meant. 

> > Also may be better ways to parse the Notification message but I don't think using POJOs here is smart
> 
> Why?

Uh, I think I was thinking like generally, since Notifications have different structures, it may be difficult to extend the parsing to all kinds, but I realize that Notifications are all generally the same structure of `Meta` and `Message` so it probably isn't that hard to parse.

andrewazores:
> Sorry, I meant that the NotificationsModule is in java.io.cryostat.messaging.notifications.NotificationsModule and the report cache is in java..io.cryostat.net.reports.ActiveRecordingReportCache so I think the NotificationsModule can't see the cache as a class since it's in a different package and that's what I meant.

I see. Though, why does the `NotificationsModule` need visiblity of the `ActiveRecordingReportCache`? The `ActiveRecordingReportCache` is provided by the `ReportsModule`, so the `ReportsModule` can do the `@Binds @IntoSet` to bind the cache as a notification listener. This way the `NotificationListener` interface just needs to be publicly visible so that the `ReportsModule` and cache class can see it. Dagger will take care of generating the `Set<NotificationListener>` for injection for whatever callers want it.

> Uh, I think I was thinking like generally, since Notifications have different structures, it may be difficult to extend the parsing to all kinds, but I realize that Notifications are all generally the same structure of Meta and Message so it probably isn't that hard to parse.

Makes sense. Yea, I think your insight that our notifications all have a consistent form of Meta and Message, where both of those are also generally just flat map structures, is the key.

The `MessagingServer` and `WsMessage` classes predate the current unified `Notification` system that we have, which is why the server and `WsMessage` are so open-ended and generic. In practice the only kind of messages we ever send anymore are `Notification`s that come through the `NotificationFactory`, so perhaps the `WsMessage` can be removed and simply replaced with `Notification`. `WsMessage` is an empty definition anyway and the only class that extends it is `Notification`, so this typing simplification should be pretty straightforward to perform, and then provides for better type safety elsewhere since you can start relying on messages to be of type `Notification`. And `Notification` already does look a lot like that `NotificationListener<T>`, `onMessage(String category, T message)` we talked about earlier.

src/main/java/io/cryostat/messaging/notifications/NotificationListener.java
maxcao13: I'm not sure how to get past these compile time errors 

If I do this 
```
public interface NotificationListener<T> {
  void onMessage(String category, T message);
}
...
...
...

    private final Vertx vertx;
    private final Set<NotificationListener<?>> listeners;

listeners.forEach(listener -> {
            listener.categories().forEach(category -> {
                vertx.eventBus().consumer(category, message -> listener.onMessage(category, message.body()));
            });
        });
```
It won't like it since `message.body()` is of type Object and onMessage expects type `<T>`, but since the `Set` is a set of `NotificationListener<?>`, it won't accept any casting at runtime. 

So instead should `onMessage` be like 
`void onMessage(String category, JsonObject msg);`
And we parse it in the receiver?

```
    NotificationPublisher(Vertx vertx, Set<NotificationListener<?>> listeners) {
        this.vertx = vertx;
        this.listeners = listeners;

        listeners.forEach(listener -> {
            listener.categories().forEach(category -> {
                vertx.eventBus().consumer(category, message -> {
                    JsonObject msg = JsonObject.mapFrom(message.body());
                    listener.onMessage(category, msg);
                });
            });
        });
    }
```

andrewazores: That is an option but it's really unfortunate if we have to serialize and deserialize JSON for this. I would probably rather say forget the Vertx EventBus in that case and just implement a plain listener/callback pattern in plain Java, in memory.

It's a common pattern, and we have an example already here:

https://github.com/cryostatio/cryostat/blob/5fae352f0fd66454bab564844be5943c83fba7c0/src/main/java/io/cryostat/platform/PlatformClient.java#L51

https://github.com/cryostatio/cryostat/blob/5fae352f0fd66454bab564844be5943c83fba7c0/src/main/java/io/cryostat/platform/internal/AbstractPlatformClient.java#L67

This is for the `TargetDiscoveryEvent`, but it should generalize to `Notification<T>` fairly easily. The `Set<String> categories()` part would still be a problem, so maybe we just drop that concept for now.

andrewazores: Thanks. Sorry for wasting your effort on the EventBus - hopefully you at least learned some new patterns or techniques through that experience.

maxcao13:
Is this sort of what it should look like? Sorry, I'm sort of having trouble adapting this to the example you linked, mostly because of the `Notification<T>`.

maxcao13:
Before tests and removing prints, thoughts?

src/main/java/io/cryostat/MainModule.java
```diff
@@ -149,6 +151,13 @@ JMXServiceURL.class, new GsonJmxServiceUrlAdapter(logger))
         return builder.create();
     }
 
+    @Provides
+    @Singleton
+    public static ObjectMapper provideObjectMapper() {
+        return new ObjectMapper()
```
andrewazores, line 157: I'm still not clear on why we need Jackson's `ObjectMapper`. What is this doing that the `gson` instance we're already using elsewhere doesn't/can't?

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -52,34 +53,46 @@
...
             FileSystem fs,
             TargetConnectionManager targetConnectionManager,
+            NotificationSource notificationSource,
             @Named(ReportsModule.REPORT_GENERATION_TIMEOUT_SECONDS) long generationTimeoutSeconds,
+            ObjectMapper oMapper,
             Logger logger) {
         this.reportGeneratorServiceProvider = reportGeneratorServiceProvider;
         this.fs = fs;
         this.targetConnectionManager = targetConnectionManager;
+        this.notificationSource = notificationSource;
```
andrewazores, line 92: This field looks like it's unused. I think it would make sense for the cache to register itself as a listener here in the constructor, rather than performing that registration within the `@Provides` method in the module.

src/main/java/io/cryostat/MainModule.java
maxcao13: I don't think gson is able to parse Java Maps (unless I'm wrong), which are the underlying type of the Notifications (or at least the one we care about) into POJOs. 
```
    public void writeMessage(Notification message) {
        notificationSource.notifyListeners(message);
        String json = gson.toJson(message);
         ...
    }
```
That being said, I could change the type of 
`notifyListeners` to take in json `String` instead of `Notification` and change `writeMessage`

to
```
        String json = gson.toJson(message);
        notificationSource.notifyListeners(json);
```
but that sort of gets rid of the generic `T` if I just pass json String.

But then we use gson to parse that into the POJOs if that's better.

andrewazores: I'm quite sure Gson handles Maps.

`Map<X, Y> map = gson.fromJson(jsonString, new TypeToken<Map<X, Y> {}.getType())`

should do it.

But why is this needed exactly? You're performing a deserialization of a Notification's `.message()` content from JSON into a Java Map. How/where/when did the Notification's `Map` message become JSON to begin with?

Can we just avoid making it JSON to begin with here? It should only need to be made into JSON when we're going to write it out to the WebSocket connection and broadcast it over the network. This `NotificationListener` system is just passing object references within our same Java process so JSON serialization should not be needed or used.

maxcao13: That makes sense, sorry about that.

maxcao13: Wait, sorry it's a lot of questions but, 

> 
> to
> 
> ```
>         String json = gson.toJson(message);
>         notificationSource.notifyListeners(json);
> ```
this String json was already in the code being passed through the socket, I wasn't using it. I was just saying that to use `Gson` to parse the `Notification<T>`, I have to somehow use something to turn the `T` into something readable and that's where the jackson `ObjectMapper` came in, which doesn't have the extra step of converting the `T` into a String first, and then converting that String back into an `ActiveRecordingStopMatcher` object. It just converts the `T`, which is implicitly assumed to be a Java `Map<String, Object>>, into that resulting object.

Sorry, I realize that my wording was confusing, I wasn't actually using json at all.

maxcao13: Oh, I'm not really using Jackson to do unchecked type casting, I'm just using it to easily convert the Map `notification.getMessage()` into something more readable which is the object
```
    private static class ActiveRecordingStopMatcher {
        private String target;
        private String name;
...
```
because otherwise, if I didn't use some sort of thing like Jackson, in 
```
    @Override
    public void callback(Notification<Map<String, Object>> notification) {
        String category = notification.getCategory();
```
I have to do 
`String targetId = notification.getMessage().get("targetId");`
`Object recording = notification.getMessage().get("recording");`
`String recordingName = ((Map<String, String>)recording).get("name"); \\Some way to cast Object into Map anyways
`invalidateCacheEntry(new ConnectionDescriptor(targetId), recordingName);`
`etc...`

maxcao13: To me, it makes more sense to convert the `Map` into an object which is easily readable using Jackson annotations and `ObjectMapper`, and I don't think there's any other way around it since `Notification.getMessage()` in this case will always be of type `Map<String, Object>` otherwise, I have to do the manual Map field getting, like above.

andrewazores: > Notification.getMessage() in this case will always be of type Map<String, Object>

"this case" meaning for this cache's listener, right? Other listeners on other notification categories could be expecting different notification message formats.

I'm just not quite convinced it's worth tying this implementation to the Jackson dependency just to avoid the "manual Map field getting" and the explicit cast from Object to Map. That's pretty standard and conventional, and portable if/when we migrate to another framework that may not already include Jackson as a dependency.

maxcao13: Okay, sounds good, makes sense to me thanks for the feedback. ^^

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -52,41 +53,50 @@
...
 import io.cryostat.net.ConnectionDescriptor;
 import io.cryostat.net.TargetConnectionManager;
 
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.Scheduler;
 
-class ActiveRecordingReportCache {
+class ActiveRecordingReportCache implements NotificationListener<Map<String, Object>> {
+    private static final String STOP_NOTIFICATION_CATEGORY = "ActiveRecordingStopped";
```
andrewazores, line 68: This constant is already defined in the `RecordingTargetHelper` class, so this one should be removed and replaced with just references to that original definition.

src/main/java/io/cryostat/messaging/notifications/NotificationSource.java
```diff
@@ -0,0 +1,68 @@
...
+ * SOFTWARE.
+ */
+package io.cryostat.messaging.notifications;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+@SuppressWarnings("unchecked")
+public class NotificationSource {
```
andrewazores, line 45: Is there any way this can be refactored and merged with the `NotificationFactory`?

src/main/java/io/cryostat/messaging/notifications/NotificationListener.java
```diff
@@ -0,0 +1,42 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package io.cryostat.messaging.notifications;
+
+public interface NotificationListener<T> {
+    void callback(Notification<T> notification);
```
andrewazores, line 41: A more descriptive name would be nice, something like `onNotification` perhaps? It's also less likely to collide with method names from other interfaces that way.

src/main/java/io/cryostat/net/reports/ActiveRecordingReportCache.java
```diff
@@ -109,14 +119,7 @@ Future<String> get(
...
-        RecordingDescriptor key = new RecordingDescriptor(connectionDescriptor, recordingName);
-        boolean hasKey = cache.asMap().containsKey(key);
-        if (hasKey) {
-            logger.trace("Invalidated active report cache for {}", recordingName);
-            cache.invalidate(key);
-        } else {
-            logger.trace("No cache entry for {} to invalidate", recordingName);
-        }
+        boolean hasKey = invalidateCacheEntry(connectionDescriptor, recordingName);
         return hasKey;
```
andrewazores, line 123: It doesn't look like there is any meaningful difference between `delete()` and the new `invalidateCacheEntry()` methods, is there? The implementation was just moved from the first to the second, and all the first does now is call the second and return its return value.

andrewazores (commented):
Looking good, nice and clean now. Just a few more notes.

src/main/java/io/cryostat/messaging/notifications/NotificationSource.java
maxcao13: What kind of merging? From the way I look at it, the `NotificationFactory` is it's own factory to make `Notification.Builder`s and the `NotificationSource` is just a source point to add/remove listeners to and notify and classes that implement `NotificationListener`, and they seem sort of separate, so I'm unsure of the best way to cleaning refactor them into one class.

andrewazores: I was just thinking that if the definitions of the `NotificationSource` and `NotificationFactory` were combined, then `Notification#send()` could directly notify the listeners. It would avoid passing the Notification, cast to a `WsMessage`, into the `MessagingServer`, just for that to then check if the `WsMessage instanceof Notification` and notify the listeners if so. `MessagingServer` would retain its simple responsibility of emitting WebSocket messages to connected clients, and not also become responsible for passing events between components within the Cryostat process.

maxcao13: Is what I recently pushed sort of what you were thinking?

andrewazores: Sure, that looks like a good approach to achieve what I was describing before. Nice way to do it.

src/main/java/io/cryostat/messaging/MessagingServer.java
```diff
@@ -216,6 +221,9 @@ public void start() throws SocketException, UnknownHostException {
     }
 
     public void writeMessage(WsMessage message) {
+        if (message instanceof Notification) {
+            notificationSource.notifyListeners((Notification<?>) message);
```
andrewazores, line 225: This can be removed now since the `Notification#send()` call is already doing this, so listeners will actually end up getting the notification event twice.

andrewazores (commented):
Looks good. I have some further refactoring ideas in mind but I think I'll just play around with that myself after we get this merged, and see what I end up liking.

Issue #947, opened 2022-05-17
[Task] Unit tests for GraphQL


andrewazores:
Related to https://github.com/cryostatio/cryostat/issues/879

GraphQL handler implementations are severely lacking unit tests.

PR #1022, opened 2022-07-12
feat(graphql): granular permissions

### LINKED ISSUE #1021: [Task] Implement per-subquery permissions for GraphQL
Currently, GraphQL permissions checking is all handled up front by the main `GraphQLPostHandler`, which is the API entrypoint. This endpoint simply defines the expected set union of permissions for all of our implemented `Fetcher` and `Mutator` classes.

1. This is fragile, because adding a new `Mutator` might imply a new permission required for users executing this mutation. This requirement will not be captured by the `GraphQLPostHandler` and is very easy to miss during review.
2. This doesn't allow for granular user permissions, ex. read-only user roles, when interacting via the GraphQL API. Our standard HTTP API does implement granular permissions per endpoint, so it would make sense for GraphQL's fetchers and mutations to also have granular permissions in the same way.

Related to #947 
Related to #760 (granular permissions will be useful for namespace isolation)

andrewazores:
Fixes #1021

- chore(permissions): extract PermissionedAction interface
- chore(graphql): mark all fetchers/mutators with resource actions
- feat(graphql): implement permissions check for each graphql fetcher/mutator
- fix(graphql): ensure GraphQLContext is not null by copying DataFetchingEnvironment
- chore(graphql): apply spotless formatting

This implements granular permissions for each `fetcher`/`mutator` within the GraphQL endpoint implementation. This is achieved by removing the endpoint-level resource actions set union - by default, only a basic "token is valid" check will be performed - and delegating further permissions checks to each subsequent or nested `fetcher`/`mutator` that is executed afterward. In this implementation, each such subquery that is performed will result in a permissions check being performed, checking that the contextual user has the required permissions for the subquery's action. For complex GraphQL queries with deep nesting, this might add noticeable overall request latency since there is an authentication server request round-trip-time at each level of nested query. This might be something to optimize in the future by pre-computing the set union by recursing through the nested queries up front and performing a single permissions check all at once before actually executing the queries, but this needs either some deeper knowledge of GraphQL to pull off or a more complex implementation of `AbstractPermissionedDataFetcher` that encodes the nested query relationships.

Overall this ends up implementing something very similar to how our HTTP API permissions checks work with the `AbstractAuthenticatedRequestHandler`/`AbstractV2RequestHandler`, but on the `DataFetcher`s.

Unit tests are still overall sorely lacking - see #947. There is some basic integration test coverage, at least, which caught an error during implementation where the `DataFetchingEnvironment` passed to subqueries contained a `null` `GraphQLContext`.

### DIFF
--- a/src/main/java/io/cryostat/net/security/PermissionedAction.java
+++ b/src/main/java/io/cryostat/net/security/PermissionedAction.java
@@ -35,49 +35,10 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package io.cryostat.net.web.http;
+package io.cryostat.net.security;
 
-import io.cryostat.net.security.PermissionedAction;
-import io.cryostat.net.web.http.api.ApiVersion;
+import java.util.Set;
 
-import io.vertx.core.Handler;
-import io.vertx.core.http.HttpMethod;
-import io.vertx.ext.web.RoutingContext;
-
-public interface RequestHandler extends Handler<RoutingContext>, PermissionedAction {
-    /** Lower number == higher priority handler */
-    static final int DEFAULT_PRIORITY = 100;
-
-    static final String ALL_PATHS = "*";
-
-    default int getPriority() {
-        return DEFAULT_PRIORITY;
-    }
-
-    ApiVersion apiVersion();
-
-    default String basePath() {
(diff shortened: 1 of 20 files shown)


andrewazores:
@ebaron ping

ebaron:
> @ebaron ping

Hey Andrew, sorry I completely missed this PR. I'll take a look today.

ebaron (approved):
Looks good!

PR #1025, opened 2022-07-18
fix(jwt): use stored credentials when available

### LINKED ISSUE #1024: [Task] JWT creation still expects client to passthrough jmxauth claim
Since https://github.com/cryostatio/cryostat/pull/1000 , credentials are stored with a `matchExpression` and it is generally expected that the client must first create a JMX Credentials definition through https://github.com/cryostatio/cryostat/pull/1019 in order to interact with target JVMs that have JMX Auth enabled. Previously, clients would usually include JMX credentials along with API requests and Cryostat would pass those through to the target JVM at JMX connection time.

There are API endpoints that use JWT, primarily for the web-client for an improved frontend UX, which still expect the client to provide a `jmxauth` JWT claim at token creation time. This is then interpreted later by Cryostat as a set of JMX Auth Basic credentials to be used and passed through to the target JVM. However, the web-client no longer knows about JMX credentials and never sends those along with API requests, instead relying on the backend stored credentials to be automatically attached to requests as needed. This work was already done in #1000 / #1019 , but it was missed for the JWT implementation(s).

Clients should still be able to specify the `jmxauth` claim explicitly to force/override JMX credentials for passthrough, but if none are provided (and the client otherwise passes the requisite authz checks), Cryostat should look up any stored credentials and attach ones that match the target JVM.

This can be reproduced easily enough:
1. Run Cryostat as of https://github.com/cryostatio/cryostat/commit/882af72f7380a87b70a152616d684d75c8c77093 via `smoketest.sh`
2. Open web-client
3. Delete any stored credentials
4. Go to Events view
5. Select one of the target applications that requires JMX authentication
6. Enter JMX credentials. You should be able to see the list of TARGET-provided templates
7. Click the three-dots menu on one of the templates and try to Download it
8. The request will fail with `HTTP 427`

andrewazores:
Fixes #1024

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/AbstractJwtConsumingHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/AbstractJwtConsumingHandler.java
@@ -37,6 +37,7 @@
  */
 package io.cryostat.net.web.http.api.v2;
 
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.SocketException;
 import java.net.URI;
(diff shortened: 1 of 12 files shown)


jan-law (approved):
Looks good

Issue #937, opened 2022-05-13
[Story] Pluggable Discovery Database


andrewazores:
Parent: https://github.com/cryostatio/cryostat/issues/936

To serve the needs of parent 3, we should add some form of database container to the Cryostat deployment. This needs to be able to hold information about target applications as well as their parent nodes in the deployment graph - so in OpenShift, from the Namespace, to Deployment(Config), to Pod, to Replica, to Endpoint, or some similar variations.

Once we determine what type of database this will be (ex SQL vs NoSQL) we should solidify on a design that allows Cryostat to be configured to point to and connect to that database, including with optional credentials and authentication. The database container should be able to be one supplied in advance by the end user, or one set up by the Operator if the user does not provide their own.

- [x] https://github.com/cryostatio/cryostat/issues/941
- [x] https://github.com/cryostatio/cryostat/issues/942
- [x] https://github.com/cryostatio/cryostat/issues/943

andrewazores:
Closed by #1032 / #1028

PR #1026, opened 2022-07-18
feat(discovery): pluggable discovery services API

### LINKED ISSUE #939: [Story] Pluggable Discovery API
Parent: https://github.com/cryostatio/cryostat/issues/936

Internally we currently have the `MergingPlatformClient` that serves this goal, but only for the hardcoded and included implementations within Cryostat. We need to generify this and expose it as an authenticated API so that API clients can perform their own platform-specific discovery of environment and target nodes and then publish that information to Cryostat, where Cryostat will then store it in its internal database. This API should allow for Create, Read, Update, Delete semantics on the node resources. This will serve parent 1 and 5, and should address portions of 2 as well.

- [x] Implement Pluggable Discovery API (draft: #1026)
- [x] #1049
- [x] #1058
- [x] #1061

andrewazores:
Closes #939


andrewazores:
@jan-law @hareetd @tthvo @maxcao13 please browse through this and leave your comments.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -0,0 +1,234 @@
...
+
+            Set<TargetNode> previousLeaves = findLeavesFrom(previousTree);
+            Set<TargetNode> currentLeaves = findLeavesFrom(currentTree);
+
+            Set<TargetNode> added = new HashSet<>(currentLeaves);
+            added.removeAll(previousLeaves);
+
+            Set<TargetNode> removed = new HashSet<>(previousLeaves);
+            removed.removeAll(currentLeaves);
+
```
maxcao13, line 111: It looks like to me that the `currentLeaves` set will always be bigger than the `previousLeaves` set because the `currentTree`'s children are instantiated from the previous tree's children and the `children` parameter of `update` are always added to the `currentTree.subtree`.

andrewazores:
```
previous = [0, 1, 2]
current = [1, 2]

added = copy([1, 2])
added = added - [0, 1, 2] // [ ] empty set

removed = copy([0, 1, 2])
removed = removed - [1, 2] // [ 0 ]
```

```
previous = [0, 1]
current = [0, 1, 2]

added = copy([0, 1, 2])
added = added - [0, 1] // [ 2 ]

removed = copy([0, 1)
removed = removed - [0, 1, 2] // [ ] empty set
```

> because the currentTree's children are instantiated from the previous tree's children and the children parameter of update are always added to the currentTree.subtree.

`currentTree` is the updated tree given to us by the plugin service, but it is not "instantiated from the previous tree's children" - the previous subtree root's name and nodeType are copied, but the children are only the ones given by the update. The algorithm here is detecting what elements were added/removed and using that to emit LOST/FOUND discovery notifications.


```java
            PluginInfo updatedInfo = new PluginInfo(map.get(id), children);
```

I think this is the potentially confusing line, since this looks like a plain copy constructor. The first argument is the `PluginInfo` in storage from the last update (or from registration time), from which the name and nodeType are copied. The second argument is the set of children from the current update.

src/main/java/io/cryostat/net/web/http/api/v2/AbstractJwtConsumingHandler.java
```diff
@@ -164,8 +173,12 @@ private JWT validateJwt(RoutingContext ctx)
     protected ConnectionDescriptor getConnectionDescriptorFromJwt(RoutingContext ctx, JWT jwt)
             throws ParseException {
         String targetId = ctx.pathParam("targetId");
-        // TODO inject the CredentialsManager here to check for stored credentials
         Credentials credentials = null;
+        try {
+            credentials = credentialsManager.getCredentialsByTargetId(targetId);
+        } catch (ScriptException | IOException e) {
+            logger.error(e);
+        }
```
jan-law, line 181: Is this code from a rebase on top of https://github.com/cryostatio/cryostat/pull/1025 ?

src/main/java/io/cryostat/net/web/http/api/v2/DiscoveryPostHandler.java
```diff
@@ -0,0 +1,128 @@
...
+            throws Exception {
+        String body = params.getBody();
+        Set<AbstractNode> nodes =
+                gson.fromJson(body, new TypeToken<Set<AbstractNode>>() {}.getType());
+        // TODO validate the nodes more thoroughly, all branches should terminate in leaves, no
+        // fields should be null, etc.
+        Set<AbstractNode> previous =
+                storage.update(this.uuidFromString.apply(params.getPathParams().get("id")), nodes);
+
+        return new IntermediateResponse<Set<AbstractNode>>().body(previous);
```
jan-law, line 126: Why is the updated subtree returned in the response called `previous`?

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
maxcao13: Yeah, I was getting confused at the constructor part
```
            this.subtree =
                    new EnvironmentNode(original.subtree.getName(), original.subtree.getNodeType());
            subtree.addChildren(children);
```
I thought `subtree.addChildren()` was adding to existing children somehow, but I understand now thanks!

src/main/java/io/cryostat/net/web/http/api/v2/DiscoveryDeregistrationHandler.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public HttpMethod httpMethod() {
+        return HttpMethod.DELETE;
+    }
+
+    @Override
+    public String path() {
+        return basePath() + "discovery/:d";
```
hareetd, line 92: Should be `"discovery/:id"`

src/main/java/io/cryostat/net/web/http/api/v2/HttpApiV2Module.java
```diff
@@ -182,6 +182,28 @@ abstract RequestHandler bindTargetCredentialsDeleteHandler(
...
+    abstract RequestHandler bindDiscoveryRegistrationHandler(DiscoveryRegistrationHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindDiscoveryRegistrationBodyHandler(
+            DiscoveryRegistrationBodyHandler handler);
+
+    @Binds
+    @IntoSet
+    abstract RequestHandler bindDiscoveryDeregistrationBodyHandler(
```
hareetd, line 196: Should be `bindDiscoveryDeregistrationHandler`

hareetd (commented):
The code changes make sense. 

Besides the file @jan-law already commented on, some of the other files such as the `RecordingGetHandler` also seem to contain changes from #1025.

src/main/java/io/cryostat/net/web/http/api/v2/DiscoveryPostHandler.java
andrewazores: Yea, not sure honestly. It's probably just not worth sending anything back to the client (plugin( here - I don't imagine it would have any use for it, so it's best to just save the bandwidth. I'll take it out.

src/main/java/io/cryostat/net/web/http/api/v2/AbstractJwtConsumingHandler.java
andrewazores: Yes, looks like it. I'll try to clean up the history.

andrewazores: Done, I rebased the `939-story-pluggable-discovery-api` upstream branch so it's up-to-date with upstream main, and then rebase this feature branch of mine onto the `939-` branch. History should be clean now and the diff shouldn't contain any chunks from #1025 or other unrelated changes.

andrewazores:
In the last commit as of now (`add quarkus-test sample...`) I update `smoketest.sh` to include a new sample application deployment, built from https://github.com/andrewazores/quarkus-test/commit/5c05ec33eb67aefffd65b4bc88df2c550dd99fe7 . This is the same tiny, does-almost-nothing Quarkus-based application as we already have, except this one implements the plugin side of the API. On startup it repeatedly tries to register itself with the configured Cryostat instance, and once it succeeds it updates Cryostat with a singleton set of nodes just containing a reference to itself. Running this updated `smoketest.sh` should result in a `quarkus-test-1` target appearing with JMX port 9097, and you should be able to connect and interact with that target in all the usual ways. It still has JMX enabled as per usual but it is only discoverable via the pluggable discovery API - it does not have JDP enabled.

andrewazores:
I'm marking this as ready for review. In seems to be in a workable state, I believe it matches the API design doc, and we have an example of a basic plugin implemented and communicating over the new API to advertise itself. There is no support for the callback URL flow yet (on the server or in the example plugin) because that doesn't make any sense until we are persisting the discovery trees, which will come in #937.

smoketest.sh
```diff
@@ -47,6 +47,22 @@ function runDemoApps() {
...
+    if [ -z "$CRYOSTAT_DISABLE_SSL" ]; then
+        local protocol="https"
+    else
+        local protocol="http"
+    fi
+    podman run \
+        --name quarkus-test-plugin \
+        --pod cryostat-pod \
+        --env org.acme.CryostatService/mp-rest/url="${protocol}://localhost:${webPort}" \
+        --rm -d quarkus-test:0.0.3
```
hareetd, line 66: The image link is incomplete, it should be `quay.io/andrewazores/quarkus-test:0.0.3`

hareetd (commented):
Tested it out using `smoketest` and it's working well. Here are my logs showing the relevent discovery API requests:

```
Jul 27, 2022 6:43:13 PM io.cryostat.core.log.Logger info
INFO: (127.0.0.1:41818): POST /api/v2.2/discovery 200 140ms
Jul 27, 2022 6:43:13 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"TargetJvmDiscovery","type":{"type":"application","subType":"json"},"serverTime":1658947393},"message":{"event":{"kind":"FOUND","serviceRef":{"connectUrl":"service:jmx:rmi:///jndi/rmi://localhost:9097/jmxrmi","alias":"code-with-quarkus","labels":{},"annotations":{"platform":{},"cryostat":{}}}}}}
Jul 27, 2022 6:43:13 PM io.cryostat.core.log.Logger info
INFO: (127.0.0.1:41818): POST /api/v2.2/discovery/027b64d1-eab3-4c7e-8de0-73b0d4df0fe2 200 5ms
```
Awesome work!

src/main/java/io/cryostat/discovery/AbstractNodeTypeAdapter.java
```diff
@@ -0,0 +1,256 @@
...
+import io.cryostat.platform.discovery.NodeType;
+import io.cryostat.platform.discovery.TargetNode;
+import io.cryostat.util.PluggableTypeAdapter;
+
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+import dagger.Lazy;
+
+public class AbstractNodeTypeAdapter extends PluggableTypeAdapter<AbstractNode> {
```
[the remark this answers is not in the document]
andrewazores: Ah, it's for this line basically:

https://github.com/andrewazores/cryostat/blob/555f38edbf06a71aae57c3f1915f02f13fc49590/src/main/java/io/cryostat/net/web/http/api/v2/DiscoveryPostHandler.java#L121

Without this type adapter, gson doesn't know how to deserialize the JSON into an `AbstractNode` - because it's Abstract, so it could be either an `EnvironmentNode` or a `TargetNode`, or possibly something else. Gson just doesn't have enough information about the types to figure this out automatically.

It probably will end up being used for serializing subtrees into JSON through gson to stick them into the database though, you're right on there.

smoketest.sh
andrewazores: Ah yes, thanks.

src/main/java/io/cryostat/discovery/AbstractNodeTypeAdapter.java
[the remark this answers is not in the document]
andrewazores: > It probably will end up being used for serializing subtrees into JSON through gson to stick them into the database though, you're right on there.

Well - Gson already does know how to do the serialization side for us. It does that in the existing `GET /api/v2.1/discovery` handler, for example. It's easier for Gson to serialize structures like this because it just uses reflection to read the names and types of fields, and then creates the JSON structure that way. So the fact that the `EnvironmentNode` or `TargetNode` it finds is an extension of an abstract class doesn't really matter.

There's still a benefit to having the serialization implemented by this adapter, which is that it doesn't use reflection, and so it should be higher performance than Gson's default serialization strategy. For large subtrees, or deployments with many discovery plugins and many subtrees, this probably results in better API performance.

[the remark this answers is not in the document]
hareetd: Ah. makes sense. Appreciate the explanation.

andrewazores:
> Tested it out using `smoketest` and it's working well. Here are my logs showing the relevent discovery API requests:
> 
> ```
> Jul 27, 2022 6:43:13 PM io.cryostat.core.log.Logger info
> INFO: (127.0.0.1:41818): POST /api/v2.2/discovery 200 140ms
> Jul 27, 2022 6:43:13 PM io.cryostat.core.log.Logger info
> INFO: Outgoing WS message: {"meta":{"category":"TargetJvmDiscovery","type":{"type":"application","subType":"json"},"serverTime":1658947393},"message":{"event":{"kind":"FOUND","serviceRef":{"connectUrl":"service:jmx:rmi:///jndi/rmi://localhost:9097/jmxrmi","alias":"code-with-quarkus","labels":{},"annotations":{"platform":{},"cryostat":{}}}}}}
> Jul 27, 2022 6:43:13 PM io.cryostat.core.log.Logger info
> INFO: (127.0.0.1:41818): POST /api/v2.2/discovery/027b64d1-eab3-4c7e-8de0-73b0d4df0fe2 200 5ms
> ```
> 
> Awesome work!

Try `podman stop` on the new plugin container, too.

I haven't tested it out in OpenShift but it should do the same thing.

maxcao13:
I'm having trouble getting past authentication on `smoketest.sh` with this branch vs. `main` where it's fine. Does any of my config need to be changed?

I get this even though that file does exist with a `user:sha64password`
```
Jul 27, 2022 8:14:12 PM io.cryostat.core.log.Logger warn
WARNING: HTTP 401: HTTP Authorization Failure
io.vertx.ext.web.handler.HttpException: Unauthorized
Jul 27, 2022 8:14:12 PM io.cryostat.core.log.Logger info
INFO: (127.0.0.1:37484): POST /api/v2.2/discovery 401 1ms
Jul 27, 2022 8:14:17 PM io.cryostat.core.log.Logger warn
WARNING: User properties file "/opt/cryostat.d/conf.d/cryostat-users.properties" does not exist
```

andrewazores:
In https://github.com/cryostatio/cryostat/pull/1026/commits/555f38edbf06a71aae57c3f1915f02f13fc49590 I enabled the `BasicAuthManager` by default, since it makes sense that the user as well as API clients should be authenticating by default in our testing setup, since that more closely resembles a real deployment.

The file should actually exist at ex. `$HOME/workspace/cryostat/conf/cryostat-users.properties` when you're using `smoketest.sh`. That directory gets mounted into the container at the `/opt/cryostat.d/conf.d` path.

```bash
$ cat conf/cryostat-users.properties 
user:d74ff0ee8da3b9806b18c877dbf29bbde50b5bd8e4dad7a3a725000feb82e8f1
```

That's `user:pass` credentials. Used like:

```bash
$ https :8181/api/v1/targets
HTTP/1.1 401 Unauthorized
X-WWW-Authenticate: Basic
content-encoding: gzip
content-length: 83
content-type: text/plain

HTTP Authorization Failure caused by HttpException: Unauthorized

$ https :8181/api/v1/targets Authorization:"Basic $(echo -n user:pass | base64)"
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 304

[ ... ]
```

maxcao13:
w/ podman stop
```
Jul 27, 2022 8:41:23 PM io.cryostat.core.log.Logger info
INFO: Outgoing WS message: {"meta":{"category":"TargetJvmDiscovery","type":{"type":"application","subType":"json"},"serverTime":1658954483},"message":{"event":{"kind":"LOST","serviceRef":{"connectUrl":"service:jmx:rmi:///jndi/rmi://localhost:9097/jmxrmi","alias":"code-with-quarkus","labels":{},"annotations":{"platform":{},"cryostat":{}}}}}}
Jul 27, 2022 8:41:23 PM io.cryostat.core.log.Logger info
INFO: (127.0.0.1:41772): DELETE /api/v2.2/discovery/6947d634-15ff-4499-9d85-0116cda83386 200 4ms
```

Really great job!

hareetd:
`podman stop` works for me as well

andrewazores:
I'm going to go ahead and merge this into the upstream feature branch (https://github.com/cryostatio/cryostat/tree/939-story-pluggable-discovery-api). I'll then open a new PR for that branch into `main`. Further work in this area will continue with PRs against the upstream feature branch.

hareetd (approved):

jan-law (approved):

maxcao13 (approved):

smoketest.sh
```diff
@@ -47,6 +48,23 @@ function runDemoApps() {
         --pod cryostat-pod \
         --rm -d quay.io/andrewazores/quarkus-test:0.0.2
 
+    if [ -z "$CRYOSTAT_WEB_PORT" ]; then
+        local webPort="$(xpath -q -e 'project/properties/cryostat.itest.webPort/text()' pom.xml)"
+    else
+        local webPort="8181"
```
tthvo, line 54: Not sure but shouldn't `webPort` be set to `$CRYOSTAT_WEB_PORT`?

PR #1027, opened 2022-07-27
chore(notifications): refactor for simplicity

andrewazores:
Related to #1020
Related to #921

- chore(notifications): bind listeners at build time
- chore(messaging): remove unused WsMessage class
- chore(messaging): ws server receives messages via internal notification listener


maxcao13 (approved):
It looks great, no complaints and it all makes sense.

PR #1028, opened 2022-07-28
feat(discovery): pluggable discovery services API

### LINKED ISSUE #938: [Task] Pluggable Discovery In-Place Model Updates
Depends on https://github.com/cryostatio/cryostat/issues/937
Parent: https://github.com/cryostatio/cryostat/issues/936

To serve parent 4 and 6, once we have a database of environment and target nodes, we should use the platform capabilities to listen for changes to these nodes and use those notifications to dynamically update the database contents. For example, in OpenShift, if one of our Endpoints objects disappears, then it should be removed from the database. The parent Pod that contained the Endpoints should be checked, and if it has no other child Endpoints, the Pod should also be removed. Then the Pod's parent Deployment(Config) should be checked and removed if it has no other Pod children.

### LINKED ISSUE #939: [Story] Pluggable Discovery API

andrewazores:
Closes #939
Closes #938 (the subtree update model and built-in discovery mechanisms already work this way)

Includes:
- [x] #1026
- [x] #1029
- [x] #1032


andrewazores:
Ready for review. This is just merging this long-lived upstream feature branch into upstream main. The component PRs of the feature have already been reviewed separately and merged into this branch.

maxcao13 (approved):
Everything looks good from my view testing with smoketest. I have nothing to add, good work!

PR #1029, opened 2022-07-28
feat(discovery): persistence and callback URL handling

andrewazores:
Related to #1026
Related to #938
Related to #939

- feat(discovery): prepare to implement URL callback handling
- don't persist or restore plugins without callbacks
- POST plugin callback URLs on startup to check presence
- run discovery plugin with restart policy

### DIFF
--- a/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
@@ -37,7 +37,11 @@
  */
 package io.cryostat.discovery;
 
+import java.io.BufferedReader;
+import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -50,11 +54,14 @@ import javax.inject.Provider;
 
 import io.cryostat.core.log.Logger;
 import io.cryostat.core.net.discovery.JvmDiscoveryClient.EventKind;
+import io.cryostat.core.sys.FileSystem;
 import io.cryostat.platform.ServiceRef;
 import io.cryostat.platform.discovery.AbstractNode;
 import io.cryostat.platform.discovery.BaseNodeType;
 import io.cryostat.platform.discovery.EnvironmentNode;
 import io.cryostat.platform.discovery.TargetNode;
+
+import com.google.gson.Gson;
 
 @Deprecated
 /**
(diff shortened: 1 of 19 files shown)


maxcao13 (commented):
Running `smoketest.sh` and sending a ctrl+c terminate anytime after creating these containers in the pod waits 10 seconds, and then podman resorts to sending a SIGKILL signal. Is this intended as a function of using ctrl+c? 

```
^C+ cleanup
+ podman pod stop cryostat-pod
WARN[0010] StopSignal SIGTERM failed to stop container vertx-fib-demo-3 in 10 seconds, resorting to SIGKILL 
WARN[0010] StopSignal SIGTERM failed to stop container vertx-fib-demo-2 in 10 seconds, resorting to SIGKILL 
WARN[0010] StopSignal SIGTERM failed to stop container vertx-fib-demo-1 in 10 seconds, resorting to SIGKILL 
121f62e55c1e02410e48c12c06d38d8665314c5d874cae84864b6645dc040c22
+ podman pod rm cryostat-pod
121f62e55c1e02410e48c12c06d38d8665314c5d874cae84864b6645dc040c22
```

andrewazores:
That's because I changed the exit trap function from sending a `podman kill` to `podman stop`. That allows the containers to receive a `SIGTERM` and try to gracefully shut themselves down, and if they haven't done so within the default 10 second grace period then they are `SIGKILL`ed. Unfortunately that test container doesn't gracefully shut down on `SIGTERM` right now, but I can go update it so that it does.

maxcao13 (approved):
Looks great, no problems from what I can see.

tthvo:
@andrewazores I ran into Auth issue as @maxcao13 when running smoketest. My `cryostat-users.properties` seems not be generated. It seems this file was generated in `devserver.sh` but not in `smoketest.sh`.

Original comments: https://github.com/cryostatio/cryostat/pull/1026#issuecomment-1197319136

andrewazores:
I guess that file might be worth automatically generating in the `run.sh` or `smoketest.sh` now that the BasicAuthManager is in the standard setup.

hareetd (approved):
Works well and the code changes make sense.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -65,18 +75,100 @@ public class DiscoveryStorage extends AbstractPlatformClientVerticle {
...
+    }
+
+    private void removePlugin(UUID uuid, URI uri) {
+        map.remove(uuid);
+        try {
+            fs.deleteIfExists(persistencePath.resolve(uuid.toString()));
+        } catch (IOException ioe) {
+            logger.error(ioe);
+        }
+        logger.info("Stale discovery service {} removed", uri);
```
tthvo, line 148: Should this log should be called on error tho?

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -65,18 +75,100 @@ public class DiscoveryStorage extends AbstractPlatformClientVerticle {
...
+    }
+
+    @Override
+    public void stop() {
```
tthvo, line 152: I wonder in the case of immediate kill like `SIGKILL`, does this method also get called to clean up or will we lose all discovery tree? Not sure I word that right :D

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
andrewazores: I think so - it's still removed from the in-memory map even if the file I/O operation fails.

This is changing a lot in #1031 anyway since there are no more persistence files and in-memory map there, just a proper database connection.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
andrewazores: Nope, if the OS `SIGKILL`s you then by definition you don't get to do any clean up, and anything that was only in-memory is lost.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
tthvo: oh yeh sure makes sense :D

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
tthvo: Are we planning to do a backup by some intervals? Or this loss is okay?

andrewazores: In #1031 with the real database, every update sent to us is handled by a database transaction that persists changes as they happen. No intervals needed, and losses are about as minimized as they can be.

tthvo: Oh okayy understood! I guess I will check out PR with the real database.

tthvo (approved):

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
andrewazores: This file-based persistence PR is really just an implementation stub stepping toward #1031 anyway - the file backing is there just so I could implement and test the storage startup behaviour where it checks for already-known callback URLs and tests them.

We have other places that also use file-based storage and update that per-request too, like the JMX Credentials storage, so no interval-based strategy is needed either.

andrewazores: Well, don't get too far ahead on #1031. That one is still in progress. This one is meant to be fleshing out the client-facing API and internal implementation for the behaviours, but the file-based storage is a temporary measure as a stand-in for a database.

PR #1031, opened 2022-08-03
feat(discovery): use hibernate, h2database, postgresql

### LINKED ISSUE #942: [Task] Add Services Discovery Database
Parent: https://github.com/cryostatio/cryostat/issues/937
Depends on https://github.com/cryostatio/cryostat/issues/941

Once we have determined what kind of database to use, we should add configuration to Cryostat to allow it to connect to such a database instance. This probably consists of adding environment variables to provide Cryostat with the database URL and credentials. We also need to support SSL/TLS on this database connection.

### LINKED ISSUE #943: [Task] Store existing Discovered Targets in Database
Parent: https://github.com/cryostatio/cryostat/issues/937
Depends on https://github.com/cryostatio/cryostat/issues/942

Once we have a database available, the existing target discovery information we have should be stored and retrieved from that database, rather than being done entirely in-memory. Custom Targets should be migrated from storage as JSON files on disk to storage in this database.

andrewazores:
Depends on #1029
Based on top of #1029
Related to #1028
Related to #937
Related to #938
Related to #939
Fixes #942
Fixes #943

- feat(discovery): pluggable discovery services API (#1026)
- feat(discovery): prepare to implement URL callback handling
- don't persist or restore plugins without callbacks
- POST plugin callback URLs on startup to check presence
- run discovery plugin with restart policy
- improve shutdown handling if verticles fail to start up
- refactor startup for cleanliness
- don't register self if builtin discovery disabled
- respond 400 if UUID invalid
- add handler unit tests, minor fixups
- use newer vertx-fib-demo that handles SIGTERM
- feat(db): add in-memory h2 database w/ Hibernate
- extract StorageModule
- extract AbstractDao
- make database parameters configurable

By default, H2 database in-memory will be used. There are various new environment variables that can be set to control various aspects of JPA/Hibernate for connecting to a database. The easy high-level way to test out each is to pass an argument to `smoketest.sh.

```bash
$ sh smoketest.sh # use h2 in-memory
$ sh smoketest.sh h2file # use h2 with file-based persistence, will write into the Cryostat conf dir
$ sh smoketest.sh postgres # spin up a PostgreSQL container using podman and connect to that remotely
```


andrewazores:
Just noticed this PR is opened against `main`, but it should be opened against #1026 . I'll close this and re-open against that.

PR #1032, opened 2022-08-03
feat(discovery): use hibernate, h2database, postgresql

### LINKED ISSUE #942: [Task] Add Services Discovery Database

### LINKED ISSUE #943: [Task] Store existing Discovered Targets in Database

andrewazores:
Part of #1028
Related to #1028
Related to #937
Related to #938
Related to #939
Fixes #942
Fixes #943

- feat(db): add in-memory h2 database w/ Hibernate
- extract StorageModule
- extract AbstractDao
- make database parameters configurable
- format sql in logs
- use jsonb for subtrees
- use file-based h2 by default
- minor cleanup
- rough setup for postgres db and switching from h2 in-memory, h2 file, and postgres remote
- minor postgres fixes

By default, H2 database in-memory will be used. There are various new environment variables that can be set to control various aspects of JPA/Hibernate for connecting to a database. The easy high-level way to test out each is to pass an argument to `smoketest.sh.

```bash
$ sh smoketest.sh # use h2 in-memory
$ sh smoketest.sh h2file # use h2 with file-based persistence, will write into the Cryostat conf dir
$ sh smoketest.sh postgres # spin up a PostgreSQL container using podman and connect to that remotely
```

As this implementation matures and stabilizes I think we should add a simple startup check that the expected h2 file-based location is accessible and writable, and default to h2:file. If not then log the problem and fall back to h2:mem. The environment variables would still be used to override this setting so that h2:mem can be forced, or a connection to an external Postgres instance can be established. The h2:mem database is obviously always private to the Cryostat instance, and the h2:file can be assumed to be as well, so it is OK for Cryostat (via hibernate) to initialize these databases and set up schemas etc. For an external Postgres instance it is likely unsafe for Cryostat/hibernate to modify the database schema, if the user Cryostat connects as even has permissions to do so, so it will be up to the end deploying user to ensure the Postgres instance is correctly configured. There is some more work we can do here to create schema scripts and things that are useful in development as well as for a reference for Postgres users.


maxcao13 (commented):
I'm wondering if these hibernate logs printed should be telling us anything more? I'm not sure if they are telling anything useful...
```
Hibernate: 
    /* select
        generatedAlias0 
    from
        PluginInfo as generatedAlias0 */ select
            plugininfo0_.id as id1_0_,
            plugininfo0_.callback as callback2_0_,
            plugininfo0_.realm as realm3_0_,
            plugininfo0_.subtree as subtree4_0_ 
        from
            PluginInfo plugininfo0_
```

src/main/java/io/cryostat/discovery/PluginInfo.java
```diff
@@ -0,0 +1,141 @@
...
+    public void setCallback(URI callback) {
+        this.callback = callback;
+    }
+
+    public void setSubtree(String subtree) {
+        this.subtree = subtree;
+    }
+
+    @Override
+    public int hashCode() {
```
maxcao13, line 115: I'm curious at the significance of this hashCode function. Does the prime make sure of the uniqueness of a PluginInfo?

andrewazores: It's auto-generated.

https://stackoverflow.com/questions/3613102/why-use-a-prime-number-in-hashcode

andrewazores:
> I'm wondering if these hibernate logs printed should be telling us anything more? I'm not sure if they are telling anything useful...
> 
> ```
> Hibernate: 
>     /* select
>         generatedAlias0 
>     from
>         PluginInfo as generatedAlias0 */ select
>             plugininfo0_.id as id1_0_,
>             plugininfo0_.callback as callback2_0_,
>             plugininfo0_.realm as realm3_0_,
>             plugininfo0_.subtree as subtree4_0_ 
>         from
>             PluginInfo plugininfo0_
> ```

It may not be useful to look at right now, but it's just logging the database queries being made. Should something go wrong then the logs would probably be handy to have :-)

tale
```diff
@@ -0,0 +1,2057 @@
+[m        plugininfo0_.subtree as subtree4_0_0_
```
maxcao13, line 1: Should this file be commited in the final push or is it for test viewing?

src/main/java/io/cryostat/discovery/PluginInfo.java
maxcao13: Ahh, I see. Interesting.

tale
andrewazores: Not even sure where that came from tbh, thanks for noticing it.

maxcao13:
I guess this is unrelated but, is it a feature that automated rules don't start when we haven't authenticated a JVM with JMX auth? For example, when I run smoketest with `CRYOSTAT_DISABLE_JMX_AUTH=false` (default), any automated rules won't start, even after I auth with `user=smoketest,password=smoketest`. But when I set the env to true, it will.

andrewazores:
Yea, that sounds unrelated. File a separate issue with reproduction steps please. Rules should run regardless of JMX auth, but it sounds like perhaps a failed JMX connection results in the rule no longer evaluating on its future scheduled times.

maxcao13:
Also, seems the postgres functionality requires `psql` to be installed. Should be documented somewhere later I think.

andrewazores:
`psql` wouldn't be needed by the Cryostat image at runtime, just needed if you're a developer running `smoketest.sh` and want to test out a Postgres setup that way. We could still document that, I guess.

smoketest.sh
```diff
@@ -10,16 +10,54 @@ function runCryostat() {
...
+    podman run \
+        --name postgres \
+        --pod cryostat-pod \
+        --env POSTGRES_USER=postgres \
+        --env POSTGRES_PASSWORD=abcd1234 \
+        --mount type=bind,source="$(dirname $0)/conf/postgres",destination=/var/lib/postgresql/data/pgdata,relabel=shared \
+        --mount type=bind,source="$(dirname $0)/src/test/resources/postgres",destination=/docker-entrypoint.initdb.d,relabel=shared \
+        --env PGDATA=/var/lib/postgresql/data/pgdata \
+        --rm -d "${image}:${version}"
+    sleep 5
```
maxcao13, line 57: Is the sleep there because it gives time for the pod to start running the postgres instance before connecting?

andrewazores: Exactly, there are probably smarter and more reliable ways to do this but it seemed to work well enough for this script, for now.

maxcao13:
When making and deleting custom targets, it seems custom targets are still stored in json on disk. Is this right? According to this PR, it should be changed https://github.com/cryostatio/cryostat/issues/943.

Seems like the web-client side sends a post to v2/targets but that handler hasn't been updated yet.

andrewazores:
Ah yes, good catch! I got too focused on the JDP and KubeApi discovery mechanisms and brain farted on the custom targets. I can address that easily enough. Would you prefer I do that in this PR or as another follow-up?

maxcao13:
> Ah yes, good catch! I got too focused on the JDP and KubeApi discovery mechanisms and brain farted on the custom targets. I can address that easily enough. Would you prefer I do that in this PR or as another follow-up?

This PR is fine, no worries :-)

src/main/java/io/cryostat/storage/AbstractDao.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public final T save(T t) {
+        Objects.requireNonNull(t);
+        EntityTransaction transaction = entityManager.getTransaction();
+        try {
+            transaction.begin();
+            entityManager.persist(t);
+            transaction.commit();
+            entityManager.detach(t);
```
hareetd, line 72: We detach here so that any changes made to the returned object won't be synchronized to the database, promoting a safer, more deliberate workflow, correct?

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -63,173 +56,142 @@
...
-        }
-        logger.info("Stale discovery service {} removed", uri);
+        List<Future> futures =
+                dao.getAll().stream()
+                        .map(
+                                plugin -> {
+                                    UUID key = plugin.getId();
+                                    URI uri = plugin.getCallback();
+                                    if (uri == NO_CALLBACK) {
+                                        removePlugin(key, key);
```
hareetd, line 99: I think this should be `removePlugin(key, uri)`.

smoketest.sh
```diff
@@ -10,16 +10,54 @@ function runCryostat() {
...
+    if [ "$1" = "postgres" ]; then
+        JDBC_URL="jdbc:postgresql://cryostat:5432/cryostat"
+        JDBC_DRIVER="org.postgresql.Driver"
+        HIBERNATE_DIALECT="org.hibernate.dialect.PostgreSQL95Dialect"
+        JDBC_USERNAME="postgres"
+        JDBC_PASSWORD="abcd1234"
+        HBM2DDL="update"
+    elif [ "$1" = "h2file" ]; then
+        JDBC_URL="jdbc:h2:file:/opt/cryostat.d/conf.d/h2;INIT=create domain if not exists jsonb as other"
+        HBM2DDL="update"
```
hareetd, line 23: Could you explain what happens "behind the scenes" when we set `HBM2DDL="update"`? Does this mean any time we make changes to the schema defined in `src/test/resources/postgres/postgre.sql`, the database instance deployed in `smoketest` will be updated accordingly?

hareetd (commented):
Just had a few comments/questions, will test using `smoketest` as well.

src/main/java/io/cryostat/storage/AbstractDao.java
andrewazores: Right. I think it would be more surprising as a developer to find that the POJO instance returned here still has some underlying database connection and mutations may incur transaction costs.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
andrewazores: In this case the `uri` is `null` (`NO_CALLBACK`). The API handler for registering a plugin does not allow null URIs, so the only way this can happen is for registrations performed in-process by the `BuiltInDiscovery`. It's meaningless for Cryostat to perform a callback startup check to itself, so this null special case is written in here.

smoketest.sh
andrewazores: The `update` value is only passed in here when we're running in `h2:file` mode, not when we're connecting to a Postgres instance. It means that Hibernate will try to update the database schema for the file-based H2 database instance. The schema is generated by Hibernate based on the `persistence.xml` and the various `Properties` values set in the `StorageModule` (which act as overrides on the `persistence.xml` config). Hibernate inspects all the classes named in the config, which right now is only the `PluginInfo` class, and generates a schema from that object definition.

For Postgres, the schema is hand-written in that `postgre.sql` file that you identified, and it is applied "automatically" in the developer testing workflow [here](https://github.com/cryostatio/cryostat/pull/1032/files#diff-cce023726064db3aa55e2ba6abf9be5551fc6ad70919029d36f9a4a3a1c33f9fR58). This isn't a proper database migration script, it's just an initialization really, so for the development workflow there would still likely be a requirement to either manually drop the existing database table(s) or just blow away the storage directory before being able to re-apply any changes to the `postgre.sql`.

hareetd: > The schema is generated by Hibernate based on the persistence.xml and the various Properties values set in the StorageModule (which act as overrides on the persistence.xml config). Hibernate inspects all the classes named in the config, which right now is only the PluginInfo class, and generates a schema from that object definition. If we add a new field to the PluginInfo for example then on the next rebuild/re-run (in h2:file mode both before and after) Hibernate will add a new corresponding column to the table.

Ah, this makes a lot more sense. Thanks!

maxcao13 (commented):
Seems like the new quarkus-test jvm with port 9097 that is discovered by the new Discovery plugin is not correctly picked up by the Automated Rule match expressions, unless I'm missing something.

 e.g.

image

andrewazores:
That's just because the `quarkus-test` discovery plugin isn't publishing itself with any annotations, so your `.annotations.cryostat` in the match expression is an empty map.

https://github.com/andrewazores/quarkus-test/blob/f83dbc85c67e1f54330547f9dbe7e8761e9a0811/src/main/java/org/acme/AppLifecycle.java#L45

This structure corresponds to https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/platform/discovery/TargetNode.java . The `quarkus-test` plugin just doesn't populate anything for labels or annotations right now. You can see this in the match expression evaluator in the bottom right of your screenshot:

```json
"labels": {},
"annotations": {
  "platform": {},
  "cryostat": {}
}
```

hareetd (approved):
Target detection, retrieval, and handling is working well in all three `smoketest` database setups.

maxcao13:
I'm not sure if this is just a function of smoketest being a dev testing deployment of cryostat or not, but the CustomTargets do not persist after restarting the application.

andrewazores:
Try `sh smoketest.sh h2file` or `sh smoketest.sh postgres`. The default is an in-memory h2 database instance which doesn't actually persist anything, so it's always completely empty when you restart. The file-backed h2 database and a proper Postgres instance both have real persistence and your custom target definitions should appear on a restart.

maxcao13:
> Try `sh smoketest.sh h2file` or `sh smoketest.sh postgres`. The current script default is an in-memory h2 database instance which doesn't actually persist anything, so it's always completely empty when you restart. The file-backed h2 database and a proper Postgres instance both have real persistence and your custom target definitions should appear on a restart.

I think I did tried on all 3 different instances and it did the same thing, but if it's working on your end its more than likely my problem again, :/ . I will retry again.

andrewazores:
Hmm. It does work for me with `h2file`. @tthvo @hareetd can you test that?

hareetd:
It isn't working for me either with `h2file`. To double-check, the expected workflow is that once a custom target definition is posted to `v2/targets` (regardless of whether such a custom target actually exists or not), it's added to the `CustomTargetPlatformClient`, resulting in a `TargetDiscoveryEvent` being issued, which should trigger an update of the platform's discovery tree in persistant storage, correct?

maxcao13:
Yeah, all 3 for me still aren't persisting. I've just been (w/ web-client running `npm run start:dev` on :9000)
1. creating a custom target 
2. ctrl+C 
3. rerun `CRYOSTAT_DISABLE_SSL=true CRYOSTAT_DISABLE_JMX_AUTH=true CRYOSTAT_CORS_ORIGIN=http://localhost:9000 sh smoketest.sh h2file`

And the target does not comeback after restarting.

andrewazores:
Weird. I tried deleting the `conf/h2.mv.db` and `conf/h2.trace.db` files and starting from scratch and now I also don't see the definition getting persisted. Maybe something to do with a schema change I made in the middle somewhere here? I'll dig in.

andrewazores:
Oh, I think I see the problem, though I'm not sure why deleting the `h2.*.db` files had anything to do with it. When the `BuiltInDiscovery` (replacement for `MergingPlatformClient` and bridge between `DiscoveryStorage` and other `PlatformClient`s) shuts down it deregisters all of the realms that were created for the old platform clients, which removes the whole subtree from the database. On startup it registers a new realm for each of the platform clients and immediately repopulates the database by querying each platform client for its discovery tree and adding that as a subtree for the realm. Since the `CustomTargetPlatformClient` no longer has any storage of its own other than the database, this just ends up meaning that custom targets no longer have any sort of effective persistent storage. Funnily enough, it would persist as expected if the Cryostat server crashed and shut down uncleanly.

Bringing back separate persistent storage for custom targets isn't right, so the actual solution will have to revamp how the `BuiltInDiscovery` does the startup/shutdown registration/deregistration/reregistration of platform clients.

andrewazores:
Ah wait, deleting custom targets that are loaded from the database no longer works. You can only delete custom targets that are defined in the current run.

smoketest.sh
```diff
@@ -10,16 +10,54 @@ function runCryostat() {
...
+        --name postgres \
+        --pod cryostat-pod \
+        --env POSTGRES_USER=postgres \
+        --env POSTGRES_PASSWORD=abcd1234 \
+        --mount type=bind,source="$(dirname $0)/conf/postgres",destination=/var/lib/postgresql/data/pgdata,relabel=shared \
+        --mount type=bind,source="$(dirname $0)/src/test/resources/postgres",destination=/docker-entrypoint.initdb.d,relabel=shared \
+        --env PGDATA=/var/lib/postgresql/data/pgdata \
+        --rm -d "${image}:${version}"
+    sleep 5
+    PGPASSWORD=abcd1234 psql -h localhost -p 5432 -U postgres -f src/test/resources/postgres/postgre.sql
```
tthvo, line 58: Wondering if we can run this command from inside the postgres container like `podman exec -t` so that we don't need to install `psql`?

andrewazores: Good call. I did a little reading and the latest commit handles this much more nicely.

andrewazores:
> Ah wait, deleting custom targets that are loaded from the database no longer works. You can only delete custom targets that are defined in the current run.

Solved in latest commit.

tthvo:
I just run into a situation where killing `quarkus-test(service:jmx:rmi:///jndi/rmi://cryostat:9097/jmxrmi)` with `podman kill` does not remove the target. I guess the target didn't have enough time to notify Cryostat since receiving SIGKILL. I wonder how we go ahead and handle this?

andrewazores:
`SIGKILL` will always terminate the process/container without leaving it time to do any clean up, like notifying Cryostat that it's going away. `podman stop` will let it have up to 10 seconds (default) to perform such clean up.

The callback URL that plugins provide at registration time is meant to be the mechanism for handling this. Currently Cryostat only checks those URLs and prunes unreachable plugins when it starts up, but we can make this happen periodically too.

pom.xml
```diff
@@ -26,6 +26,9 @@
   <node.version>v14.17.0</node.version>
   <yarn.version>v1.22.13</yarn.version>
 
+  <postgres.image>docker.io/postgres</postgres.image>
+  <postgres.version>13-bullseye</postgres.version>
```
tthvo, line 30: Any specific reason to switch to `-bulleye`?

andrewazores: https://hub.docker.com/_/postgres

No good anchor link on the page so Ctrl+F for "Arbitrary --user Notes"

tl;dr the default image is based on Alpine Linux, `-bullseye` is Debian. The Debian version of the image has some extra magic built in that lets it run under your host system UID/GID, so mounting the init script as a shared volume owned by your user works properly.

tthvo: Ahh makes sense! Thanks :D

maxcao13 (approved):
Everything looks good to me. Great work!

tthvo:
Somehow i cannot delete any of the targets other than custom targets. Is that intentional?

postgres_fail

andrewazores:
Yes, you can only delete custom targets. It's unfortunate that the UI still allows you to click the button on other targets discovered via automatic means, but that predates this PR.

tthvo:
Ah makes sense!

tthvo (approved):
Looks great!!!!

andrewazores:
> Yes, you can only delete custom targets. It's unfortunate that the UI still allows you to click the button on other targets discovered via automatic means, but that predates this PR.

That said, with the new-ish discovery tree API, the UI could use that to both populate the list (by just looking for leaves of the tree and listing them) as well as toggle the delete button (if the selected leaf was part of the Custom Targets realm subtree). That's a bit inefficient and annoying to re-implement on the frontend, though.

Maybe we can just add ex. `target.annotations.cryostat["realm"]="Custom Targets"` to the `ServiceRef`s. This way the standard target listing API still contains that information about the realm and it doesn't need to be determined by descending through the whole discovery tree.

PR #1034, opened 2022-08-04
feat(version): log application version at startup

### LINKED ISSUE #1033: [Task] Print application version information on startup
We have some application version information that is available on the `GET /health` endpoint, which we also display in the web-client. It would be useful to have this also logged at startup.

andrewazores:
Fixes #1033


tthvo (approved):
Looks good! Will this come into pluggable discovery PR too since this part is refactored there?

andrewazores:
Yep, I'll have to rebase the pluggable discovery branches on top of `main` after this is merged.

PR #1037, opened 2022-08-04
fix(k8s): replace Watch with Informer

andrewazores:
Fixes #1036

update fabric8 kubernetes-client from 5.4.1 to 5.12.3 (latest 5.x) and replace direct Watch implementation with a fabric8 Informer. The Informer handles reconnecting the Watch on connection failures or when the API server responds with HTTP 410 (GONE), which can happen when too much time has elapsed between update notifications (such as during API server outages). The Informer also maintains an internal Store modelling the resource, in this case Endpoints, and resynchronizes this againt the API server periodically. This allows us to compute the list of ServiceRefs or the DiscoveryTree by constructing the result from the Informer's Store, rather than making an API call every time. Finally, the Informer's logic for added/updated/removed events is improved over the plain Watch and allows us to detect Pod scaling events correctly, as well as creation/deletion events.

For ease of testing, images are available at `quay.io/andrewazores/cryostat:k8s-rewatch-13` and `quay.io/andrewazores/cryostat-operator:k8s-rewatch-13`.

Issue #625, opened 2021-08-05
[Task] Update fabric8-client to 5.5.0+ and remove KUBERNETES_AUTH_TRYKUBECONFIG_SYSTEM_PROPERTY


andrewazores:
> I figured it out. There is an OkHttp Interceptor added in the DefaultOpenShiftClient to retry requests that fail authorization by finding a token or asking the server for a new one:
https://github.com/fabric8io/kubernetes-client/blob/7e747ca61c0097562622eb53b9f008cc6d5e82fb/openshift-client/src/main/java/io/fabric8/openshift/client/internal/OpenShiftOAuthInterceptor.java

> There's a special case for RBAC review objects since these can return 200 or 201 even when unauthorized. These are set to always retry. As a consequence, I believe each SelfSubjectAccessReview will be sent twice unfortunately. This request doesn't get resent if there is no username/password or token in the OpenShiftConfig object, which should be the case in our tests, but not in practice.

> The difference we're seeing between 5.4.1 and 5.6.0 is due to this PR: https://github.com/fabric8io/kubernetes-client/pull/3158. The test was failing because the OpenShiftClient found a token to use within our ~/.kube/config. If you rename that file, the test will pass.

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat/pull/599#discussion_r683007125_

andrewazores:
Downstream builds currently have 5.4.1 as the latest version, though 5.6.0 is available for upstream. Once 5.5.0+ is available downstream then this issue should be addressed.

PR #1042, opened 2022-08-16
fix(k8s-rbac): update permission mapping between k8s and cryostat

### LINKED ISSUE #979: [Task] Select default permissions that do not use FlightRecorder/Recording
After #635, we can modify which Cryostat permissions map to which Kubernetes RBAC permissions. Now we should modify the default Kubernetes permissions from their current values. Since the FlightRecorder and Recordings APIs will soon be removed, we have to pick some alternatives. Ideally these would not depend on CRDs shipped with the operator, but that's not a hard requirement, it would simply allow for more flexibility.

tthvo:
Fixes #979 
Related to https://github.com/cryostatio/cryostat-operator/pull/440

Use alternatives for mapping between k8s resource to cryostat since flightRecorder/Recordings are deprecated.

### DIFF
--- a/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
+++ b/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
@@ -9,5 +9,5 @@
 # ResourceType to Kubernetes resource kind mapping:
 TARGET=pods,services
 RECORDING=pods,pods/exec,cryostats.operator.cryostat.io
-CERTIFICATE=pods,cryostats.operator.cryostat.io
-CREDENTIALS=pods,cryostats.operator.cryostat.io
+CERTIFICATE=pods,secrets,cryostats.operator.cryostat.io
+CREDENTIALS=pods,secrets,cryostats.operator.cryostat.io
--- a/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
+++ b/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
@@ -9,5 +9,5 @@
 # ResourceType to Kubernetes resource kind mapping:
 TARGET=pods,services
 RECORDING=pods,pods/exec,cryostats.operator.cryostat.io
-CERTIFICATE=pods,secrets,cryostats.operator.cryostat.io
-CREDENTIALS=pods,secrets,cryostats.operator.cryostat.io
+CERTIFICATE=pods,cryostats.operator.cryostat.io
+CREDENTIALS=pods,cryostats.operator.cryostat.io
--- a/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
+++ b/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
@@ -3,12 +3,11 @@
 # Kubernetes resource kinds:
 # DEPLOYMENTS="deployments.apps"
 # PODS="pods"
+# SERVICES="services"
 # CRYOSTATS="cryostats.operator.cryostat.io"
-# FLIGHTRECORDERS="flightrecorders.operator.cryostat.io"
-# RECORDINGS="recordings.operator.cryostat.io"
 
 # ResourceType to Kubernetes resource kind mapping:
-TARGET=pods,endpoints,deployments.apps
-RECORDING=pods,pods/exec
-CERTIFICATE=deployments.apps,pods,cryostats.operator.cryostat.io
-CREDENTIALS=cryostats.operator.cryostat.io
+TARGET=pods,services
+RECORDING=pods,pods/exec,cryostats.operator.cryostat.io
+CERTIFICATE=pods,cryostats.operator.cryostat.io
+CREDENTIALS=pods,cryostats.operator.cryostat.io
--- a/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
+++ b/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
@@ -8,7 +8,7 @@
 # RECORDINGS="recordings.operator.cryostat.io"
 
 # ResourceType to Kubernetes resource kind mapping:
-TARGET=flightrecorders.operator.cryostat.io
-RECORDING=recordings.operator.cryostat.io
+TARGET=pods,deployments.apps
+RECORDING=pods,pods/exec
 CERTIFICATE=deployments.apps,pods,cryostats.operator.cryostat.io
 CREDENTIALS=cryostats.operator.cryostat.io
--- a/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
+++ b/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
@@ -8,7 +8,7 @@
 # RECORDINGS="recordings.operator.cryostat.io"
 
 # ResourceType to Kubernetes resource kind mapping:
-TARGET=pods,deployments.apps
+TARGET=pods,endpoints,deployments.apps
 RECORDING=pods,pods/exec
 CERTIFICATE=deployments.apps,pods,cryostats.operator.cryostat.io
 CREDENTIALS=cryostats.operator.cryostat.io
--- a/src/test/java/io/cryostat/net/openshift/OpenShiftAuthManagerTest.java
+++ b/src/test/java/io/cryostat/net/openshift/OpenShiftAuthManagerTest.java
@@ -147,14 +147,12 @@ class OpenShiftAuthManagerTest {
     void setup() throws IOException {
         client = Mockito.spy(client);
         tokenProvider = new TokenProvider(client);
-        MultiMap headers = MultiMap.caseInsensitiveMultiMap();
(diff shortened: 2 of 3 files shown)


src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
```diff
@@ -8,7 +8,7 @@
 # RECORDINGS="recordings.operator.cryostat.io"
 
 # ResourceType to Kubernetes resource kind mapping:
-TARGET=flightrecorders.operator.cryostat.io
-RECORDING=recordings.operator.cryostat.io
+TARGET=pods,deployments.apps
```
tthvo, line 11: For targets, I wonder if `endpoints` is also necessary? My thinking is that discovery mechanism needs access endpoint resource so it should be declared here.

andrewazores: Makes sense to me.

tthvo:
Marked as draft to update tests and check with running cluster.

src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties
ebaron: I just stumbled across this note that write access to Endpoints objects is going to be forbidden in most scenarios in Kubernetes 1.22+: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#write-access-for-endpoints

This may make `CREATE`, `UPDATE`, `DELETE` operations on targets difficult if we include Endpoints.

andrewazores: Fair enough. One way around would be to differentiate between normally discovered targets (readonly) and custom targets (all verbs), but that feels like a bit of a hack.

tthvo: For that security issue, I think we could remove it. I was looking into a replacement but I wonder if a more "
abstract" object like `service` is a good alternative?

ebaron: Service seems like a reasonable substitute since they're so closely coupled.

ebaron: These are all somewhat arbitrarily chosen, it would be good to allow users to configure this permissions mapping file from the operator. I'm not sure if we have an existing issue for that.

tthvo: Hmm I cant see one but I only remember Andrew mentioned it in this issue https://github.com/cryostatio/cryostat/issues/979#issuecomment-1146325376. In this case, we will allow some config in Cryostat for this mapping, right?

andrewazores: The file you edited in this PR is that mapping file, see #662 . There may be a bit more work required to make this something that the Operator can replace at runtime, but the idea is that the Operator could just mount a replacement file overtop of this default one in the container image.

ebaron: I filed an issue in the operator for this feature: https://github.com/cryostatio/cryostat-operator/issues/439

tthvo: Nicee! I will dig more into this auth flow and look at the new issue too!

tthvo:
Currently, this is my idea for refining the mappings:

``` bash
# Access pods (i.e application) and establish connections (i.e. using service instead of endpoints)
TARGET=pods,services 
# Access Cryostat to operate recordings
RECORDING=pods,pods/exec,cryostats.operator.cryostat.io
# Mount certificates from secrets into Cryostat
CERTIFICATE=pods,secrets,cryostats.operator.cryostat.io 
# Access Cryostat to add credentials (i.e. via webUI I think)
CREDENTIALS=pods,cryostats.operator.cryostat.io 
```
Any thoughts @andrewazores @ebaron?

ebaron:
> Currently, this is my idea for refining the mappings:
> 
> ```shell
> # Access pods (i.e application) and establish connections (i.e. using service instead of endpoints)
> TARGET=pods,services 
> # Access Cryostat to operate (i.e. execute) recordings
> RECORDING=pods,pods/exec,cryostats.operator.cryostat.io
> # Mount certificates from secrets into Cryostat
> CERTIFICATE=pods,secrets,cryostats.operator.cryostat.io 
> # Access Cryostat to add credentials (i.e. via webUI I think)
> CREDENTIALS=pods,cryostats.operator.cryostat.io 
> ```
> 
> Any thoughts @andrewazores @ebaron?

I think this is good. While the certificates themselves may not be sensitive, the act of telling Cryostat to trust them should be privileged. I would also add Secrets for credentials as well, since that is a common use case for Secrets in Kubernetes.

tthvo:
@andrewazores  I believe #625 is resolved since #1037? Just come across a FIXME to disable reading `kubeconfig` in `OpenShiftAuthManagerTest`.

andrewazores:
@tthvo thanks: https://github.com/cryostatio/cryostat/pull/1057

src/test/java/io/cryostat/net/openshift/OpenShiftAuthManagerTest.java
```diff
@@ -358,8 +348,7 @@ void shouldNotValidateTokenWithInsufficientPermissions() throws Exception {
                 Matchers.instanceOf(PermissionDeniedException.class));
         PermissionDeniedException pde = (PermissionDeniedException) ExceptionUtils.getRootCause(ee);
         MatcherAssert.assertThat(pde.getNamespace(), Matchers.equalTo(NAMESPACE));
-        MatcherAssert.assertThat(
-                pde.getResourceType(), Matchers.equalTo("recordings.operator.cryostat.io"));
+        MatcherAssert.assertThat(pde.getResourceType(), Matchers.equalTo("pods/exec"));
```
tthvo, line 351: I just noticed that with this, we can only check the first resource denied. That means in our mock for `RECORDING` should only specify one k8s resource unless we have some orders. But I think it does not matter that much.

tthvo:
@ebaron @andrewazores Ready for review :D

For tests, I just removed references in deprecated apis. There was no issue with subresource being used in tests (I thought it would be part of the `getResource` but using `getSubResource` is the right way).

src/test/java/io/cryostat/net/openshift/OpenShiftAuthManagerTest.java
andrewazores: I think that's fine. Maybe in the future it's worth refactoring things so that all of the requested permissions are checked and the set of failures are all reported, but the important thing is that unauthorized requests don't get completed.

andrewazores (dismissed):
LGTM

ebaron (dismissed):
Seems good to me. This should probably be committed close to https://github.com/cryostatio/cryostat-operator/pull/440, otherwise the OAuth tokens will not have the correct set of permissions.

tthvo:
Right, I will update https://github.com/cryostatio/cryostat-operator/pull/440 oauth role with these default permissions and tests it.

tthvo:
Since including `Secrets` in mapping is denied due escalation security issue, I am thinking of its "closely" resembling resource `ConfigMap` though I am not sure if it is good alternative.

*See  original https://github.com/cryostatio/cryostat-operator/pull/440#issuecomment-1227443550*

ebaron:
> Since including `Secrets` in mapping is denied due escalation security issue, I am thinking of its "closely" resembling resource `ConfigMap` though I am not sure if it is good alternative.
> 
> _See original [cryostatio/cryostat-operator#440 (comment)](https://github.com/cryostatio/cryostat-operator/pull/440#issuecomment-1227443550)_

Config Maps are similar, but not used for sensitive data, while Secrets are. I can't think of any other reasonable substitute though, so I'd say either use Config Maps or remove it entirely.

tthvo:
Right, I would just remove secrets instead of adding ConfigMap as it could means unprotected data which is not the case for certificates and credentials.

tthvo:
I found some `ConfigMap` that actually contains certificate data. I think in the case of certificates, we could use `ConfigMap`?
```bash
oc get cm
NAME                       DATA   AGE
auth-properties            1      24h
kube-root-ca.crt           1      65d
openshift-service-ca.crt   1      65d
```

ebaron:
> I found some `ConfigMap` that actually contains certificate data. I think in the case of certificates, we could use `ConfigMap`?

I don't think we gain much from doing that. Certificates can be stored in Config Maps because they aren't sensitive, but for Cryostat, the certificate permission allows the user to tell Cryostat to trust a certificate. This should be somewhat of a privileged action, and I don't think creating/updating config maps in the namespace captures it. Permission to create/update the Cryostat CR is better, because that captures more of an admin role.

tthvo:
Oh right! Make sense! Thanks a lot for explaining! I will finish up features to add custom role on top of default and clean up tests before marking this ready,

ebaron:
Thanks, once this is ready, I'll test this one along with the operator PR and we should (hopefully) be done.

ebaron (approved):
Works well for me!

PR #1060, opened 2022-08-30
test(graphql): add graphql unit tests

### LINKED ISSUE #947: [Task] Unit tests for GraphQL

maxcao13:
Fixes #947


maxcao13:
I have written tests for the fetcher and mutator implementations and surface level tests for the handlers. I have a few questions overall.

1. Why doesn't the `AllArchivedRecordingFetcher`implement `AbstractPermissionedDataFetcher`? I'm assuming we need permissions to view targets and recordings from the AllArchives view just like the other fetchers.
2. Is the GraphiQL handler for devs to be able to _graphically_ see our GraphQL queries and responses? If so, does it still make sense to test it?
3. Speaking of the handlers, I'm not sure how to test the `GraphQLGetHandler` and `Post` handlers 'correctly' since we don't override the `handle` method of the vert.x `RequestHandler`s and most of the graphql implementation is very abstracted and internal to vertx and graphql itself. Does it still make sense to try and test the handler endpoints as unit tests and get actual responses, feasibly? I was thinking the integration tests also does this anyways.

andrewazores:
1. Good question. It probably should. Hareet wrote that fetcher around the same time that I wrote the abstract permissioned base class, so I think I just missed including this class when I rebased my work on top.
2. Correct, it gives a REPL-like page where you can tinker with GraphQL queries. There's no need to test it.
3. These can have very basic unit tests just to ensure that they have the correct HTTP verb, correct path, and require authorization for example. There is no need to try to test them any deeper than that since it just ends up delegating to a Vert.x implementation and as you say, we exercise that already with the integration tests too.

maxcao13:
I've fixed everything that I can see and most of the GraphQL stuff should be covered.

Thoughts or anything I've missed?

andrewazores (approved):

andrewazores:
Awesome, thanks for doing all of this.

PR #1062, opened 2022-09-12
feat(discovery): nonunique plugin realms

### LINKED ISSUE #1061: [Task] Discovery plugin names should have unique ID generated and appended
In verbal discussion a few days ago, the idea came up that rather than plugins being required to provide their own unique registration name, they should simply provide a display name. This display name can be stored for display in API results and used in UI and need not be unique. To generate a unique ID for plugin registrations, Cryostat should use a sanitized form of the display name with a UUID appended.

andrewazores:
Fixes #1061

- feat(discovery): realm names need not be unique
- embed registration ID in realm node labels
- use updated quarkus-test-plugin
- extract configuration property

Rather than creating a separate "display name" property, or appending to the plugin-provided realm name, this simply sets the registration UUID to a `REALM` label on the node. This way the existing realm property and node "name" property can be used for display purposes, but the UUID is available for querying via label selection and to differentiate nodes with the same name.

```bash
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)"
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 578

{
    "data": {
        "result": {
            "children": [
                {
                    "children": [],
                    "labels": {
                        "REALM": "da1ec8a9-3797-4873-a4d8-207b22e772e1"
                    },
                    "name": "Custom Targets",
                    "nodeType": "Realm"
                },
                {
                    "children": [
                        {
                            "labels": {},
                            "nodeType": "JVM",
                            "target": {
                                "alias": "/deployments/quarkus-run.jar",
                                "annotations": {
                                    "cryostat": {
                                        "HOST": "cryostat",
                                        "JAVA_MAIN": "/deployments/quarkus-run.jar",
                                        "PORT": "10000",
                                        "REALM": "JDP"
                                    },
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:10000/jmxrmi",
                                "labels": {}
                            }
                        },
                        {
                            "labels": {},
                            "nodeType": "JVM",
                            "target": {
                                "alias": "io.cryostat.Cryostat",
                                "annotations": {
                                    "cryostat": {
                                        "HOST": "cryostat",
                                        "JAVA_MAIN": "io.cryostat.Cryostat",
                                        "PORT": "9091",
                                        "REALM": "JDP"
                                    },
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi",
                                "labels": {}
                            }
                        },
                        {
                            "labels": {},
                            "nodeType": "JVM",
                            "target": {
                                "annotations": {
                                    "cryostat": {
                                        "HOST": "cryostat",
                                        "JAVA_MAIN": "es.andrewazor.demo.Main",
                                        "PORT": "9093",
                                        "REALM": "JDP"
                                    },
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi",
                                "labels": {}
                            }
                        },
                        {
                            "labels": {},
                            "nodeType": "JVM",
                            "target": {
                                "annotations": {
                                    "cryostat": {
                                        "HOST": "cryostat",
                                        "JAVA_MAIN": "es.andrewazor.demo.Main",
                                        "PORT": "9094",
                                        "REALM": "JDP"
                                    },
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9094/jmxrmi",
                                "labels": {}
                            }
                        },
                        {
                            "labels": {},
                            "nodeType": "JVM",
                            "target": {
                                "annotations": {
                                    "cryostat": {
                                        "HOST": "cryostat",
                                        "JAVA_MAIN": "es.andrewazor.demo.Main",
                                        "PORT": "9095",
                                        "REALM": "JDP"
                                    },
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9095/jmxrmi",
                                "labels": {}
                            }
                        },
                        {
                            "labels": {},
                            "nodeType": "JVM",
                            "target": {
                                "alias": "/deployments/quarkus-run.jar",
                                "annotations": {
                                    "cryostat": {
                                        "HOST": "cryostat",
                                        "JAVA_MAIN": "/deployments/quarkus-run.jar",
                                        "PORT": "9096",
                                        "REALM": "JDP"
                                    },
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9096/jmxrmi",
                                "labels": {}
                            }
                        }
                    ],
                    "labels": {
                        "REALM": "8ea31747-bb8f-456f-8874-c61d48b63ce1"
                    },
                    "name": "JDP",
                    "nodeType": "Realm"
                },
                {
                    "children": [
                        {
                            "labels": {},
                            "name": "service:jmx:rmi:///jndi/rmi://cryostat:9097/jmxrmi",
                            "nodeType": "JVM",
                            "target": {
                                "alias": "quarkus-test",
                                "annotations": {
                                    "cryostat": {},
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9097/jmxrmi",
                                "labels": {}
                            }
                        }
                    ],
                    "labels": {
                        "REALM": "a232d7f6-f878-46e6-8bdb-67598e0bfef4"
                    },
                    "name": "quarkus-test",
                    "nodeType": "Realm"
                },
                {
                    "children": [
                        {
                            "labels": {},
                            "name": "service:jmx:rmi:///jndi/rmi://cryostat:9197/jmxrmi",
                            "nodeType": "JVM",
                            "target": {
                                "alias": "quarkus-test",
                                "annotations": {
                                    "cryostat": {},
                                    "platform": {}
                                },
                                "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9197/jmxrmi",
                                "labels": {}
                            }
                        }
                    ],
                    "labels": {
                        "REALM": "8c0ce236-305b-47c9-8367-d17c463d7162"
                    },
                    "name": "quarkus-test",
                    "nodeType": "Realm"
                }
            ],
            "labels": {},
            "name": "Universe",
            "nodeType": "Universe"
        }
    },
    "meta": {
        "status": "OK",
        "type": "application/json"
    }
}
```


andrewazores:
@maxcao13 ready for review

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -260,10 +272,15 @@ public List<ServiceRef> listDiscoverableServices() {
     }
 
     public List<ServiceRef> listDiscoverableServices(String realm) {
```
maxcao13, line 274: This method seems to list any services with the same name as the realm, and one place it seems to be in use is in the `CustomTargetPlatformClient`. Will there be a bug if someone names their plugin "Custom Targets"?

andrewazores: True, if a plugin is registered that uses that same realm name then it could cause a conflict. I can get around this by providing the registration ID to the built-in `PlatformClient` implementations as a parameter to the `load` call.

smoketest.sh
```diff
@@ -102,12 +102,26 @@ function runDemoApps() {
...
         --env ORG_ACME_JMXPORT="9097" \
-        -d quay.io/andrewazores/quarkus-test:0.0.6
+        -d quay.io/andrewazores/quarkus-test:0.0.8
+
+    podman run \
+        --name quarkus-test-plugin-2 \
+        --pod cryostat-pod \
+        --restart unless-stopped \
+        --env QUARKUS_HTTP_PORT=10011 \
+        --env JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dcom.sun.management.jmxremote.port=9197 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false" \
```
maxcao13, line 118: What's the difference between these `quarkus-test-plugin-2`? This one has extra java options args.

Also there are 2 plugin in the same port in the comments.
```
    # 9096: quarkus-test RJMX
    # 9096: quarkus-test-plugin RJMX
```

andrewazores: `quarkus-test-plugin-2` is running the same code as `quarkus-test-plugin`, but the `JAVA_OPTIONS` run it with a different JMX port than the container's default (9197 vs the default of 9097 actually, that comment is outdated).

This is just to test the case of two identical plugins registering themselves with the same realm name.

andrewazores: The defaults are here btw:

https://github.com/andrewazores/quarkus-test/blob/e91afb589e961f157c472e24e57b5977ed73cf1b/src/main/docker/Dockerfile.jvm#L44

maxcao13: Ah, I see thanks!

andrewazores:
Oh, wait. Gah. This won't work either. The custom target "plugin" will be registering with a new ID on each startup and not retaking ownership of the data the previous instance published last lifecycle. You can see old custom target definitions, create new ones, and delete those new ones, but can't delete old ones.

Hmm. Needs more thought. I really don't want to make the names of the built-in plugins special cases where only those are enforced as unique.

maxcao13:
> Oh, wait. Gah. This won't work either. The custom target "plugin" will be registering with a new ID on each startup and not retaking ownership of the data the previous instance published last lifecycle. You can see old custom target definitions, create new ones, and delete those new ones, but can't delete old ones.
> 
> Hmm. Needs more thought. I really don't want to make the names of the built-in plugins special cases where only those are enforced as unique.

Would it make sense to have unique special case ids for the build-in "plugins"? Or does that sort of remove the purpose of them?

andrewazores:
> > Oh, wait. Gah. This won't work either. The custom target "plugin" will be registering with a new ID on each startup and not retaking ownership of the data the previous instance published last lifecycle. You can see old custom target definitions, create new ones, and delete those new ones, but can't delete old ones.
> > Hmm. Needs more thought. I really don't want to make the names of the built-in plugins special cases where only those are enforced as unique.
> 
> Would it make sense to have unique special case ids for the build-in "plugins"? Or does that sort of remove the purpose of them?

I thought of that too, but that would need some other bigger overhauls. The database itself is what generates the ID, so to handle the built-in plugins' IDs as special cases would mean they need to be generated and then persisted somewhere we can reference those again for the next startup. Or we would need to generate the IDs in application code and hardcode the IDs for the built-in plugins only and generate random ones for others.

The solution I settled on is to query the database by realm name again, but also to filter those by plugins that have a null callback URL. Plugins cannot provide a null/blank callback URL when they register through the external API, only the built-ins can do that through the internal registration interface, so this lets us restore our registration at startup time.

andrewazores:
Rebased.

maxcao13:
> The solution I settled on is to query the database by realm name again, but also to filter those by plugins that have a null callback URL. Plugins cannot provide a null/blank callback URL when they register through the external API, only the built-ins can do that through the internal registration interface, so this lets us restore our registration at startup time.

That's clever! Everything LGTM now.

maxcao13 (approved):

PR #1047, opened 2022-08-18
fix(archives): add target-specific archive API endpoints

hareetd:
Fixes #1046 
Depends on https://github.com/cryostatio/cryostat-web/pull/499

Summary:

1. I've added five new `beta` target-specific archived recording handlers: `RecordingDeleteHandler`, `RecordingUploadPostHandler`, `RecordingGetHandler`, `RecordingGetWithJwtHandler`, `ReportGetHandler` and `ReportGetWithJwtHandler`. They're essentially equivalent to their `v1`/`v2` counterparts, except with an added `sourceTarget` request parameter that expects the JMX connection URL of the source target for the archived recording in question.

2. The `WebServer` now points to the `beta`  `RecordingGetHandler` and `ReportGetHandler` endpoints when the `WebServer#getArchivedDownloadUrl` and `WebServer#getArchivedReportUrl` methods are called, respectively. This means archived recording-related notifications will now contain target-specific download URLs. Notifications have also been updated so that the `target` field is equal to `"uploads"` rather than the empty string for uploaded recordings. This is because for the new target-specific handlers the `sourceTarget` parameter cannot be set to the empty string in place of a `connectUrl` for uploaded recordings. Since the `"uploads"` string is the logical alternative to use for this parameter, in order to maintain consistency, notifications also explicitly identify a recording as having originated as an upload directly into Cryostat's archives. 

3. The implementation should still preserve the functionality of the older, non-target-specific `v1`/`v2` handlers. I didn't have time to manually test but unit and integration tests are all passing. In the case where such an endpoint is called, the archives are searched using only the recording name, which is still error-prone for the reasons discussed in https://github.com/cryostatio/cryostat-web/issues/445. Perhaps these handlers should now be marked as deprecated?

4. For the cached report path, if no `sourceTarget` is provided the report is placed inside a new `default` subdirectory which is itself inside the temporary directory provided by the `WebServer`. Uploaded recordings are placed inside an `uploads` subdirectory and target-specific recordings inside a subdirectory named according to the target's Base32 encoded `connectUrl`. See my comment below on issues with creating these subdirectories.  

5. The web-client in https://github.com/cryostatio/cryostat-web/pull/499 has been updated to use the new `beta` handlers in place of the old ones (I might have missed something though, so please double-check). Unfortunately, the View Report functionality is broken as it keeps timing out with the report not being delivered. However, I'm experiencing this behavior on the main branch as well, so could be a general issue or an error on my part. The report dropdown generation works as expected. 

[...]

PR #1066, opened 2022-09-14
feat(graphql): add aggregate information for active recordings in the GraphQL API

### LINKED ISSUE #1030: [Task] Add aggregate information for active recordings in the GraphQL API
In #924 the [GraphQL schema](https://github.com/cryostatio/cryostat/blob/22380ca0bd76689b3de162fb94487f1124d3775c/src/main/resources/types.graphqls#L86) for archived recordings was changed to include aggregate information alongside the actual recording list itself. 

```
type Recordings {
    active(filter: ActiveRecordingFilterInput): [ActiveRecording!]!
    archived(filter: ArchivedRecordingFilterInput): Archived!
}

...

type Archived {
    data: [ArchivedRecording!]!
    aggregate: AggregateInfo!
}

type AggregateInfo {
    count: Long!
}
```
Active recordings should mirror this change for consistency purposes. Adding more aggregate information fields alongside the count for both types of recordings would be nice as well.

maxcao13:
Fixes #1030


andrewazores:
I think this will depend on a -web PR to be created as well: https://github.com/cryostatio/cryostat-web/pull/473#issuecomment-1223383346

andrewazores:
Related: https://github.com/cryostatio/cryostat/pull/1077

@maxcao13 can you resume work on this PR so we can be sure it gets into this release alongside the linked one above? Otherwise the API structure becomes asymmetric between active and archived.

maxcao13:
The PR doesn't included the new 'size' aggregate info, do you want me to include it here?

maxcao13:
Also, a query like this:
```bash
$ http :8181/api/v2.2/graphql Authorization:"Basic $(echo user:pass | base64)" query="query { targetNodes { name recordings { active { aggregate { count } } }
 } }"
```
results in 
```
{
    "data": null,
    "errors": [
        {
            "extensions": {
                "classification": "DataFetchingException"
            },
            "locations": [
                {
                    "column": 28,
                    "line": 1
                }
            ],
            "message": "Exception while fetching data (/targetNodes[4]/recordings) : Failed to retrieve RMIServer stub: javax.naming.CommunicationException [Root exception is java.rmi.ConnectIOException: non-JRMP server at remote endpoint]",
            "path": [
                "targetNodes",
                4,
                "recordings"
            ]
        }
    ]
}
```
if a target is unreachable (in my case 9095 with no SSL cert), but this is fixed in the archive endpoint PR #1047, by the try catch in the `ActiveRecordingFetcher`.

This is a query on reachable targets using smoketest (for me who doesn't know how to validate the SSL for 9095).
```bash
$ http :8181/api/v2.2/graphql Authorization:"Basic $(echo user:pass | base64)" query="query { environmentNodes(filter: { name: \"quarkus-test\" }) { name nodeType descendantTargets { recordings { active { data { name state metadata { labels } } } archived { data { name metadata { labels } } } } } } }"
```

andrewazores:
> The PR doesn't included the new 'size' aggregate info, do you want me to include it here?

I'm not sure if we can easily do that right now - the [JDK Recording class](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/Recording.html#getSize()) has an accessor for this, but the [JMC IRecordingDescriptor](https://github.com/cryostatio/cryostat-core/blob/main/src/main/java/org/openjdk/jmc/rjmx/services/jfr/IRecordingDescriptor.java) does not. The information [is available over JMX](https://docs.oracle.com/en/java/javase/12/docs/api/jdk.management.jfr/jdk/management/jfr/FlightRecorderMXBean.html) with [an accessor like the Recording one](https://docs.oracle.com/en/java/javase/12/docs/api/jdk.management.jfr/jdk/management/jfr/RecordingInfo.html#getSize()), but to use this we would need to update `-core`. This could be something to propose for JMC to add upstream, too.

In theory this would be a nice addition but in practice I don't think we should go for it, not right now.

src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
```diff
@@ -94,54 +99,69 @@ public List<GraphRecordingDescriptor> getAuthenticated(DataFetchingEnvironment e
...
+        return active;
+    }
+
+    static class Active {
+        List<GraphRecordingDescriptor> data;
+        AggregateInfo aggregate;
+    }
+
+    static class AggregateInfo {
+        Long count;
```
andrewazores, line 165: Can this be the primitive `long`?

src/test/java/itest/GraphQLIT.java
```diff
@@ -460,8 +461,11 @@ void testDeleteMutation() throws Exception {
         query.put(
                 "query",
                 "query { targetNodes(filter: { annotations: \"PORT == 9093\" }) { recordings {"
-                    + " active { name doDelete { name } } archived { data { name doDelete { name }"
-                    + " } aggregate { count } } } } }");
+                        + " active { data { name doDelete { name }"
+                        + " } aggregate { count } }"
+                        + " archived { data { name doDelete { name }"
+                        + " } aggregate { count } }"
+                        + " } } }");
```
andrewazores, line 468: Do the sample queries in https://github.com/cryostatio/cryostat/tree/main/docs/graphql also need updating similarly to this change?

andrewazores:
> (for me who doesn't know how to validate the SSL for 9095).

https://github.com/cryostatio/cryostatio.github.io/pull/70

+

https://github.com/andrewazores/vertx-fib-demo/blob/master/src/main/extras/app/resources/vertx-fib-demo.cer

src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
andrewazores, line 165:
```suggestion
        long count;
```

andrewazores (approved):

andrewazores (approved):

Issue #880, opened 2022-04-06
[Task] Machine-processable automated analysis results endpoint


andrewazores:
Depends on https://github.com/cryostatio/cryostat-core/issues/128

The machine-processable automated analysis results should be exposed in the HTTP API as an endpoint's JSON response.

maxcao13:
Is it better to make new classes like `[Target]ReportEvalGetHandler` and do that or to add some option in the existing handlers if returning JSON, (as as form, path, or query) in the old v2 `[Target]ReportGetHandler`s?

maxcao13:
When do I know if I should update the v1 vs. v2 handlers or both? In my mind, this is a newer feature so it would make more sense to just update the v2 ReportGetHandler and leave the v1 version as an older version however I notice that that v2 version is not documented in the HTTP API yet.

andrewazores:
> When do I know if I should update the v1 vs. v2 handlers or both?

I guess that kind of decision falls on me :-)

In the case of the JWT-consuming V2 handlers, those are the ones primarily used by the web-client, and their V1 equivalents would still be preferred for programmatic clients. So for this feature, both should be updated.

> I notice that that v2 version is not documented in the HTTP API yet.

Ah...

Yea, that's something else that should be fixed.

maxcao13:
I'm looking at [vert.x docs ](https://vertx.io/docs/vertx-web/java/#_routing_based_on_mime_types_acceptable_by_the_client)and I see that the way they do produces and consumes is by using the route. 

The code is what I tried in the `ReportGetHandler` but I don't think it works like that.
`
        ctx.currentRoute().produces("application/json").produces("text/html");
        String acceptableContentType = ctx.getAcceptableContentType();
`
I've also tried just manually parsing the `ctx.headers.accept()` but I'm not sure if there is a cleaner way to do this directly in each handler.

So should I be changing stuff in [WebServer.java ](https://github.com/maxcao13/cryostat/blob/main/src/main/java/io/cryostat/net/web/WebServer.java) then? Like changing the specific routes `[Target]ReportGetHandler` to produce both `application/json` and `text/html`?

maxcao13:
If a client adds a header "Accept: image/webp" but the `[Target]ReportGetHandler` doesn't accept this type, what should happen? Right now it is simply returning a 404 HTTP Error message, but I'm wondering if a bad accept should just return the default content (html), as this is what most websites do (though I realize this endpoint is not a website).

[...]

PR #1072, opened 2022-09-16
feat(webserver): implement content negotiation and deprecated endpoints

### LINKED ISSUE #1016: [Story] RequestHandlers should implement "produces" for Content-Type negotiation
> Should I be supporting multiple Accept headers like `Accept: text/html;application/json;*/*` ? I was originally going to but I realized since this is going to be an internally used endpoint, there wouldn't be much reason to if we just document that you must have at most 1 Accept header to each request.
> 
> How it is right now, if there are mutliple, 406 is thrown since we are looking at the raw acceptHeader.
> 
> And I notice that when I press "View Report" on an ActiveRecording, it will send add an AcceptHeader like `  text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8` because I'm guessing the default behaviour of Accept headers in the `downloadReport` function in the cryostat-web Api.service.
> 
> What should happen here?
> 
>     1. Support multiple headers using `ctx.parsedHeaders().accept()` and do a bunch of manual parsing by checking q values and having an implicit order #(text/html always priority over application/json (or something))` (I don't think vert.x has a way of doing this easily)
> 
>     2. Change the way the downloadReport function works with headers only.
> 
>     3. something else...?

I think Vert.x does support this:

https://vertx.io/docs/vertx-web/java/#_routing_based_on_mime_types_acceptable_by_the_client

but we would need to set up the router in `WebServer` to use `.produces()` on route definitions, so that's out of scope for this particular PR. Ideally we do fully support this, but that's a larger change to make later on.

I'd rather not have requests to this endpoint outright fail if multiple `Accept`s are given, or if `Accept` has multiple listed content types. What do you think of the following behaviour?

1. Check the `Accept` header(s)
2. If there is exactly one, and its value is a known supported type, respond with that
3. In any other scenario, respond with HTML

This way if the client didn't provide an `Accept`, or provided an invalid value, or provided multiple, or provided the weighted list, we just respond with a default HTML formatted doc (for now). This suits our own frontend's needs if we can set the `Accept` header there manually/explicitly to only HTML or only JSON, and also allows some reasonable degree of flexibility as a public API with particular constraints. If we have time to come back around and properly support content type routing then we can fix this endpoint up to fully support the content type negotiation semantics and `Accept` header values.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/1009#issuecomment-1175053670_

andrewazores:
Fixes #1016

redefine existing mimeType() method as produces() which provides a List of HttpMimeType. Existing implementations simply wrap their old return values into a singleton List. The default implementation is an empty list, which means that the webserver ignores content negotiation for such handlers (ie. v1 handlers). For compliant handlers (all v2+), content negotiation will be used if the client specifies an Accept request header. If no Accept request header is given then the behaviour is unchanged from before. If the Accept header is given then the handler will only be called if an acceptable content type can be negotiated, falling back to the general 'GET /'-style handler if negotiation fails. There are no such examples yet, but any handlers which produce() multiple mime types should use ctx.getAcceptableContentType() to check what format they should respond with.

The basic hook for consumes() on the other end of content negotiation is also included, but there are no implemented examples.


andrewazores:
All of the actual meaningful change in this PR is contained within WebServer, RequestHandler, and AbstractV2RequestHandler classes/interfaces. The rest are essentially all redefining `mimeType()` to `produces()`.

andrewazores:
To test, pick a v2 endpoint like `/api/v2.1/discovery` and try the following:

```bash
$ sh smoketest.sh
$ # in another terminal do:
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" # should get a JSON response
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" Accept:text/plain # should get the web-client index.html because content negotiation failed
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" Accept:application/json # should get a JSON response
```

The content negotiation behaviour of falling back to the web-client index.html doesn't seem great at the moment, so I'm looking into how to make that into a `406` or `415` response instead.

andrewazores:
Generally fixed in the last commit.

```bash
$ sh smoketest.sh
$ # in another terminal do:
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" # should get a JSON response
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" Accept:text/plain # should get a 406 because content negotiation failed (handler only produces application/json)
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" Accept:application/json # should get a JSON response
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" Accept:application/* # should get a JSON response
$ https :8181/api/v2.1/discovery Authorization:"Basic $(echo user:pass | base64)" Accept:*/* # should get a JSON response
$ https :8181/api/v2.1/nosuchhandler # 404
$ https :8181/api/v2.1/nosuchhandler Accept:application/json # should be a 404 but currently is also a 406
```

andrewazores:
Not sure why the itests keep hanging in CI. They run fine on my machine...

maxcao13 (commented):
Everything looks good, thanks for doing this!

I'm wondering about vert.x `consumes` since you've included it in the PR along with `produces` - is including consumes info in relevant endpoints something that should be done separately?

andrewazores:
I can add it on here as well if you'd like. It's only really relevant for `POST` endpoints, I think, so there aren't too many.

andrewazores:
As an example, I modified the `AuthTokenPostHandler` with:

```java
    @Override
    public List<HttpMimeType> consumes() {
        return List.of(HttpMimeType.MULTIPART_FORM, HttpMimeType.URLENCODED_FORM);
    }
```

Then we have the following behaviours:

Negotiation failure when POSTing JSON:
```
$ http -v :8181/api/v2.1/auth/token Authorization:"Basic $(echo user:pass | base64)" resource=/api
POST /api/v2.1/auth/token HTTP/1.1
Authorization: Basic dXNlcjpwYXNzCg==
Connection: keep-alive
Content-Length: 20
Host: localhost:8181
User-Agent: HTTPie/3.2.1

{
    "resource": "/api"
}


HTTP/1.1 415 Unsupported Media Type
content-length: 0
content-type: text/plain
vary: origin
```

Success when POSTing urlencoded form:
```
 $ http -v -f :8181/api/v2.1/auth/token Authorization:"Basic $(echo user:pass | base64)" resource=/api
POST /api/v2.1/auth/token HTTP/1.1
Accept: */*
Authorization: Basic dXNlcjpwYXNzCg==
Connection: keep-alive
Content-Length: 15
Content-Type: application/x-www-form-urlencoded; charset=utf-8
Host: localhost:8181
User-Agent: HTTPie/3.2.1

resource=%2Fapi

HTTP/1.1 200 OK
content-encoding: gzip
content-length: 498
vary: origin

{
    "data": {
        "result": {
            "resourceUrl": "/api?token=eyJjdHkiOiJKV1QiLCJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIn0..wXBKbx-uEILvAtmH.gkEYMmDvuT_ok5RjWBSV_gXY3vf2x_R_W3-VMFgSH4bxVy8cjzULlEiUeUjwm3AT1GrymiNymLJxSakT3l1jJAXOL3vWSA_iTkb-N87_UwLcuBicGpSPxnDPblS2ZshTIxfRTzY2fLeZwQs9CRNAFpa_5ZmwAW13BwG_X95rOyqjz9N3-wHBfRTcO5Cf6AdEWI-0PPhqAtvpERbQGqZ83CC6I5vGjMyGH90dY1BGvS6XRbiRJyVhaS6VIXXseOXw9-yJ8cm4RbwgLGpXwQPgBgoYLu9EYuQZVr6qQNX3ou4XTQYg9YKMZeWteNSNwAJi8PGZUGl63HkaJzCoCdJwwTXPA5NsFKqFvSGLWi0gQp7TeruHPydVJetRNlEAlpw.7dGZQJNOI1sGpLrU66FVfg"
        }
    },
    "meta": {
        "status": "OK",
        "type": "application/json"
    }
}
```

and success when POSTing multipart form:
```
$ http -v --multipart :8181/api/v2.1/auth/token Authorization:"Basic $(echo user:pass | base64)" resource=/api
POST /api/v2.1/auth/token HTTP/1.1
Accept: */*
Authorization: Basic dXNlcjpwYXNzCg==
Connection: keep-alive
Content-Length: 131
Content-Type: multipart/form-data; boundary=3093fe0a8c374db5b59bd7fbc1ba38bc
Host: localhost:8181
User-Agent: HTTPie/3.2.1

--3093fe0a8c374db5b59bd7fbc1ba38bc
Content-Disposition: form-data; name="resource"

/api
--3093fe0a8c374db5b59bd7fbc1ba38bc--


HTTP/1.1 200 OK
content-encoding: gzip
content-length: 498
vary: origin

{
    "data": {
        "result": {
            "resourceUrl": "/api?token=eyJjdHkiOiJKV1QiLCJlbmMiOiJBMjU2R0NNIiwiYWxnIjoiZGlyIn0..CskNInMd_NXqdkRb.Ap94JzWLlM9xAsp08MdA9S3Av2ik9Cv40wrAjLr6uJaIbh_iWSEUvXbL9SYK2NFUjaV7EVaVvqLG4vqhbeXicmVfEwfLDsZlTx5GoRKxQFAhlTktVeU86MbhnyQK02Xupx5FXSYc4yJPunyQS10HoA1Kt0P62QZcWdxTRKBkvAKiTM38C2hbeaNnSP5Y8Qn3QSJPB7NN4aSteRwIpS_KrtguCaH9ZDiqe5TN0u2LkIZng77lMwhDnv7ccCQazqnyIJvQrXoya71C8fesIL7C3eW9RE94PbsymTepFS1Td8o3eGLbeVs7KxuNL6pexH-Or1Pjaw3ZsEcAZp0YCMSkZy_2RULldW9kZfwHfTH25gHlA4hMJL9dFvDBAt2CEW8.L8nkHxFLYlgR_ErMxU88Yg"
        }
    },
    "meta": {
        "status": "OK",
        "type": "application/json"
    }
}
```

src/main/java/io/cryostat/net/web/http/api/v1/TemplatesPostHandler.java
```diff
@@ -114,6 +115,11 @@ public boolean isAsync() {
         return false;
     }
 
+    @Override
```
maxcao13, line 118: Is there a reason this is the only v1 endpoint that has content negotiation?

src/test/java/io/cryostat/net/web/http/api/v2/RuleGetHandlerTest.java
```diff
@@ -106,8 +107,9 @@ void shouldHaveExpectedRequiredPermissions() {
         }
 
         @Test
-        void shouldHavePlaintextMimeType() {
-            MatcherAssert.assertThat(handler.mimeType(), Matchers.equalTo(HttpMimeType.JSON));
+        void shouldProducePlaintext() {
```
maxcao13, line 110: shouldProduceJson ?

src/main/java/io/cryostat/net/web/http/api/v1/TemplatesPostHandler.java
andrewazores: No, I think I probably did that one by accident. There's no reason that v1 handlers can't or shouldn't also implement content negotiation, though. I'll back this one out for now and handle all the v1 ones together in a follow-up commit to this branch tomorrow.

maxcao13 (approved):
LGTM! I will probably resume to https://github.com/cryostatio/cryostat/issues/880 after the next release.

PR #1079, opened 2022-09-22
test(smoketest): publish different HTTP ports for each fib demo app

### LINKED ISSUE #1078: [Task] `smoketest.sh` publishes all `vertx-fib-demo` on same HTTP port

andrewazores:
Fixes #1078

### DIFF
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -73,7 +73,7 @@ function runDemoApps() {
 
     podman run \
         --name vertx-fib-demo-2 \
-        --env HTTP_PORT=8081 \
+        --env HTTP_PORT=8082 \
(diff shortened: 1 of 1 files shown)


maxcao13:
What's the reason for this change? I'm not too familiar with using the HTTP ports on those apps.

andrewazores:
Just to be sure you can actually interact with each of the three applications separately. They all implement a little application that computes Fibonacci numbers and exposes them on a very simple web app that you can reach through these ports. They should all be separate so that you can properly open each web app to compute Fibonacci numbers, and separately start recordings on each with their separated JMX ports too.

maxcao13 (approved):

PR #1083, opened 2022-09-27
fix(credentials): use database

### LINKED ISSUE #1082: [Task] Use database for JMX credentials

andrewazores:
Fixes #1082

- fix(credentials): keep JMX credentials in memory
- remove credentials from metadata test
- remove credentials from JVM ID test
- update UUID field definition for postgres/hibernate5 compat
- replace in-memory "store" with db connection
- clean up dao and deletion query
- correct unit tests
- set JMX DB password in itest
- Revert "clean up dao and deletion query"
- Revert "Revert "clean up dao and deletion query""
- use bouncycastle algo

This is still not an ideal way to handle credentials. The target applications' JMX servers either use no auth or use Basic auth, and Cryostat needs to be able to talk to JMX servers "unattended" (for things like automated rules), so this necessitates some form of cached or stored authorization. The implementation has used stored credentials for a while now. This change improves the situation by adding a new table to the Discovery Database (now just a general purpose database) (see https://github.com/cryostatio/cryostat/issues/937) and using that for storage rather than manually managed JSON files dumped to disk. It also improves the situation by encrypting the credentials while they are at rest, and we don't hold them long in memory, so the attack surface/window is a bit narrower. The secret used to encrypt the credentials is provided by the user, so the goal is that when deployed in Kubernetes this will be generated by the Operator as a Secret and supplied to the Cryostat instance in a consistent way.

Users can still use Cryostat without using any stored credentials whatsoever while keeping JMX authentication enabled on their applications. This means automated rules cannot be used, and the current version of the web-client will not be useful[0]. But, a user can make API requests including our old `X-JMX-Authorization` header and include the JMX auth credentials there to accompany every request made to a target application. These header credentials are completely transient and only in Cryostat's memory as long as necessary to open the JMX connection. This way the user can still start, stop, download, and analyze JFR using Cryostat's HTTP API.

[0] older versions of the web-client would cache the JMX credentials locally within the user's browser and provide those credentials with API requests via the aforementioned `X-JMX-Authorization` header. Keeping the credentials in browser memory is also not good. Maybe we should re-implement something on the browser client side to always prompt for JMX credentials as needed rather than automatically making a request to store them on the backend.


src/main/java/io/cryostat/configuration/ConfigurationModule.java
```diff
@@ -77,36 +70,24 @@ static Path provideConfigurationPath(Logger logger, Environment env) {
...
+                matchExpressionEvaluator,
+                discovery,
+                dao,
+                notificationFactory,
+                logger);
+    }
+
+    @Provides
+    @Singleton
+    static StoredCredentialsDao providePluginInfoDao(EntityManager em, Logger logger) {
```
maxcao13, line 90: Should be `provideStoredCredentialsDao`?

src/main/java/io/cryostat/configuration/CredentialsManager.java
```diff
@@ -67,137 +60,61 @@
...
+                return id;
             }
         }
-        throw new FileNotFoundException();
+        throw new IllegalArgumentException();
     }
 
-    public Credentials getCredentialsByTargetId(String targetId)
-            throws JsonSyntaxException, JsonIOException, IOException, ScriptException {
+    public Credentials getCredentialsByTargetId(String targetId) throws ScriptException {
```
maxcao13, line 117: One problem that I was sometimes running into was that during my work on the Archives and JvmMapping PRs, on the `MetadataManager` startup, the code tests connections to targets to see if active recordings are still valid and uses `Credentials` to access them through JMX. But sometimes since the task was async, the `platformClient` is not started fast enough to `listDiscoverableServices` and that's why sometimes there are `Credential` errors. So it seems the `DiscoveryStorage` should be loaded first but the reason I changed the verticles to start in order was because the `TargetJvmDiscovery` ws notifications should happen after the `MetadataManager` startup. Maybe to solve this, if possible add a different `getCredentialsByTargetId` that only uses the Dao to query without `listDiscoverableServices` but not having to pass an entire `serviceRef`?

src/main/java/io/cryostat/net/web/http/api/v2/CredentialDeleteHandler.java
```diff
@@ -136,11 +134,8 @@ public IntermediateResponse<Void> handle(RequestParameters params) throws ApiExc
                     .send();
 
             return new IntermediateResponse<Void>().statusCode(200);
-        } catch (FileNotFoundException e) {
+        } catch (IllegalArgumentException e) {
```
maxcao13, line 137: What's the rationale here for throwing the unchecked exception and catching it?

andrewazores: Cleaned up in latest commit.

src/main/java/io/cryostat/configuration/ConfigurationModule.java
andrewazores: Thanks, fixed.

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: I'm not sure if I follow. I think this would mean adding a method that basically checks for credentials defined with a match expression specifically like `target.connectUrl == some-specific-targetId`? That's conceptually different from what the current implementation does, which is looking for a target that matches the targetId given and then checking if there are any matchExpression-based credentials that fit the corresponding ServiceRef - and they can be matching based on any property of the ServiceRef, not only the connectUrl. Without using a ServiceRef object or listDiscoverableServices or any other information from the DiscoveryStorage I'm not sure how we can query for any stored credentials that match a target by connectUrl alone. It would only work for certain credentials that are only defined for one specific target.

If I'm understanding the problem correctly it seems that the root cause is that the MetadataManager wants to do its pruning immediately at startup based on what it can see on disk, but it needs to cross-reference those against discovered targets from the DiscoveryStorage. Depending on what discovery plugins are active, the DiscoveryStorage may end up reloading an empty state and waiting for async network processes to occur before repopulating the database - this is always what will happen with the DefaultPlatformClient/JDP for example. If you're running in OpenShift with the KubernetesApiPlatformClient we can mitigate this by ensuring the storage is repopulated with a startup Kubernetes API query so that the storage isn't considered as "started" until after we have loaded target data into it. This isn't currently done but it should be easy enough to add.

Another angle on a fix would be for the MetadataManager to check what data is has on disk and queue up jobs to validate that data. When a matching target does appear then the corresponding job is executed and we connect to the target to check if the recordings we expect are still there. If the target never appears within some time interval then we assume it's never going to and we clear the data instead. This way the metadata prune is a little more lazily evaluated instead of eagerly and it is more tolerant of delayed target reappearances, which are never entirely avoidable.

andrewazores: Anyway, that was very long-winded. We can have a meeting to talk through those details and I can try to draw up some illustrations to depict the data flow and dependencies as I'm picturing it.

maxcao13: > I'm not sure if I follow. I think this would mean adding a method that basically checks for credentials defined with a match expression specifically like `target.connectUrl == some-specific-targetId`? That's conceptually different from what the current implementation does, which is looking for a target that matches the targetId given and then checking if there are any matchExpression-based credentials that fit the corresponding ServiceRef - and they can be matching based on any property of the ServiceRef, not only the connectUrl. Without using a ServiceRef object or listDiscoverableServices or any other information from the DiscoveryStorage I'm not sure how we can query for any stored credentials that match a target by connectUrl alone. It would only work for certain credentials that are only defined for one specific target.
> 
Ah, I see then I was thinking about this differently than how it actually is.

> If I'm understanding the problem correctly it seems that the root cause is that the MetadataManager wants to do its pruning immediately at startup based on what it can see on disk, but it needs to cross-reference those against discovered targets from the DiscoveryStorage. Depending on what discovery plugins are active, the DiscoveryStorage may end up reloading an empty state and waiting for async network processes to occur before repopulating the database - this is always what will happen with the DefaultPlatformClient/JDP for example. Other mechanisms like Custom Targets and KubernetesApiPlatformClient are able to load their plugin information to the discovery database on startup (Custom Targets by just reloading what was already in the DB, Kubernetes by performing a k8s API query for Endpoints and building the subtree) so I think they're already less susceptible to this class of problem.
> 
> Another angle on a fix would be for the MetadataManager to check what data is has on disk and queue up jobs to validate that data. When a matching target does appear then the corresponding job is executed and we connect to the target to check if the recordings we expect are still there. If the target never appears within some time interval then we assume it's never going to and we clear the data instead. This way the metadata prune is a little more lazily evaluated instead of eagerly and it is more tolerant of delayed target reappearances, which are never entirely avoidable. This way the MetadataManager and CredentialsManager can both start up early, the MetadataManager does not have an immediate direct coupling/dependency on credentials being available (which can sometimes require the DiscoveryStorage to be ready too), and then we can start the DiscoveryStorage after. Once the DiscoveryStorage starts emitting notifications that's what triggers the MetadataManager to start validating its findings, but at this point the DiscoveryStorage is obviously already ready, which means it's also ready for the CredentialsManager to reference it. I think any kind of discovery plugin, even ones that "trickle" new target updates like JDP, would work in this scheme since the metadata validation is done on-demand as each target appears.
> 

That makes sense and solves the problem, sure that can be changed in a new issue I will make.

> The last hole in this scheme is that for platform mechanisms that don't "trickle" and simply repopulate fully on startup (like custom targets or kubernetes), you don't get any notifications for those targets. If the MetadataManager does not start until after the DiscoveryStorage has completed startup though, then this case should be handled by just having the MetadataManager call `listDiscoverableServices` on its own startup and hook that into the same validation job queue processing as the target discovery events go into as described above.

The MetadataManager completes before the DiscoveryStorage starts however, so this would not work unfortunately. Seems like a way to do this would just to send a callback to the MetadataManager when the DiscoveryStorage is ready to do the queue processing job again, but I'm wondering about a way to do this without, like you said, a direct dependency on credentials and the DiscoveryStorage to be ready.

maxcao13: > Anyway, that was very long-winded. We can have a meeting to talk through those details and I can try to draw up some illustrations to depict the data flow and dependencies as I'm picturing it.

I think it makes sense to me, if something gets too confusing I will let you know, thanks!

andrewazores: > The MetadataManager completes before the DiscoveryStorage starts however, so this would not work unfortunately. 

Does it need to? If the metadata validation was done lazily then the MetadataManager could be started after the DiscoveryStorage. Both instances could be constructed, with the MetadataManager registering a listener for target discovery events and add them to its queue of observed targets. Then the DiscoveryStorage can be started and allowed to complete, then the MetadataManager can be started. The MetadataManager scans its storage and creates a map of `connectUrl` to metadata-to-be-validated, then calls `listDiscoverableServices` once and adds those results to the queue of observed targets. Then it starts a task that takes elements from the observed targets queue and checks if it matches an entry in the validation map. If it does, it connects to the target and checks if the stored metadata is still accurate, performing the migration as needed and removing the validation map entry. If the target from the queue doesn't match anything in the validation map just skip it. Then, separately, drop entries from the validation map if they become too old and stale.

I think key data structures to use here are probably a Caffeine cache for the validation map and something like a BlockingQueue for the observed targets queue. The target discovery callback would place elements into queue with `add()`, and the metadata validation tasks would remove elements from the queue using a loop and `take()`. Once the validation map is empty the validation task can terminate itself.

maxcao13: Right, I was thinking about on startup, at least on the JDP side of things, the BuiltInDiscovery which is part of the DiscoveryStorage is a verticle which sends notifications that the MetadataManager wont be able to see if it starts afterwards, but I see that if we just have the MetadataManager start after the DiscoveryStorage, then a call to `listDiscoverabeServices` should cover those notifications we missed anyways. 

> I think key data structures to use here are probably a Caffeine cache for the validation map and something like a BlockingQueue for the observed targets queue. The target discovery callback would place elements into queue with add(), and the metadata validation tasks would remove elements from the queue using a loop and take(). Once the validation map is empty the validation task can terminate itself.

Sure I'll look into it, looks like its back in the saddle with these types of issues. :-)

andrewazores: Makes sense. Yea, the MetadataManager can register a target discovery listener and begin receiving updates before it itself actually goes through Verticle startup. That plus the `listDiscoverableServices` upon startup should ensure that the MetadataManager doesn't miss any targets.

maxcao13:
image

Not sure if this is intended, but if you add credentials to targets with match expression e.g. `target.alias == 'es.andrewazor.demo.Main'`, any automated rules set on those will start recordings on them, even if they didn't need credentials i.e. vertx.fib.demo 9093. In this case, the active recording that was already started due to an automated rule was replaced.

maxcao13:
However using this with the web front-end PR feels good and the credentials are correctly saved/not saved in the database depending on the setting.

andrewazores:
> Not sure if this is intended, but if you add credentials to targets with match expression e.g. target.alias == 'es.andrewazor.demo.Main', any automated rules set on those will start recordings on them, even if they didn't need credentials i.e. vertx.fib.demo 9093. In this case, the active recording that was already started due to an automated rule was replaced.

This isn't exactly "intended", but it's an expected outcome (and not a result of this PR). When a rule activates on a target it will try to start a new active recording with a name reflecting the rule's name, using the rule's defined event template. If there is already a recording with that same name then it is restarted. Thanks to @lkonno's recent work, adding credentials also causes rules to re-trigger for those targets matched by the credentials, since you may have been adding credentials to fix a rule that was originally created and should have triggered on a target where credentials were missing. There is no special case handling for targets that match the credential but do not need credentials. It should be a pretty contrived scenario in practice, and I think this result is generally harmless.

> However using this with the web front-end PR feels good and the credentials are correctly saved/not saved in the database depending on the setting.

That PR and this one are actually independent of each other though obviously related to the same concepts. You should be able to use either of the PRs separately and both would work as expected, since this one changes an implementation detail of how the credentials are stored, and the web one changes a detail of how the frontend sends JMX credentials to that storage.

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: Oh, one more thought along that line. There might be a good way to use the Vertx eventbus here to do the target discovery communication between storage and metadata manager. Rather than forking off a separate worker thread or something, the metadata manager could just listen for discovery events over the event bus, and the storage can publish those events on the bus. This way the verticles' worker threads handle that async job nicely already. This might change the scheme of registering an early listener hook before starting the verticle however - it might mean the metadata manager actually has to start first so it can register its event bus listener before the storage starts and potentially begins to publish events. The storage could also publish a message indicating that its startup has completed, and the metadata manager can listen for that message and use that to trigger calling `listDiscoverableServices`.

maxcao13: Yeah makes sense, I'll try stuff out and see what works and what doesn't, thanks for the ideas!

maxcao13 (approved):
Looks good to me!

ebaron (dismissed):
Looks good. We'll need to generate credentials for this in the operator and Helm chart.

maxcao13 (approved):

ebaron:
> Looks good. We'll need to generate credentials for this in the operator and Helm chart.

I filed https://github.com/cryostatio/cryostat-operator/issues/465 and https://github.com/cryostatio/cryostat-helm/issues/37.

src/main/java/io/cryostat/configuration/CredentialsManager.java
```diff
@@ -134,70 +159,29 @@ public static String targetIdToMatchExpression(String targetId) {
...
+        StoredCredentials saved = dao.save(new StoredCredentials(matchExpression, credentials));
         emit(CredentialsEvent.ADDED, matchExpression);
-        return nextId++;
+        return saved.getId();
     }
 
-    public int removeCredentials(String matchExpression)
-            throws IOException, MatchExpressionValidationException {
+    public int removeCredentials(String matchExpression) throws MatchExpressionValidationException {
+        // TODO refactor this to do a proper query on the DAO
```
tthvo, line 171: Is this still a `TODO`?

src/main/java/io/cryostat/configuration/CredentialsManager.java
```diff
@@ -206,22 +190,16 @@ public Credentials getCredentialsByTargetId(String targetId)
...
-            throws JsonSyntaxException, JsonIOException, IOException, ScriptException {
-        for (String pathString : this.fs.listDirectoryChildren(credentialsDir)) {
-            Path path = credentialsDir.resolve(pathString);
-            try (BufferedReader br = fs.readFile(path)) {
-                StoredCredentials sc = gson.fromJson(br, StoredCredentials.class);
-                if (matchExpressionEvaluator.applies(sc.getMatchExpression(), serviceRef)) {
-                    return sc.getCredentials();
-                }
+    public Credentials getCredentials(ServiceRef serviceRef) throws ScriptException {
+        for (StoredCredentials sc : dao.getAll()) {
```
tthvo, line 194: I am wondering if we could convert this to a query in DAO instead of iterating all rows? Not sure if there is any optimization in the database now but we could add an index on this field if it's frequently used?

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: Yes, I tried to solve this before but ran into some issue that I don't fully remember now. I *think* the problem was that doing the removal with a delete query in the DAO wouldn't return back the ID, so the return value from this method would need to change and that would change the behaviour of or break the original TargetCredentialsDeleteHandler.

The target-specific API handlers should be deprecated since they are now superseded by the matchExpression credentials API, so once those are removed we can actually just remove this method altogether and skip the TODO.

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: Index on which field? This method tests the stored credentials's match expression against the given ServiceRef. The match expression test can't be done in the database so I think we're pretty much stuck with querying for all rows.

tthvo: oh I mean the `matchExpression` field in the the StoredCredentials table. Something just similar to `deleteByMatchExpression` in DAO but for get?

src/main/java/io/cryostat/configuration/CredentialsManager.java
tthvo: Right, make senses! Thanks for explaining :D

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: No, it's not that easy unfortunately. `deleteByMatchExpression` deletes a row by String-equality on the match expression column. This method queries by semantic evaluation of the match eexpression, ie the match expression actually gets run as a piece of code.

andrewazores: Although now that you point it out, `deleteByMatchExpression` is actually the query we were talking about in the other conversation thread, and it isn't used anywhere because of that `TODO`. I'll remove it. I may as well mark the API handlers as deprecated while I'm at it.

tthvo: ohhh right! Okayyy I see now! Thanks for explaining again :D

tthvo (approved):
Looks great to me :D

andrewazores:
Rebased, deleted some dead code (https://github.com/cryostatio/cryostat/pull/1083#discussion_r988172550), and marked some methods and API handlers as deprecated.

andrewazores:
@tthvo re-review please, just the last commit since nothing else needed changing in the rebase. The only conflict was in `run.sh` because Max applied some shellcheck fixes and I added the `CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD` in the same location of the file.

tthvo:
Oh some ITest failed...

andrewazores:
> Oh some ITest failed...Probably we should change expected status code for deprecated routes in tests.

Oh, yea, I should have set `forRemoval=false` on the deprecation annotations so that the endpoints are still functional but some headers are added to help indicate to clients that they'll be going away later.

Issue #773, opened 2021-12-03
[Story] Enable/disable Automated Rules


andrewazores:
Currently, for a client to activate an Automated Rule, they must do `POST /api/v2/rules` to create the rule definition, which becomes effective immediately. To disable this rule they must do `DELETE /api/v2/rules/:ruleName`, which removes the rule definition and prevents it from activating in the future. The `?clean=true` query parameter can be included on the `DELETE` request to stop and remove all `ACTIVE` recordings that were created by the Automated Rule.

Would it also make sense to add an `enabled: boolean` state to Automated Rule definitions? This could be accompanied by a `PATCH /api/v2/rules/:ruleName` handler for modifying the `enabled` status. When a rule is first `POST`ed it may include `enabled: true` in the definition so that it is activated immediately. If `enabled: false` or if `enabled` is not included in the definition then the Automated Rule is still created but is left inactive, and the `RuleProcessor` will not apply the rule to any current or future targets until the rule is enabled.

This somewhat mimics the API for Active Recordings, where a Recording is in a `RUNNING` state when created and a `PATCH` request can be made to either snapshot or stop the recording. Stopping an Active Recording is somewhat analogous to disabling an Automated Rule.

The first advantage here is that end users would be able to define an Automated Rule and have it persisted in Cryostat's storage, but not necessarily to have it actually triggering at all times. This is perhaps simpler on the user's end than the `POST/DELETE` flow currently required, and would help to ensure that users don't accidentally `DELETE` a rule to disable it and then accidentally re-`POST` a different rule definition when trying to re-enable it.

The second advantage, which is perhaps more compelling, is that this would make it very easy and straightforward for Cryostat to ship with some basic and generic Automated Rule definitions out of the box as JSON file assets embedded in the container image, just like the `cryostat.jfc` Event Template. Some examples:

`monitoring.json`
```json
{
  "name":"monitoring",
  "description":"Enable the 'default' Continuous monitoring event template on all targets",
  "matchExpression": true,
  "eventSpecifier":"template=Continuous,type=TARGET",
  "enabled": false
}
```

`openshift-monitoring.json`
```json
{
  "name":"openshift-monitoring",
  "description":"Enable the 'default' Continuous monitoring event template on all targets with OpenShift annotation 'cryostat.io/monitoring: true`. Copy JFR data to archives every 5 minutes and maintain 1 hour of archives.",
  "matchExpression": "target.annotations.platform['cryostat.io/monitoring']==true",
[...]

PR #1084, opened 2022-09-27
feat(rules): enable/disable automated rules

lkonno:
Fixes [#773](https://github.com/cryostatio/cryostat/issues/773)

`http PATCH :8181/api/v2/rules/demo?enabled=false`

When a rule created disabled is enabled, the rule is activated and it starts recording.


andrewazores:
Looks mostly good from a reading. I haven't tested it out yet but judging by the test changes I think it would work as expected.

One change I would like to see is to use the request body for the `PATCH` rather than using a query parameter. This is a little more conventional and also matches how we stop or archive active recordings:

https://github.com/cryostatio/cryostat/blob/d55889e2266986b88c84c295f6b18240befcabd2/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingPatchHandler.java#L117

Here we get the request body as a simple String, which is either `"save"` (copy to archives) or `"stop"` (`RUNNING` active recording becomes `STOPPED`).

So rather than `PATCH /api/v2/rules/:ruleName?enabled=state` it would be simply `PATCH /api/v2//rules/:ruleName` with a request body content. I think the natural thing to do for the body content is to expect it as a JSON object like `{ "enabled": false }`. After parsing the JSON the rest of the request would be handled the same way you're currently doing it.

src/main/java/io/cryostat/rules/RuleRegistry.java
```diff
@@ -179,6 +179,16 @@ public void deleteRules(ServiceRef serviceRef) throws IOException {
         }
     }
 
+    public void enableRule(Rule rule, boolean enabled) {
+        String name = rule.getName();
+        if (enabled && !rule.isEnabled()) {
+            getRule(name).get().setEnabled(enabled);
+            emit(RuleEvent.ADDED, rule);
```
andrewazores, line 186: If I'm reading this correctly, the intent of this method is so that (re)enabling a rule, which can also happen when it's first created, emits an ADDED event. But disabling a rule does not emit an event. Enabling an already-enabled rule does not emit an event.

Would it make sense for this to be "symmetric" and also emit a REMOVED event when an enabled rule becomes disabled?

Or perhaps there can simply be an UPDATED/MODIFIED event too? Since the event contains the Rule itself there is no need for the event type to reflect the enabled state, that can just be read from the Rule itself.

andrewazores (commented):
It might be worth adding an integration test here. I think it can be pretty simple:

```
- create a rule definition with enabled=false
- start an external target container that matches the rule's matchExpression
- query the list of recordings on that target and assert that the list is empty
- patch the rule to enable it
- sleep a few seconds
- query the recording list again and assert that it has one recording with the expected name
```

src/main/java/io/cryostat/rules/RuleRegistry.java
lkonno: Created the UPDATED event where the rules are activated or deactivated when enabled/disabled.

lkonno:
> It might be worth adding an integration test here. I think it can be pretty simple:
> 
> ```
> - create a rule definition with enabled=false
> - start an external target container that matches the rule's matchExpression
> - query the list of recordings on that target and assert that the list is empty
> - patch the rule to enable it
> - sleep a few seconds
> - query the recording list again and assert that it has one recording with the expected name
> ```

Created a new test in the AutoRulesIT.

src/main/java/io/cryostat/rules/RuleProcessor.java
```diff
@@ -154,8 +154,10 @@ public void onEvent(Event<RuleEvent, Rule> event) {
                                                 .filter(
                                                         serviceRef ->
                                                                 registry.applies(
-                                                                        event.getPayload(),
-                                                                        serviceRef))
+                                                                                event.getPayload(),
+                                                                                serviceRef)
+                                                                        && event.getPayload()
```
andrewazores, line 159: Let's put the `isEnabled()` check before the `applies()` check. That way if the rule is disabled we can skip evaluating the match expression and save some CPU time.

src/main/java/io/cryostat/rules/RuleRegistry.java
```diff
@@ -179,9 +179,18 @@ public void deleteRules(ServiceRef serviceRef) throws IOException {
         }
     }
 
+    public void enableRule(Rule rule, boolean enabled) {
+        String name = rule.getName();
+        if ((enabled && !rule.isEnabled()) || (!enabled && rule.isEnabled())) {
```
andrewazores, line 184: Is this equivalent to `if (enabled != rule.isEnabled())`?

src/main/java/io/cryostat/rules/Rule.java
```diff
@@ -140,6 +142,14 @@ public int getMaxSizeBytes() {
         return this.maxSizeBytes;
     }
 
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public void setEnabled(boolean enabled) {
```
andrewazores, line 149: Can these two methods be package-private? It could trip someone up to see `setEnabled` as a visible method here, but calling it doesn't actually have the desired effect, which is really done by `RuleRegistry#enableRule`

src/main/java/io/cryostat/rules/RuleRegistry.java
lkonno: Yes >.<

src/main/java/io/cryostat/rules/Rule.java
lkonno: `RuleRegistry#enableRule` uses the `setEnable`. Do you think that it is fine to set to `protect` instead of `private`?

lkonno: ack, thanks!

andrewazores (approved):
Looks good, thanks!

PR #1088, opened 2022-09-30
fix(rules): correct archive scanning on startup

### LINKED ISSUE #1087: [Task] After restart, Automated Rules fail to trim old archived copies
Example reproduction:

1. Start `smoketest.sh`
2. Create a rule with the following definition:
```json
{
  "name": "myrule",
  "description": "",
  "matchExpression": "target.alias==\"es.andrewazor.demo.Main\"",
  "eventSpecifier": "template=Profiling,type=TARGET",
  "archivalPeriodSeconds": 60,
  "initialDelaySeconds": 60,
  "preservedArchives": 5,
  "maxAgeSeconds": 300,
  "maxSizeBytes": 0
}
```
3. Wait a few minutes for archived recordings to accumulate for the `vertx-fib-demo:9093` sample target. The archives should accumulate up to 5 copies, after which the oldest one gets removed when a new one is added.
4. `Ctrl-c` the `smoketest.sh`
5. Restart `smoketest.sh`
6. The rule should re-activate on the sample target and begin accumulating more archived copies. However, the archive will accumulate 6, 7, ... 10 copies before trimming, at which point the 6th will be removed to make space for the 11th, then the 7th for the 12th, etc.
7. Repeatedly stopping and restarting `smoketest.sh` adds another sliding window of 5 copies on top

andrewazores:
Fixes #1087

I think this bug occurred in #1047 , looking at context like this: https://github.com/cryostatio/cryostat/pull/1047/files#diff-763906edf56f8525d84844f8e2b01a98e15bf4afe7cb959159b3a378e448fd17L275

but I haven't bisected the commit history to verify this. Just a hunch, but the fix seems simple enough anyway.

### DIFF
--- a/src/main/java/io/cryostat/rules/PeriodicArchiver.java
+++ b/src/main/java/io/cryostat/rules/PeriodicArchiver.java
@@ -37,9 +37,9 @@
  */
 package io.cryostat.rules;
 
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayDeque;
 import java.util.List;
+import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


maxcao13 (approved):
Makes sense, probably should've realized the PeriodicArchiver would have been affected as well. LGTM

tthvo (approved):
Looks good to me!

andrewazores:
> Makes sense, probably should've realized the PeriodicArchiver would have been affected as well. LGTM

No worries, I should've caught it too.

PR #1089, opened 2022-09-30
feat(graphql): add jvmId to ServiceRef and expose id to GraphQL queries

### LINKED ISSUE #1076: [Task] Expose `jvmId` through GraphQL query in target metadata
@maxcao13 in doing this I have just noticed that the `jvmId` comes through in the HTTP REST API `metadata`, but it isn't available to query via GraphQL. Would you mind taking a look at that?

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/964#issuecomment-1252439523_

maxcao13:
Fixes #1076 
Depends on #1105


maxcao13:
I'm having trouble resolving a testing error where basically:

I have a JvmIdHelper instance in the discovery `AbstractNodeTypeAdapter` since the read method creates a `ServiceRef` object. But during tests where `Gson` is not mocked and tries to read EnvironmentNodes, the read method inside the `AbstractNodeTypeAdapter` will be called and will eventually call this line
```
jvmId = jvmIdHelper.get().getJvmId(connectUrl.toString());
```
And since it's not mocked, the call will go through everything and eventually call a CredentialsManager which runs into a RuntimeException 
```
java.lang.RuntimeException: java.nio.file.AccessDeniedException: /opt/cryostat.d/conf.d/credentials
        at io.cryostat.configuration.ConfigurationModule.provideCredentialsManager(ConfigurationModule.java:109)
...
...
```
Ideas on how I should resolve this?

andrewazores:
It sounds like the tests should use a real Gson that is provided with mock type adapters, so that the type adapter(s) can also return mocks or at least return controlled instances of things like ServiceRefs.

There is already a method for using Gson in that way:

https://github.com/cryostatio/cryostat/blob/88321b810eff0ccd72b6c6c85592bdf0ad11f332/src/main/java/io/cryostat/MainModule.java#L145

but it may be a little cumbersome to use, since you have to add type adapters for everything Gson might encounter in the tests that are not standard types it supports out of the box. In this case it sounds like that would probably just be the `AbstractNodeType` adapter, though, so not too bad.

maxcao13:
Thoughts? Here's some things fixed, thoughts, and questions:

1. The jvmId isn't being persisted to the db because it messes up the gson/json validation of the jvmId field for the smoketest plugins being registered. Should that be changed?
2. Fixed a bug where a connectUrl FileNotFound exception is thrown if a target is discovered but didn't already have a archives folder created for it.
3. The mocking of the gson Adapters is a little ugly but it works, is there a better way to refactor it?
4. JvmIds are calculated during runtime so if integration tests need to reference these, I made a new helper class called JvmIdWebRequest to help with that. 
5. And the original GraphQL bug is fixed, a query like this can be used
```
 query {
    targetNodes {
        name
        nodeType
        labels
        target {
            jvmId
            alias
            serviceUri
        }
    }
}
```
6. And I removed the ability for the metadata to include the jvmId inside when using REST http, since that was a bug.

src/main/java/io/cryostat/discovery/AbstractNodeTypeAdapter.java
```diff
@@ -190,10 +197,18 @@ public AbstractNode read(JsonReader reader) throws IOException {
                         }
                     }
                     reader.endObject();
-                    target = new ServiceRef(connectUrl, alias);
+
+                    // To prevent Gson recursion, we need to manually construct the ServiceRef with
+                    // a null jvmId and use that ServiceRef object to get the jvmId with correct
+                    // credentials without calling platformClient.listDiscoverableServices() which
+                    // would recurse through this adapter again to getJvmId
+                    target = new ServiceRef(null, connectUrl, alias);
```
andrewazores, line 205: Hmm. Generally these deserializers and adapters just read the JSON content and build the right object instance directly out of that. In this case I would expect that either the jvmId is ignored on deserialization and we lazy-initialize it later on as needed (not during JSON deserialization), or the jvmId is taken at face value if it is in the serialized JSON form and we just assume that's correct. I haven't thought too deep into this but I would tend toward the lazy-init solution. This way we aren't just blindly trusting that the hash ID supplied by clients is correct, but we also aren't doing a heavyweight network operation during a simple JSON deserialization.

I'm imagining something like a method on the JvmIdHelper:

```java
public ServiceRef resolveId(ServiceRef sr) {
  if (sr.getJvmId() != null) return sr;
  // use tcm to compute ID
  return new ServiceRef(id, ... );
}
```

This might be annoying to use though because almost anywhere that has a `ServiceRef` would need to then also have the `JvmIdHelper` and try to resolve the IDs at every public method boundary.

Or maybe it's enough to handle that resolution inside the `DiscoveryStorage#update` method? Everything flows into the storage model through there, and everything else that queries it gets data flowing out from there.

andrewazores: Scrolling through the rest of the diff, I do see that the `JvmIdHelper` has been added to the `PlatformClient` implementations that hook into the `BuiltInDiscovery`. I think this can be refactored as I was explaining above and all that shared logic around resolving the hash ID for a ServiceRef can be pulled into `DiscoveryStorage`, rather than being handled in each platform client plus the custom targets API handler - they're all pushing that data into the same storage anyway.

If the hash ID resolution fails then the storage can examine the failure reason:
1. If it's a JMX SSL failure, just log a warning and complete the plugin update with a null ID for that ServiceRef
2. If it's a JMX Auth failure, same
3. If it's a network exception like the hostname specified in the connectUrl could not be resolved, or the connection timed out, etc. then fail the update. Either by removing that single element that could not be reached, or probably just by throwing an exception and letting that bubble up to whatever plugin tried to publish this update.

andrewazores:
> JvmIds are calculated during runtime so if integration tests need to reference these, I made a new helper class called JvmIdWebRequest to help with that.

Since the ID is now embedded in the `ServiceRef` it's visible by querying ex. `GET /api/v1/targets` as well as in GraphQL, and with a GraphQL query the client can get that structure specific to a single target too using the query filter. So maybe we can just drop the `JvmIdGetHandler`? If the `JvmIdWebRequest` is still useful for tests then it can use a GraphQL request under the hood instead.

maxcao13:
> > JvmIds are calculated during runtime so if integration tests need to reference these, I made a new helper class called JvmIdWebRequest to help with that.
> 
> Since the ID is now embedded in the `ServiceRef` it's visible by querying ex. `GET /api/v1/targets` as well as in GraphQL, and with a GraphQL query the client can get that structure specific to a single target too using the query filter. So maybe we can just drop the `JvmIdGetHandler`? If the `JvmIdWebRequest` is still useful for tests then it can use a GraphQL request under the hood instead.

Sure, I will remove the JvmIdGet handler and use GraphQL instead.

andrewazores:
I had the idea for something like a `GET /api/vx/targets/:targetId` handler for a long time, but I always struggled with what kind of information that should return (alias and connectUrl sure, but what else? Active recordings? Archives? And later, metadata?) so it never got implemented. That GraphQL query seems to be the solution I had wanted all along :-)

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -228,10 +235,31 @@ private Map<String, String> mergeLabels(
 
     public List<? extends AbstractNode> update(
             UUID id, Collection<? extends AbstractNode> children) {
+        var updatedChildren = new ArrayList<AbstractNode>();
+        for (AbstractNode child : Objects.requireNonNull(children, "children")) {
```
andrewazores, line 239: The subtree being pushed here may not be "flat", with the collection being a direct list of `TargetNodes`. It could be nested like in a k8s environment where it does something like namespace -> deployment -> replica -> endpoints -> serviceref. I think you should use `findLeavesFrom()` on each of the `children` here, then merge all of those results into a single `Collection` and then iterate over those and do the JVM ID retrieval.

maxcao13: Right, forgot about that. Thanks!

maxcao13: Just realized something, I would have to mutate the actual ServiceRef objects themselves such that the structure of the children isn't messed up when calling `dao.update()`, is that okay to do or is there a better way?

andrewazores: You mean to remove the invalid ones from the tree, where the JVM ID retrieval fails?

maxcao13: I mean I think if I did something like this
```
    public List<? extends AbstractNode> update(
            UUID id, Collection<? extends AbstractNode> children) {
        var updatedChildren = new HashSet<AbstractNode>();
        for (AbstractNode child : findAllLeavesFrom(Objects.requireNonNull(children, "children"))) {
            if (child instanceof TargetNode) {
                ServiceRef ref = ((TargetNode) child).getTarget();
                 ref = jvmIdHelper.get().resolveId(ref);
                child = new TargetNode(child.getNodeType(), ref, child.getLabels());
            }
            updatedChildren.add(child);
        }
...
        plugin = dao.update(id, Objects.requireNonNull(updatedChildren));
```
I don't think it would work since now the updateChildren is flattened again, so if I wanted to update the jvmIds along the way, I would have to, iterate the children leaves again but **not** make a new Collection and modify the leaves' ServiceRef objects directly with something like TargetNode.setJvmId(). Then this way we can just update the dao as normal with children
`plugin = dao.update(id, children);` since the structure wasn't flattened. Another way to do would I guess to reconstruct the tree structure as we are iterating so that mutations are not needed.

andrewazores:
I've marked this as dependent on #1105. It may not be directly so, but I think these two are very likely to conflict with each other and need some rebasing work. I think #1105 is higher priority to get merged before release than this is since that's more of a foundational performance fix, and this is a nice-to-have feature that mostly dovetails in with 2.3 planning more so than required for 2.2.

maxcao13:
Not sure why itests fail here, they work locally...

maxcao13:
Ah, looks like another CI problem, I think. The repeated-integration-tests and mvn verify try to pull the image with tag `2.3.0-snapshot` https://github.com/cryostatio/cryostat/actions/runs/3316176363/jobs/5478379389#step:7:419, but as we know when loading the image with podman load, the tags aren't preserved so I think it actually pulls from quay. That's annoying.

andrewazores:
Ah right, makes sense. I think `mvn -Dcryostat.imageVersion=latest verify` would force itests to run with the `latest` tagged image, for example, so it should be pretty easy to modify the `repeated-integration-tests.bash` and CI configs to override the tagged image as needed. I'll have to do that in #1133 as well to set the PR-scoped CI image to be the one tested in the same workflow, I think.

maxcao13:
Well, since tests pass now, looks like #1153 worked. :-)

src/test/java/itest/util/http/JvmIdWebRequest.java
```diff
@@ -0,0 +1,115 @@
...
+                "query",
+                String.format("query { targetNodes(filter: { name: \"%s\" }) { ", targetId)
+                        + " target { jvmId } } }");
+        webClient
+                .post("/api/v2.2/graphql")
+                .putHeader(
+                        "X-JMX-Authorization",
+                        "Basic "
+                                + Base64.getUrlEncoder()
+                                        .encodeToString("admin:adminpass123".getBytes()))
```
andrewazores, line 80: The basic credentials should probably be passed into this method so that the calling test method can specify what the required passthrough credentials are, if any.

src/test/java/itest/util/http/JvmIdWebRequest.java
```diff
@@ -0,0 +1,115 @@
...
+    // String.format("service:jmx:rmi:///jndi/rmi://%s:9091/jmxrmi", Podman.POD_NAME)
+    public static String jvmIdRequest(String targetId)
+            throws InterruptedException, ExecutionException, TimeoutException {
+        CompletableFuture<TargetNodesQueryResponse> resp = new CompletableFuture<>();
+
+        JsonObject query = new JsonObject();
+        query.put(
+                "query",
+                String.format("query { targetNodes(filter: { name: \"%s\" }) { ", targetId)
+                        + " target { jvmId } } }");
```
andrewazores, line 73: Can the whole query string be combined as a single string and the `.format()` applied on that, rather than applying formatting to the first half and concatenating the last half?

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -148,7 +157,14 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
                 }
             }
             Map<AnnotationKey, String> cryostatAnnotations = new HashMap<>();
-            ServiceRef serviceRef = new ServiceRef(uri, alias);
+
+            String jvmId = null;
+            try {
+                jvmId = jvmIdHelper.getJvmId(uri.toString());
+            } catch (JvmIdGetException e) {
+                logger.warn("Couldn't put jvmId in serviceRef for {}", uri.toString());
```
andrewazores, line 165: I think this should actually result in an error response, probably in the `4xx` range.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -226,12 +234,51 @@ private Map<String, String> mergeLabels(
...
+            Collection<? extends AbstractNode> children) {
+        List<AbstractNode> modifiedChildren = new ArrayList<>();
+        for (AbstractNode child : children) {
+            if (child instanceof TargetNode) {
+                ServiceRef ref = ((TargetNode) child).getTarget();
+                try {
+                    ref = jvmIdHelper.get().resolveId(ref);
+                } catch (Exception e) {
+                    logger.warn("Failed to resolve jvmId for node {}", child.getName());
+                    logger.info(e.getCause().toString());
```
andrewazores, line 247: Likewise, I'm also hesitant about allowing these to pass and storing them into the database with null JVM IDs. Since we're already reaching out to the target to get the ID we may as well take the opportunity to be defensive about the data model and reject updates that attempt to publish targets we don't think are valid.

maxcao13: Mmm... what if the error is of SecurityException i.e. invalid credentials or ssl error? 

> Scrolling through the rest of the diff, I do see that the `JvmIdHelper` has been added to the `PlatformClient` implementations that hook into the `BuiltInDiscovery`. I think this can be refactored as I was explaining above and all that shared logic around resolving the hash ID for a ServiceRef can be pulled into `DiscoveryStorage`, rather than being handled in each platform client plus the custom targets API handler - they're all pushing that data into the same storage anyway.
> 
> If the hash ID resolution fails then the storage can examine the failure reason:
> 
>     1. If it's a JMX SSL failure, just log a warning and complete the plugin update with a null ID for that ServiceRef
> 
>     2. If it's a JMX Auth failure, same
> 
>     3. If it's a network exception like the hostname specified in the connectUrl could not be resolved, or the connection timed out, etc. then fail the update. Either by removing that single element that could not be reached, or probably just by throwing an exception and letting that bubble up to whatever plugin tried to publish this update.

You said it's fine before, so I am just confirming because I think what I've already written accounts for that. Unless I'm missing something.

andrewazores: You're right, what you've written does implement the behaviour I described before, but I'm second guessing what I said...

Give me a bit of time to think this through and figure out which way I want it to go. On the one hand, the SSL/Auth failures tell us that something really is there but we just aren't able to complete the task, so it seems valid to allow the update and include this in the discovery database. On the other hand, the ID is meant to be a unique identifier for an instance of a target, so nullable IDs seem likely to break things in unexpected ways in the future.

I'm also wondering to do about IDs for non-JMX Cryostat Agents. I've been thinking that such a Cryostat Agent can register itself with a `connectUrl` starting with `cryostat-agent://` rather than the typical `service:rmi:jmx://`. Cryostat would not be able to connect to the agent over JMX, but in this case we can assume and assert that the same agent must also be registered as a Discovery plugin (publishing only itself during the update we're processing now), which means we can enforce that the `connectUrl` has the `cryostat-agent` URI scheme rather than `http`, but uses the same host/IP:port as it used for its `callbackUrl`. But, since we want everything to have IDs and we can't necessarily connect back to these agents over JMX to retrieve the ID, we would need to allow the agent to also supply its own ID.

Or maybe we implement an alternate `JFRConnection` that is actually a facade overtop of HTTP connections to an Agent, and we implement a subset of capabilities on the Agent's integrated webserver. One of the supported operations could be retrieving the ID.

andrewazores:
Re-triggering test runs.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
maxcao13: > SSL/Auth
Hmm... not sure but maybe just preventing them from being added to the db may be better since I assume if we can't connect to the target for the ID because of the those errors, then we aren't able to perform any actions anyways on them? I'm wondering about the potential user cases for if someone doesn't provide SSL or something, but still wants to persist that target to the db...

> non-JMX Cryostat Agents
I think I like having the alternate JFRConnection over the other method because we wouldn't have to force extra functional compatibility from outside agents, then the fallback is always just some http attributes about the agent and maybe some other id given by the agent itself -> and making some id out of that, perhaps. So sort of both ideas at once. The cryostat-agent URI scheme maybe can then be used to identify using between the two different implementations to choose which "JFRConnection" to use.

andrewazores: > Hmm... not sure but maybe just preventing them from being added to the db may be better since I assume if we can't connect to the target for the ID because of the those errors, then we aren't able to perform any actions anyways on them? I'm wondering about the potential user cases for if someone doesn't provide SSL or something, but still wants to persist that target to the db...

It might be useful for the user to still be able to see that they exist but are not connectable - we do that right now in Cryostat anyway, you can see the SSL-enabled or JMX-Auth-enabled targets in the API responses or web UI, and you just get a prompt when you try to perform an action on them if the SSL cert is untrusted or there are no good auth credentials stored. If we drop those targets from plugin publishes or reject the update entirely then the targets would never appear in the API responses or UI and the user would lose that ability to interactively provide credentials via X-JMX-Authorization passthrough headers etc. It'd be a pretty fundamental change to the interactive use cases for Cryostat. It seems like the "right", pure way to handle things, but it might just be too overly defensive to feel understandable to end users.

I think I'm leaning toward keeping the behaviour that I outlined before and which you've implemented here. Be specifically lenient about the SSL/Auth failure scenarios and allow null IDs only in that case, which indicates a "known" target that we are unable to connect to. if the problem was the SSL cert then it's taken care of when the user adds the cert and restarts Cryostat. If the problem was JMX Auth then the nulled ID should eventually get updated when something triggers Cryostat to retry connecting to that target. We wouldn't have that behaviour right now, though, because the null ID would be there from the time of the plugin update and would remain there until something caused the DiscoveryStorage to re-evaluate that target. We would need some hook to update those IDs in the DiscoveryStorage database after the fact, for example if the credentials are added late and an automated rule is triggered and causes Cryostat to successfully connect to the target.

Starting to look like we'll need that broken out relationship between Targets in the database as their own entities, and the discovery subtrees referencing those. This would allow us to update the JVM ID on a Target record individually, without having to figure out which discovery plugin subtree provided the target, update the one node, rebuild the tree, and store the subtree again.

>  think I like having the alternate JFRConnection over the other method because we wouldn't have to force extra functional compatibility from outside agents, then the fallback is always just some http attributes about the agent and maybe some other id given by the agent itself -> and making some id out of that, perhaps. So sort of both ideas at once. The cryostat-agent URI scheme maybe can then be used to identify using between the two different implementations to choose which "JFRConnection" to use.

Agreed, I think this imaginary "CryostatAgentConnection" that can do some subset of the usual JFRConnection things is probably the best way to go. I think we could pretty easily implement this within Cryostat by hooking it into the `TargetConnectionManager` and checking the URI scheme. If it's `cryostat-agent://` then we build a CryostatAgentConnection, otherwise we fall back to the current behaviour. The CryostatAgentConnection would just fire off HTTP requests to a small handful of API endpoints exposed by the Cryostat Agent to allow for things like retrieving a JVM ID (which could just be a UUID generated by the Cryostat Agent's JVM and held for the lifecycle), and maybe in the future also for listing, starting/stopping/deleting, and retrieving active flight recordings.

maxcao13: > It might be useful for the user to still be able to see that they exist but are not connectable - we do that right now in Cryostat anyway, you can see the SSL-enabled or JMX-Auth-enabled targets in the API responses or web UI, and you just get a prompt when you try to perform an action on them if the SSL cert is untrusted or there are no good auth credentials stored. If we drop those targets from plugin publishes or reject the update entirely then the targets would never appear in the API responses or UI and the user would lose that ability to interactively provide credentials via X-JMX-Authorization passthrough headers etc. It'd be a pretty fundamental change to the interactive use cases for Cryostat. It seems like the "right", pure way to handle things, but it might just be too overly defensive to feel understandable to end users.
> 

Ah okay, that makes sense. 

>  without having to figure out which discovery plugin subtree provided the target, update the one node, rebuild the tree, and store the subtree again.
> 

Funny thing is that I've done this and added the hook for credentials being added to trigger a DiscoveryStorage.updateNode call in this pr already :-) . But it is definitely not the most efficient. Having a Target database like the the Credentials and Plugins may be nice...

> Agreed, I think this imaginary "CryostatAgentConnection" that can do some subset of the usual JFRConnection things is probably the best way to go. I think we could pretty easily implement this within Cryostat by hooking it into the `TargetConnectionManager` and checking the URI scheme. If it's `cryostat-agent://` then we build a CryostatAgentConnection, otherwise we fall back to the current behaviour. The CryostatAgentConnection would just fire off HTTP requests to a small handful of API endpoints exposed by the Cryostat Agent to allow for things like retrieving a JVM ID (which could just be a UUID generated by the Cryostat Agent's JVM and held for the lifecycle), and maybe in the future also for listing, starting/stopping/deleting, and retrieving active flight recordings.

Sounds good and makes sense to me, what should be the plan for now then?

andrewazores: > Funny thing is that I've done this and added the hook for credentials being added to trigger a DiscoveryStorage.updateNode call in this pr already :-) . But it is definitely not the most efficient. Having a Target database like the the Credentials and Plugins may be nice...

Oh nice. We can probably move forward with what you already have then, so long as we get to the Target database stuff before the next release since it should be a rather large real-world performance improvement to do it that way.

> Sounds good and makes sense to me, what should be the plan for now then?

I have a bunch of other stuff on my plate for now so I'll come back around and re-review this PR. The next step after would be to pick up work on the Cryostat Agent again. Its implementation of a Discovery Plugin is pretty well done but could still probably use a bit of polishing.

The next thing it should have is the JFR Harvester that collects JFR data and periodically pushes it to Cryostat. That's already working to an extent, but we need a good way to identify this agent push as a sort of sourceTarget so that the data can go into the archives corresponding to the specific agent instance. This does tie in to what we're talking about with the CryostatAgentConnection since this could be what we use to pull an ID from the agent, and the agent can also use that as its sourceTarget path parameter when it's pushing JFR files, so maybe we implement the CryostatAgentConnection part first and then finish off work on the JFR Harvester part last.

andrewazores:
@maxcao13 rebase this please and hopefully I'll get a chance to review and test it tomorrow or Friday. The null ID handling you have is fine for now as discussed above.

maxcao13:
Just a warning before testing, the changes will mess up the database if you checkout another branch after testing since this PR adds jvmIds to the ServiceRef objects which get pushed to the db. The only way I know to solve it is to just delete `h2.mv.db` afterwards, but it removes any credentials as well.

andrewazores:
So upgrading works seamlessly, downgrading is broken? I think that's fine.

andrewazores:
```
$ http -v :8181/api/v2.2/graphql Authorization:"Basic $(echo user:pass | base64)" query="{ targetNodes { target { serviceUri jvmId } name } }"
POST /api/v2.2/graphql HTTP/1.1
Authorization: Basic dXNlcjpwYXNzCg==
Connection: keep-alive
Content-Length: 65
Host: localhost:8181
User-Agent: HTTPie/3.2.1

{
    "query": "{ targetNodes { target { serviceUri jvmId } name } }"
}


HTTP/1.1 200 OK
content-encoding: gzip
content-length: 386
vary: origin

{
    "data": {
        "targetNodes": [
            {
                "target": {
                    "jvmId": "amLkDPDb0Po8sTz_UddB5HvTtGtjqgDxNPVhlwrtE44=",
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:10000/jmxrmi"
                }
            },
            {
                "target": {
                    "jvmId": null,
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi"
                }
            },
            {
                "target": {
                    "jvmId": "V4jM9Fu5lvvZDEJ2hQyL5aDerHqUW4a-5pe_bvc9juM=",
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9093/jmxrmi"
                }
            },
            {
                "target": {
                    "jvmId": null,
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9094/jmxrmi"
                }
            },
            {
                "target": {
                    "jvmId": null,
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9095/jmxrmi"
                }
            },
            {
                "target": {
                    "jvmId": "VkDn4nD-Ouaz8Z8gOjGcfA91F9p1ymO45mQ_b0NLdRE=",
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9096/jmxrmi"
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9097/jmxrmi",
                "target": {
                    "jvmId": "hwRtIOihg_Q36nnGc_YK7T6JidWqn6GHjwl8z387dSI=",
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9097/jmxrmi"
                }
            },
            {
                "name": "service:jmx:rmi:///jndi/rmi://cryostat:9197/jmxrmi",
                "target": {
                    "jvmId": "yFQJdbDielDysrNYIisKWsgmysfm5JSc9DpqU7sI9hg=",
                    "serviceUri": "service:jmx:rmi:///jndi/rmi://cryostat:9197/jmxrmi"
                }
            }
        ]
    }
}
```

Looking generally good. However, the `9091` is Cryostat itself as usual, and I have JMX credentials stored in the keyring for Cryostat. The `jvmId: null` here seems unexpected.

andrewazores:
> Just a warning before testing, the changes will mess up the database if you checkout another branch after testing since this PR adds jvmIds to the ServiceRef objects which get pushed to the db. The only way I know to solve it is to just delete h2.mv.db afterwards, but it removes any credentials as well.

Also on that note, rather than running and then deleting the `h2.mv.db` every time, may as well just do `sh smoketest.sh h2mem`.

maxcao13:
> Looking generally good. However, the `9091` is Cryostat itself as usual, and I have JMX credentials stored in the keyring for Cryostat. The `jvmId: null` here seems unexpected.

Not sure, why but I would get ConcurrentModificationExceptions when adding credentials to 9091, so I just replace the listener set to a concurrent one, and it seems to work...

andrewazores:
Interesting, I guess I can see how that could make sense.

Might a CopyOnWriteArraySet be better suited? It would be a very small difference either way, I think.

andrewazores:
Working very nicely from manual testing. I'll get back to code review next week.

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -165,6 +176,8 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
                 throw new ApiException(400, "Duplicate connectUrl");
             }
             return new IntermediateResponse<ServiceRef>().body(serviceRef);
+        } catch (JvmIdGetException e) {
+            throw new ApiException(404, "Couldn't put connect to target: " + e.getTarget());
```
andrewazores, line 180: "Couldn't put connect"?

andrewazores (approved):

PR #1090, opened 2022-09-30
fix(metadata): change RecordingMetadataManager startup process

### LINKED ISSUE #1085: [Task] Allow MetadataManager to start metadata validation without relying on DiscoveryStorage
If I'm understanding the problem correctly it seems that the root cause is that the MetadataManager wants to do its pruning immediately at startup based on what it can see on disk, but it needs to cross-reference those against discovered targets from the DiscoveryStorage. Depending on what discovery plugins are active, the DiscoveryStorage may end up reloading an empty state and waiting for async network processes to occur before repopulating the database - this is always what will happen with the DefaultPlatformClient/JDP for example. Other mechanisms like Custom Targets and KubernetesApiPlatformClient are able to load their plugin information to the discovery database on startup (Custom Targets by just reloading what was already in the DB, Kubernetes by performing a k8s API query for Endpoints and building the subtree) so I think they're already less susceptible to this class of problem.

Another angle on a fix would be for the MetadataManager to check what data is has on disk and queue up jobs to validate that data. When a matching target does appear then the corresponding job is executed and we connect to the target to check if the recordings we expect are still there. If the target never appears within some time interval then we assume it's never going to and we clear the data instead. This way the metadata prune is a little more lazily evaluated instead of eagerly and it is more tolerant of delayed target reappearances, which are never entirely avoidable. This way the MetadataManager and CredentialsManager can both start up early, the MetadataManager does not have an immediate direct coupling/dependency on credentials being available (which can sometimes require the DiscoveryStorage to be ready too), and then we can start the DiscoveryStorage after. Once the DiscoveryStorage starts emitting notifications that's what triggers the MetadataManager to start validating its findings, but at this point the DiscoveryStorage is obviously already ready, which means it's also ready for the CredentialsManager to reference it. I think any kind of discovery plugin, even ones that "trickle" new target updates like JDP, would work in this scheme since the metadata validation is done on-demand as each target appears.

The last hole in this scheme is that for platform mechanisms that don't "trickle" and simply repopulate fully on startup (like custom targets or kubernetes), you don't get any notifications for those targets. If the MetadataManager does not start until after the DiscoveryStorage has completed startup though, then this case should be handled by just having the MetadataManager call `listDiscoverableServices` on its own startup and hook that into the same validation job queue processing as the target discovery events go into as described above.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/pull/1083#discussion_r981906589_

maxcao13:
Fixes #1085

### DIFF
--- a/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
@@ -84,6 +84,8 @@ public class DiscoveryStorage extends AbstractPlatformClientVerticle {
     private final Logger logger;
     private long timerId = -1L;
 
+    public static final String DISCOVERY_STARTUP_ADDRESS = "discovery-startup";
+
     DiscoveryStorage(
             VerticleDeployer deployer,
             Duration pingPeriod,
@@ -108,7 +110,8 @@ public class DiscoveryStorage extends AbstractPlatformClientVerticle {
                         cf ->
                                 deployer.deploy(builtin.get(), true)
                                         .onSuccess(ar -> future.complete())
-                                        .onFailure(t -> future.fail((Throwable) t)))
+                                        .onFailure(t -> future.fail((Throwable) t))
+                                        .eventually(m -> getVertx().eventBus().send(DISCOVERY_STARTUP_ADDRESS, "Discovery storage deployed")))
                 .onFailure(future::fail);
 
         this.timerId = getVertx().setPeriodic(pingPeriod.toMillis(), i -> pingPrune());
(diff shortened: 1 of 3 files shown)


andrewazores:
Nice, pretty compact change but this already ensures that migration doesn't happen until after discovery startup is completed, and also should push the migration work off onto a verticle worker thread.

Do you intend to also implement the lazy evaluation stuff we talked about previously or do you think this change is already sufficient?

maxcao13:
> Nice, pretty compact change but this already ensures that migration doesn't happen until after discovery startup is completed, and also should push the migration work off onto a verticle worker thread.
> 
> Do you intend to also implement the lazy evaluation stuff we talked about previously or do you think this change is already sufficient?

I think it's sufficient enough. As it is right now, I think the notifications that come in which the Manager hooks into already sort of allows it to do lazy evaluation along with the eager startup metadata pruning, so in fact maybe there is no need to do startup pruning at all but I guess it doesn't hurt to have redundancy in case notifications aren't seen correctly somehow.

andrewazores (approved):

PR #1091, opened 2022-10-02
chore(build): add shellcheck script checking to build

### LINKED ISSUE #1040: [Task] Use `shellcheck` to analyze bundled scripts
There are currently two shell scripts included in the Cryostat image:
- https://github.com/cryostatio/cryostat/blob/main/src/main/extras/app/entrypoint.sh
- https://github.com/cryostatio/cryostat/blob/main/src/container/include/truststore-setup.sh

These should have `shellcheck` (or some equivalent linting tool) applied to them as part of the build process, the same way we use `spotless` and `spotbugs` on the Java sources.

There are other shell scripts around for development use like `run.sh`/`smoketest.sh`, `devserver.sh`, `generate-dev-certs.sh`, and the `repeated-*-tests.bash` which would also benefit from this.

Bonus chore: update or just remove `build-images.sh`.

maxcao13:
Fixes #1040 

`mvn shellcheck:check`

Any problems with configuration or disagreements with the shellcheck analysis on our .sh scripts before I resolve them?


maxcao13:
Also, not sure why yarn testing is failing all of a sudden...

tthvo:
I think the tests for `DatePicker` component tried to mock the system time with jest but it didn't mock it as expected. Now, it's October so the tests fail. I will look into it :D

tthvo:
See

> I think the tests for `DatePicker` component tried to mock the system time with jest but it didn't mock it as expected. Now, it's October so the tests fail. I will look into it :D

See https://github.com/cryostatio/cryostat-web/pull/531 for fix. Sorry about that :D

andrewazores:
Config looks good, and shellcheck's reports on `mvn shellcheck:check` look good. Some of the problems it presents may not really be fixable by simply cleaning up some styling, ex. it points out that we use arrays in `entrypoint.sh` but POSIX sh doesn't support arrays. So this section of the script either needs some rewriting, or the script should actually indicate that `bash` or some other similar shell should be used instead.

maxcao13:
Modified everything instead of specifying some other shell to be used, hopefully functionality stays the same. There actually was a variable that wasn't being used correctly and shellcheck was able to show that.

The only thing I suppressed was the usage of local https://www.shellcheck.net/wiki/SC3043.

src/container/include/truststore-setup.sh
```diff
@@ -2,16 +2,16 @@
 
 set -e
 
-function genpass() {
-    echo "$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)"
+genpass() {
+    printf '%s' "$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)"
```
andrewazores, line 6: I guess this whole line can probably be replaced with just `< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32` and strip out the rest. It should naturally print to stdout this way, no need to wrap it into a string and format that to print with printf/echo.

src/main/extras/app/entrypoint.sh
```diff
@@ -2,31 +2,31 @@
...
+    printf "| %-40s |\n" "$(date)"
     echo   "|                                          |"
     printf "| %-40s |\n" "$@"
     echo   "+------------------------------------------+"
 }
 
-function genpass() {
-    echo "$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)"
+genpass() {
+    printf '%s' "$(< /dev/urandom tr -dc _A-Z-a-z-0-9 | head -c32)"
```
andrewazores, line 14: same here re. removing printf

maxcao13:
I think there's somewhere in the script which messed up from my changes and that's why the build is failing, I will try to see what's wrong.

maxcao13:
Sort of confused why the only problem is the converting from an array to a really long FLAGS variable and passing the flags into the 
```
exec java \
    "${FLAGS}" \
    -cp "$CLASSPATH" \
    @/app/jib-main-class-file \
    "$@"
```

maxcao13:
Problem was that I was storing options as a variable but that's not allowed as seen here https://unix.stackexchange.com/questions/459367/using-shell-variables-for-command-options. I don't think there's a good way to solve this without manually passing the options in to `entrypoint.sh`. Using the positional parameter array '@' also doesn't work since we still need to pass the the actual arguments in the same command. Should this just stay ignored, should we change the script to .bash, or is there some fix that I haven't thought about?

andrewazores:
There is a `bash` in the base image:

```
podman run --rm -it --entrypoint=/bin/bash cryostat:latest
```

so I think we're safe to switch the entrypoint over. Rename the filename extension and change the shebang line and I think that's all that's needed.

andrewazores:
I'd probably revert any changes to `entrypoint.sh` to this point, then change it over to using Bash, and then re-apply shellcheck. Some of the things changed/removed under the assumption of POSIX sh might be perfectly acceptable using Bash. Things like removing the extraneous `echo`s in `genpass` can stay changed, though.

repeated-integration-tests.bash
```diff
@@ -64,12 +64,9 @@ while [ "${runcount}" -lt "${runs}" ]; do
     timestamp="$(date -Iminutes)"
     client_logfile="$DIR/target/${POD_NAME}-${timestamp}.client.log"
     server_logfile="$DIR/target/${POD_NAME}-${timestamp}.server.log"
-    mkdir -p "$(dirname $client_logfile)"
-    mkdir -p "$(dirname $server_logfile)"
-    >"${client_logfile}"
```
maxcao13, line 69: I'm wondering what these redirections were for before? Shellcheck said they were redirecting to nowhere and removing them didn't seem to change any functionality, but I'm not sure if there was some subtle change.

andrewazores: It redirects an empty stream into the named file. This has the effect of creating the file if it doesn't already exist, or if it did exist then it truncates its contents to the empty stream's contents.

```bash
$ cat foo
cat: foo: No such file or directory
$ >foo
$ cat foo
$ echo hello > foo
$ cat foo
hello
$ >foo
$ cat foo
$ rm foo
$ >foo
$ cat foo
```

andrewazores (commented):
Looks like the permissions on the entrypoint script changed:

` 100755 → 100644 `

Not sure if this matters, it may depend on the platform that's running the image? Best to keep it the same as it was.

repeated-integration-tests.bash
maxcao13: Should I revert it?

andrewazores: I think `mkdir` lines would only do anything if you tried to `bash repeated-integration-tests.bash` without a `$cryostat_dir/target` directory existing, ie you haven't run a build before. The empty redirection would be very unlikely to ever do anything because the file it targets has the current timestamp in its filename, and the later bits that write logs to that file should be fine with creating the file and appending to it without it already existing.

So I guess it's okay to remove them, but it seems harmless to keep these checks even if shellcheck thinks they're ineffective.

andrewazores (commented):
Seems good but I'll wait for ebaron's feedback too.

pom.xml
```diff
@@ -833,6 +834,44 @@
...
+        <failBuildIfWarnings>true</failBuildIfWarnings>
+        <binaryResolutionMethod>download</binaryResolutionMethod>
+        <releaseArchiveUrls>
+          <Linux-amd64>
+              https://github.com/koalaman/shellcheck/releases/download/v0.8.0/shellcheck-v0.8.0.linux.x86_64.tar.xz
+          </Linux-amd64>
+          <Mac_OS_X-x86_64>
+              https://github.com/koalaman/shellcheck/releases/download/v0.8.0/shellcheck-v0.8.0.darwin.x86_64.tar.xz
+          </Mac_OS_X-x86_64>
+        </releaseArchiveUrls>
```
ebaron, line 863: Is there a reason for using the download method instead of the [embedded](https://github.com/giraciopide/shellcheck-maven-plugin#how-it-works) method? The embedded method might make our downstream builds easier since there's a proxy involved.

ebaron (commented):
This is a pretty cool tool. I just have one build related comment.

pom.xml
maxcao13: No real reason other than that it was the default. Will change, thanks for the feedback!

andrewazores (approved):

ebaron (approved):

andrewazores:
Wonder what's the holdup on CI. Maybe GH infra is just overloaded right now?

maxcao13:
> Wonder what's the holdup on CI. Maybe GH infra is just overloaded right now?

Seems so. https://www.githubstatus.com/incidents/smn1qtqvbsb6?utm_ts=1664982269

maxcao13:
Hmm... not sure what's the problem actually...

andrewazores (approved):

PR #1096, opened 2022-10-05
test(db): use h2:file db mode by default

### LINKED ISSUE #1094: [Task] Development and CI runs should use `h2:file` mode by default
`h2:file` is more representative of a real Cryostat deployment than `h2:mem` is, since persistence would generally be wanted for the JMX credentials keyring at least and also for the discovery database (especially custom targets). The Operator will deploy in `h2:file` mode when there is an associated PVC and storage space to put the file on:

https://github.com/cryostatio/cryostat-operator/issues/451

`h2:mem` should remain a configurable option and should also be easily selectable for development since it may also have its own uses for "ephemeral" Cryostat instances.

andrewazores:
Fixes #1094
Depends on #1083

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -514,6 +514,18 @@
               <argument>--mount</argument>
               <argument>type=tmpfs,target=/opt/cryostat.d/probes.d</argument>
               <argument>--env</argument>
+              <argument>JDBC_URL="jdbc:h2:file:/opt/cryostat.d/conf.d/h2;INIT=create domain if not exists jsonb as varchar"</argument>
+              <argument>--env</argument>
+              <argument>JDBC_DRIVER=org.h2.Driver</argument>
+              <argument>--env</argument>
+              <argument>JDBC_USERNAME=sa</argument>
+              <argument>--env</argument>
+              <argument>JDBC_PASSWORD=""</argument>
+              <argument>--env</argument>
+              <argument>HIBERNATE_DIALECT=org.hibernate.dialect.H2Dialect</argument>
+              <argument>--env</argument>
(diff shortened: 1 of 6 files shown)


maxcao13:
> Along with setting the `h2:file` mode to the default in `smoketest.sh` and the `pom.xml` itest setup, I also made a few minor changes to some itests themselves to disable credentials handling in the `AutoRulesIT`, at least temporarily. I'm not sure exactly why but that test is flaky with the interaction between credentials, target discovery, and rule activation, though in practice with manual testing I can never reproduce the failure.

Is there any useful logs of these runs that I may be able to look at?

andrewazores:
I didn't bother preserving any of the ones I observed locally, but this CI run looks like it's the same thing:

https://github.com/cryostatio/cryostat/actions/runs/3198632408/jobs/5223422453

maxcao13 (dismissed):
Hmm, that's weird - not too sure as of now without diving too deep, but anyways this PR looks good!

andrewazores:
Sorry, I just went back and checked on the JPA entity detachment thing and found that the change was no longer making any difference. It may have just seemed that way by coincidence yesterday when I was testing it since it was only affecting the flaky test(s) to begin with.

andrewazores:
I also included a change to the `repeated-integration-testing.bash` script to reduce the JDP polling times and not require multiple verifications of the count. Those default settings are overly cautious and cause tests to run a fair bit slower than really necessary, so using that development re-testing script to experiment with speeding JDP things up could be interesting.

maxcao13 (approved):
That makes testing a lot faster, right looks good!

Issue #1064, opened 2022-09-13
[Story] Option to automatically archive manually-created recordings on stop


andrewazores:
When creating a recording manually a user may specify various properties such as the data maxAge, maxSize, and the duration before the recording is automatically stopped by the target JVM.

When defining an Automated Rule a user may specify these same properties along with an archivalPeriod and a number of archived properties to preserve.

There is a middle ground option where a user may wish to create a recording manually one time for a specific profiling or troubleshooting purpose, but it's inconvenient to start the recording for a fixed duration and then have to wait around for that duration before proceeding to click the Archive button to copy the data into Cryostat's storage.

In the web-client UI there should be a checkbox or toggle like "Archive on Stop". This control should be disabled and ignored if the "Continuous" checkbox is marked. If "Archive on Stop" is marked then when the recording duration elapses, Cryostat should re-connect to the target JVM and copy the recording to archives. If the connection fails or the recording no longer exists then simply log the failure.

- [x] cryostatio/cryostat#1100
- [x] cryostatio/cryostat-web#541

tthvo:
I think we can close this one too?

andrewazores:
Closed, completed by #1100 and  https://github.com/cryostatio/cryostat-web/issues/541

PR #1097, opened 2022-10-05
feat(recording): option to automatically archive manually-created recordings on stop

lkonno:
Fixes [#1064](https://github.com/cryostatio/cryostat/issues/1064)

`curl --data "recordingName=foo2&duration=5&events=template=ALL&archiveOnStop=true" localhost:8181/api/v1/targets/localhost/recordings`


src/main/java/io/cryostat/jmc/serialization/HyperlinkedSerializableRecordingDescriptor.java
```diff
@@ -51,6 +51,7 @@ public class HyperlinkedSerializableRecordingDescriptor extends SerializableReco
     protected String downloadUrl;
     protected String reportUrl;
     protected Metadata metadata;
+    protected Boolean archiveOnStop;
```
andrewazores, line 54: I think this can be the primitive `boolean`

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
```diff
@@ -201,6 +201,14 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
                                                     attrs.get("metadata"),
                                                     new TypeToken<Metadata>() {}.getType());
                                 }
+                                boolean archiveOnStop = false;
+                                if (attrs.contains("archiveOnStop")) {
+                                    Pattern bool = Pattern.compile("true|false");
```
andrewazores, line 206: `Pattern` seems like overkill here :-)

`attr.equals("true") || attr.equals("false")` would do about the same, right?

andrewazores (commented):
Very cool! Just a few comments.

src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
lkonno: The `Patter` is being used in the `toDisk` `if` too, so I will replace it in both `if`s.

andrewazores:
Looks great, I'll test this out manually soon - just running some other unrelated integration tests at the moment.

andrewazores (approved):
Works perfectly!

Issue #1103, opened 2022-10-06
[Task] Automated Rule enabled/disabled state is not persisted


andrewazores:
When I create an Automated Rule I can set its "enabled" status. After creation I can also toggle the enabled status and this works as expected. However, if I shut down and restart Cryostat, the Automated Rule always comes back with the same enabled state as it was created with, not with the last state that it had before shutdown.

To reproduce, try creating a rule in the enabled state, which performs some noticeable action (matchExpression matches at least one target, starts a recording, maybe copies to archives frequently). Disable the rule. Then, shut down and restart Cryostat. Expected: the rule is disabled. Actual: the rule is back to being enabled.

PR #1107, opened 2022-10-07
fix(rules): persisting the rule enable/disable status

lkonno:
Fixes [#1103](https://github.com/cryostatio/cryostat/issues/1103)


src/main/java/io/cryostat/rules/RuleRegistry.java
```diff
@@ -183,6 +183,17 @@ public void enableRule(Rule rule, boolean enabled) {
         String name = rule.getName();
         if (enabled != rule.isEnabled()) {
             getRule(name).get().setEnabled(enabled);
+            Path destination = rulesDir.resolve(rule.getName() + ".json");
+            try {
+                this.fs.writeString(
```
andrewazores, line 188: Since this is pretty much the same thing that is done in `addRule()`, could you extract this to a helper method? Something like `private void persistRule(Rule rule)` probably, which just writes the JSON form of the rule to the corresponding path like we have here.

andrewazores (commented):
Looks great, just one small refactor request.

src/main/java/io/cryostat/rules/RuleRegistry.java
lkonno: Done!

andrewazores (approved):

PR #1110, opened 2022-10-11
feat(archives): add functionality to view archived files that have no recognizable targetId/serviceUri

### LINKED ISSUE #1109: [Task] add functionality to view archived files that have no recognizable targetId/serviceUri
Backend of https://github.com/cryostatio/cryostat-web/issues/546

maxcao13:
Fixes #1109 
Depends on https://github.com/cryostatio/cryostat/pull/1105

### DIFF
--- a/src/main/java/io/cryostat/net/reports/ArchivedRecordingReportCache.java
+++ b/src/main/java/io/cryostat/net/reports/ArchivedRecordingReportCache.java
@@ -71,6 +71,35 @@ class ArchivedRecordingReportCache {
         this.logger = logger;
     }
 
+    Future<Path> getFromPath(String subdirectoryName, String recordingName) {
+        CompletableFuture<Path> f = new CompletableFuture<>();
+        Path dest = null;
+        try {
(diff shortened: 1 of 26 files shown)


maxcao13:
There's definitely a lot of refactoring that should be done here...

andrewazores:
> There's definitely a lot of refactoring that should be done here...

This is my fault for missing the design flaw and having to put this high priority item on the release list so late in the cycle. It's definitely not on you for being the one to pick up the item.

> I didn't think it was right to use old handlers to do this, but there is a lot of refactoring for me to do later.

Correct, this stuff should rightfully be all its own separate sub-API. You made the right call adding them all under `beta` because they're all likely to change somewhat for the next release, or even be dropped altogether if we can find a better way to reconcile this model with the standard `:sourceTarget` API endpoints. For that reason I'm also not too picky about the implementation details here, so long as it achieves the goal and is fairly well self-contained.

andrewazores:
Looks good. Please add some basic unit tests and maybe a couple of integration tests and we're good to go.

andrewazores (approved):
Great work @maxcao13 , and thanks for taking it on and getting it done so quickly.

PR #1114, opened 2022-10-13
fix(discovery): leave Custom Targets enabled when builtins are disabled

### LINKED ISSUE #1113: [Task] `DISABLE_BUILTIN_DISCOVERY` env var should not disable Custom Targets
The env var should allow the user to disable the KubeApi, KubeEnv, or Default (JDP) discovery mechanisms. Custom Targets only "discovers" targets when the user explicitly adds them, does not have a conditionally-enabled UI, and in fact the POST API requests would even still respond with successful codes. This behaviour is inconsistent and also probably surprising to the user. If this env var is set, all other discovery mechanisms should be disabled but Custom Targets should remain normally functional.

andrewazores:
Fixes #1113

- use non-sa user by default for h2 setups
- fix(discovery): leave Custom Targets enabled when built-ins are disabled
- refactor to clean up instanceof check

`- use non-sa user by default for h2 setups` is not directly related to this PR but comes from discussions here:
https://github.com/cryostatio/cryostat-operator/pull/474#discussion_r994036680

In describing how things worked I realized I had left `sa` (system administrator) as the default JDBC user, which was not intentional and is not necessary. Even for the H2 database scenarios,
Cryostat should not require the use of a system administrator account.

To test, try `CRYOSTAT_DISABLE_BUILTIN_DISCOVERY=true sh smoketest.sh` before and after this PR. In either case, JDP targets should not be discovered, only the quarkus-test targets that self-publish via the discovery plugin API. With this PR applied you should also be able to define custom targets either through the web-client or by API request, whereas without this PR applied this functionality will be broken and defining a custom target will appear to be a no-op (200 success status code but not notification and no custom target visible in UI).

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ Cryostat can be configured via the following environment variables:
 * `CRYOSTAT_JDP_ADDRESS`: the JDP multicast address to send discovery packets. Defaults to `224.0.23.178`.
 * `CRYOSTAT_JDP_PORT`: the JDP multicast port to send discovery packets. Defaults to `7095`.
 * `CRYOSTAT_CONFIG_PATH`: the filesystem path for the configuration directory. Defaults to `/opt/cryostat.d/conf.d`.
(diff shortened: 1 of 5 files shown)


tthvo (approved):
Looks good to me :D

PR #1117, opened 2022-10-16
build(pom): set version to 2.2.0

andrewazores:
Related to #1118


ebaron (commented):
This should probably be something like `2.2.0-BETA` or something otherwise the CI will push a `2.2.0` version to Quay.

ebaron (approved):

PR #1126, opened 2022-10-18
feat(jmc-agent): add handler for retrieving available probe templates

Josh-Matsuoka:
Small PR to add a handler for retrieving currently available probe templates, this is needed for the Agent Plugin frontend. Also fixes an issue with the notification category for the delete handler and tweaks the way the TargetProbesGetHandler outputs the active probes for the frontend.

Keeping as Draft until I add tests for the new handler

Depends on https://github.com/cryostatio/cryostat-core/pull/156


src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateGetHandler.java
```diff
@@ -0,0 +1,157 @@
...
+                    .send();
+        } catch (Exception e) {
+            throw new ApiException(400, e.getMessage(), e);
+        }
+        return new IntermediateResponse<String>().body(probeTemplates.toString());
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return EnumSet.of(ResourceAction.DELETE_PROBE_TEMPLATE);
```
andrewazores, line 150: I think this should be `ResourceAction.READ_PROBE_TEMPLATE`, which looks like it needs to be added to the enum.

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateGetHandler.java
```diff
@@ -0,0 +1,157 @@
...
+        List<String> probeTemplates = new ArrayList<>();
+        try {
+            for (ProbeTemplate p : this.probeTemplateService.getTemplates()) {
+                probeTemplates.add(
+                        "{\"name\" : \""
+                                + p.getFileName()
+                                + "\" , "
+                                + "\"xml\" : \""
+                                + p.serialize().replaceAll("\n", "").replaceAll("\"", "")
+                                + "\"}");
```
andrewazores, line 132: Is this different from ex. `gson.toJson(p)`?

src/main/java/io/cryostat/net/web/http/api/v2/TargetProbesGetHandler.java
```diff
@@ -112,16 +116,25 @@ public IntermediateResponse<String> handle(RequestParameters requestParams) thro
...
                 connection -> {
                     connection.connect();
+                    List<String> response = new ArrayList<String>();
                     AgentJMXHelper helper = new AgentJMXHelper(connection.getHandle());
                     String probes = helper.retrieveEventProbes();
+                    if (probes != null && !probes.isBlank()) {
+                        ProbeTemplate template = new ProbeTemplate();
+                        template.deserialize(new ByteArrayInputStream(probes.getBytes()));
+                        for (Event e : template.getEvents()) {
+                            response.add(e.toString());
```
andrewazores, line 126: Can this handler return a `List<Event>` and allow the JSON serialization of the response (via Gson) to perform the output formatting, rather than calling `toString()` directly on each element here?

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateGetHandler.java
Josh-Matsuoka: It should, good catch, I've updated this now.

src/main/java/io/cryostat/net/web/http/api/v2/TargetProbesGetHandler.java
Josh-Matsuoka: Fixed this to use Gson to serialize.

src/main/java/io/cryostat/net/web/http/api/beta/ProbeTemplateGetHandler.java
andrewazores: Yea perfect, a `TypeAdapter` is exactly what I would suggest for a situation like that.

Josh-Matsuoka:
Sure,

For some testing steps you can add the agent to cryostat by copying the agent jar into the clientlib directory, making sure it gets copied into the image by run.sh 

+    --mount type=bind,source="$(dirname "$0")/clientlib/agent.jar",destination=/clientlib/agent.jar,relabel=shared \

then adding it to the startup flags in entrypoint.sh

+    "-javaagent:/clientlib/agent.jar"

For an example configuration:

<jfragent>
	
	<config>
		
		<classprefix>__JFREvent</classprefix>
		
		<allowtostring>true</allowtostring>
		
		<allowconverter>true</allowconverter>
	</config>
    <events>
        <event id="demo.jfr.test1">
            <label>DemoEvent</label>
            <description>Event for the agent plugin demo</description>
            <path>demo</path>
            <stacktrace>true</stacktrace>
            <class>io.cryostat.net.web.http.api.v1.TargetTemplatesGetHandler</class>
            <method>
                <name>handleAuthenticated</name>
		<descriptor>(Lio/vertx/ext/web/RoutingContext;)V</descriptor>
            </method>
            
            <location>ENTRY</location>
        </event>
    </events>
</jfragent>

Will add probes to cryostat and add an event to the TargetTemplatesGetHandler

Josh-Matsuoka:
`<jfragent>
	
	<config>
		
		<classprefix>__JFREvent</classprefix>
		
		<allowtostring>true</allowtostring>
		
		<allowconverter>true</allowconverter>
	</config>
    <events>
        <event id="demo.jfr.test1">
            <label>DemoEvent</label>
            <description>Event for the agent plugin demo</description>
            <path>demo</path>
            <stacktrace>true</stacktrace>
            <class>io.cryostat.net.web.http.api.v1.TargetTemplatesGetHandler</class>
            <method>
                <name>handleAuthenticated</name>
		<descriptor>(Lio/vertx/ext/web/RoutingContext;)V</descriptor>
            </method>
            
            <location>ENTRY</location>
        </event>
    </events>
</jfragent>`

andrewazores:
Could you update `run.sh` to mount the probes directory somewhere persistent/not tmpfs? A subdirectory of `conf` is probably fine.

andrewazores:
Sorry, I think I missed this when the initial merge was done too, but this can/should be fixed now:

```java
        return connectionManager.executeConnectedTask(
                getConnectionDescriptorFromParams(requestParams),
                connection -> {
                    connection.connect();
                    AgentJMXHelper helper = new AgentJMXHelper(connection.getHandle());
                    helper.defineEventProbes(probeTemplateService.getTemplate(probeTemplate));
```

In the `TargetProbe*Handler`s, I think these `connection.connect();` lines should be removed. The `JFRConnection` instance passed into the callback is already open and connected when the callback is run.

Josh-Matsuoka:
Updated to remove the unnecessary connect calls. I'll open a separate issue for the upload notification.

andrewazores:
Nice, that removal of `connection.connect()` actually revealed a bug, since I don't think anywhere else in Cryostat actually uses `connection.getHandle()` like that.

```
SEVERE: HTTP 500: Cannot invoke "org.openjdk.jmc.rjmx.IConnectionHandle.getServiceOrDummy(java.lang.Class)" because "connectionHandle" is null
io.vertx.ext.web.handler.HttpException: Internal Server Error
Caused by: io.cryostat.net.web.http.api.v2.ApiException: Cannot invoke "org.openjdk.jmc.rjmx.IConnectionHandle.getServiceOrDummy(java.lang.Class)" because "connectionHandle" is null
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:112)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:68)
	at io.vertx.core.impl.ContextImpl.lambda$null$0(ContextImpl.java:159)
	at io.vertx.core.impl.AbstractContext.dispatch(AbstractContext.java:100)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$1(ContextImpl.java:157)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
Caused by: java.lang.NullPointerException: Cannot invoke "org.openjdk.jmc.rjmx.IConnectionHandle.getServiceOrDummy(java.lang.Class)" because "connectionHandle" is null
	at io.cryostat.core.agent.AgentJMXHelper.<init>(AgentJMXHelper.java:68)
	at io.cryostat.net.web.http.api.v2.TargetProbesGetHandler.lambda$handle$0(TargetProbesGetHandler.java:113)
	at io.cryostat.net.TargetConnectionManager.executeConnectedTask(TargetConnectionManager.java:163)
	at io.cryostat.net.web.http.api.v2.TargetProbesGetHandler.handle(TargetProbesGetHandler.java:109)
	at io.cryostat.net.web.http.api.v2.AbstractV2RequestHandler.handle(AbstractV2RequestHandler.java:102)
	... 10 more
```

The problem is here, in `-core`'s JFRConnection class:

```java
    public synchronized IConnectionHandle getHandle() {
        return this.handle;
    }

    public synchronized IFlightRecorderService getService()
            throws ConnectionException, IOException, ServiceNotAvailableException {
        if (!isConnected()) {
            connect();
        }
        IFlightRecorderService service = serviceFactory.getServiceInstance(handle);
        if (service == null || !isConnected()) {
            throw new ConnectionException(
                    String.format(
                            "Could not connect to remote target %s",
                            this.connectionDescriptor.createJMXServiceURL().toString()));
        }
        return service;
    }
```

This could either be fixed in `-core` by checking for a connection before returning from `getHandle()`, or it could be fixed in Cryostat's `TargetConnectionManager` by calling `connect()` on the connection instance before invoking the connected task on it. I'm inclined to say that doing it in `-core` is the better fix.

Josh-Matsuoka:
Fixed the getHandle issue in the core PR, and made a quick adjustment here to accompany it

andrewazores (approved):

PR #1127, opened 2022-10-18
test(itest): reduce JDP wait times

andrewazores:
Related to #1068

default values have been thoroughly tested over a long time and appear to be overly cautious. JDP "settling" doesn't seem to happen and is more likely to indicate a broken implementation than a natural effect.

### DIFF
--- a/repeated-integration-tests.bash
+++ b/repeated-integration-tests.bash
@@ -38,9 +38,6 @@ STARTFLAGS=(
     "failsafe:integration-test"
     "failsafe:verify"
     "-DfailIfNoTests=true"
-    "-Dcryostat.itest.jdp.poll.period=2500"
-    "-Dcryostat.itest.jdp.poll.count=1"
-    "-Dcryostat.itest.jdp.poll.timeout=17500"
 )
 
 if [ -n "$2" ]; then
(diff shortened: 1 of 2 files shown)


andrewazores:
Oh and btw, my `bash repeated-integration-tests.bash *25*` that I ran a week or two ago with 100% success rate was using the shortened settings that are being removed from the dev script and made the global defaults here. So I consider it quite safe :-)

tthvo (approved):
Looks good and fast! Is number 20000 round from 17500?

```java
static final int DISCOVERY_BASE_MS =
          Integer.parseInt(System.getProperty("cryostat.itest.jdp.poll.timeout", "20000"));
```

andrewazores:
The total timeout is really just arbitrary. It's the total amount of time that the test execution will wait for target discovery to complete before just timing out and failing. Normally JDP targets appear just about instantly and take ~7.5 seconds to disappear (that's just the default timing on the JDP broadcast packet system), so anything beyond 7.5 seconds plus a bit of margin for time skew is also overly cautious. Whether the timeout is 7.5 seconds plus 10 or 12.5 for the additional buffer is mostly inconsequential, it's just there to account for possible hiccups or infra overload during the test run.

tthvo:
> The total timeout is really just arbitrary. It's the total amount of time that the test execution will wait for target discovery to complete before just timing out and failing. Normally JDP targets appear just about instantly and take ~7.5 seconds to disappear (that's just the default timing on the JDP broadcast packet system), so anything beyond 7.5 seconds plus a bit of margin for time skew is also overly cautious. Whether the timeout is 7.5 seconds plus 10 or 12.5 for the additional buffer is mostly inconsequential, it's just there to account for possible hiccups or infra overload during the test run.

Right! Makes sense! Thanks a lot for the explanations :D

maxcao13 (approved):
Makes sense to me, this is good to cut wait times way shorter if the values have been tested for a long time like you said!

Issue #1118, opened 2022-10-16
[Meta] Upstream 2.2 release


andrewazores:

ebaron:
All related tasks are complete

PR #1132, opened 2022-10-18
build(frontend): skip frontend tests, require frontend updated by CI only

andrewazores:
Related to #1118

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -19,4 +19,7 @@ pull_request_rules:
       review:
         type: REQUEST_CHANGES
         message: >-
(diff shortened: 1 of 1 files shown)


andrewazores:
@ebaron not sure what the best way to test out this config update is. It won't run on one of our forks by default unless we go enable Mergify for our own accounts, so it seems like it needs to be done here first.

ebaron (approved):

andrewazores:
Cool: https://github.com/cryostatio/cryostat/pull/1089#pullrequestreview-1147845547

ebaron:
@andrewazores can we backport this? the frontend tests are holding up our downstream builds.

andrewazores:
Sure.

PR #1136, opened 2022-10-20
fix(archives): fixed archive migration and metadata handling from all-archives view

### LINKED ISSUE #1135: [Task] Archives incorrectly migrating
To recreate:

1. Git checkout cryostat/cryostat-v2.1
2. `$ podman pull quay.io/cryostat/cryostat:2.1.0 # clear cache! `
3. `$ CRYOSTAT_IMAGE="quay.io/cryostat/cryostat:2.1.0" sh smoketest.sh`
4. create and archive a recording
5. Go back to current commit, and there will be an IOException and the recording is not migrated correctly

I think the migration is ignoring the connectUrl files.

maxcao13:
Fixes #1135

I think this needs documentation since users will wonder where their recordings went after migrating from 2.1.0 to 2.2.0.

### DIFF
--- a/src/main/java/io/cryostat/recordings/JvmIdHelper.java
+++ b/src/main/java/io/cryostat/recordings/JvmIdHelper.java
@@ -38,6 +38,7 @@
 package io.cryostat.recordings;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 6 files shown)


andrewazores (approved):
Seems to work as expected and startup completes even faster, nice. Hard to read and mentally step through these changes to verify the fix that way, to be honest, but it seems to make sense.

andrewazores:
@dfitzmau the change summary here is just an upgrade note for users coming from 2.1.0 (or older). When these users upgrade to 2.2.0, any recordings that were previously visible in their Archives view will now be listed under `Archives > All Archives > lost`. This indicates that the target application from which the recordings is sourced has been "lost", ie Cryostat doesn't know exactly where they came from, so they are grouped together and stored in that special location.

PR #1139, opened 2022-10-20
ci(mergify): dismiss reviews when PRs update

andrewazores:
Related to #1132
Related to #1118

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -15,6 +15,7 @@ pull_request_rules:
       - base=main
     actions:
       dismiss_reviews:
+        changes_requested: [ 'mergify', 'Mergifyio' ]
 
   - name: block pull requests that manually update web-client submodule
     conditions:
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -10,6 +10,12 @@ pull_request_rules:
         assignees:
           - "{{ author }}"
 
+  - name: dismiss stale reviews
+    conditions:
+      - base=main
+    actions:
+      dismiss_reviews:
+
   - name: block pull requests that manually update web-client submodule
     conditions:
       - or:


ebaron (commented):
Does this config change apply to only the web-client rule?

andrewazores:
https://docs.mergify.com/actions/dismiss_reviews/

It would apply to everything, though when a PR is updated GitHub already dismisses stale reviews.

I do wonder if the "Update branch" button on the GitHub UI would also trigger this Mergify action, though. It seems like it probably would, which is not desirable.

andrewazores:
Ah, hold on. Reading that docs page more carefully reveals the answer...

andrewazores:
Well there's this:

https://dashboard.mergify.com/github/cryostatio/repo/cryostat/config-editor

but it doesn't say much useful about this new rule. It just indicates that it would dismiss reviews on https://github.com/cryostatio/cryostat/pull/1126, for example, but it doesn't actually confirm that it would only dismiss its own.

ebaron (approved):
Seems harmless enough when constrained to its own reviews.

andrewazores:
Yea, the problem seems to be that mergify is a GitHub Application, not an actual user account, so its "username" isn't match by the rule I added in this PR. I wonder if having it review PRs that do *not* contain `web-client` changes and leave a review comment (no approval or change request) would clear the previous status?

andrewazores:
^ probably not:

https://github.com/cryostatio/cryostat/pull/624

Requesting changes and then leaving a review comment leaves the request status in place. Without being able to have mergify dismiss its own reviews specifically, it seems our options are probably either to have it dismiss all reviews when PRs update (which seems OK to me and really not much different from how things already are), or to have Mergify automatically approve PRs that don't modify `web-client`. I can modify the repo settings to require 2 approvals to merge, with the assumption that one of those would be Mergify, but that wouldn't be exactly set in stone.

Or, we can use a CODEOWNERS file:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

and see if it's possible to set mergify as the owner for `web-client`.

This is getting way more complex than I had hoped it to be.

Perhaps rather than Mergify issuing a "changes requested" review when the `web-client` is touched, it should just add a comment and a label to the PR to draw attention to that for the reviewer to notice. If the PR is updated and `web-client` is not touched then the label is removed.

PR #1141, opened 2022-10-21
build(jib): upgrade jib plugin to 3.3.0

### LINKED ISSUE #1128: [Task] Update jib plugin
[INFO] A new version of jib-maven-plugin (3.3.0) is available (currently using 3.1.1). Update your build configuration to use the latest features and fixes!
[INFO] https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/CHANGELOG.md

tthvo:
Fixes #1128 

Upgraded `jib` to latest version (i.e. `3.3.0`). There is a new [feature](https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/CHANGELOG.md#added-3) specified multi-platform build, which is pretty interesting but not sure we need it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
   <org.codehaus.mojo.build.helper.plugin.version>3.3.0</org.codehaus.mojo.build.helper.plugin.version>
   <com.mycila.license.maven.plugin.version>4.0</com.mycila.license.maven.plugin.version>
   <org.owasp.dependency.check.version>6.1.5</org.owasp.dependency.check.version>
-  <com.google.cloud.tools.jib.maven.plugin.version>3.1.1</com.google.cloud.tools.jib.maven.plugin.version>
+  <com.google.cloud.tools.jib.maven.plugin.version>3.3.0</com.google.cloud.tools.jib.maven.plugin.version>
 
   <com.google.dagger.version>2.34.1</com.google.dagger.version>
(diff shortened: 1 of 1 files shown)


andrewazores:
> There is a new [feature](https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/CHANGELOG.md#added-3) specified multi-platform build, which is pretty interesting but not sure we need it.

That might be useful for downstream actually so we can have product builds for both x86 and ARM, mostly. Upstream can just leave it default, at least for now.

ebaron:
> @ebaron if this makes sense for 2.2.0 downstream, I think this is safe enough to backport too.

Downstream doesn't build with Jib, but you can backport if there's some benefit to the upstream 2.2 builds.

andrewazores:
Are there any other specific reasons that downstream builds with custom Dockerfiles rather than using jib? Is the Dockerfile route better supported by some of the CVE rebuild automation perhaps?

andrewazores (dismissed):

ebaron:
> Are there any other specific reasons that downstream builds with custom Dockerfiles rather than using jib? Is the Dockerfile route better supported by some of the CVE rebuild automation perhaps?

Building from Dockerfiles is the only option for container images: https://osbs.readthedocs.io/en/latest/users.html#writing-a-dockerfile

andrewazores (approved):

PR #1144, opened 2022-10-21
fix(ci): fix broken ci job for pushing to quay

### LINKED ISSUE #1143: [Task] Fix broken CI job for pushing to quay
> @maxcao13 looks like `push-to-quay` might be broken:
https://github.com/cryostatio/cryostat/actions/runs/3300643166/jobs/5445469164

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/1138#issuecomment-1287477374_

maxcao13:
Fixes #1143

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -166,6 +166,10 @@ jobs:
     env:
       CRYOSTAT_IMG: quay.io/cryostat/cryostat
     steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+        fetch-depth: 0
     - name: Tag images
       id: tag-image
       env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -169,8 +169,10 @@ jobs:
     - uses: actions/download-artifact@v3
       with:
         name: cryostat
+      if: ${{ github.event_name == 'push' && github.repository_owner == 'cryostatio' }}
     - name: Load cryostat image
       run: podman load -i cryostat.tar
+      if: ${{ github.event_name == 'push' && github.repository_owner == 'cryostatio' }}
     - name: Tag images
       id: tag-image
       env:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -166,10 +166,11 @@ jobs:
     env:
       CRYOSTAT_IMG: quay.io/cryostat/cryostat
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/download-artifact@v3
       with:
-        submodules: true
-        fetch-depth: 0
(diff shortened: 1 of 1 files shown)


andrewazores:
Why does the git branch need to be checked out? I thought the fix would be to `download-artifact` to get the container image into the registry within that job and then push it further on to the Quay registry. Although, this happens across different events (the rest are on PR, this one on Push which comes after a PR Merge), so I wonder if the same artifact will actually be available. Maybe the push-to-quay job needs to be split out into its own separate Push-only config and do the entire process over again on its own, if the artifact is not retrievable across jobs/events.

maxcao13:
> Why does the git branch need to be checked out? I thought the fix would be to `download-artifact` to get the container image into the registry within that job and then push it further on to the Quay registry. Although, this happens across different events (the rest are on PR, this one on Push which comes after a PR Merge), so I wonder if the same artifact will actually be available. Maybe the push-to-quay job needs to be split out into its own separate Push-only config and do the entire process over again on its own, if the artifact is not retrievable across jobs/events.

Whoops, I misread the error message, I thought the error was because of `Run if [ "$GITHUB_REF" == "refs/heads/main" ]; then`

Right, that should be the fix, I wonder if artifacts are available across workflows as well.

andrewazores:
Unfortunately there isn't a great way to test these things out without just merging them. https://github.com/nektos/act would be nice but it still doesn't support podman. Otherwise, you could test it out by modifying the config to point at your quay username/org instead of the upstream Cryostat one, and then generate a quay API token and add it to your fork's repo settings as a secret, and then see if the workflow works as expected when you push to your fork's main.

I say we just merge this and see if it works, and if not revert this last job to run the full cycle on its own.

andrewazores (approved):

andrewazores (changes requested):
Actually, hold on.

andrewazores (approved):

andrewazores (approved):

andrewazores:
I also just noticed:

https://github.com/cryostatio/cryostat/actions/runs/3301133592/jobs/5446321262

the spotbugs job, because of how spotbugs works, actually ends up building the web-client. It should be run with `-Dheadless` to skip building the web-client since the spotbugs analysis only applies to the compiled bytecode of the server.

andrewazores:
https://github.com/cryostatio/cryostat/actions/runs/3301133592/jobs/5446368758#step:4:13

The `IMAGE_VERSION` env var is empty.

Looks like:

```yaml
    needs: [integration-tests, spotless, spotbugs, shellcheck]
    ...
      env:
        IMAGE_VERSION: ${{ needs.get-pom-properties.outputs.image-version }}
```

is probably the problem? I think `get-pom-properties` needs to be added to `needs`.

PR #1145, opened 2022-10-21
fix(credentials): automatically use stored JMX credentials in V2 API handlers

### LINKED ISSUE #1142: [Task] V2 API endpoints do not use stored credentials

andrewazores:
Fixes #1142

JWT-based handlers and API V1 handlers all automatically check for stored JMX credentials because the abstract base classes implement this behaviour. The abstract V2 base class did not, so handlers either needed to handle it manually, or were restricted to requiring the client to pass the `X-JMX-Authorization` header. This change updates the V2 abstract base class to bring it in line with the others.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/AbstractV2RequestHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/AbstractV2RequestHandler.java
@@ -47,6 +47,9 @@ import java.util.concurrent.Future;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import javax.script.ScriptException;
+
+import io.cryostat.configuration.CredentialsManager;
 import io.cryostat.core.net.Credentials;
 import io.cryostat.net.AuthManager;
 import io.cryostat.net.ConnectionDescriptor;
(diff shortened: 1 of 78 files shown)


tthvo (commented):
Looks good to me! I will wait for @maxcao13 to sign off.

src/main/java/io/cryostat/net/web/http/api/v2/AbstractV2RequestHandler.java
```diff
@@ -119,46 +125,52 @@ protected Future<Boolean> validateRequestAuthorization(String authHeader) throws
...
-            String proxyAuth = params.getHeaders().get(JMX_AUTHORIZATION_HEADER);
-            Matcher m = AUTH_HEADER_PATTERN.matcher(proxyAuth);
-            if (!m.find()) {
-                params.getHeaders().set(JMX_AUTHENTICATE_HEADER, "Basic");
-                throw new ApiException(427, "Invalid " + JMX_AUTHORIZATION_HEADER + " format");
-            } else {
-                String t = m.group("type");
-                if (!"basic".equals(t.toLowerCase())) {
+        try {
+            Credentials credentials = credentialsManager.getCredentialsByTargetId(targetId);
```
maxcao13, line 129: If the credentials here were found by the `CredentialsManager`, is there a need to go through this whole if statement? If there was no such JMX_AUTH_HEADER, then would the ConnectionDescriptor be constructed anyways?

andrewazores: This is supported intentionally to allow the client to override stored credentials by passing the `X-JMX-Authorization` header regardless. The ConnectionDescriptor will be constructed in any case, whether without credentials, with credentials from the header, or with stored credentials.

But, since checking for stored credentials is now a much more expensive operation than it used to be, I'll go through and refactor things so that we only perform that query if the header wasn't provided.

maxcao13 (approved):
Looks good to me then!

PR #1153, opened 2022-10-24
fix(ci): add ITEST_IMG_VERSION arg to repeated-tests bash script

### LINKED ISSUE #1152: [Task] CI pulls from quay.io instead of from local podman repository
> Ah, looks like another CI problem, I think. The `repeated-integration-tests` and `mvn verify` try to pull the image with tag `2.3.0-snapshot` https://github.com/cryostatio/cryostat/actions/runs/3316176363/jobs/5478379389#step:7:419, but as we know when loading the image with podman load, the tags aren't preserved so I think it actually pulls from quay. That's annoying.

_Originally posted by @maxcao13 in https://github.com/cryostatio/cryostat/issues/1089#issuecomment-1289676657_

maxcao13:
Fixes #1152 

Also, I didn't mention this before but there is a mismatch of the itest output filename logs between the bash script and the mvn build executions. 

That's why in the split-container-pr #1138, I had to change the `run: ls -1dt target/cryostat-itest-*.log | head -n1 | xargs cat` -> `cryostat-itests`. 

The pom.xml here on this line: 

https://github.com/cryostatio/cryostat/blob/132459fbefed742f00c26058b184b2889f790aa3/pom.xml#L633

outputs the file as `cryostat-itest-timestamp.log` while the script does `cryostat-itests-timestamp.log` since `cryostat.itest.containerName="cryostat-itest"`. 

Which should be changed here?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -108,7 +108,7 @@ jobs:
       with:
         step: restore
     - name: Run integration tests
-      run: POD_NAME=cryostat-itests CONTAINER_NAME=cryostat-itest bash repeated-integration-tests.bash
+      run: POD_NAME=cryostat-itests CONTAINER_NAME=cryostat-itest ITEST_IMG_VERSION=latest bash repeated-integration-tests.bash
     - name: Print itest logs
       if: ${{ failure() }}
       run: ls -1dt target/cryostat-itests-*.log | head -n1 | xargs cat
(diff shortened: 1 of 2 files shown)


andrewazores:
I don't think it matters much either way. I suppose the `pom.xml` is the more "correct" one and the script and CI configs should be adjusted to match.

andrewazores (approved):

Issue #1159, opened 2022-10-25
Add configuration to skip building base image


ebaron:
On environments without an image builder, I would like to be able to skip building the base image without disabling exec-maven-plugin completely.

PR #1160, opened 2022-10-25
build(pom): optionally skip building base image

ebaron:
The following should succeed to build the Java artifacts for Cryostat:
`mvn -DskipBaseImage=true -DskipITs=true -Djib.skip=true -DimageBuilder=false clean verify`

Fixes: #1159

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
   <skipTests>false</skipTests>
   <skipITs>${skipTests}</skipITs>
   <skipUTs>${skipTests}</skipUTs>
+  <skipBaseImage>false</skipBaseImage>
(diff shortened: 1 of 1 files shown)


andrewazores:
Looks good.

```
mvn -Dheadless -DskipBaseImage=true -DskipITs=true -Djib.skip=true -DimageBuilder=false clean verify
```

Source and resource generation, compilation, and unit tests all run. Spotbugs, spotless, and shellcheck also run.

Inspecting `target/` looks like all the expected artifacts are present. Build logs and `podman images` show that no containers were built.

andrewazores (approved):

PR #1179, opened 2022-10-26
fix(graphql): expose archiveOnStop property in doStartRecording settings

### LINKED ISSUE #1178: [Task] `archiveOnStop` property is not accessible from GraphQL
Related to #1097

andrewazores:
Fixes #1178

Manual testing steps:

```bash
$ http -v :8181/api/v2.2/graphql Authorization:"Basic $(echo user:pass | base64)" query="{ targetNodes(filter: { name: \"service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi\" }) { doStartRecording(recording: { name:\"foo\", template:\"Profiling\", templateType:\"TARGET\", duration: 10 }) { name } } }"
```

(will probably need to separately define JMX keyring credentials for the Cryostat target)
This query starts a recording without specifying or reading `archiveOnStop` and should work identically before and after the PR.

```bash
$ http -v :8181/api/v2.2/graphql Authorization:"Basic $(echo user:pass | base64)" query="{ targetNodes(filter: { name: \"service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi\" }) { doStartRecording(recording: { name:\"bar\", template:\"Profiling\", templateType:\"TARGET\", duration: 10 }) { name archiveOnStop } } }"
```

This query starts a recording without specifying `archiveOnStop`, but then tries to read the value back from the started recording. This should fail on `main` and pass (`false` value) on this PR.

```bash
$ http -v :8181/api/v2.2/graphql Authorization:"Basic $(echo user:pass | base64)" query="{ targetNodes(filter: { name: \"service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi\" }) { doStartRecording(recording: { name:\"baz\", template:\"Profiling\", templateType:\"TARGET\", duration: 10, archiveOnStop: true }) { name archiveOnStop } } }"
```

This query starts a recording specifying `archiveOnStop` and reads the value back. This should also fail on `main` and pass (`true` value) on this PR.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java
@@ -165,7 +165,8 @@ class StartRecordingOnTargetMutator
                             desc,
                             ws.getDownloadURL(conn, desc.getName()),
                             ws.getReportURL(conn, desc.getName()),
-                            metadata);
+                            metadata,
+                            archiveOnStop);
                 });
     }
 }
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java
@@ -137,6 +137,11 @@ class StartRecordingOnTargetMutator
                     if (settings.containsKey("maxSize")) {
                         builder = builder.maxSize((Long) settings.get("maxSize"));
                     }
+                    boolean archiveOnStop = false;
+                    if (settings.containsKey("archiveOnStop")) {
+                        Boolean v = (Boolean) settings.get("archiveOnStop");
+                        archiveOnStop = v != null && v;
+                    }
                     Metadata m = new Metadata();
                     if (settings.containsKey("metadata")) {
                         m =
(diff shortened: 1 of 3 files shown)


src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java
```diff
@@ -137,6 +137,11 @@ public HyperlinkedSerializableRecordingDescriptor getAuthenticated(
                     if (settings.containsKey("maxSize")) {
                         builder = builder.maxSize((Long) settings.get("maxSize"));
                     }
+                    boolean archiveOnStop = false;
+                    if (settings.containsKey("archiveOnStop")) {
+                        Boolean v = (Boolean) settings.get("archiveOnStop");
+                        archiveOnStop = v != null && v;
```
tthvo, line 143: I think here we don't have to check null? Or is there special use here?

tthvo (commented):
Looks good! Worked well! Just a question..

maxcao13 (approved):
Looks good to me, but I notice there seems to be a new bug with the UI not showing attatched metadata when recordings are created through graphQL? I get this on main as well, wondering if others have the same problem. The labels reappear after refreshing cause it seems the metadata was correctly created, but the UI didn't seem to notice it at first.

src/main/java/io/cryostat/net/web/http/api/v2/graph/StartRecordingOnTargetMutator.java
andrewazores: Just being defensive. Since the `settings` is a `Map<String, Object>`, the `.get()` is a boxed `Boolean`. If it's a `null` value then unboxing that to the primitive `boolean` would NPE.

tthvo: Oh oh right! That makes sense! Thanks :D

andrewazores:
@maxcao13 possible, I haven't tested it but both the HTTP and GraphQL requests should end up following the same code paths. Maybe one of the branches along the path doesn't emit websocket notifications when it should.

tthvo (approved):

PR #1181, opened 2022-10-27
ci(dependabot): create dependabot update config

### LINKED ISSUE #1180: [Task] Create dependabot config for automatic dependency version updates
This config should also be copied to cryostat-core, cryostat-reports, jfr-datasource, cryostat-agent, cryostat-web, cryostat-operator, cryostat-grafana-dashboard

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

andrewazores:
Related to #1180

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+pom.xml @andrewazores @ebaron
+src/container/ @andrewazores @ebaron
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,4 @@ updates:
     labels:
       - "dependencies"
       - "chore"
+      - "safe-to-test"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: "maven"
-    directory: "/" # Location of package manifests
+    directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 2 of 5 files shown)


andrewazores:
I tested this on my fork and got PRs like these:
https://github.com/andrewazores/cryostat/pull/13
https://github.com/andrewazores/cryostat/pull/14

This would automatically get picked up and run by the CI builder to ensure update compatibility, plus the PR CI image would be published for reviewers to perform additional manual checks.

andrewazores:
The `pom.xml` dependency updates are definitely something we have to be careful about and manually review. Luckily Dependabot is smart enough that we can tell it to ignore major or minor versions or not notify at all for specific dependencies, so I think this would be helpful for upstream to at least notify us that there is a new version to potentially evaluate and see if there is anything corresponding for downstream.

Otherwise, I guess we need to define some process for downstream checking for updated dependencies and then pulling in the corresponding upstream updates, but this can be a little tricky if the downstream dependency builds also include backports that may not be in the upstream versions...

ebaron (approved):
The update notifications are indeed helpful. I'd be fine with this as long as we remember to check for downstream compatibility before approving Dependabot's PRs.

andrewazores:
@ebaron I added a `CODEOWNERS` file that should trigger GitHub to automatically request reviews from both you and I, and require at least one of us to sign off, whenever `pom.xml` or anything within `src/container/` is modified by a PR. This should help ensure that we don't forget to check downstream availability. I think as long as one of us notices that the PR is from Dependabot and has the `dependencies` label it should be pretty self-explanatory, and we can revert a commit afterward if we do make a mistake.

andrewazores:
We can probably build some further automation too, like having Dependabot apply a `needs-downstream-check` label and an Action that blocks merge until that label is removed.

ebaron (approved):
That's a useful feature. Thanks!

ebaron:
We share a lot of dependencies with Quarkus, I wonder if we could simplify things by importing Quarkus's BOM.

PR #1187, opened 2022-11-01
fix(devserver): correct devserver setup for other changes

andrewazores:
Related to #1160
Related to #1083

- don't compile tests if tests are skipped
- set JMX keyring password and skip building base image


maxcao13 (approved):
Looks good to me!

PR #1189, opened 2022-11-03
fix(discovery): allow targets restarting quickly using JDP to render correctly on the -web

maxcao13:
Fixes https://github.com/cryostatio/cryostat-web/issues/493

Before fixing tests, was there a reason that the notifications were in the previous order?

The only "issue" is that the LOST target will eventually notify Cryostat that it was lost but nothing will be done, (no notifications emitted since the subtree will be the same), but this is just unnecessary work. 

I will test this on my own through OpenShift to make sure that it doesn't break that functionality.


andrewazores:
> Before fixing tests, was there a reason that the notifications were in the previous order?

Not that I can remember...

maxcao13:
Works as expected on OpenShift.

andrewazores (approved):

PR #1193, opened 2022-11-04
fix(api): fix GET /api serialization and webserver JFR event

andrewazores:
Fixes #1191
Fixes #1192

- fix(api): GET /api returns properly serialized endpoint verb
- chore(log): replace request logging with Vertx impl
- correct event name and move into logging handler

PR #1202, opened 2022-11-06
chore(graphql): cleanup, refactor, async execution

### LINKED ISSUE #1201: [Task] GraphQL async execution
Refactor GraphQL fetchers/mutators to support [async execution](https://www.graphql-java.com/documentation/execution/#asynchronous-execution). This should hopefully improve throughput for concurrent requests.

andrewazores:
Fixes #1201
Related to #1214 / #1193
Depends on https://github.com/cryostatio/cryostat-web/pull/611


andrewazores:
Oh, I guess this does actually need https://github.com/cryostatio/cryostat-web/pull/611 in that case. I worked on them together but it seemed they should work independently. The request being sent has the raw GraphQL `query { ... }` string as the POST request body, which the handler doesn't seem to know what to do with.

andrewazores:
Most recent commit touches up the HTTP request logging a little.

```
INFO: 127.0.0.1 - - [Tue, 8 Nov 2022 00:02:31 GMT] 46ms "GET /api/beta/fs/recordings HTTP/1.1" 200 354 bytes "http://localhost:9000/" "Mozilla/5.0 (X11; Linux x86_64; rv:106.0) Gecko/20100101 Firefox/106.0"
```

The `46ms` is new. Our old custom format built-in to the WebServer class included that detail and it was lost when I switched over to extracting it to a separate Handler and delegating to a Vert.x default implementation. That default implementation also includes the content-length of the response in raw bytes, and I've touched it up to use `FileUtils.byteCountToDisplaySize` from commons-io so that it's a bit more readable.

src/main/java/io/cryostat/net/web/http/api/v2/graph/GraphModule.java
```diff
@@ -93,26 +102,8 @@ public abstract class GraphModule {
     @Provides
     @Singleton
     static GraphQL provideGraphQL(
-            NodeTypeResolver nodeTypeResolver,
-            RecordingTypeResolver recordingTypeResolver,
-            RootNodeFetcher rootNodeFetcher,
-            EnvironmentNodesFetcher environmentNodesFetcher,
-            TargetNodesFetcher targetNodesFetcher,
```
maxcao13, line 100: I just realized something, why doesn't `EnvironementNodeRecurseFetcher` show up here? Is it being used somewhere else somehow?

src/main/java/io/cryostat/net/web/http/generic/RequestLoggingHandler.java
```diff
@@ -114,6 +123,74 @@ public void handle(RoutingContext ctx) {
         this.delegate.handle(ctx);
     }
 
+    // this is ripped from the LoggerHandler's `DEFAULT` format, but with the addition of the
```
maxcao13, line 126: Just had a thought, I think it may be nice to have color coding in the terminal for better readability. For example just using something like https://stackoverflow.com/questions/5762491/how-to-print-color-in-console-using-system-out-println here in your Logger formatter implementation. Thoughts? I'm not sure how it would work for other Logger implementations currently, but it could be something for me to look into if it's not straight-forward.

src/main/java/io/cryostat/net/web/http/api/v2/graph/GraphModule.java
andrewazores: I guess this was previously used for some earlier query that ended up removed. It doesn't look like it is actually used anymore at all. The implementation does something like `descendantTargets`, but it would be `descendantEnvironments` instead which doesn't seem particularly useful. I'll remove it.

src/main/java/io/cryostat/net/web/http/generic/RequestLoggingHandler.java
andrewazores: Sure, that could be a nice enhancement. It would probably go along with refactoring our `Logger` stuff entirely, though. Injecting the `Logger` instance with Dagger means that everything uses the same single instance, which is a tiny performance gain but comes at the cost of not being able to tell which class is logging the message easily. More commonly, a class gets a Logger instance like ex. `private static final Logger LOG = LoggerFactory.getLogger(MyClass.class)`, and then the `LOG` instance can use `MyClass` as a prefix for the messages it logs, and the log config can even set up different default log levels on a per-class basis etc. If you're interested in taking on that work then adding in colourized output for the class name prefix and a timestamp might make it worthwhile.

Just note however that those colour escape codes are generally only meaningful and useful when the output is being viewed directly in a terminal and not when the logs are being written to a file. There are other tools that an end user could use to strip the colour codes out of the logs but it might be more annoying to deal with. For that reason I would suggest making the colour-coding an opt-in feature by env var or system property so we can just enable it during `devserver.sh`/`run.sh`/`smoketest.sh` runs and leave it off by default elsewhere.

PR #1214, opened 2022-11-07
fix(logging): fix broken log config

### LINKED ISSUE #1213: [Task] Broken logging
Since https://github.com/cryostatio/cryostat/pull/1193

andrewazores:
Fixes #1213


tthvo:
spotless?

tthvo (commented):
Ah didn't notice this before. Looks like old configs are overwritten with null? 

Just a question on this again. As we are applying config from `logging.properties`, should the new config takes higher precedence?

tthvo (approved):
Looks good to me! Logs are back :D

PR #1257, opened 2022-11-22
feat(targets): dryrun query params for testing connectUrl and credentials

tthvo:
Related to https://github.com/cryostatio/cryostat-web/issues/592

This added the back-end support for test connection UI via `dryrun` query parameter. I chose another status code for JMX Auth Failure here is to avoid returning `427` that triggers openning up Auth Modal in the front-end. Another way is to re-use 427 but there must be a way for the front-end to know which requests are dryruns (i.e. maybe a custom header?)

**Manual Tests**

Using `localhost:9091` here to go around duplicate target error:

- With credentials
```bash
https --form POST https://localhost:8181/api/v2/targets?dryrun=true \
connectUrl='service:jmx:rmi:///jndi/rmi://localhost:9091/jmxrmi' \
alias="self" \
username="smoketest" \
password="smoketest" \
--verify=no \
--auth user:pass \
--auth-type basic
```

For non-dryrun requests (saving target), the workaround could be the front-end first posts a credentials with the matchExpression that matches this target. Then, cached when computing jvmid can grab it from there.

- Without credentials
```bash
https --form POST https://localhost:8181/api/v2/targets?dryrun=true \
connectUrl='service:jmx:rmi:///jndi/rmi://localhost:9091/jmxrmi' \
alias="self" \
--verify=no \
--auth user:pass \
--auth-type basic
```

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
@@ -196,7 +196,7 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
             return new IntermediateResponse<ServiceRef>().body(serviceRef);
         } catch (JvmIdGetException e) {
             if (AbstractAuthenticatedRequestHandler.isJmxAuthFailure(e)) {
-                throw new ApiException(406, "Credentials Not Acceptable", e);
+                throw new ApiException(406, "Credentials Not Provided/Not Acceptable", e);
             }
             if (AbstractAuthenticatedRequestHandler.isUnknownTargetFailure(e)) {
                 throw new ApiException(404, "Target Not Found", e);
--- a/src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
@@ -54,6 +54,7 @@ import io.cryostat.configuration.CredentialsManager;
 import io.cryostat.core.log.Logger;
 import io.cryostat.core.net.Credentials;
 import io.cryostat.discovery.DiscoveryStorage;
+import io.cryostat.messaging.notifications.NotificationFactory;
 import io.cryostat.net.AuthManager;
 import io.cryostat.net.security.ResourceAction;
 import io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler;
@@ -64,6 +65,7 @@ import io.cryostat.platform.ServiceRef.AnnotationKey;
 import io.cryostat.platform.internal.CustomTargetPlatformClient;
 import io.cryostat.recordings.JvmIdHelper;
 import io.cryostat.recordings.JvmIdHelper.JvmIdGetException;
+import io.cryostat.rules.MatchExpressionValidationException;
 import io.cryostat.util.URIUtil;
 
 import com.google.gson.Gson;
@@ -79,6 +81,7 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
     private final JvmIdHelper jvmIdHelper;
     private final CustomTargetPlatformClient customTargetPlatformClient;
     private final Logger logger;
+    private final NotificationFactory notificationFactory;
 
     @Inject
     TargetsPostHandler(
@@ -88,12 +91,14 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
             DiscoveryStorage storage,
             JvmIdHelper jvmIdHelper,
             CustomTargetPlatformClient customTargetPlatformClient,
+            NotificationFactory notificationFactory,
             Logger logger) {
         super(auth, credentialsManager, gson);
         this.storage = storage;
         this.jvmIdHelper = jvmIdHelper;
         this.customTargetPlatformClient = customTargetPlatformClient;
         this.logger = logger;
+        this.notificationFactory = notificationFactory;
     }
 
     @Override
@@ -160,6 +165,11 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
                 }
             }
 
+            MultiMap queries = params.getQueryParams();
+            boolean dryRun =
+                    StringUtils.isNotBlank(queries.get("dryrun"))
+                            && Boolean.valueOf(queries.get("dryrun"));
+
             String username = attrs.get("username");
             String password = attrs.get("password");
             Optional<Credentials> credentials =
@@ -167,10 +177,24 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
                             ? Optional.empty()
                             : Optional.of(new Credentials(username, password));
 
-            MultiMap queries = params.getQueryParams();
-            boolean dryRun =
-                    StringUtils.isNotBlank(queries.get("dryrun"))
-                            && Boolean.valueOf(queries.get("dryrun"));
+            if (!dryRun && credentials.isPresent()) {
+                String matchExpression = String.format("target.connectUrl == '%s'", connectUrl);
+                int id = credentialsManager.addCredentials(matchExpression, credentials.get());
+                notificationFactory
+                        .createBuilder()
+                        .metaCategory("CredentialsStored")
+                        .metaType(HttpMimeType.JSON)
+                        .message(
+                                Map.of(
+                                        "id",
+                                        id,
+                                        "matchExpression",
+                                        matchExpression,
+                                        "numMatchingTargets",
+                                        1))
+                        .build()
+                        .send();
+            }
 
             String jvmId = jvmIdHelper.getJvmId(uri.toString(), !dryRun, credentials);
             ServiceRef serviceRef = new ServiceRef(jvmId, uri, alias);
@@ -212,6 +236,8 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
             throw new ApiException(400, "Invalid connectUrl", use);
         } catch (IOException ioe) {
             throw new ApiException(500, "Internal Error", ioe);
+        } catch (MatchExpressionValidationException e) {
+            throw new ApiException(400, e);
         }
     }
 }
--- a/src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
@@ -156,10 +156,16 @@ class TargetsPostHandler extends AbstractV2RequestHandler<ServiceRef> {
                     throw new ApiException(400, "Duplicate connectUrl");
                 }
             }
-            Map<AnnotationKey, String> cryostatAnnotations = new HashMap<>();
 
-            String jvmId = jvmIdHelper.getJvmId(uri.toString());
+            MultiMap queries = params.getQueryParams();
+            boolean dryRun =
+                    StringUtils.isNotBlank(queries.get("dryrun"))
+                            && Boolean.TRUE.equals(Boolean.valueOf(queries.get("dryrun")));
+
+            String jvmId = jvmIdHelper.getJvmId(uri.toString(), !dryRun);
             ServiceRef serviceRef = new ServiceRef(jvmId, uri, alias);
+
+            Map<AnnotationKey, String> cryostatAnnotations = new HashMap<>();
             for (AnnotationKey ak : AnnotationKey.values()) {
                 // TODO is there a good way to determine this prefix from the structure of the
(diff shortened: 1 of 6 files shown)


tthvo:
Hey @maxcao13, could you take a quick review at these changes to see if this is the proper way to perform this operation?

I am not very familiar with jvmID setup so there might be some misunderstanding. I will test once these changes are good to go.

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -171,13 +187,27 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
...
+                boolean v = customTargetPlatformClient.addTarget(serviceRef);
+                if (!v) {
+                    throw new ApiException(400, "Duplicate connectUrl");
+                }
             }
             return new IntermediateResponse<ServiceRef>().body(serviceRef);
         } catch (JvmIdGetException e) {
-            throw new ApiException(404, "Couldn't connect to target: " + e.getTarget());
+            if (AbstractAuthenticatedRequestHandler.isJmxAuthFailure(e)) {
+                throw new ApiException(406, "Credentials Not Provided/Not Acceptable", e);
```
maxcao13, line 199: I don't think using 406 is the best idea. Usually, I believe status 406 is reserved for throwing an error when the the content-type the user requests does not match what the endpoint can accept. Maybe using 401 Unauthorized is better?

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -156,10 +159,23 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
...
+            String password = attrs.get("password");
+            Optional<Credentials> credentials =
+                    StringUtils.isBlank(username) || StringUtils.isBlank(password)
+                            ? Optional.empty()
+                            : Optional.of(new Credentials(username, password));
+
+            MultiMap queries = params.getQueryParams();
+            boolean dryRun =
+                    StringUtils.isNotBlank(queries.get("dryrun"))
+                            && Boolean.TRUE.equals(Boolean.valueOf(queries.get("dryrun")));
```
maxcao13, line 173: Does this need the `Boolean.TRUE.equals()` ?

src/main/java/io/cryostat/recordings/JvmIdHelper.java
```diff
@@ -165,13 +171,19 @@ private CompletableFuture<String> computeJvmId(String targetId) throws ScriptExc
...
     public String getJvmId(ConnectionDescriptor connectionDescriptor) throws JvmIdGetException {
-        return getJvmId(connectionDescriptor.getTargetId());
+        return getJvmId(connectionDescriptor.getTargetId(), true, Optional.empty());
     }
 
     public String getJvmId(String targetId) throws JvmIdGetException {
+        return getJvmId(targetId, true, Optional.empty());
+    }
+
+    public String getJvmId(String targetId, boolean saveToCached, Optional<Credentials> credentials)
```
maxcao13, line 181: I would maybe rename the boolean to `cache`, or `saveToCache`.

maxcao13 (commented):
Currently, I run into an error when doing dryrun=false since credentials are not being persisted and when `customTargetPlatformClient` calls `addTarget`, I believe some discoveryListeners are getting notified and call some path to `computeJvmId`. The problem is that the computation needs credentials and it gets them from the CredentialsManager and at this point, there is no matching credentials in the backend. 

So yes a potential workaround would be to persist credentials first, and then addTarget. There would still need to be a bit of refactoring with the way credentials work however, because the `addCredentials` method emits a notification when being added, which would likely cause more side-effects.

tthvo:
> Currently, I run into an error when doing dryrun=false since credentials are not being persisted and when customTargetPlatformClient calls addTarget, I believe some discoveryListeners are getting notified and call some path to computeJvmId. The problem is that the computation needs credentials and it gets them from the CredentialsManager and at this point, there is no matching credentials in the backend.

> So yes a potential workaround would be to persist credentials first, and then addTarget. There would still need to be a bit of refactoring with the way credentials work however, because the addCredentials method emits a notification when being added, which would likely cause more side-effects.

Ahh currently on main, this causes the target creation form to fail if credentials are required. So I think the users have to manually create credentials with correct matching expression, then go back to create target. I am planning to allow users to enter credentials on target creation.

Do you think its now a better idea to allow `/api/v2/targets` to have some functionality to add credentials too? Perhaps, there, we can skip the notifications?

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
tthvo: Just wondering if status code 401 confict with other places? Like cases of missing`Authentication: `?

tthvo: How about an unused 4xx status code that is closer to 427? How about 432?

tthvo: Using 420 for now.

maxcao13:
> Do you think its now a better idea to allow `/api/v2/targets` to have some functionality to add credentials too? Perhaps, there, we can skip the notifications?

I imagine if a user wants to only use the command line and not the web-client, it would be difficult to add the target and they would be confused why they cannot add it.

So I'm unsure if having a endpoint do two different things (add target and post credentials) is correct from a REST perspective or other programming principles. So I will defer to @andrewazores when he gets back :) 

It would be a lot easier if we could though.

andrewazores:
> So I'm unsure if having a endpoint do two different things (add target and post credentials) is correct from a REST perspective or other programming principles. So I will defer to @andrewazores when he gets back :)

It just depends on what we consider to be a "resource" from the perspective of the API. Right now we have a hard separation between a "target" and "credentials", but conceptually the credentials are meaningless without a target, and any given target may only be useful if related credentials also exist (as we plainly see in this PR). Since the credentials are expression-based and can potentially match against multiple targets it makes sense to keep that as a separate resource, but I think creating a target definition with its associated credentials up-front also makes sense. In this case it should simply default to constructing a matchExpression that requires an exact connectUrl match.

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -156,10 +159,23 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
                     throw new ApiException(400, "Duplicate connectUrl");
                 }
             }
-            Map<AnnotationKey, String> cryostatAnnotations = new HashMap<>();
 
-            String jvmId = jvmIdHelper.getJvmId(uri.toString());
+            String username = attrs.get("username");
+            String password = attrs.get("password");
+            Optional<Credentials> credentials =
+                    StringUtils.isBlank(username) || StringUtils.isBlank(password)
```
andrewazores, line 166: I wonder if the target JVM would actually accept a `jmxremote.password` file defining a user with an empty password. Maybe we need to be more lenient for that case. Something to investigate anyway, out of scope of this particular PR. It could be tested by modifying Cryostat's own `entrypoint.sh` something like this:

```diff
diff --git a/src/main/extras/app/entrypoint.bash b/src/main/extras/app/entrypoint.bash
index 185d0340..4f9cd524 100755
--- a/src/main/extras/app/entrypoint.bash
+++ b/src/main/extras/app/entrypoint.bash
@@ -25,8 +25,10 @@ function createJmxCredentials() {
     fi
 
     echo -n "$CRYOSTAT_RJMX_USER $CRYOSTAT_RJMX_PASS" > "$PWFILE"
+    echo -n "NOPASS " >> "$PWFILE"
     chmod 400 "$PWFILE"
     echo -n "$CRYOSTAT_RJMX_USER readwrite" > "$USRFILE"
+    echo -n "NOPASS readwrite" >> "$USRFILE"
     chmod 400 "$USRFILE"
 }
```

but then there would also be required changes to the Credentials API and/or frontend to allow blank passwords to be stored.

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -171,13 +187,27 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
...
+                boolean v = customTargetPlatformClient.addTarget(serviceRef);
+                if (!v) {
+                    throw new ApiException(400, "Duplicate connectUrl");
+                }
             }
             return new IntermediateResponse<ServiceRef>().body(serviceRef);
         } catch (JvmIdGetException e) {
-            throw new ApiException(404, "Couldn't connect to target: " + e.getTarget());
+            if (AbstractAuthenticatedRequestHandler.isJmxAuthFailure(e)) {
+                throw new ApiException(420, "JMX Credentials Not Provided or Not Valid", e);
```
andrewazores, line 199: This sounds like it's the same or very similar to what `427` is already used for?

tthvo: Oh yess it is actually same as 427 but if we return 427, on the front-end, the auth modal will pop up. I think this should not happen as the credentials are just for testing. So I chose another one like 420. What do you think?

tthvo:
Converting to draft to add a step that adds credentials in.

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
andrewazores: Hmm. That's a little unfortunate. Is there any reasonable way to keep the backend emitting a `427` but refactor the frontend so that that status code can be caught and ignored, or handled with a different error handler which doesn't cause the modal to appear?

tthvo: I think there could be a way. Maybe on the front-end, we could check the `dryrun` query-parameter is available and set to true? I could try to test it out with the `url` field of the returned response.

andrewazores: That's possible, though it means that `Api.service` that handles the `427` is going to become a bit tightly coupled to this specific notion of a `dryrun` query on a request, which probably only happens in this one place. Better if an alternate error handler function can get passed from the target creation modal into the Api.service when the request is being made so that the knowledge of the dryrun parameter is kept encapsulated within the component that makes those specific requests with that specific query param.

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
tthvo: I open #1270 to track this :D

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
tthvo: Right, I added some new changes for front-end error handler here: https://github.com/cryostatio/cryostat-web/pull/700

src/main/java/io/cryostat/net/web/http/api/v2/TargetsPostHandler.java
```diff
@@ -156,10 +164,42 @@ public IntermediateResponse<ServiceRef> handle(RequestParameters params) throws
...
+
+            String username = attrs.get("username");
+            String password = attrs.get("password");
+            Optional<Credentials> credentials =
+                    StringUtils.isBlank(username) || StringUtils.isBlank(password)
+                            ? Optional.empty()
+                            : Optional.of(new Credentials(username, password));
+
+            if (!dryRun && credentials.isPresent()) {
+                String matchExpression = String.format("target.connectUrl == '%s'", connectUrl);
```
andrewazores, line 181: https://github.com/cryostatio/cryostat/blob/bcf8181fd0ea1ae73ee4dad582bbc9bb948491d8/src/main/java/io/cryostat/configuration/CredentialsManager.java#L157

tthvo: Right! Updated. Any background on why this method is deprecated?

andrewazores: It's related to the change from target-specific to matchExpression-based JMX credentials. Before your usage of it here it was only used by old, deprecated 2.1 credentials API endpoints, so that those endpoints could still exist but feed into the new matchExpression system underneath.

You can un-deprecate the method since this is another valid usecase for it.

tthvo: Make sense! Thanks for explaining! I undid the deprecation now.

andrewazores (approved):

PR #1265, opened 2022-11-30
chore(smoketest): correct conflicting discovery plugins

### LINKED ISSUE #1253: [Task] smoketest quarkus-test-plugin should be removed gracefully with SIGTERM
Terminating `smoketest.sh` results in having to wait 10 seconds for a quarkus-test-plugin target to terminate, but eventually resorts to be sigkilled.
```
^CNov 16, 2022 11:26:39 PM io.cryostat.core.log.Logger info
INFO: cryostat shutting down...
+ cleanup
+ podman pod stop cryostat-pod
WARN[0010] StopSignal SIGTERM failed to stop container quarkus-test-plugin in 10 seconds, resorting to SIGKILL 
6daf47526ad0266747d946783a406d2addd48baec7a75237c73d308142f7fb79
+ podman pod rm cryostat-pod
6daf47526ad0266747d946783a406d2addd48baec7a75237c73d308142f7fb79
```

andrewazores:
Fixes #1253

The previous setup had two different `quarkus-test` app instances configured to use their own built-in Discovery Plugin implementation in application code, plus another third instance configured to use an early build of the Cryostat Agent to act as the associated Discovery Plugin. These configs created conflicts where two plugins tried to register with the same information, and also resulted in the containers being unable to shut down cleanly because the Agent code does not yet properly handle being improperly initialized and missing some of the expected environment variables/system properties.

These things could be fixed by correcting the configs for the applications' plugin implementations and fixing the broken bad state handling in the agent, but it was quicker, easier, and probably more valuable to just switch the overall setup to using two app instances, both using the Agent implementation with correct configurations and disabled app-level implementations.

### DIFF
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -148,20 +148,6 @@ runDemoApps() {
         --env ORG_ACME_JMXPORT="9097" \
         -d quay.io/andrewazores/quarkus-test:0.0.10
 
-    podman run \
-        --name quarkus-test-plugin-2 \
-        --pod cryostat-pod \
-        --restart unless-stopped \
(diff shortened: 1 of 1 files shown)


maxcao13 (approved):
Problem is fixed, and looks good to me :) just needs a rebase.

Issue #1278, opened 2022-12-06
[Bug] Labels attached to an active recording within an agent-equipped target get removed after plugin re-registration


maxcao13:
## Current Behavior:
There is a bug now however with metadata labels. If you have an Active Recording with labels in an agent equipped discovered target, and the target is pinged to re-register, the associated active metadata files are deleted. 

I investigated a bit and I think it's because when the `DiscoveryPostHandler` calls `DiscoveryStorage#update`, it calls `JvmdIdHelper#resolveId` on all target children of the plugin which always puts into the `ids` LoadingCache. I think when this happens, the `RecordingMetadataManager` is hooked in by the `IdEvent.INVALIDATED` event which is emitted everytime an id is "removed" from the cache (however this is a bug, because the jvmId will actually be the same, so it is replacing itself in the cache with itself), and event is emitted, the metadata manager calls `#removeLostTargetMetadata` which deletes all ActiveRecording metadata associated with the "lost" target.

I believe a simple if check inside the `JvmIdHelper#resolveId`
```java
            future.thenAccept(
                    id -> {
                        if (this.ids.synchronous().get(sr.getServiceUri().toString()).equals(id)) {
                            return;
                        }
                        this.ids.synchronous().put(sr.getServiceUri().toString(), id);
}); # or some simpler prettier logic
```
would probably fix this.

_Originally posted by @maxcao13 in https://github.com/cryostatio/cryostat-agent/pull/9#pullrequestreview-1205485514_
      

## Expected Behavior:
The labels should stay after plugin re-registration.

## Steps To Reproduce:
1. Run the steps outlined in https://github.com/cryostatio/cryostat-agent/pull/9
2. Create an active recording in one of the quarkus test targets.
3. Wait 5 minutes for the plugin to re-register itself. 
4. Refresh the page and the labels will be gone.

## Environment:
- OS: Fedora 36
- Environment: smoketest.sh with JDP
- Version 2.3.0-snapshot 
- Commit: https://github.com/cryostatio/cryostat/commit/4c06c4858259c6b19f0aa9db55d923faa0861233

andrewazores:
There is a variable that can be set to reduce the waiting period on repro step 3:

https://github.com/cryostatio/cryostat/blob/4c06c4858259c6b19f0aa9db55d923faa0861233/src/main/java/io/cryostat/discovery/DiscoveryModule.java#L75

PR #1281, opened 2022-12-06
fix(jvmid): only replace cached entries if changed

andrewazores:
- allow discovery ping period to be set
- fix(jvmid): only replace cached entries if changed

# Welcome to Cryostat! 👋

Fixes: #1278

## Description of the change:
Adds a check after JVM ID computation to only replace the cached entry if the computed ID has changed.

## Motivation for the change:
The cache emits an event when an entry is updated, even if the value associated with the given key is equal to the previous value.
This results in other internal components triggering on the "update". When the cache is updated because a discovery plugin
re-registers, the target applications published by the plugin are refreshed by the ID cache. The false-positive cache update
notifications cause the targets to be considered lost and immediately refound, which clears other state about the known active
recordings present on the targets. Avoiding the false-positive cache update allows the state to be preserved.

## How to manually test:
1. Run `CRYOSTAT_DISCOVERY_PING_PERIOD=15000 sh smoketest.sh...` from before this PR
2. Select one of the `quarkus-test` applications and manually start a recording. Verify that the default expected recording labels are present
3. Select a different target application (or no selection) and wait 15+ seconds
4. Re-select the original `quarkus-test` application and observe that the recording is still present, but the labels are empty
5. Run `CRYOSTAT_DISCOVERY_PING_PERIOD=15000 CRYOSTAT_IMAGE=ghcr.io/... sh smoketest.sh` with this PR's test image
6. Repeat steps 2-4. Verify that the recording labels are preserved.


maxcao13 (approved):
Code looks good, fixes the bug, and ping period is correctly configurable :-)

Issue #1252, opened 2022-11-16
[Task] Multiple similar targets can be found under /targets


maxcao13:
Not sure how to replicate, but sometimes multiple of the same targets can be found when querying the backend.

I assume it has something to do with https://github.com/cryostatio/cryostat/pull/1089 and the TargetNode db since only the jvmIds are different. Seems like targets aren't being cleaned up correctly somehow.

e,g,

```
$ http -v :8181/api/v1/targets 'Authorization: Basic dXNlcjpwYXNzCg'
```
```
[
    {
        "alias": "io.cryostat.Cryostat",
        "annotations": {
            "cryostat": {
                "HOST": "cryostat",
                "JAVA_MAIN": "io.cryostat.Cryostat",
                "PORT": "9091",
                "REALM": "JDP"
            },
            "platform": {}
        },
        "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi",
        "jvmId": "71i3ivTX0ucC9BzsZTMd3ySSPX3q3azC-sgrs5-yV54=",
        "labels": {}
    },
    {
        "alias": "quarkus-test",
        "annotations": {
            "cryostat": {},
            "platform": {}
        },
        "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9197/jmxrmi",
        "jvmId": "owjqR0mWhLKVLrezbSAi_hYWd8MNV8KeBEGvYks6aYg=",
        "labels": {}
    },
    {
        "alias": "quarkus-test",
        "annotations": {
            "cryostat": {},
            "platform": {}
        },
        "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9097/jmxrmi",
        "jvmId": "80YtD5qmf5nD2GAxBZcwMObUHACaB8XdPIBYilG3oXQ=",
        "labels": {}
    },
    {
        "alias": "io.cryostat.Cryostat",
        "annotations": {
            "cryostat": {
                "HOST": "cryostat",
                "JAVA_MAIN": "io.cryostat.Cryostat",
                "PORT": "9091",
                "REALM": "JDP"
            },
            "platform": {}
        },
        "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi",
        "jvmId": "43p-u3RxnzZ7K42EQG39k7T6JoRoin7a5sR56WZKRjM=",
        "labels": {}
    },
    {
        "alias": "quarkus-test",
[...]

PR #1324, opened 2023-01-08
fix(targets): remove jvmId from ServiceRef equals and hash

maxcao13:
# Welcome to Cryostat! 👋
* [ ] Signed the last commit: `git commit --amend --signoff`

Fixes: #1252 

## Description of the change:
Removes jvmId from ServiceRef equals and hash.

## Motivation for the change:
This is because if a target restarts, the jvmId will be different and Cryostat will think it's actually a new target and will add a new TargetNode into the database, resulting in two of the "same" target, one being the old lost target, and the other being the new restarted target. This was changed in https://github.com/cryostatio/cryostat/pull/1189 but it needs to be reverted because of the above reason. 

## How to manually test:
1. Run the PR.
2.  `    podman ps --format "{{.Pod}} || {{.ID}} || {{.RunningFor}} || {{.Names}}"`
3.  Recognize the vertx-fib-demo-1 id and run this:
```bash
podman kill "THE_DEMO_CONTAINER_ID" && 
podman run --name vertx-fib-demo-1 --env HTTP_PORT=8081 --env JMX_PORT=9093 --pod cryostat-pod --rm -d quay.io/andrewazores/vertx-fib-demo:0.9.1
```
4.  `http --verify=no -v :8181/api/v1/targets 'Authorization: Basic dXNlcjpwYXNzCg'` and see there is no extra 9093 target and that Cryostat never sees the original issue in https://github.com/cryostatio/cryostat-web/issues/493

### DIFF
--- a/src/main/java/io/cryostat/platform/ServiceRef.java
+++ b/src/main/java/io/cryostat/platform/ServiceRef.java
@@ -135,7 +135,6 @@ public class ServiceRef {
         }
         ServiceRef sr = (ServiceRef) other;
         return new EqualsBuilder()
-                .append(jvmId, sr.jvmId)
                 .append(serviceUri, sr.serviceUri)
                 .append(alias, sr.alias)
                 .append(labels, sr.labels)
@@ -146,7 +145,6 @@ public class ServiceRef {
     @Override
     public int hashCode() {
         return new HashCodeBuilder()
-                .append(jvmId)
                 .append(serviceUri)
                 .append(alias)
                 .append(labels)


andrewazores:
> This is because if a target restarts, the jvmId will be different and Cryostat will think it's actually a new target and will add a new TargetNode into the database, resulting in two of the "same" target, one being the old lost target, and the other being the new restarted target. 

That sounds like correct behaviour to me. If a target disappears and a new copy (replica) of it appears, that *is* a new target, with a different jvmID and a different actual conceptual identity (which the ID is meant to embody).

The duplicated target definition in the database is actually a symptom of the discovery plugin that provides the target definition behaving badly and not removing definitions for targets when they are lost. In this case it sounds like it's simply because JDP has that intrinsic lag time between a target becoming unreachable and emitting the `LOST` event. If you repeat similar testing using other better-behaved discovery plugins, like Custom Targets or OpenShiftAPI, the duplicate target definition problem shouldn't manifest, or at least not noticeably.

maxcao13:
> The duplicated target definition in the database is actually a symptom of the discovery plugin that provides the target definition behaving badly and not removing definitions for targets when they are lost.

Hmm... I missed that in the original issue then. You're right, there's no problems with me testing now with postgres and h2file. I guess the issue is resolved and I'll close this PR.

Issue #1301, opened 2022-12-22
[Task] Add API endpoint for uploading recordings by JVM ID


andrewazores:
JMX pull-based targets as well as Agent push applications are both associated with unique JVM IDs that are computed the same way. These use various factors to produce what should be a unique ID across all the JVMs that a given Cryostat instance may interact with. This ID is what Cryostat internally uses to subdivide the JFR archives storage into buckets corresponding to each Target JVM. An API endpoint should be added that uses the JVM ID as ex. a path parameter and accepts a single JFR file upload, requiring a standard Authorization check, verifying the upload is a valid JFR file, etc., and then placing the file within the archives subdirectory. The upload should also accept an optional map of labels to associate with the recording. This would allow clients to (re-)upload JFR files to be connected with a specific existing target, as well as allow the Agent to push files into storage linked to its target definition.

Depends on #1300 
Depends on https://github.com/cryostatio/cryostat-agent/pull/4

PR #1325, opened 2023-01-09
feat(archives): allow target-specific archived recording uploads with jvmId parameter

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #1301 

## Description of the change:
Allows recordings to be uploaded to specific target archives from jvmId. Also factors some methods out of `RecordingsPostHandler` and other constant Strings.

## Motivation for the change:
See #1301 

## How to manually test:
1. Run PR normally with smoketest.
2. Check jvmId -> `JVM_ID` of some target. e.g. vertx-fib-demo-1 (cryostat:9093) from TargetSelect dropdown (won't work on targets that need auth -> will be null <-- this needs fixing)
3. have some archived recording relative path saved -> `RECORDING_PATH`
4. Run
```bash
$ https -f --verify=no -v POST :8181/api/beta/recordings/JVM_ID"Authorization: Basic dXNlcjpwYXNzCg==" recording@RECORDING_PATH labels='{"foo": "bar"}'
```
5. Notice recording is uploaded and with labels as well.
6. (There is some front-end bug with duplicate recordings? Refreshing page, fixes this)


src/main/java/io/cryostat/net/web/http/api/beta/RecordingsFromIdPostBodyHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Inject
+    RecordingsFromIdPostBodyHandler(
+            AuthManager auth,
+            CredentialsManager credentialsManager,
+            @Named(MainModule.RECORDINGS_PATH) Path recordingsPath,
+            FileSystem fs,
+            Logger logger) {
+        super(auth, credentialsManager, logger);
+        Path fileUploads = recordingsPath.resolve("file-uploads");
```
andrewazores, line 73: I think `"file-uploads"` should be extracted as a constant somewhere.

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -890,6 +913,27 @@ public void validateSourceTarget(String sourceTarget)
         }
     }
 
+    // Returns serviceUri associated with jvmId
+    public String validateJvmId(String jvmId) throws JvmIdDoesNotExistException {
```
andrewazores, line 917: Method name seems a little misleading, this isn't "validating" the ID so much as doing a reverse lookup.

andrewazores: On that note I also have a reverse lookup implementation that looks like it provides a similar contract here:

https://github.com/cryostatio/cryostat/blob/2c5ac7f91c0a891c32cbd96509f823c3668e1563/src/main/java/io/cryostat/recordings/JvmIdHelper.java

from https://github.com/cryostatio/cryostat/pull/1188

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -1019,6 +1063,204 @@ private long getArchivedTime(String recordingName) {
...
+
+                        handler.handle(makeFailedAsyncResult(t));
+                        return;
+                    }
+
+                    handler.handle(makeAsyncResult(null));
+                });
+    }
+
+    public FileUpload getTempFileUpload(Collection<FileUpload> fileUploads) {
```
andrewazores, line 1198: This method seems misplaced, it should live closer to the HTTP handling code I think since it's dealing specifically with file uploads and how those are encoded in a multipart form.

andrewazores:
> Check jvmId -> JVM_ID of some target. e.g. vertx-fib-demo-1 (cryostat:9093) from TargetSelect dropdown (won't work on targets that need auth -> will be null <-- this needs fixing)

I followed the same test procedure but using Cryostat itself. After I added backend JMX auth credentials to the keyring, `https :8181/api/v2.1/discovery` showed a non-null `jvmId` for the Cryostat target. I think it's working as expected as long as credentials are stored?

maxcao13:
> > Check jvmId -> JVM_ID of some target. e.g. vertx-fib-demo-1 (cryostat:9093) from TargetSelect dropdown (won't work on targets that need auth -> will be null <-- this needs fixing)
> 
> I followed the same test procedure but using Cryostat itself. After I added backend JMX auth credentials to the keyring, `https :8181/api/v2.1/discovery` showed a non-null `jvmId` for the Cryostat target. I think it's working as expected as long as credentials are stored?

I think the problem I see has something to do with the React states (or caching?). If I add credentials to a target, and then check the jvmId right after, I will see the old id. Even if I refresh the page, it will stay like that. I have to explicitly switch targets with the `TargetSelect` and then switch back, and it will update correctly. I've noticed this even happens across Cryostat builds where, old jvmIds will persist on my webpage from previous builds. This might be another caching problem even, I'm not sure.

andrewazores:
Hmm. Something to look into further then. Not directly relevant to this PR anyway.

Any thoughts @tthvo ?

tthvo:
> Hmm. Something to look into further then. Not directly relevant to this PR anyway.

ahh it was actually a problem with the way we save selected target to local storage:

1. `selectTargetFromCache` is outdated (i.e. no awareness of jvmId). Cached target in local storage will be used instead of fetched one so displayed jvmId is outdated.

      https://github.com/cryostatio/cryostat-web/blob/3e2ff76e89cc9b519c2b795d0e45fa41a1eae07a/src/app/TargetSelect/TargetSelect.tsx#L122

2. `TargetSelect` has no means of fetching new target information unless auto-refresh is enabled. So, adding credentials to target (i.e. which computes jvmId in some steps of the current queried endpoint) won't update target info on front-end with new jvm id.

I think a quick way is to save only serviceUrl instead of the whole target in local storage and trigger target info queries whenever auth modal is submitted?

maxcao13:
> I think a quick way is to save only serviceUrl instead of the whole target in local storage and trigger target info queries whenever auth modal is submitted?
This is sort of how it already works in the back-end so I think this would make sense. It also makes sense to only have this work if the credentials are actually valid, so it would piggyback of this change https://github.com/cryostatio/cryostat-web/issues/593

andrewazores (approved):

andrewazores:
@maxcao13 any interest in modifying the -agent to push its harvested files to this new endpoint?

maxcao13:
> @maxcao13 any interest in modifying the -agent to push its harvested files to this new endpoint?

Sure, sounds exciting!

andrewazores:
https://github.com/cryostatio/cryostat-agent/issues/25

Issue #1258, opened 2022-11-23
[Task] Support for editting labels on targets


tthvo:
It looks like the users cannot edit labels of target. Most targets now have empty labels. It would be nice to be able to support this.

```json
{
  "alias": "io.cryostat.Cryostat",
  "annotations": {
      "cryostat": {
          "HOST": "cryostat",
          "JAVA_MAIN": "io.cryostat.Cryostat",
          "PORT": "9091",
          "REALM": "JDP"
      },
      "platform": {}
  },
  "connectUrl": "service:jmx:rmi:///jndi/rmi://cryostat:9091/jmxrmi",
  "jvmId": "71i3ivTX0ucC9BzsZTMd3ySSPX3q3azC-sgrs5-yV54=",
  "labels": {}
}
```

tthvo:
Maybe we could extend to allow users to edit any details (i.e. alias or others if later added). I think the connectUrl can remain fixed.

andrewazores:
This one I'm not totally sure about. Maybe there should be an `annotations.user` subfield, and `labels` could be split into `labels.platform` and `labels.user` as well. I don't think the user should be able to remove or modify the annotations or labels that are copied from the platform (ex. OpenShift labels). This would make it too easy for a user to desync their targets in Cryostat's view from what the platform represents them as, and it would also mean we need some way to track and re-apply these changes if the target database is updated.

Since these target definitions can be published by discovery plugins and we can't assume we have control of those plugins and the labels they publish, these "user labels" need to be stored separately and correlated to targets by name/ID.

Also, any changes to these labels would necessitate re-valuating matchExpressions for rules and credentials against the updated target and its new labels.

PR #1327, opened 2023-01-11
fix(discovery): should send notifications on credential store update

tthvo:
# Welcome to Cryostat! 👋

Fixes:  https://github.com/cryostatio/cryostat-web/issues/805
Depends on https://github.com/cryostatio/cryostat-core/pull/172

## Description of the change:

- [x] Fixed Discovery Storage to always send WS notifications when new Jvmids are successfully computed.
- [x] Updated tests to check if notifications are sent (also apply fixes to check nested nested Environement node).

## Motivation for the change:

Front-end should be able to display newly computed jvmIds.

## How to manually test:
1. `Run CRYOSTAT_IMAGE=quay.io... sh smoketest.sh...`
2.  Choose any target with authentication (and no backend-saved credentials). See jvmId is null.
3. Add credentials.
4. See the updated jvmIds.

### DIFF
--- a/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
@@ -328,80 +328,23 @@ public class DiscoveryStorage extends AbstractPlatformClientVerticle {
         logger.trace("Discovery Update {} ({}): {}", id, plugin.getRealm(), updatedChildren);
         EnvironmentNode currentTree = gson.fromJson(plugin.getSubtree(), EnvironmentNode.class);
 
-        Set<TargetNode> previousLeaves = findLeavesFrom(originalTree);
-        Set<TargetNode> currentLeaves = findLeavesFrom(currentTree);
+        Set<ServiceRef> previousRefs = getRefsFromLeaves(findLeavesFrom(originalTree));
+        Set<ServiceRef> currentRefs = getRefsFromLeaves(findLeavesFrom(currentTree));
 
-        Set<TargetNode> updated = getUpdatedLeaves(previousLeaves, currentLeaves);
-        updated.stream()
-                .map(TargetNode::getTarget)
-                .forEach(sr -> notifyAsyncTargetDiscovery(EventKind.MODIFIED, sr));
+        Set<ServiceRef> updated = ServiceRef.getUpdatedRefs(previousRefs, currentRefs);
+        updated.stream().forEach(sr -> notifyAsyncTargetDiscovery(EventKind.MODIFIED, sr));
 
-        Set<TargetNode> added =
-                removeAllUpdatedLeaves(
-                        getAddedOrUpdatedLeaves(previousLeaves, currentLeaves), updated);
-        added.stream()
-                .map(TargetNode::getTarget)
-                .forEach(sr -> notifyAsyncTargetDiscovery(EventKind.FOUND, sr));
+        Set<ServiceRef> added =
+                ServiceRef.removeAllUpdatedRefs(
+                        ServiceRef.getAddedOrUpdatedRefs(previousRefs, currentRefs), updated);
+        added.stream().forEach(sr -> notifyAsyncTargetDiscovery(EventKind.FOUND, sr));
 
-        Set<TargetNode> removed =
-                removeAllUpdatedLeaves(
-                        getRemovedOrUpdatedLeaves(previousLeaves, currentLeaves), updated);
-        removed.stream()
-                .map(TargetNode::getTarget)
-                .forEach(sr -> notifyAsyncTargetDiscovery(EventKind.LOST, sr));
+        Set<ServiceRef> removed =
+                ServiceRef.removeAllUpdatedRefs(
+                        ServiceRef.getRemovedOrUpdatedRefs(previousRefs, currentRefs), updated);
+        removed.stream().forEach(sr -> notifyAsyncTargetDiscovery(EventKind.LOST, sr));
 
         return currentTree.getChildren();
-    }
-
-    public Set<TargetNode> getAddedOrUpdatedLeaves(
-            Set<TargetNode> previousLeaves, Set<TargetNode> currentLeaves) {
-        Set<TargetNode> added = new HashSet<>(currentLeaves);
-        added.removeAll(previousLeaves);
-        return added;
-    }
-
-    public Set<TargetNode> getRemovedOrUpdatedLeaves(
-            Set<TargetNode> previousLeaves, Set<TargetNode> currentLeaves) {
-        Set<TargetNode> removed = new HashSet<>(previousLeaves);
-        removed.removeAll(currentLeaves);
-        return removed;
-    }
-
-    public Set<TargetNode> getUpdatedLeaves(
-            Set<TargetNode> previousLeaves, Set<TargetNode> currentLeaves) {
-        Set<TargetNode> added = getAddedOrUpdatedLeaves(previousLeaves, currentLeaves);
-        Set<TargetNode> removed = getRemovedOrUpdatedLeaves(previousLeaves, currentLeaves);
-        Set<TargetNode> updated = new HashSet<>();
-
-        // Manual set intersection since ServiceRef also compares jvmId
-        for (TargetNode addedTNode : added) {
-            ServiceRef atServiceRef = addedTNode.getTarget();
-            for (TargetNode removedTNode : removed) {
-                if (Objects.equals(
-                        atServiceRef.getServiceUri(), removedTNode.getTarget().getServiceUri())) {
-                    updated.add(addedTNode);
-                }
-            }
-        }
-
-        return updated;
-    }
-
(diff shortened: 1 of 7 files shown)


tthvo:
DiscoveryStorage previously skip notifications so I enable it now. Just one issue is that because we are emitting LOST event so web-console will de-select the target. Maybe in the future, we can have an UPDATE event instead after #1258?

tthvo:
> front-end probably can use some hacks to cache and wait for the later FOUND event

Though, this is likely gonna cause trouble since other notifications might come in between. Not a good idea then.

> > If that succeeds, then a WebSocket notification should be emitted with the updated targets' state. The frontend can figure out how to merge that updated state into its current model, or it can simply drop its model and query to replace it entirely.

https://github.com/cryostatio/cryostat-web/pull/806#issuecomment-1378710497

Since we have only LOST and ADD for now, any guide on this tho? Maybe a new event UPDATE like above?

andrewazores:
Yes, I think there needs to be an UPDATED/MODIFIED even kind added.

tthvo:
Ready to review again :D

Need to build with:
- web: https://github.com/cryostatio/cryostat-web/pull/812
- core: https://github.com/cryostatio/cryostat-core/pull/172

This should allow target list to update with the new MODIFIED notification and aovid deselecting it abruptly.

andrewazores:
https://github.com/cryostatio/cryostat-core/pull/172 has been merged and should be available momentarily in `v2.18.0`.

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -318,41 +319,98 @@ private List<AbstractNode> modifyChildrenWithJvmIds(
...
+            Set<TargetNode> previousLeaves, Set<TargetNode> currentLeaves) {
+        Set<TargetNode> added = getAddedOrUpdatedLeaves(previousLeaves, currentLeaves);
+        Set<TargetNode> removed = getRemovedOrUpdatedLeaves(previousLeaves, currentLeaves);
+        Set<TargetNode> updated = new HashSet<>();
+
+        // Manual set intersection since ServiceRef also compares jvmId
+        for (TargetNode addedTNode : added) {
+            ServiceRef atServiceRef = addedTNode.getTarget();
+            for (TargetNode removedTNode : removed) {
+                if (new EqualsBuilder()
```
andrewazores, line 381: This could just be `Objects.equals(atServiceRef.getServiceUri(), removedTNode.getTarget().getServiceUri())`.

maxcao13: This is enforcing that a target is modified and two TargetNodes are considered a modification if the ServiceRef serviceUri is the same. I'm wondering does this make sense and is enough? e.g. Do other ServiceRef attributes(not jvmId) need to be compared as well?

tthvo: I am thinking maybe later we allow editting other attributes, for example, alias or labels (not sure how tho)?

src/main/java/io/cryostat/rules/RuleProcessor.java
```diff
@@ -231,6 +231,8 @@ public synchronized void accept(TargetDiscoveryEvent tde) {
             case LOST:
                 deactivate(null, tde.getServiceRef());
                 break;
+            case MODIFIED:
```
maxcao13, line 234: Hmm... I would expect that Rule starting would be broken since there is nothing in this case now, but I just tried it and it worked fine... something I am missing?

tthvo: For MODIFIED state, I am guessing the ServiceRef does not change so it should not affect Rule?

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
andrewazores: ^ I don't think so, those are provided by discovery plugins (even builtins) and I don't want that to be a two-way communication between the plugin and Cryostat to coordinate overriding attributes like labels. Two-way communication to establish and maintain the plugin registration is enough, the rest should just be a one-way data flow from plugin to Cryostat. The other way to do it would be for Cryostat to store these modifications and apply them as overlays atop of whatever data the plugins have pushed, but I'd rather avoid that too. This also opens a whole can of worms for rule activation, credentials matching, etc. that use match expressions and where modifications to targets should cascade into re-evaluations of any resources that have match expressions.

tthvo: Ahh right I remember you explained this that issue too. Seems to cause some trouble. Do you also want to compare other attributes too (except jvmId)?

andrewazores: The targets' URLs should already be unique and distinct from each other on any given update of a plugin's subtree - if there are two leaf nodes with identical URLs then that's a discovery plugin bug. So for the purpose of detecting added/changed/removed elements in the set of leaf nodes, the URL can be considered as a unique ID. It may happen for some discovery plugins, corresponding to some application platforms, that target properties like labels or alias are actually configurable or otherwise runtime dynamic (but this would be implemented by the plugin itself, not through any interface on Cryostat), and so these things are not part of the target's identity.

As I'm saying this I realize that this means that discovery mechanisms like JDP which have an intrinsic lag time between an application going offline and the registration that this has occurred will have to be careful about how they publish discovery subtrees. If the target reappears with the same URL before the discovery mechanism has detected its prior disappearance, then a naive discovery implementation may cause this to register as a `MODIFIED` event followed by a `LOST` event.

tthvo: >  If the target reappears with the same URL before the discovery mechanism has detected its prior disappearance, then a naive discovery implementation may cause this to register as a MODIFIED event followed by a LOST event.

Oh this problem is pretty severe. Is this something Cryostat is encoutering or already implemented to safeguard against?

andrewazores: Well, the original safeguard against this class of problem was the `LOST/FOUND` event pair emitted by the original discovery mechanisms. Now that `MODIFIED` is an option, the discovery mechanisms probably need to handle it in some more intelligent way if this is a possibility for them. I think it's probably only a likely problem for JDP right now.

andrewazores: While we're here, the other built-in discovery mechanisms like `KubeApiPlatformClient` also need updating to emit `MODIFIED` events instead of the `LOST/FOUND` pair.

src/main/java/io/cryostat/rules/RuleProcessor.java
tthvo: Make sense thanks :D

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
tthvo: Ahh make sense, thanks :D I refactored and added MOFIDIFED for `KubeApiPlatformClient` now. Stuck in test but forgot to reset latch :D

src/main/java/io/cryostat/platform/ServiceRef.java
```diff
@@ -218,4 +220,52 @@ public enum AnnotationKey {
         REALM,
         ;
     }
+
+    public static Set<ServiceRef> getAddedOrUpdatedRefs(
```
andrewazores, line 224: What do you think of an interface like this?

```java
Set<ServiceRef> added = ServiceRef.compare(previous).to(curremt).added();
Set<ServiceRef> changed = ServiceRef.compare(previous).to(current).changed();
Set<ServiceRef> removed = ServiceRef.compare(previous).to(current).removed();
```

It's basically just a builder pattern. The implementation would be something like:

```
class ServiceRef {
  ServiceRef.Compare compare(Collection previous) {
    return new Compare(previous);
  }

  static class Compare {
    private Collection previous, current;
    private Compare(Collection previous) {
      this.previous = previous;
    }
    public to(Collection current) {
      this.current = current;
    }

    public added() {
        Set result = new HashSet(current);
        result.removeAll(previous);
        return result;
    }

    ...
  }
}
```

src/main/java/io/cryostat/platform/ServiceRef.java
```diff
@@ -218,4 +220,74 @@ public enum AnnotationKey {
...
+        }
+
+        private Collection<ServiceRef> removeAllUpdatedRefs(
+                Collection<ServiceRef> src, Collection<ServiceRef> updated) {
+            Collection<ServiceRef> tnSet = new HashSet<>(src);
+
+            // Manual removal since ServiceRef also compares jvmId
+            for (ServiceRef srcRef : src) {
+                for (ServiceRef updatedRef : updated) {
+                    if (Objects.equals(srcRef.getServiceUri(), updatedRef.getServiceUri())) {
```
andrewazores, line 284: Maybe this for-loop can be extracted to a private helper method since it looks like the loop implementation is the same here and above.

andrewazores (approved):

Issue #1306, opened 2022-12-22
[Task] Policies for pruning target-specific uploads


andrewazores:
When JFR files are uploaded to the target-specific archives location, there should be some kind of policy definition that Cryostat evaluates each time a file is uploaded and causes one or more previously uploaded files to be deleted in FIFO order. The policy could consider factors such as a maximum number of files, or maximum combined size on disk, or maximum age (time since file creation).

These policies would mostly be applicable for files uploaded by Agents, but could be applied broadly.

How should these policies be defined? As a global Cryostat configuration? As a resource definition akin to Automated Rules with a matchExpression? As a request parameter specified by the client (Agent) performing the latest upload?

maxcao13:
About this, I think having some sort of Automated Rule - like configuration json files would make sense to apply to each target, and having a default global configuration set by env vars or web-client settings at the same time also makes sense. Then the config files would override any matching targets and if multiple config files match the same target, there can be some sort of ordering that the user can apply as well maybe?

andrewazores:
That would be very powerful, but it's also complex. Somewhat complex to implement but also complex for a user to figure out what they're actually building out of their policies. I think let's start somewhere more basic with only one layer of policies, and if there is user demand in the future for a layered approach then it can be implemented later.

Maybe let's tackle it by listing out pros and cons.

Method 1: Policy definition (akin to Automated Rules)
pros:
- dynamic at runtime, user does not need to restart Cryostat to change policies
- different policies can be applied for different targets or classes of targets
- can build a nice UI for them using features already present and planned from Automated Rules and JMX Credentials
cons:
- necessitates evaluation of policies' (at least one of them, possibly all of them) matchExpressions each time a file is pushed, can become expensive since they may need to be loaded from disk/db and the json engine needs to be run

Method 2: Startup Configuration ie env vars
pros:
- simple to implement, simple to understand
- cheap to apply when files are pushed since the behaviour is hardcoded and only a small number of primitive parameters can be varied (and don't vary at runtime)
- can build a readonly UI to show the user what the current policies are and include instructions for how to modify them by setting env vars and restarting
cons:
- inflexible, requires redeployment to modify
- inflexible for different targets or classes of targets, same policy applies to all. This is particularly relevant and restrictive for the multitenant Cryostat story.

[...]

PR #1333, opened 2023-01-16
feat(uploads): add maxFiles configuration policy to jvmId uploaded recordings

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #1306

## Description of the change:
This change allows any client which posts a recording to the target-specific upload handler `RecordingsFromIdPostHandler` to specify the maxFiles parameter which keeps a set amount of archived recording files per target, depending on the given parameter. If the parameter is not specified, then it is defaulted to the env variable `CRYOSTAT_PUSH_MAX_FILES`. If that env var is not specified, the global maxFiles value is `Integer.MAX_INT`.

## Motivation for the change:
See #1306 for discussion.

## How to manually test:
1. Run normally...
2. On a target e.g. (cryostat:9093), start a recording and archive it (turn off archiveOnStop to make it easier).
3. Find the target's `jvmId`.
4. Then ->
```bash
$ https --verify=no -f -v POST :8181/api/beta/recordings/${jvmId} recording@testfolder/jfr/quarkus-test-agent_default_20230109T001426Z.jfr maxFiles=1 --auth=user:pass 
```
Notice the extra new paramter `maxFiles`. If we set it to 1, it will delete all archived files within the target except the one that was included in this upload form. If set to anything above 1, it will keep the old archived recording, plus the recent uploaded one that is included in the form.

Experiment with settings, (e.g. without a `maxFiles`, this should default to the environment variable config that was either set, or defaulted to `Integer.MAX_INT`)


maxcao13:
Thoughts before tests?

src/main/java/io/cryostat/net/web/http/api/beta/RecordingsFromIdPostHandler.java
```diff
@@ -170,6 +178,11 @@ public void handleAuthenticated(RoutingContext ctx) throws Exception {
             throw new ApiException(503, "Recording saving not available");
         }
 
+        int maxFiles = Integer.parseInt(ctx.request().getParam("maxFiles", globalMaxFiles));
```
tthvo, line 181: We need to catch exception where the `maxFiles` are not a number? like `maxFiles=abc`?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -1063,6 +1068,47 @@ public long getArchivedTimeFromTimestamp(String timestamp) {
...
+        }
+        List<String> recordings = fs.listDirectoryChildren(subdirectoryPath);
+        recordings.remove(CONNECT_URL);
+        if (recordings.size() <= maxUploads) {
+            return false;
+        }
+
+        List<Future<ArchivedRecordingInfo>> toDelete =
+                recordings.stream()
+                        .peek(n -> logger.trace("pruneTargetUploads: Recording: {}", n))
```
tthvo, line 1086: Should this log be called after filtering?

src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
```diff
@@ -581,17 +580,13 @@ public boolean deleteReports(String subdirectoryName, String recordingName) {
...
-                        logger.trace("Deleted report {}", reportName);
+                    if (fs.exists(reportPath)) {
+                        fs.deleteIfExists(reportPath);
                     }
                 }
             }
-            return deleted;
+            return true;
         } catch (IOException e) {
-            logger.warn(e);
```
tthvo, line 594: Any reasons behind removing warnings here and above?

maxcao13: The exceptions will happen if there were never any reports generated in the first place, and thus the files aren't created. This would probably happen often in the case of the agent pushing, and the exception it logs is rather big. I can replace it with a smaller warning.

tthvo: Ahh make sense! Thanks for explaining :D

tthvo (commented):
Looks good to me! 

I was just seeing some front-end bug that renders duplicated archives. Try uploading an archive while in archive table view and there will be duplicate recordings.

Below, a single uploaded notifications renders an additional of 3 rows.

Screenshot from 2023-01-16 20-20-10

Probably another hook deps issue that eslint misses.

src/main/java/io/cryostat/recordings/RecordingsModule.java
```diff
@@ -87,6 +87,12 @@ static long provideJmxConnectionTimeoutSeconds(Environment env) {
         return Math.max(1, Long.parseLong(env.getEnv(Variables.JMX_CONNECTION_TIMEOUT, "3")));
     }
 
+    @Provides
+    @Named(Variables.PUSH_MAX_FILES_ENV)
+    static String providePushMaxFiles(Environment env) {
```
andrewazores, line 92: The `Integer.parseInt` could be done here so that the sites where the value is injected can just use it as-is.

andrewazores: Oh I see, the injection site actually uses it as a String for a default parameter on the request attributes map. I think it still makes conceptual sense for this value to be converted and injected as an Integer here, and then if the callsite needs it in String form then it can do that conversion...

tthvo:
Spotless?

andrewazores (commented):
Just needs a note added to the README describing the new variable.

tthvo (approved):
Looks great to me :D Worked well.

andrewazores (approved):

Issue #1340, opened 2023-01-25
[Bug] "View Report" failure/hang/timeout on archived recordings if not previously generated


andrewazores:
## Current Behavior:
Report generation fails, appears to hang and then time out on generation in some situations.

## Expected Behavior:
Report generation succeeds and the document is visible in a new tab.

## Steps To Reproduce:
1. `sh smoketest.sh`
2. Select a sample app and start a new recording (say 10 seconds, or any) with any template (I used Profiling). Leave "archive on stop" enabled.
3. Wait for recording to complete and be archived
4. Go to Archives > All Archives > sample target > recording > click three dots menu, "view report". Appears to hang processing and then time out and fail.
5. Go to Archives > All Targets > sample target > recording > three dots "view report" and observe similar behaviour. Logs show:

```
...

Jan 25, 2023 4:19:53 PM org.slf4j.impl.JDK14LoggerAdapter fillCallerData
WARNING: Thread Thread[vert.x-eventloop-thread-2,5,main] has been blocked for 27467 ms, time limit is 2000 ms
io.vertx.core.VertxException: Thread blocked
	at java.base@17.0.5/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@17.0.5/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:252)
	at java.base@17.0.5/java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1866)
	at java.base@17.0.5/java.util.concurrent.ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3463)
	at java.base@17.0.5/java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3434)
	at java.base@17.0.5/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1939)
	at java.base@17.0.5/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
	at app//io.cryostat.net.reports.ArchivedRecordingReportCache.getFromPath(ArchivedRecordingReportCache.java:99)
	at app//io.cryostat.net.reports.ReportService.getFromPath(ReportService.java:58)
	at app//io.cryostat.net.web.http.api.beta.ReportGetFromPathWithJwtHandler.handleWithValidJwt(ReportGetFromPathWithJwtHandler.java:148)
[...]

PR #1341, opened 2023-01-25
fix(reports): fix "view report" failure timing out on jwt report get handlers

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #1340 

## Description of the change:
This change makes some jwt report get handlers non-async. Having them as `async` was causing a timeout due to the vertx event loop being blocked while waiting for a report response from the reports sidecar.

## Motivation for the change:
See #1340 
It makes sense that we wouldn't want these handlers async since we are actively waiting on a http response ourselves and are planning to return it.

## How to manually test:
1. Run cryostat normally with smoketest
2. Create a recording and archive it after a few seconds. 
3. Press view report on the kebab  (this uses the jwt handler) (Do not view a report within the iframe as this caches the report).
4. Notice that the report is generated without a timeout as opposed to before.


andrewazores:
Makes sense. itests seem to be flaky failing for unrelated reasons. Will test and verify as soon as images are available.

andrewazores (approved):

Issue #1092, opened 2022-10-04
[Task] Automated Rules should be JvmID-aware


andrewazores:
https://github.com/cryostatio/cryostat-web/issues/455

Depends on https://github.com/cryostatio/cryostat/pull/1089

PR #1347, opened 2023-01-31
fix(rules): refactor rules to become aware of jvmIds

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #1092 

## Description of the change:
Rules are not aware of jvmIds and will do operations more than once on targets if there exist more than one jvmId unique target within the discovery storage. This PR allows rules to do only do operations on jvmId unique targets with non-null jvmIds. This should work since the discovery storage is updated with non-null jvmIds whenever credentials are posted correctly and targets with the same jvmId are the same jvm. 

However there is one problem in that when Cryostat restarts, the custom targets that are persisted never update their own jvmId when Cryostat comes back up, and so there will be a mismatch of jvmIds and the custom target will have a "stale" id. I'm not sure what is the best way to deal with this. Maybe on built-in discovery startup, test target connections for their jvmId again? I'm not too sure about the race conditions for that problem for now.

## Motivation for the change:
JvmIds are unique to jvms. We don't want rule operations, for example deleting/stopping a recording, to apply more than once on a target jvm as that is extra overhead cost.

## How to manually test:
2. Create a target "cryostat:9093"
3. Create a rule with matchExpression=="true"
4. Enable the rule. 
5. Disable the rule or delete the rule. The logger should no longer show an error.


src/main/java/io/cryostat/platform/PlatformClient.java
```diff
@@ -55,6 +59,19 @@ default void load(Promise<EnvironmentNode> promise) {
 
     List<ServiceRef> listDiscoverableServices();
 
+    default List<ServiceRef> listUniqueReachableServices() {
+        return listDiscoverableServices().stream()
+                .filter((ref) -> ref.getJvmId() != null)
+                .collect(
```
andrewazores, line 65: I think there's a form like this that might be a bit easier to read and perhaps do fewer allocations:

```java
return listDiscoverableServices().stream()
  .filter((ref) -> ref.getJvmId() != null)
  .sorted((ref1, ref2) -> ref1.getJvmId().compareTo(ref2.getJvmId())
  .toList()
  .sorted
```

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
```diff
@@ -255,7 +253,6 @@ public UUID register(String realm, URI callback) throws RegistrationException {
             // FIXME it's not great to perform this action as two separate database calls, but we
             // want to have the ID embedded within the node object. The ID is generated by the
             // database when we create the plugin registration record, and the node object is
-            // serialized into a column of that record.
```
andrewazores, line 258: ?

src/main/java/io/cryostat/configuration/CredentialsManager.java
```diff
@@ -233,19 +229,20 @@ public Set<ServiceRef> resolveMatchingTargets(int id) {
...
         }
-        return matchedTargets;
+        return getMatchedTargets(matchExpression.get(), platformClient.listDiscoverableServices());
+    }
+
+    public Set<ServiceRef> resolveMatchingTargets(String matchExpression) {
+        return getMatchedTargets(matchExpression, platformClient.listDiscoverableServices());
+    }
+
+    public Set<ServiceRef> resolveMatchingUniqueTargets(String matchExpression) {
```
maxcao13, line 244: I'm not actually using this anywhere, I was going to use it for the `CredentialsListener` on `ADDED`, so that when credentials are added, we still check the db for the "most significant" target if there are multiple discoverable targets with the same jvmId. But this has a problem in that the `ADDED` event is also listened by the `DiscoveryStorage` to update the DiscoveryTree with the new non-null jvmId so it becomes a race condition. Would it make sense to try and maybe use the vertx event bus to notify the Credentials listener on the `RuleProcessor` when the db is finished persisting? Or maybe another way?

src/main/java/io/cryostat/discovery/DiscoveryStorage.java
maxcao13: Oops...

src/main/java/io/cryostat/net/web/http/api/v2/RuleDeleteHandler.java
```diff
@@ -162,7 +162,7 @@ public IntermediateResponse<Void> handle(RequestParameters params) throws ApiExc
     }
 
     private void cleanup(RequestParameters params, Rule rule) {
-        storage.listUniqueReachableServices().stream()
+        storage.listUniqueReachableServices().parallelStream()
```
andrewazores, line 165: Just leave these as `stream()` calls, I think. The inner `vertx.executeBlocking` already moves the real work off onto a worker thread, so there is no need to use an implicit/hidden executor to try to do this step in parallel.

src/main/java/io/cryostat/platform/PlatformClient.java
```diff
@@ -60,16 +58,18 @@ default void load(Promise<EnvironmentNode> promise) {
...
-                                        () ->
-                                                new TreeSet<>(
-                                                        Comparator.comparing(
-                                                                ServiceRef::getJvmId))),
-                                ArrayList::new));
+                .filter((ref) -> ref.getJvmId() != null && uniqueIds.add(ref.getJvmId()))
+                .toList();
+    }
+
+    default boolean hasDuplicateTarget(ServiceRef ref) {
```
andrewazores, line 67: Maybe call this `contains`?

src/main/java/io/cryostat/configuration/CredentialsManager.java
andrewazores: > CredentialsListener on ADDED, so that when credentials are added, we still check the db for the "most significant" target if there are multiple discoverable targets with the same jvmId

The only difference this new method would make is in the case where there are multiple definitions for the same target (same ID), *and* the credential match expression matches both of them, I think? If so then I'm not sure how much this is worth worrying about.

> Would it make sense to try and maybe use the vertx event bus to notify the Credentials listener on the RuleProcessor when the db is finished persisting? Or maybe another way?

That would probably work but I would rather try to find some solution that doesn't involve adding more hooks to components and cascading events like this.

src/main/java/io/cryostat/platform/PlatformClient.java
```diff
@@ -55,6 +57,21 @@ default void load(Promise<EnvironmentNode> promise) {
...
     List<ServiceRef> listDiscoverableServices();
 
+    default List<ServiceRef> listUniqueReachableServices() {
+        Set<String> uniqueIds = new HashSet<>();
+        return listDiscoverableServices().stream()
+                .filter((ref) -> ref.getJvmId() != null && uniqueIds.add(ref.getJvmId()))
+                .toList();
+    }
+
+    default boolean containsDuplicateTarget(ServiceRef ref) {
```
andrewazores, line 67: Oh, I meant just name the method simply `contains`, ie define it as `default boolean contains(ServiceRef ref)`.

src/main/java/io/cryostat/configuration/CredentialsManager.java
maxcao13: Yeah, this is a small edge case where you would need identical targets saved, but no credentials for either, and after having a rule that matches both and saving a matching credential for both, the only thing that would happen is that it would start the same recording twice, which doesn't log an error it seems. Seems like JFR just overwrites the active recording so it's not a big issue.

andrewazores: Yea, I think this is a small enough edge case to just not really worry about unless we find a scenario where it actually breaks something.

andrewazores (approved):

PR #1351, opened 2023-02-01
chore(dev): use JAVA_HOME instead of hardcoded path

### LINKED ISSUE #1353: [Bug] `certs/generate-dev-certs.sh` assumes `/usr/lib/jvm/java-11-openjdk`

andrewazores:
Signed-off-by: Andrew Azores <aazores@redhat.com>

# Welcome to Cryostat! 👋

Fixes #1353

## Description of the change:
Replaces a hardcoded path (assumes JDK11 installed in a Fedora-like location) with `$JAVA_HOME`

### DIFF
--- a/certs/generate-dev-certs.sh
+++ b/certs/generate-dev-certs.sh
@@ -48,7 +48,7 @@ keytool \
     -importkeystore \
     -noprompt \
     -storetype PKCS12 \
-    -srckeystore /usr/lib/jvm/java-11-openjdk/lib/security/cacerts \
+    -srckeystore "${JAVA_HOME}/lib/security/cacerts" \
     -srcstorepass changeit \
     -destkeystore "$SSL_TRUSTSTORE" \
     -deststorepass "$SSL_TRUSTSTORE_PASS"


tthvo (approved):
Looks good to me! I think I havent used this script before. What is its use case?

andrewazores:
`sh certs/generate-dev-certs.sh generate` can be used to create a self-signed SSL certificate. The `run.sh` (or `smoketest.sh` which also invokes `run.sh`) script checks for these outputs and if they are found, configures the Cryostat container to use that certificate. When you go to `https://localhost:8181` you'll be presented with a warning from your browser saying that the SSL cert could not be verified because it's self-signed, or something like that, and you can choose to add an exception and ignore the warning. Every subsequent run of `smoketest.sh`/`run.sh` will present the same SSL cert. This is a little bit closer to a real production setup where everything communicates with trusted SSL certs, as opposed to using unencrypted HTTP. You can do `sh certs/generate-dev-certs.sh clean` to undo all this.

tthvo:
Make sense to me! Thanks! I suppose if I don't generate these certs using this script, Cryostat will do it themselves somewhere in `entrypoint.sh`?

andrewazores:
Yup:

https://github.com/cryostatio/cryostat/blob/a01a6d07d55c9614aff699411d3e6956a4de115c/src/main/extras/app/entrypoint.bash#L72

This cert is generated at runtime and doesn't get persisted anywhere permanent, so it would be a different cert each time you re-run `run.sh` (or otherwise start a Cryostat container).

tthvo:
Ahh okayy! Thanks for explaining :D

PR #1359, opened 2023-02-07
feat(activerecordings): add param for restarting existing recording definitions

andrewazores:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat-web/issues/847

## Description of the change:
This change adds parameters for the HTTP API and GraphQL API endpoint/fetcher for starting new recordings. The new parameter allows the client to specify that the recording should be "restarted" (closed/removed remotely and recreated) if a recording with the same name already exists on the target. This same logic is already implemented internally and used by Automated Rules, so this simply extends it to the explicit recording creation API entrypoints.

## Motivation for the change:
This is useful for the embedded charts dashboard views since those need to be backed by an active and running recording. If this recording has been previously created but some client stopped it without removing it, this parameter can be set so that it's quicker and cleaner for the user to trigger the required recording to be restarted and allow the charts dashboard views to be re-enabled.

## How to manually test:
2. Start a new recording on a particular target
3. Try to start another new recording using the same recording name. It should fail with an HTTP 400 response due to the naming conflict.
4. Try again, this time adding the `restart=true` form attribute. The request should now succeed and result in a single recording with the specified attributes being present on the target, regardless of whether the previous recording of the same name was running or stopped.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java
@@ -173,6 +173,10 @@ public class TargetRecordingsPostHandler extends AbstractAuthenticatedRequestHan
                                         recordingOptionsBuilderFactory
                                                 .create(connection.getService())
                                                 .name(recordingName);
+                                boolean restart = false;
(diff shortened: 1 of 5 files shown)


maxcao13 (approved):
Code looks good and the params work through testing.

Issue #1365, opened 2023-02-10
[Task] Consider implicit inclusion of own namespace


ebaron:
Currently, Cryostat includes its own namespace in addition to those specified in `CRYOSTAT_K8S_NAMESPACES`. We should decide on whether we want to keep this assumption or to use only those explictly provided in `CRYOSTAT_K8S_NAMESPACES`.

PR #1374, opened 2023-02-20
fix(k8s): no implicit own-namespace watching

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1365

## Description of the change:
This removes the own-namespace from being implicitly included in the list of watched namespaces in k8s environments.

## Motivation for the change:
The end user, or managing Cryostat Operator, should explicitly list any and all namespaces that they want the Cryostat instance to watch. The implicit inclusion of Cryostat's own namespace restricted users' choices since it was not previously possible to have Cryostat *not* watch that namespace and discover itself.

## How to manually test:
1. Deploy test image in k8s environment
2. Set `CRYOSTAT_K8S_NAMESPACES` environment variable to ex. `CRYOSTAT_K8S_NAMESPACES=myns1,myns2`.
3. Using the HTTP API or web-client, check if Cryostat discovers itself. Repeat steps 2 and 3 including (or not) the Cryostat deployment namespace in the list.

### DIFF
--- a/src/main/java/io/cryostat/platform/internal/OpenShiftPlatformStrategy.java
+++ b/src/main/java/io/cryostat/platform/internal/OpenShiftPlatformStrategy.java
@@ -37,6 +37,8 @@
  */
 package io.cryostat.platform.internal;
 
+import java.util.function.BiFunction;
+
 import io.cryostat.core.log.Logger;
 import io.cryostat.core.net.JFRConnectionToolkit;
 import io.cryostat.core.sys.Environment;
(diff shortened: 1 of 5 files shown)


src/main/java/io/cryostat/platform/internal/OpenShiftPlatformStrategy.java
```diff
@@ -72,32 +65,12 @@ public int getPriority() {
...
-            return false;
-        }
-    }
-
-    @Override
-    public KubeApiPlatformClient getPlatformClient() {
-        logger.info("Selected OpenShift Platform Strategy");
-        return super.getPlatformClient();
+    protected BiFunction<OpenShiftClient, String, ?> testAvailability() {
+        return (client, ns) -> client.routes().inNamespace(ns).list();
```
ebaron, line 69: I wonder if we'd still run into problems if Cryostat can't list routes in its own namespace. Maybe this method would work instead? https://github.com/fabric8io/kubernetes-client/blob/188bd37280ca96ec2a84d5a7bb088e4387f687b3/openshift-client/src/main/java/io/fabric8/openshift/client/impl/OpenShiftClientImpl.java#L736-L742

ebaron (approved):
Looks good!

PR #1388, opened 2023-03-02
fix(credentials): store JMX session credentials in ThreadLocal

andrewazores:
# Welcome to Cryostat! 👋

Fixes: https://github.com/cryostatio/cryostat-web/issues/656

## Description of the change:
This adds a [`ThreadLocal`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ThreadLocal.html) to the `CredentialsManager` containing a `Map<String, Credentials>`. The `String` is just the `targetId` of the current request, if any, and the `Credentials` are likewise the corresponding session credentials `X-JMX-Authorization` extracted from the API request context, if any. The `CredentialsManager`, and for now (needs refactor cleanup) the `ConnectionDescriptor`, reference this `ThreadLocal` to check for any session-provided credentials. If they are present then they are used and preferred over backend stored credentials. Once the current request is completed these credentials are cleared out of the `ThreadLocal Map`. This map is only expected to contain a singleton value currently and should only ever have one key corresponding to the current request's `targetId` URL parameter, but for testing purposes and request safety I used the `Map` to ensure credentials are not accidentally spilling between calls.

This concept is pretty standard in the "thread-per-request" model used by many webservers, where a new thread is spawned for each incoming client request, and that thread handles the entire execution lifespan, finally terminating after the response has been written. Here we use Vert.x which is reactive and processes requests using a main verticle thread and a worker pool. Most of our API handlers, in particular ones that may handle passthrough credentials, are handled by a worker in the Vert.x threadpool. Which exact thread handles the request is not predictable, but and depending on the exact execution path of that request, the thread may end up dropping the request and another thread (the same or another) may pick up the request later, if the request call stack contains something like `vertx.executeBlocking()` or `targetConnectionManager.executeConnectedAsync()`, either of which push the next unit of executable code into a work queue to be picked up by a worker thread. If the request execution is handed off between threads in this way then the session credentials held in the `ThreadLocal` will not be accessible and will appear to be `null` to the thread that picks up the request. If, however, all of these asynchronous delegation calls happen late in the call stack, ie after the point where the original thread has saved the session credentials to `ThreadLocal` and then picked them back up again for the last time, then this does not matter.

## Motivation for the change:
This allows API requests using the old JMX passthrough authentication mechanism, the `X-JMX-Authorization` header, to work as expected again including for requests that trigger executions in the stack in other classes like the `JvmIdHelper` that do not have a direct call from the API handler that can pass through the `RoutingContext`/`RequestParameters` object directly so that the session credentials can be extracted at the call site. As explained above, if all of the requests are structured and executed in the right way, then the session credentials can be stored in a "global" `ThreadLocal` so that as the worker thread processes the request, the session credentials can be retrieved from any point in the call stack, without having to pass the `RoutingContext` object through as a method call argument the whole way through the stack. For reviewers reading this, this can be thought of as somewhat analogous to a React `<Context>` where pieces of state are available to descendant components anywhere below the provider in the component hierarchy. The net effect, if the above assumptions hold, is that the passthrough credentials mechanism works as expected again, without any significant refactoring of the internal interfaces to allow passing the `RoutingContext` through every layer of the call stack, and without resorting to temporary storage of session credentials in the encrypted keyring database.

## How to manually test:
2. Open the `web-client` or use `curl`/`httpie` and exercise as many API calls as possible, without using stored credentials. In the `web-client` this would mean deleting any previously stored credentials, going to `/settings` -> `Advanced` and setting the credentials storage to `Session (Browser Memory)`. Use browser dev tools to inspect outgoing network requests for ex. active recordings queries and ensure that the `X-JMX-Authorization` credentials passthrough header is included in the request and that the request succeeds.


andrewazores:
Still a draft since it needs more testing to ensure it really works as expected. Once we're confident that it does then I will go about refactoring and cleaning it up before final code review.

andrewazores:
ex.

```
$ https :8181/api/v1/targets/localhost:9093/recordings
HTTP/1.1 401 Unauthorized
X-WWW-Authenticate: Basic
content-encoding: gzip
content-length: 83
content-type: text/plain

HTTP Authorization Failure caused by HttpException: Unauthorized
```

----

````
$ https --auth-type=basic --auth=user:pass :8181/api/v1/targets/localhost:9094/recordings
HTTP/1.1 427 Client Error (427)
X-JMX-Authenticate: Basic
content-encoding: gzip
content-length: 104
content-type: text/plain

JMX Authentication Failure caused by FailedLoginException: Invalid username or password
````

----

````
$ https --auth-type=basic --auth=user:pass :8181/api/v1/targets/localhost:9094/recordings X-JMX-Authorization:"Basic $(echo -n admin:adminpass123 | base64)"
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 28

[]
````

**Comment by andrewazores** [comment]
A quick and clear reproducer for the bug is to use session credentials and then try to start a recording on a target that requires credentials, for example Cryostat itself in the smoketest scenario. Before this PR queries like listing active recordings or populating the Events page should work, but attempting to start a recording will fail and present the auth dialog again. With this PR it succeeds and the recording is created as expected.

File: src/main/java/io/cryostat/net/web/http/api/v2/graph/GraphQLPostHandler.java [hunk 1]
```diff
@@ -107,6 +114,78 @@ public void handle(RoutingContext ctx) {
                     .get()) {
                 throw new ApiException(401);
             }
+
+            String targetId = ctx.pathParam("targetId");
```
maxcao13, line 118: I'm not sure but how is the GraphQL handler able to use path params in the requests to `/api/v2.2/graphql`? Especially if we querying on multiple targets at once?

I've been trying GraphQL requests but they don't seem to work either, e.g.
```bash
http --auth-type=basic --auth=user:pass :8181/api/v2.2/graphql query=@my-query.graphql X-JMX-Authorization:"Basic $(echo -n admin:adminpass123 | base64)"
```
and `my-query.graphql`:
```graphql
query { 
    targetNodes(filter: { annotations: "PORT == 9094" }) {
        doStartRecording(recording: { 
            name: "graphql-itest", 
            duration: 30,
            template: "Profiling", 
            templateType: "TARGET", 
            metadata: {
                labels: {
                    newLabel: someValue 
                } 
            } 
        })
        { 
            name state duration 
        }
    } 
}
```
results in 
```bash
...
...
"message": "Exception while fetching data (/targetNodes[0]/doStartRecording) : org.openjdk.jmc.rjmx.ConnectionException caused by java.lang.SecurityException: Authentication failed! Invalid username or password",
...
...
```

andrewazores: Ah good catch. Bad case of copy-paste brain fart. I'll fix it up tomorrow.

andrewazores: I'm actually not sure exactly what I want to do here. This code needs to move into something like the `TargetNodeFetcher`/`TargetNodesRecurseFetcher` I think, but a single request can actually be for multiple different targets. The web-client frontend doesn't have any UI that would allow the user to supply multiple credentials in case the targets may need different credentials, but conceptually that's possible, and the resulting map could be encoded in headers like `X-JMX-Authorization targetId1=encodedCredentials1,targetId2=encodedCredentials2`.

Maybe that's too much and, at least for now, the assumption should just be that if session credentials are passed with a request then they should be valid for *all* targets affected by the request execution.

----

https://github.com/cryostatio/cryostat/blob/73f42066c48fcd4f2cc173aaec15f9a1fec258cc/src/main/java/io/cryostat/net/web/http/api/v2/graph/GraphModule.java#L124

Each stage of the fetcher/mutator pipeline actually runs async and so may be executed by a different thread like I described above, so the `ThreadLocal` won't work properly here if it's set early in the call stack like by the `GraphQLPostHandler`. But, each fetcher/mutator stage can do `        GraphQLHandler.getRoutingContext(environment.getGraphQlContext()).request().getHeader("X-JMX-Authorization");
` to directly check for session credentials, and can set the `ThreadLocal` at that point. This should work equivalently well (or not) as the rest of the solution is working for the HTTP API abstract classes.

andrewazores:
Latest commit should work with GraphQL now. It still uses the same `X-JMX-Authorization` header with a single session credential value, which right now is just presumed valid for all of the targets that the GraphQL query might be executing against. Should be enough to prove the concept. Expanding that further by changing the header value format or adding a different header for multiple mapped targetId/credentials values should not be too hard and shouldn't affect the request execution data flow.

maxcao13:
Most operations now seem to work, looks very nice!

Just some things that probably still need fixing:

1. Automated rules with session stored credentials may need fixing. If my session has credentials for a jmx-authenticated target, starting an automated rule on it will not start a recording.
2. GraphQL jvmId getting. This does not work to get jvmIds:
```bash
http --auth-type=basic --auth=user:pass :8181/api/v2.2/graphql query=@conf/my-query1.graphql X-JMX-Authorization:"Basic $(echo -n admin:adminpass123 | base64)"
```
my-query1:
```graphql
query {
    targetNodes {
        name
        target {
            jvmId
        }
    }
}
```
this results only vertx-fib-demo-1 returning a jvmId, even though we have passed in the jmx-auth header that applies to both demo-2 and demo-3.

maxcao13:
But I think everything else looks really good and feels good to use as well. As far as my testing goes, I think we can move forward with this idea.

andrewazores:
> Automated rules with session stored credentials may need fixing. If my session has credentials for a jmx-authenticated target, starting an automated rule on it will not start a recording.

I think this is just going to stay broken, pretty much intentionally. The session credentials are supposed to only be used for the current client session, which in terms of the web-client means until you close the browser tab essentially, so activating rules based on session credentials would also imply deactivating rules when the session ends, and I don't think that's really a well-defined enough concept to try to support here. The backend only knows about request lifetimes anyway, not actual client sessions, so firing rule activations as soon as a valid credential is passed through a session credential header might seem a little haphazard IMO.

> GraphQL jvmId getting. This does not work to get jvmIds:

Interesting. The GraphQL fetchers/mutators are wrapped in a piece of code that causes each stage to be performed asynchronously by the worker thread pool - I suspect the fact that `targetNodes` returns multiple values means that the next subquery is parallelized into the worker pool and so the `ThreadLocal` doesn't get correctly shared across them.

I'll see if I can figure a way around that.

andrewazores:
> Interesting. The GraphQL fetchers/mutators are wrapped in a piece of code that causes each stage to be performed asynchronously by the worker thread pool - I suspect the fact that targetNodes returns multiple values means that the next subquery is parallelized into the worker pool and so the ThreadLocal doesn't get correctly shared across them.

I think this is true but in fact the cause for the null IDs in the case you gave seems to be even simpler.

The query is looking for the jvmId field of the ServiceRef object which is fetched from the Discovery database at the beginning of the request. The request is already in progress and this object loaded out of the database and into memory by the time we check for session credentials, let alone see that they are present, connect to the target, compute the jvmId, and update the database entry.

The way around here is actually to query for the "live" JVM ID field that is nested under the mbeanMetrics query, this way the ID check is performed at the end of a query execution chain when we have a target connection open, rather than operating over stale data loaded from the database.

Maybe the direct serviceRef jvmId should be removed from the GraphQL schema in favour of this live mbeanMetrics version.

andrewazores:
```graphql
query {
    targetNodes(filter: { annotations: ["PORT in (9094, 9095)"] }) {
        name
        target {
            jvmId
            annotations {
                cryostat
            }
        }
        mbeanMetrics {
            jvmId
        }
    }
}
```

andrewazores:
I left the original ServiceRef jvmId in place because it's probably actually useful to have a way to retrieve it from the database, rather than requiring a JMX connection be opened to check it. We also already have integration tests that depend on that database query and the resulting null jvmId value for targets requiring credentials that aren't stored. This is just a caveat that should be mentioned somewhere. I guess in the GraphQL schema file makes sense?

my-query1.graphql
```diff
@@ -0,0 +1,14 @@
+query {
```
maxcao13, line 1: Remove this file?

maxcao13:
Almost done reviewing, it all looks great! Just doing some nitpicky things. :-)

maxcao13:
I don't have any other comments or review for now, it looks good and working with it in the smoketest works well. Subsequent requests that need authentication work and don't require the web-client to pop up with the addCredentials modal, unless you refresh the page, which makes sense to me.

andrewazores:
I'll save the idea of providing multiple session credentials for a single request for a followup, if there is enough time this release cycle. This at least puts things back to working how they used to and should make session credentials work as expected on all (or at least most?) of the web-client views. Any view that isn't doing GraphQL requests across multiple targets that open connections per query, basically.

maxcao13 (approved):
Sounds good to me. Good work!

andrewazores:
Ah, need to rebase to sign commits. Hang on.

maxcao13 (approved):

andrewazores:
https://github.com/cryostatio/cryostat/actions/runs/4389513949/jobs/7687219684#step:7:960

This looks like it's probably related to this PR. It would seem that the worker thread that serviced the test request was still holding on to the session credentials from a previous request, so the current request was expected to respond 427 (JMX auth failed) and instead was 200 OK. Maybe this happens because the RoutingContext endHandler is called asynchronously, and perhaps in this case of rapid enough calls, it is being called after a followup client request has been received and responded to?

maxcao13:
Makes sense to me, the same thread is definitely being used again.

Issue #1386, opened 2023-03-01
[Request] Updated credentials and rules endpoints


tthvo:
## Describe the feature

In order to support viewing `related resources` for each target in topology view, we need some endpoint that conveniently returns the required information.

- [ ] A credential endpoint that also returns the array of matching targets [1]
- [ ] A rule endpoint that does the same as [1]
- [ ] A credential endpoint that takes a credential and a target and check if the target is matched [2]
- [ ] A rule endpoint that does the same as [2]

andrewazores:
>  A credential endpoint that takes a credential and a target and check if the target is matched [2]

If the credential and the target are both ones that were retrieved from the backend already, ie not ones that the user is currently in the process of creating, then this can just be done on the frontend. The reason why is that if they have been retrieved from the backend then they have already gone through the backend's validation processes, ie. the matchExpression of the credential has gone through the full parsing and sanitization. In the other conversation where I said we should not port the matchExpression validation/evaluation to the frontend I should have clarified that. It's fine for the frontend to simply evaluate a retrieved expression against the list of targets that it has also just queried for. This is essentially already happening when you ex. create an Automated Rule and type in the matchExpression and get immediate feedback about whether the expression is valid and whether it matches the selected target - this is the frontend evaluating the expression against the target. But, in the end that rule still gets submitted to the backend and the expression is fully validated there to ensure it doesn't contain any illegal syntax like function definitions or other control flow.

>  A credential endpoint that also returns the array of matching targets [1]
> A rule endpoint that does the same as [1]

If there was an endpoint that took a matchExpression and evaluated it on the server side, returning the list of matching targets, this would also meet the criteria for both of these, right? It would also be extensible into an expression validation call that the frontend can use to replace or complement the client-side evaluation I talked about above, if the request also took an optional target parameter (or list of targets parameter).

Something like:

```
request:
POST /api/beta/matchExpressions
matchExpression: "target.alias == 'mytarget'"

response:
/api/beta/matchExpression/1
{ "matchExpression": "target.alias == 'mytarget'" }
```

```
request:
DELETE /api/beta/matchExpressions/1

response:
OK
```

```
request:
GET /api/beta/matchExpressions/1

response:
{ "matchExpression": "target.alias == 'mytarget'" }
```


```
request:
GET /api/beta/matchExpressions/1?matches=true

response:
{
  "matchExpression": "target.alias == 'mytarget'",
  "matches": [
    { ... },
    { ... }
  ]
}
```

```
[...]

PR #1389, opened 2023-03-03
feat(match-expr): add `MatchExpression` resource

maxcao13:
# Welcome to Cryostat! 👋

Fixes: https://github.com/cryostatio/cryostat/issues/1386

## Description of the change:
This change adds MatchExpressions as a top level resource to Cryostat. We can use these to query matched targets that are associated with a MatchExpression. This change adds handlers that help with associating targets with rules and targets with credentials and vice versa. 

The endpoints are 

GET /api/beta/matchExpressions (getAll stored MatchExpressions)
GET /""/matchExpression/:id (get stored MatchExpression using id)
POST /""/matchExpressions (add MatchExpression) (with form parameters `matchExpression` and `targets`)
DELETE /""/matchExpressons/:id (delete MatchExpression using id)

#### The POST works like this:
**the `matchExpression` is required**
**the `targets` is optional and if it is a valid array of strings**
- *we do not add the `matchExpression` to the database (simulating a dryrun)*
 - *and we return all the service refs that match the any targetIds in the `targets` string array*

## Motivation for the change:
See https://github.com/cryostatio/cryostat/issues/1386

## How to manually test:
1. Run a smoketest setup.
2. Try these different queries:

**addExpression**
```bash 
https --auth=user:pass -f POST :8181/api/beta/matchExpressions matchExpression="target.alias == 'es.andrewazor.demo.Main'"
```
**getExpression**
```bash
https --auth=user:pass :8181/api/beta/matchExpressions/1?matches=true
```
**getAllExpressions**
```bash
https --auth=user:pass :8181/api/beta/matchExpressions
```
**dryrun getting matchedExpressions** *(also try it without a matching target)*
```bash
https --verify=no --auth=user:pass -f POST :8181/api/beta/matchExpressions matchExpression="target.alias == 'es.andrewazor.demo.Main'" targets="[{\"alias\":\"es.andrewazor.demo.Main\",\"annotations\":{\"cryostat\":{\"HOST\":\"cryostat\",\"JAVA_MAIN\":\"es.andrewazor.demo.Main\",\"PORT\":\"9094\",\"REALM\":\"JDP\"},\"platform\":{}},\"connectUrl\":\"service:jmx:rmi:\/\/\/jndi\/rmi:\/\/cryostat:9094\/jmxrmi\",\"jvmId\":null,\"labels\":{}}]"
```
**deleteExpression**
```bash
https --auth=user:pass DELETE :8181/api/beta/matchExpressions/1
```

### DIFF
--- a/src/main/java/io/cryostat/Cryostat.java
+++ b/src/main/java/io/cryostat/Cryostat.java
@@ -53,6 +53,7 @@ import io.cryostat.messaging.MessagingServer;
 import io.cryostat.net.HttpServer;
 import io.cryostat.net.web.WebServer;
 import io.cryostat.recordings.RecordingMetadataManager;
+import io.cryostat.rules.MatchExpressionManager;
 import io.cryostat.rules.RuleProcessor;
 import io.cryostat.rules.RuleRegistry;
 
@@ -178,6 +179,8 @@ class Cryostat extends AbstractVerticle {
         MessagingServer messagingServer();
 
         RecordingMetadataManager recordingMetadataManager();
+
+        MatchExpressionManager matchExpressionManager();
 
         @Component.Builder
         interface Builder {
--- a/src/main/java/io/cryostat/Cryostat.java
+++ b/src/main/java/io/cryostat/Cryostat.java
@@ -53,7 +53,6 @@ import io.cryostat.messaging.MessagingServer;
 import io.cryostat.net.HttpServer;
 import io.cryostat.net.web.WebServer;
 import io.cryostat.recordings.RecordingMetadataManager;
-import io.cryostat.rules.MatchExpressionManager;
 import io.cryostat.rules.RuleProcessor;
 import io.cryostat.rules.RuleRegistry;
 
@@ -179,8 +178,6 @@ class Cryostat extends AbstractVerticle {
         MessagingServer messagingServer();
 
         RecordingMetadataManager recordingMetadataManager();
-
-        MatchExpressionManager matchExpressionManager();
 
         @Component.Builder
         interface Builder {
--- a/src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionDeleteHandler.java
@@ -63,7 +63,7 @@ import io.vertx.core.http.HttpMethod;
 
 public class MatchExpressionDeleteHandler extends AbstractV2RequestHandler<Void> {
 
-    private final MatchExpressionManager matchExpressionManager;
+    private final MatchExpressionManager expressionManager;
     private final NotificationFactory notificationFactory;
 
     @Inject
@@ -74,7 +74,7 @@ public class MatchExpressionDeleteHandler extends AbstractV2RequestHandler<Void>
             NotificationFactory notificationFactory,
             Gson gson) {
         super(auth, credentialsManager, gson);
-        this.matchExpressionManager = matchExpressionManager;
+        this.expressionManager = matchExpressionManager;
         this.notificationFactory = notificationFactory;
     }
 
@@ -111,16 +111,16 @@ public class MatchExpressionDeleteHandler extends AbstractV2RequestHandler<Void>
     @Override
     public IntermediateResponse<Void> handle(RequestParameters params) throws ApiException {
         int id = Integer.parseInt(params.getPathParams().get("id"));
-        Optional<MatchExpression> matchExpression = matchExpressionManager.get(id);
+        Optional<MatchExpression> matchExpression = expressionManager.get(id);
         if (matchExpression.isEmpty()) {
             return new IntermediateResponse<Void>().statusCode(404);
         }
 
         MatchExpression expr = matchExpression.get();
-        if (this.matchExpressionManager.delete(id)) {
+        if (expressionManager.delete(id)) {
             notificationFactory
                     .createBuilder()
-                    .metaCategory("CredentialsDeleted")
+                    .metaCategory("MatchExpressionDeleted")
                     .metaType(HttpMimeType.JSON)
                     .message(Map.of("id", id, "matchExpression", expr.getMatchExpression()))
                     .build()
--- a/src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionDeleteHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionDeleteHandler.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software (each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ * The above copyright notice and either this complete permission notice or at
+ * a minimum a reference to the UPL must be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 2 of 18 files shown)


src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionGetHandler.java
```diff
@@ -0,0 +1,133 @@
...
+    public IntermediateResponse<MatchedMatchExpression> handle(RequestParameters params)
+            throws ApiException {
+        String matches = params.getQueryParams().get("matches");
+        int id = Integer.parseInt(params.getPathParams().get("id"));
+        Optional<MatchExpression> opt = expressionManager.get(id);
+        if (opt.isEmpty()) {
+            return new IntermediateResponse<MatchedMatchExpression>().statusCode(404);
+        }
+        MatchExpression expr = opt.get();
+        if (StringUtils.isNotBlank(matches)) {
```
andrewazores, line 124: Should this check for 'true'/'false' rather than or in addition to being blank?

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionsGetHandler.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return EnumSet.of(ResourceAction.READ_MATCH_EXPRESSION);
+    }
+
+    @Override
+    public String path() {
+        return basePath() + "matchExpressions";
```
andrewazores, line 97: This can use the `PATH` constant

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionGetHandler.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return EnumSet.of(ResourceAction.READ_MATCH_EXPRESSION);
+    }
+
+    @Override
+    public String path() {
+        return basePath() + "matchExpressions/:id";
```
andrewazores, line 101: `PATH` constant here too

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionDeleteHandler.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Override
+    public Set<ResourceAction> resourceActions() {
+        return EnumSet.of(ResourceAction.DELETE_MATCH_EXPRESSION);
+    }
+
+    @Override
+    public String path() {
+        return basePath() + "matchExpressions/:id";
```
andrewazores, line 103: `PATH` constant

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionsPostHandler.java
```diff
@@ -0,0 +1,183 @@
...
+            throws ApiException {
+        String matchExpression = params.getFormAttributes().get("matchExpression");
+        String targets = params.getFormAttributes().get("targets");
+        if (StringUtils.isBlank(matchExpression)) {
+            throw new ApiException(400, "'matchExpression' is required.");
+        }
+        try {
+            if (StringUtils.isNotBlank(targets)) {
+                Set<ServiceRef> matched;
+                List<String> parsedTargets = this.expressionManager.parseTargets(targets);
```
andrewazores, line 147: I'm not sure about this - what are the reasons for having the client send a list of `connectUrl`s instead of sending a list of whole `ServiceRef` (Target) objects?

src/main/java/io/cryostat/rules/MatchExpression.java
```diff
@@ -0,0 +1,98 @@
...
+    private String matchExpression;
+
+    MatchExpression() {}
+
+    MatchExpression(int id, String matchExpression) {
+        this.id = id;
+        this.matchExpression = matchExpression;
+    }
+
+    MatchExpression(String matchExpression) {
```
andrewazores, line 67: Is this invoked somewhere or required by Hibernate? Having the hardcoded 0 id seems like it could cause problems if misused.

src/main/java/io/cryostat/rules/MatchExpressionManager.java
```diff
@@ -0,0 +1,190 @@
...
+
+    public Set<ServiceRef> resolveMatchingTargets(int id) {
+        Optional<MatchExpression> matchExpression = dao.get(id);
+        if (matchExpression.isEmpty()) {
+            return Set.of();
+        }
+        return resolveMatchingTargets(matchExpression.get().getMatchExpression());
+    }
+
+    public Set<ServiceRef> resolveMatchingTargets(String matchExpression) {
```
andrewazores, line 110: I think both this method and the one above could be replaced by a `resolveMatchingTargets(MatchExpression expr)`. This puts the onus on the caller to first call `get(id)` first or do some other query, but this way the manager isn't accepting/evaluating arbitrary expressions (that's what the MatchExpressionEvaluator is for) and the query potentially coming up with no results only needs to be handled in one place.

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionsPostHandler.java
maxcao13: Hmm... I did it this way because it was easier for a client to supply these since connectUrls should be unique, but it makes sense that if the web client is sending queries with known structured ServiceRef objects, that wouldn't make that much difference.

src/main/java/io/cryostat/rules/MatchExpressionManager.java
maxcao13: I may still need to have a method that accepts a raw string though because of the dry run POST handler without adding the match expression to the backend. The raw string can still have be evaluated by the `MathExpressionEvaluator` , but it bypasses the need to add it to the backend.

src/main/java/io/cryostat/rules/MatchExpression.java
andrewazores: Okay, cool. Yea, it's just an artifact of the way that the entities get persisted with the DAO via Hibernate. The id field gets assigned by the persistence layer later on, so when the application layer is creating a brand new instance to be persisted then the ID doesn't need to be set.

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionsPostHandler.java
andrewazores: > Note: It's also a little broken, anyways since connectUrls may not match a Target, jvmIds would make more sense.

This could lead to a circular dependency. The match expression being added/tested could be for a stored credential, in which case it's likely that Cryostat has never been able to connect to the target before to determine its jvmId.

maxcao13:
```bash
https --verify=no --auth=user:pass -f POST :8181/api/beta/matchExpressions matchExpression="target.alias == 'es.andrewazor.demo.Main'" targets="[{\"alias\":\"es.andrewazor.demo.Main\",\"annotations\":{\"cryostat\":{\"HOST\":\"cryostat\",\"JAVA_MAIN\":\"es.andrewazor.demo.Main\",\"PORT\":\"9094\",\"REALM\":\"JDP\"},\"platform\":{}},\"connectUrl\":\"service:jmx:rmi:\/\/\/jndi\/rmi:\/\/cryostat:9094\/jmxrmi\",\"jvmId\":null,\"labels\":{}}]"
```

That should work with a smoketest setup now.

src/main/java/io/cryostat/net/web/http/api/beta/MatchExpressionsPostHandler.java
```diff
@@ -0,0 +1,183 @@
...
+                        .statusCode(201)
+                        .addHeader(HttpHeaders.LOCATION, String.format("%s/%d", path(), id))
+                        .body(new MatchedMatchExpression(expr));
+            }
+        } catch (RollbackException e) {
+            if (ExceptionUtils.indexOfType(e, ConstraintViolationException.class) >= 0) {
+                throw new ApiException(400, "Duplicate matchExpression", e);
+            }
+            throw new ApiException(500, e);
+        } catch (MatchExpressionValidationException e) {
```
tthvo, line 179: Should we also catch any json parsing exception? For example,

```bash
 $ https --verify=no --auth=user:pass -f POST :8181/api/beta/matchExpressions matchExpression="target.alias == 'es.andrewazor.demo.Main'" targets="[{aasd}]"
HTTP/1.1 500 Internal Server Error
content-encoding: gzip
content-length: 182

{
    "data": {
        "reason": "com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Expected ':' at line 1 column 8 path $[0].aasd"
    },
    "meta": {
        "status": "Internal Server Error",
        "type": "text/plain"
    }
}
```

This should be `400`?

tthvo (commented):
Looks good! Just a comment and waiting on tests if any^^

src/main/java/io/cryostat/rules/MatchExpressionManager.java
```diff
@@ -0,0 +1,211 @@
...
+            return Set.of();
+        }
+        return resolveMatchingTargets(matchExpression.get());
+    }
+
+    public Set<ServiceRef> resolveMatchingTargets(MatchExpression expr) {
+        Set<ServiceRef> matchedTargets = new HashSet<>();
+        for (ServiceRef target : platformClient.listDiscoverableServices()) {
+            try {
+                if (matchExpressionEvaluator.get().applies(expr.getMatchExpression(), target)) {
```
andrewazores, line 115: This looks like it can be refactored in terms of `resolveMatchingTargets(String expr, Collection<ServiceRef> targets)` I think

andrewazores: Or vice-versa.

Basically just the filter function can be either set to that `contains` check or the simple function `t -> true` to pass everything.

src/main/java/io/cryostat/rules/MatchExpressionManager.java
```diff
@@ -0,0 +1,211 @@
...
+            } catch (ScriptException e) {
+                logger.error(e);
+                break;
+            }
+        }
+        return matchedTargets;
+    }
+
+    public List<ServiceRef> parseTargets(String targets) {
+        Objects.requireNonNull(targets, "Targets must not be null");
```
andrewazores, line 150: Is this still worth keeping as a member of the Manager class? Seems like it could just be inlined into the associated Handler.

src/main/java/io/cryostat/Cryostat.java
```diff
@@ -179,6 +180,8 @@ interface Client {
 
         RecordingMetadataManager recordingMetadataManager();
 
+        MatchExpressionManager matchExpressionManager();
```
andrewazores, line 183: It doesn't look like the injected instance is getting used here, so I think it can be removed.

andrewazores:
Looks good otherwise.

andrewazores (approved):

PR #1377, opened 2023-02-21
feat(discoveryplugin): look up plugin callback credentials in database

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] Signed the last commit: `git commit --amend --signoff`
_______________________________________________

Fixes: #1376

## Description of the change:
~~If a discovery plugin (ex. the Agent) has registered using a callback URL that includes userinfo (ex. `http://user:pass@example.com`),
this same userinfo should be included by Cryostat both in the request URL as well as copied to `Authorization` header when Cryostat communicates with that plugin, for example when pinging the plugin to ensure it is still live and reachable.~~

When a discovery plugin wants to register with Cryostat, it will first define a stored credential for itself. Then it will register itself using a special value for the userinfo of the plugin callback URI which references that stored credential. Cryostat looks up that credential and includes it for authentication on the plugin callback requests it makes in the future, both to validate the plugin during registration time as well as periodically later for heartbeat checks.

## Motivation for the change:
Plugins should require an authentication challenge on requests since these requests can trigger the plugin to refresh registration with the Cryostat server. Without an authentication challenge, any client may ping the plugin instance and cause it to resend registration to Cryostat. This causes undue increased load on the Cryostat server for no productive purpose.

## How to manually test:
1. Check out this PR image
2. Check out [associated Agent PR](https://github.com/cryostatio/cryostat-agent/pull/64), build agent into local m2 repo, and rebuild `quarkus-test` container image
3. Run `smoketest.sh` using this PR image and the updated `quarkus-test` with updated `-agent`
4. Verify from logs and target discovery queries that the agent has registered and published itself with generated basic credentials, and that the agent requires these credentials to be passed with requests.

PR #1393, opened 2023-03-08
fix(matchexpr): include jvmId in js binding

andrewazores:
# Welcome to Cryostat! 👋

Related to #1377
Related to #1389

## Description of the change:
This copies the `jvmId` field from the `ServiceRef` object representing a target application into the object binding passed into the JS engine when evaluating a match expression.

## Motivation for the change:
Without this, the `jvmId` property tested in the `matchExpression` will always be an undefined value, leading to expressions that do not properly test and evaluate against the intended targets. In #1377 Cryostat is still able to use the credentials to authenticate to -agents because the lookup is direct by stored credential ID rather than evaluating the associated expression, but this bug does actually manifest in the web-client since it is visible on the `/security` view that the agents' stored credentials display as matching against 0 targets.

## How to manually test:
2. Open web-client and go to `/security`
3. Observe that the two -agent instances' stored credentials should report 1 match each, and expanding the rows should show that each match is for the expected agent target.

### DIFF
--- a/src/main/java/io/cryostat/rules/MatchExpressionEvaluator.java
+++ b/src/main/java/io/cryostat/rules/MatchExpressionEvaluator.java
@@ -162,21 +162,19 @@ public class MatchExpressionEvaluator {
                     serviceRef.getCryostatAnnotations().entrySet()) {
                 cryostatAnnotations.put(entry.getKey().name(), entry.getValue());
             }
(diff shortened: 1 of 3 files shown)


tthvo (approved):
Looks good to me! This fixed the issue with serviceUrl not being matched too.

Issue #421, opened 2021-03-15
Tree view for Targets


andrewazores:
It may make sense to add/replace the current flat list of targets/ServiceRefs with a hierarchical structure to represent the complete landscape that ContainerJFR can see when it is deployed. For example, when deployed in OpenShift/Kubernetes, this could mean the Deployment, Pods and Replicas, Containers/Endpoints. The UI could render this as an interactive tree view and users could perform batch operations on all targets within a Pod by selecting the Pod node, for example, or drill down and perform an operation on a single JVM target by selecting the Endpoint.

Alexjsenn:
In terms of the data structure that should be created and maintained by the platform client, I was thinking of something like a custom tree.

```
public class Object {
    private String name;
    private List<Object> children;
    private List<String> labels;
}
```
Another possibility would be to have a map for the children, but what I am not sure about is what keys to use (names makes sense but is likely not unique).
In the case of Openshift and Kubernetes it might be ideal to have a pointer to the actual object, such as deployment or pod, or it may not be, but I'm not sure if that could be generecized as you suggested. 
This structure can then be sent to the web client to produce a visualization, as well as to allow for group operations on endpoints.

andrewazores:
I do think a tree is the right data structure for this kind of hierarchy. Your proposed node type structure there looks like it's on the right track.

What I roughly had in mind, in JSON(ish) notation, would be something like this for OpenShift/Kubernetes:

```
{
    "nodeType": "Namespace",
    "labels": [
        {
            "key": "name",
            "value": "mynamespace"
        }
    ],
    "children": [
        {
            "nodeType": "Deployment",
            "labels": [
                {
                    "key": "name",
                    "value": "sample-app-deployment", 
                },
                {
                    "key": "app",
                    "value": "sample-app"
                }
            ],
            "children": [
                {
                    "nodeType": "Pod",
                    "labels": [
                        {
                            "key": "name",
                            "value": "sample-app-abc123",
                        },
                        {
                            "key": "app",
                            "value": "sample-app"
                        }
                    ],
                    "children": [
                        { 
                            "nodeType": "Container",
                            labels: [
                                {
                                    "key": "name",
                                    "value": "sample-app-jvm",
                                },
                                {
                                    "key": "app",
                                    "value": "sample-app"
                                }
                            ],
                            "serviceRef": {
                                "jmxUrl": "service:jmx:rmi://localhost:9091/jndi/rmi://fooHost:9091/jmxrmi",
[...]

PR #1394, opened 2023-03-09
feat(discovery): Podman platform

andrewazores:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat/pull/492
Related to https://github.com/cryostatio/cryostat/issues/421

## Description of the change:
I revisited an old, old branch (June *2021* - unfortunately the original commit dates are lost now that this has been rebased and signed) where I was experimenting with talking to the Podman HTTP API from within Cryostat. At the time the Discovery tree querying work was under way, but there was not yet any notion of a persistent database for this data, and no developed system for discovery plugins other than the runtime detection of the single type platform strategy to pick up and use. At that time the original goal of getting a Podman HTTP API discovery query working was just to enable that discovery tree structure development for a contributor who was unable to run `crc` or other heftier systems that can provide a more interesting and non-flat deployment structure as compared to JDP's flat list.
Today we have much more sophistication around discovery, persistence, and modularity, and so I got the Podman HTTP API communication working and wrapped it up into a built-in discovery plugin.

## Motivation for the change:
This may be helpful because it is an additional example of a discovery plugin implementation, but it also removes dependency on JDP which may relieve some work from upstream JMC-core refactoring. JDP is probably only used in toy deployments of Cryostat like the upstream development `smoketest.sh`, so dropping it is unlikely to cause anyone any real issues so long as we have a reasonable development cycle replacement. Querying the Podman API periodically seems to do just as well for this purpose, and as an added bonus also is a little closer to other container runtimes in that we could configure multiple pods along with multiple containers within the pods and emulate something more like an OpenShift deployment (more complex topology).

In this PR the JDP platform client is still present, still named Default, and still has the lowest priority of the built-in plugins. The system will still choose the highest priority available built-in client to run, alongside Custom Targets and the plugin API, so in `smoketest.sh`/`run.sh` the Podman setup will be active rather than JDP. In integration tests the Podman socket is not active and mounted into the container, so the system continues to fall back on the old JDP discovery built-in.

## How to manually test:
2. List targets, view topology if you have that PR checked out, etc.


andrewazores:
Screenshot_2023-03-09_18-57-25

andrewazores:
This + https://github.com/cryostatio/cryostat-web/pull/891 =

image

src/main/java/io/cryostat/net/HttpServer.java
```diff
@@ -82,7 +82,10 @@ public class HttpServer extends AbstractVerticle {
                                         .setPort(netConf.getInternalWebServerPort())
                                         .addWebSocketSubProtocol("*")
                                         .setCompressionSupported(true)
-                                        .setLogActivity(true)));
+                                        .setLogActivity(true)
+                                        .setTcpFastOpen(true)
```
maxcao13, line 86: Just curious what is reason for enabling these tcp features now?

maxcao13:
image

Running the smoketest with the podman socket open results in both the Podman Realm and the JDP realm visible. Does this make sense?

maxcao13:
Hmm... weird but it seems like jvmIds are different for two equivalent JVMs in the two realms. 
image
image

andrewazores:
> image
> 
> Running the smoketest with the podman socket open results in both the Podman Realm and the JDP realm visible. Does this make sense?

Ah right, I forgot to mention. Since the built-in discovery only loads one of these plugins at runtime (plus custom targets, always), it's the Podman realm that is up to date. The JDP realm you're seeing is old data in the database from the last time you ran it and JDP was loaded.

I'm not sure what I want to do about that. Maybe the builtins that are not activated should have their realms deleted automatically?

maxcao13:
Oh, that makes a lot of sense then. Then the jvmID thing is not an issue.

andrewazores:
> I'm not sure what I want to do about that. Maybe the builtins that are not activated should have their realms deleted automatically?

External plugins that go through the API have the ping-pruning mechanism that will result in their realms being deleted if the plugins becomes unreachable however, so they do potentially suffer the same problem. And just in general, if a target disappears, that context is already lost for any archived recordings that came from it, which is why we have separate archived recording metadata that we store. So deleting the builtin plugins realms' doesn't seem too much different.

On a technical level it will take some refactoring, but I don't think it should be too hard to do.

andrewazores:
Okay, last set of commits:

1. Allows multiple platform strategies to be selected. The highest priority available strategy continues to provide the sole AuthManager implementation. All selected strategies provide PlatformClients, which are what the internal implementation calls a built-in discovery plugin still. Custom Targets is special cased and always enabled. JDP happens to have its availability method hardcoded to returning `true`. The Podman client checks for the existence of the Podman API Unix socket. These may all be available and publishing to the discovery database at once, along with any external discovery plugins (like -agents). The previous variables for selecting a specific auth manager, a specific platform strategy, and for disabling built-ins (except custom targets) are all still there and work the same way.
2. Deregisters built-in plugins if they are unavailable on startup. For example, `CRYOSTAT_PLATFORM=io.cryostat.platform.internal.PodmanPlatformStrategy sh smoketest.sh` will cause the internal JDP built-in plugin to be deregistered and its discovery realm deleted from the tree. A subsequent rerun with no `CRYOSTAT_PLATFORM` setting will allow JDP to be selected alongside Podman again and it will re-register again as a new built-in discovery plugin.

andrewazores:
Maybe `CRYOSTAT_PLATFORM` should support a comma-separated list instead of just a single platform now, too.

maxcao13:
Running smoketest just like this: `sh smoketest.sh` runs both the podman client discovery platform and the JDP discovery platform. But running it with the env var `CRYOSTAT_PLATFORM=io.cryostat.platform.internal.PodmanPlatformStrategy sh smoketest.sh` removes the stale JDP built-in targets from the database and runs the podman discovery correctly on its own.

andrewazores:
> Running smoketest just like this: `sh smoketest.sh` runs both the podman client discovery platform and the JDP discovery platform.
> 
> But running it with the env var `CRYOSTAT_PLATFORM=io.cryostat.platform.internal.PodmanPlatformStrategy sh smoketest.sh` removes the stale JDP built-in targets from the database and runs the podman discovery correctly on its own.

Yep, that sounds like it's working as expected.

maxcao13 (approved):
Looks good to me!

ebaron (dismissed):
Dependency addition looks fine

PR #1425, opened 2023-03-23
feat(discovery): add endpoint for plugins to check their registration

andrewazores:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat-agent/issues/85

## Description of the change:
This adds a `GET` endpoint that plugins can use with their existing registration ID and token to check their registration status. The API response is 200 if the plugin is registered and the token is still valid, 401 if the token is invalid, and 404 if the token is somehow valid but the plugin is not registered (this should be impossible - if the plugin is not registered then any token it still has should be invalid).

## Motivation for the change:
This is used by the `-agent` to perform its own periodic heartbeat checks on the Cryostat server at a slower cadence. If the plugin notices that it cannot reach Cryostat, or that Cryostat no longer recognizes it as a registered plugin, then the `-agent` re-enters its original unregistered state and begins trying to more quickly re-register itself with the Cryostat server it is configured for.

## How to manually test:
Patch `shometest.sh` like so:

```diff
diff --git a/smoketest.sh b/smoketest.sh
index 4aebf722..6465b53e 100755
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -179,6 +179,7 @@ runDemoApps() {
         --env CRYOSTAT_AGENT_TRUST_ALL="true" \
         --env CRYOSTAT_AGENT_AUTHORIZATION="Basic $(echo user:pass | base64)" \
         --env CRYOSTAT_AGENT_REGISTRATION_PREFER_JMX="true" \
+        --env CRYOSTAT_AGENT_REGISTRATION_CHECK_MS="15000" \
         --rm -d quay.io/andrewazores/quarkus-test:latest
 
     # copy a jboss-client.jar into /clientlib first
```

to make one of the sample agent applications perform this check more frequently so it's easier to catch.

2. Check the server logs and agent logs. There should be frequent `GET` requests to the new endpoint, which should succeed.
3. TODO figure out a decent way to set up an environment where the Cryostat server can be taken down and restarted without also taking down the agents.

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/DiscoveryRegistrationCheckHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/DiscoveryRegistrationCheckHandler.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright The Cryostat Authors
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or data
+ * (collectively the "Software"), free of charge and under any and all copyright
+ * rights in the Software, and any and all patent rights owned or freely
+ * licensable by each licensor hereunder covering either (i) the unmodified
+ * Software as contributed to or provided by such licensor, or (ii) the Larger
+ * Works (as defined below), to deal in both
+ *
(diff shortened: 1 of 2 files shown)


maxcao13 (approved):
Looks good to me!

tthvo (approved):

Issue #1431, opened 2023-03-27
[Request] Credential test endpoint


tthvo:
## Describe the feature

> Why don't we just have a connection checking endpoint that specifically does not load any stored credentials? It either uses only the session credentials header for the test, or it ignores the session credentials header and requires the credentials to test to be passed as part of the request body. We can use this endpoint and pass no credentials at first, and just try to open a JMX connection to read some basic MBean data. If that succeeds then we know the target did not require auth. Then the user can supply credentials and we can repeat the test using those.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-web/issues/920#issuecomment-1484219792_

## Any other information?

This is useful for the credential test table.

PR #1432, opened 2023-03-27
feat(credentials): add credentials test endpoint

maxcao13:
# Welcome to Cryostat! 👋

Fixes: #1431

## Description of the change:
Adds an endpoint to specifically test JMX connection  w/credentials to a target JVM at `/beta/credentials/:targetId`

The endpoint from the backend should query these targets themselves with the supplied credentials, and give a resulting output as the response body.

They should return status=`NA` for "this target does not need credentials, status=`SUCCESS`, this target is authenticated with these supplied credentials, and status=`FAILURE`, which means these credentials can not be authenticated with these credentials. 

## Motivation for the change:
See #1431

## How to manually test:
1. Run with smoketest setup.
2. Try these queries:

```bash
http --form --auth=user:pass POST localhost:8181/api/beta/credentials/service%3Ajmx%3Armi%3A%2F%2F%2Fjndi%2Frmi%3A%2F%2Flocalhost%3A9095%2Fjmxrmi/ username="admin" password="adminpass123"
```

Try on target that doesn't need credentials as well: e.g. `9093 vertx-fib-demo`

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/beta/CredentialTestPostHandler.java
+++ b/src/main/java/io/cryostat/net/web/http/api/beta/CredentialTestPostHandler.java
@@ -49,6 +49,7 @@ import io.cryostat.net.AuthManager;
 import io.cryostat.net.ConnectionDescriptor;
 import io.cryostat.net.TargetConnectionManager;
 import io.cryostat.net.security.ResourceAction;
+import io.cryostat.net.web.http.AbstractAuthenticatedRequestHandler;
 import io.cryostat.net.web.http.HttpMimeType;
 import io.cryostat.net.web.http.api.ApiVersion;
 import io.cryostat.net.web.http.api.beta.CredentialTestPostHandler.CredentialTestResult;
@@ -143,8 +144,8 @@ public class CredentialTestPostHandler extends AbstractV2RequestHandler<Credenti
                                         conn.connect();
                                         return CredentialTestResult.NA;
                                     }));
-        } catch (Exception e) {
-            if (e.getCause() instanceof SecurityException) {
+        } catch (Exception e1) {
+            if (AbstractAuthenticatedRequestHandler.isJmxAuthFailure(e1)) {
                 ConnectionDescriptor creds =
                         new ConnectionDescriptor(targetId, new Credentials(username, password));
                 try {
@@ -157,15 +158,25 @@ public class CredentialTestPostHandler extends AbstractV2RequestHandler<Credenti
                                                 return CredentialTestResult.SUCCESS;
                                             }));
                 } catch (Exception e2) {
-                    if (e2.getCause() instanceof SecurityException) {
+                    if (AbstractAuthenticatedRequestHandler.isJmxAuthFailure(e2)) {
                         return new IntermediateResponse<CredentialTestResult>()
                                 .body(CredentialTestResult.FAILURE);
                     }
-                    throw new ApiException(500, e2);
+                    throw resolveErrors(e2);
                 }
             }
-            throw new ApiException(500, e);
+            throw resolveErrors(e1);
         }
+    }
+
+    ApiException resolveErrors(Exception e) throws Exception {
+        if (AbstractAuthenticatedRequestHandler.isJmxSslFailure(e)) {
+            return new ApiException(502, "Target SSL Untrusted", e);
+        }
+        if (AbstractAuthenticatedRequestHandler.isServiceTypeFailure(e)) {
+            return new ApiException(504, "Non-JMX Port", e);
+        }
+        return new ApiException(500, "Internal Error", e);
     }
 
     static enum CredentialTestResult {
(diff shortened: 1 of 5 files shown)


src/main/java/io/cryostat/net/web/http/api/beta/CredentialTestPostHandler.java
```diff
@@ -0,0 +1,180 @@
...
+    @Override
+    public IntermediateResponse<CredentialTestResult> handle(RequestParameters params)
+            throws Exception {
+        String targetId = params.getPathParams().get("targetId");
+        String username = params.getFormAttributes().get("username");
+        String password = params.getFormAttributes().get("password");
+        if (StringUtils.isAnyBlank(targetId, username, password)) {
+            StringBuilder sb = new StringBuilder();
+            if (StringUtils.isBlank(targetId)) {
+                sb.append("\"matchExpression\" is required.");
```
tthvo, line 128: targetId/connectUrl is required?

maxcao13: ah whoops...

tthvo (commented):
Looks good to me! Any tests planned?

maxcao13:
Right, I will write some tests, almost forgot.

src/main/java/io/cryostat/net/web/http/api/beta/CredentialTestPostHandler.java
```diff
@@ -0,0 +1,176 @@
...
+                                            (conn) -> {
+                                                conn.connect();
+                                                return CredentialTestResult.SUCCESS;
+                                            }));
+                } catch (Exception e2) {
+                    if (e2.getCause() instanceof SecurityException) {
+                        return new IntermediateResponse<CredentialTestResult>()
+                                .body(CredentialTestResult.FAILURE);
+                    }
+                    throw new ApiException(400, "Invalid credentials", e2);
```
tthvo, line 164: is this error specifically due to invalid credential? Maybe something else? 

Also, should this be `500` status code? Same for line 167?

maxcao13: 167 should probably be 500, but the nested exception there has to be invalid credentials because the connectionManager is querying using the passed form credentials.

maxcao13: Hm.. yes you are right.

tthvo (dismissed):
Looks good to me!

src/main/java/io/cryostat/net/web/http/api/beta/CredentialTestPostHandler.java
```diff
@@ -0,0 +1,177 @@
...
+            return new IntermediateResponse<CredentialTestResult>()
+                    .body(
+                            tcm.executeConnectedTask(
+                                    noCreds,
+                                    (conn) -> {
+                                        conn.connect();
+                                        return CredentialTestResult.NA;
+                                    }));
+        } catch (Exception e1) {
+            if (AbstractAuthenticatedRequestHandler.isJmxAuthFailure(e1)) {
```
andrewazores, line 148: I suspect this utility function will fail in the case of Agent HTTP connections where we have the wrong credentials, but maybe it does happen to work still. Could you test it out? I think this could be done with a procedure like:

```bash
$ sh smoketest.sh h2mem # just to make sure we have a clean database
# open web-client, let everything settle
# delete the stored credentials for both of the agents
# try this test endpoint against the agent. you won't know what the credentials are anyway, so the result should always be a `FAILURE` response.
# you could try rebuilding an agent for testing that prints out its required credentials on startup or something so you can make sure that the success path also works
```

andrewazores: Anyway, this normally should not happen anyway because the Agent should always publish its stored credentials to Cryostat, particularly after https://github.com/cryostatio/cryostat-agent/pull/86 , so it's unlikely this gets hit in practice unless someone is trying to hit it.

tthvo: Is this the test fail we are talking about?

Screenshot from 2023-03-27 21-13-01

tthvo: I am seeing this:

Screenshot from 2023-03-27 21-43-14

andrewazores: NullPointerException is surprising, but I think in general there should be some way we can detect these failures similarly to how we detect the JMX failures. It won't necessarily be 100% robust since it relies on the contents of a stack trace, but it's reasonably stable and better than just throwing raw error messages at the user.

maxcao13:
I've updated the pr.

I believe the failure comes from here https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/AgentClient.java#L234

It seems the AgentConnection `connect`, throws `ConnectionException` which cannot be used to figure out whether the nested exception in the throwable chain was a specific exception, so instead I use the exception message and catching the ConnectionException within the CredentialTest handler.

maxcao13:
I'm not sure what should happen with agent testing credentials though. As of now, if the credentials the agent uses are deleted, any tests will result in an `invalid credentials` message, which makes sense. But if agent credentials exist in the backend, the credentials test will end up as N/A, because the `AgentClient` `invoke` method will always use backend credentials if they exist. 

Do these interactions make sense?

tthvo (approved):

andrewazores (approved):

PR #1451, opened 2023-04-13
fix(itests): Fixing integration tests after jmc core upgrade

Josh-Matsuoka:
Fixes the integration tests after upgrading to jmc-core 8.2.0, there's an extra rule being run (the heap dump rule) so this adjusts the integration tests to check for it.

Depends on https://github.com/cryostatio/cryostat-core/pull/201


andrewazores:
After https://github.com/cryostatio/cryostat-core/pull/201 is merged, Dependabot should open a PR on this repo (and `-reports`) to update the dependency. At that point we'll see the failing tests in the generated PR. I think I can just cherry-pick the commit from your PR here into that Dependabot PR so that the update and the test fix come in one changeset.

PR #1457, opened 2023-04-18
fix(archive-directories): fix incorrect data in archived directory get response

maxcao13:
# Welcome to Cryostat! 👋

Related https://github.com/cryostatio/cryostat/pull/1110

## Description of the change:
Prior to this bug fix, the web ui would render the wrong data. (It was rendering the base32 encoding of the jvmId hash in the tooltip instead of the hash itself.)

Fixed: 
image
image

This should change nothing else other than the http response from the associated get handler.

### DIFF
--- a/docs/HTTP_API.md
+++ b/docs/HTTP_API.md
@@ -2431,24 +2431,6 @@ The handler-specific descriptions below describe how each handler populates the
 | Upload a recording from archive to the Grafana datasource                 | [`RecordingUploadPostHandler`](#RecordingUploadPostHandler-1)                           |
 
 ### Miscellaneous
-* #### `JvmIdGetHandler`
-
-    ##### synopsis
-    Get the unique jvmId for a target JVM. This is a unique identifier for the JVM, and is used to map targetIds their corresponding JVMs.
-
-    ##### request
-    `GET /api/beta/targets/:targetId`
-
-    ##### response
-    `200` - The result is the jvmId.
-
-    `500` - The requested target JVM cannot be reached.
-
-    ##### example
-    ```
-    $ curl http://localhost:8181/api/beta/targets/localhost:0
-    {"meta":{"type":"text/plain","status":"OK"},"data":{"result":"bIeUvB77jlm7MpIAPVhDtqesTNX9a63zLW8IUZFfUug="}}
-    ```
 ### Recordings in Target JVMs
 * #### `TargetRecordingMetadataLabelsPostHandler`
 
--- a/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
+++ b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java
@@ -748,7 +748,7 @@ public class RecordingArchiveHelper {
                                         })
                                 .filter(Objects::nonNull)
                                 .collect(Collectors.toList());
-                directories.add(new ArchiveDirectory(targetId, subdirectoryName, temp));
+                directories.add(new ArchiveDirectory(targetId, jvmId, temp));
             }
             future.complete(directories);
         } catch (ArchivePathException | IOException | InterruptedException | ExecutionException e) {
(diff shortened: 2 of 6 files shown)


tthvo (commented):
Make sense! Seems like some tests need fixing?

maxcao13:
Oh, looks like there's more to this than I thought...

maxcao13:
This requires a front-end fix now, I will push that sometime later.

andrewazores:
> This requires a front-end fix now, I will push that sometime later.

Once that PR is opened, please give it a "Depends on" this one.

andrewazores (approved):

andrewazores (approved):
Looks good. Though, re-reading it now - isn't the directory name just the base32 encoded form of the JVM hash ID? In that case, do we really need to store both of them in the `ArchiveDirectory` structure, or should `RecordingArchiveHelper` line 751 just base32 decode the `subdirectoryName` before passing it in to be stored as the `jvmId`?

tthvo:
Might be wrong but

https://github.com/cryostatio/cryostat/blob/c80a81f6f98ff59936ce4b3ae522cc4bf4280c5b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java#L715

does that and just need to pass in jvmId instead of subdirectoryName on line 571? Wondering what are the use cases of that field tho? Seems like front-end is using just `connectUrl` field to represent the directory.

maxcao13:
> Looks good. Though, re-reading it now - isn't the directory name just the base32 encoded form of the JVM hash ID? In that case, do we really need to store both of them in the `ArchiveDirectory` structure, or should `RecordingArchiveHelper` line 751 just base32 decode the `subdirectoryName` before passing it in to be stored as the `jvmId`?

Hmm, the directory name is just the base32 encoded form but I wanted to have both in the object just for convenience. It makes thinking about it now, that if the requester wants to change subdirectory -> jvmId and back and forth, then they could do it themselves.

I will just keep the field as `jvmId`, and now the front-end fix doesn't need to depend on this pr.

maxcao13:
> Might be wrong but
> 
> https://github.com/cryostatio/cryostat/blob/c80a81f6f98ff59936ce4b3ae522cc4bf4280c5b/src/main/java/io/cryostat/recordings/RecordingArchiveHelper.java#L715
> 
> does that and just need to pass in jvmId instead of subdirectoryName on line 571? Wondering what are the use cases of that field tho? Seems like front-end is using just `connectUrl` field to represent the directory.

`jvmId` is only used as a field for visual on the UI at this point I think.

tthvo:
>  if the requester wants to change subdirectory -> jvmId and back and forth, then they could do it themselves.

Do we document this step anywhere??

> jvmId is only used as a field for visual on the UI at this point I think.

Oh sorry I mean the `subDirectoryName` field :))

maxcao13:
> > if the requester wants to change subdirectory -> jvmId and back and forth, then they could do it themselves.
> 
> Do we document this step anywhere??

Not very well documented, but not really sure if clients outside of the backend need to know this. Users shouldn't be using the archived directories get handler to get the recordings outside of the web-ui.

> 
> > jvmId is only used as a field for visual on the UI at this point I think.
> 
> Oh sorry I mean the `subDirectoryName` field :))

I will just remove it.

tthvo (commented):
Look good and make sense to me!

andrewazores:
> Users shouldn't be using the RecordingGetFromPath related handlers to get the recordings outside of the web-ui.

That's not necessarily true. Any publicly-accessible API is an API that users can depend on, and if you have enough users, someone *will* depend on. Even if the endpoint is labelled as under `/beta`, it's better to be careful and considered about things so that the risk of change is minimized - both for our sakes in maintaining the web-client, but also for any third-party API consumers.

The base32-encoded form is an internal implementation detail about how we store the recordings on disk. The actual piece of information used to identify which target the recording belongs to is the JVM ID, so it makes sense that we report those directly in the JSON API responses without any encoding.

tthvo:
Right that makes sense! Thanks for explaining :))

maxcao13:
I got it now, thanks for the correction!

andrewazores (approved):

Issue #1469, opened 2023-04-24
[Task] Create testing scripts for more scenarios


andrewazores:
`smoketest.sh` is a good quick setup to exercise general basic functionality, but we should also have either 1) more options/flags for that script or 2) more scripts, to exercise things in different ways and various scenarios.

1. Many more target applications via JDP/Podman
2. Many more target applications via Cryostat Agent
3. Many JMX-enabled targets with high-frequency Automated Rules applicable
4. Target applications that are periodically destroyed and recreated
5. Target applications that are periodically stopped and restarted (long enough for the discovery system to lose them)
6. Target applications staying up, Cryostat periodically stopped and restarted, or destroyed and recreated
7. Multiple Cryostat instances deployed in the same environment, ie two in the same Podman pod, and interactions like both having Automated Rules that hit the same set of other targets (and each other)
8. Invalid targets, either by custom targets entries or Podman annotations with wrong URLs, etc.

andrewazores:
https://github.com/cryostatio/cryostat/issues/750

It might be easier to implement at least some of these scenarios as container compose files rather than shell scripts.

PR #1480, opened 2023-04-27
chore(tests): use docker-compose with additional test scripts

maxcao13:
# Welcome to Cryostat! 👋
* [ ] Signed the last commit: `git commit --amend --signoff`

Fixes: https://github.com/cryostatio/cryostat/issues/1469
Uses the Compose Specification and an container engine (podman/docker) to run different Cryostat testing environments.

Requires either 
* [podman-compose 1.0.7+](https://github.com/containers/podman-compose#installation) or [Docker Engine](https://docs.docker.com/engine/install/)

I'm not sure what extra steps need to be done to enable `podman-compose` to work with these changes, because I needed to change some configuration with my machine to get the network between the containers to work, but forgot what to document. Maybe nothing needs to be changed, but I need to know if there are any problems first.

There are still some outgoing issues that I'm not sure how to resolve, or need extra guidance:
- I'm unable to enable SSL for the Cryostat started by the compose file, not sure why. `CRYOSTAT_DISABLE_SSL` is false, but I can still access the web-client at `http:localhost:8181` and unable to at `https://...`
- Having two instances of Cryostat, one of them will take the lock the h2 database and the two also share `conf`, `archives`, and everything else since the same directories/files are mounted. What should happen instead?
- I can't use the compose feature of being able to scale/have replicas of the cryostat agents because of the `callback` environment variable that needs to explicitly be set on itself. This is different from a K8s environment when the Pod is able to give each container their own IP and each container knows it. Any ideas? This is what I am referring to:

```yaml
  quarkus-agent:
    image: quay.io/andrewazores/quarkus-test:latest
...
    environment:
      - CRYOSTAT_AGENT_CALLBACK=http://quarkus-agent:9977/
     # how would the agents known their own callback ahead of time here without creating 20 extra services?
```

## How to test:
1. Checkout this pr `gh pr checkout 1480` if you use github cli.
2. Run `bash compose-test.bash` or `CONTAINER_ENGINE=docker bash compose-test.bash` if you want to use Docker. Run these with arguments as well listed in the README.
3. While the compose containers are up, try running `bash test-restart.bash` to restart cryostat while the other containers stay up.


andrewazores:
> I'm unable to enable SSL for the Cryostat started by the compose file, not sure why. CRYOSTAT_DISABLE_SSL is false, but I can still access the web-client at http:localhost:8181 and unable to at https://...

Looks like this is just a bug. In the `entrypoint` script, this env var is checked for the value `"true"`:

https://github.com/cryostatio/cryostat/blob/f9a7f00f5576a2a81a22ff06a47f0ce20523a5f5/src/main/extras/app/entrypoint.bash#L144

But in the application code that sets up the server's SSL, it just looks for whether the env var is a not-blank string:

https://github.com/cryostatio/cryostat/blob/f9a7f00f5576a2a81a22ff06a47f0ce20523a5f5/src/main/java/io/cryostat/net/SslConfiguration.java#L66

I think the `entrypoint` has it right, the variable should be set to `"true"` to actually turn off the behaviour.

> Having two instances of Cryostat, one of them will take the lock the h2 database and the two also share conf, archives, and everything else since the same directories/files are mounted. What should happen instead?

I think it would be fine to set up the second instance to just have tmp dir mounts for conf/archives/etc. and to use an in-memory H2 database. I think the general purpose of having the duplicate instance would just be for testing what behaviours happen when there are two instances with overlapping discovery regions, overlapping automated rules, etc., so that configuration would be enough for this kind of testing. The primary instance's storage mounts would be the ones to use for easily examining the contents of files that are written and such.

> I can't use the compose feature of being able to scale/have replicas of the cryostat agents because of the callback environment variable that needs to explicitly be set on itself. This is different from a K8s environment when the Pod is able to give each container their own IP and each container knows it. Any ideas? This is what I am referring to:

Yea, I thought something like this could be the case. I can't find anything about compose having an equivalent feature to the k8s Downward API. It will be ugly, but the best bet might be to use the script to generate additional YAML service definitions for agent replicas. This could be done by just having something like a `compose-agent.yaml.in` with some placeholder values in it, then doing an inline patch with `sed` to fill in the placeholder values, and then pipe that patched content into the compose command. This way the bash script knows how many agent instances it is creating and it can loop that many times, using the index to set environment variables on each instance as it goes. You won't get the unique IP this way but I think you should be able to use the containers' unique names as hostnames instead, ex. `cryostat_quarkus-agent_$i`.

andrewazores:
> I'm not sure what extra steps need to be done to enable podman-compose to work with these changes, because I needed to change some configuration with my machine to get the network between the containers to work, but forgot what to document. Maybe nothing needs to be changed, but I need to know if there are any problems first.

FWIW I'm using `podman-compose` and I've been able to `bash compose-test.bash` with no other changes, but I also already have had `podman-compose` set up and working for quite a long time now and don't remember if I had to make any changes when I first set it up.

andrewazores:
Anyway - this is looking good! Once we have it well-sorted out then I think `smoketest.sh` and `run.sh` can just become calls to `compose-test.bash` with particular env vars or flags set.

On the note above about using the script to generate individual YAML files for containers - I think this technique of having a compose YAML for each different kind of container might be generally useful. I think you can do ex. `podman-compose -f compose-cryostat.yaml -f compose-cryostat-reports.yaml -f compose-quarkus-agent.yaml ...` to include different pieces for different scenarios. This technique can also be used to [override configuration](https://docs.docker.com/compose/extends/), so I think for example the base `compose-cryostat` could set up the container to have no `-reports` configuration, and the file that starts the `cryostat-reports` instance can also override configuration on the main Cryostat container to add that env var pointing to the `-reports` container, if that makes sense.

And then I *think* this would play well into using `kompose` to turn these each into k8s YAMLs so that the same files can be used for both development testing in compose as well as CI and e2e testing in `kind`.

maxcao13:
I'll leave the continuation of this to @tthvo until I get back from PTO.

tthvo:
Having a bit of trouble when pushing to Max fork. I think Max might have set some branch protection there.

```bash
To github.com:maxcao13/cryostat.git
 ! [remote rejected]   testing-scripts -> testing-scripts (permission denied)
```

andrewazores:
That's probably just the default branch protection preventing people from just writing over other peoples' repos. When you open a PR against a repo then I think the base repo's maintainers gain the ability to push to the head branch (since I do that on occasion to PRs opened against Cryostat - this would also be a required ability for maintainers to be able to perform rebases or merges into the PR head branch).

You could probably create a separate tracking branch on your fork for the work you carried on and push it there. I think I could then pull it and push that back onto Max's branch, thereby updating this PR in a roundabout way...

andrewazores:
@maxcao13 are you still working on this or is it ready for review?

maxcao13:
It's ready.

compose-test.bash
```diff
@@ -0,0 +1,276 @@
...
+        do
+            current_yaml="$(sed "s/$section/$section-${j}/g" "compose-$section.yaml")"
+            current_yaml="${current_yaml//---/}"; current_yaml="${current_yaml//services:/}"
+            merged_yaml+="$current_yaml"
+        done
+    fi
+done
+
+if $non_zero_replicas; then
+    echo "$merged_yaml" > compose-merged-tmp.yaml
```
andrewazores, line 242: Maybe this file should be written to `/tmp`? ex. using `mktemp`.

Or if not, then perhaps just add a cleanup function that removes it on exit?

compose-cryostat.yaml
```diff
@@ -0,0 +1,203 @@
...
+      - CRYOSTAT_LOG_DB_QUERIES=true
+      - CRYOSTAT_DEV_MODE=true
+    expose:
+      - "9091"
+    ports:
+      - "8282:8282"
+    profiles: ["duplicate"]
+
+  # Testing apps
+  scaled-app:
```
andrewazores, line 151: I'm not sure if I am just doing it wrong, but it seems like these always get included in the deployment for me, even without `-i`/`CT_EN_INVALID`. I am using podman and podman-compose, and just trying ex. `bash compose-test.bash` or `bash compose-test.bash -a 5 -j 5`

andrewazores:
After running the `bash compose-test.bash` and then Ctrl-C'ing it, I see log output like:

```
...
podman rm cryostat-reports
cryostat-reports
exit code: 0
podman rm jfr-datasource
Error: container 50b5a274c2443a62c5e9d48699ce2c2128e9120db6df974999ede972b3bf1446 has dependent containers which must be removed before it: 2367db3e6d879cb2bba77217330bf16aa69d9a7b94f27ad38ef97ed45a524bee: container already exists
exit code: 125
podman rm jmxquarkus
...
```

and `podman ps -a` shows that the `jfr-datasource` container has exited but has not been removed.

compose-cryostat.yaml
maxcao13: Oh, I haven't added those to be able to be configured off yet, thanks for noticing.

maxcao13:
> After running the `bash compose-test.bash` and then Ctrl-C'ing it, I see log output like:
> 
> ```
> ...
> podman rm cryostat-reports
> cryostat-reports
> exit code: 0
> podman rm jfr-datasource
> Error: container 50b5a274c2443a62c5e9d48699ce2c2128e9120db6df974999ede972b3bf1446 has dependent containers which must be removed before it: 2367db3e6d879cb2bba77217330bf16aa69d9a7b94f27ad38ef97ed45a524bee: container already exists
> exit code: 125
> podman rm jmxquarkus
> ...
> ```
> 
> and `podman ps -a` shows that the `jfr-datasource` container has exited but has not been removed.

I removed a `depends_on` declaration in the `grafana` config which depended on `jfr-datasource` to fix this.

compose-test.bash
```diff
@@ -0,0 +1,276 @@
...
+        i)
+            CT_EN_INVALID=true
+            ;;
+        p)
+            export JDBC_URL="jdbc:postgresql://cryostat:5432/cryostat"
+            export JDBC_DRIVER="org.postgresql.Driver"
+            export HIBERNATE_DIALECT="org.hibernate.dialect.PostgreSQL95Dialect"
+            export JDBC_USERNAME="postgres"
+            export JDBC_PASSWORD="abcd1234"
+            export HBM2DDL="update"
```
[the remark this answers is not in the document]
maxcao13: Not yet, will do.

compose-test.bash
maxcao13: Fixed by writing a cleanup function.

maxcao13:
Resolved all review comments, ready again.

andrewazores:
Thanks, I will take another look first thing Monday morning.

compose-test.bash
```diff
@@ -0,0 +1,309 @@
...
+export COMPOSE_PROJECT_NAME
+export COMPOSE_FILE
+
+# handle compose engine
+if [[ -z "${CONTAINER_ENGINE}" ]]; then
+    CONTAINER_ENGINE="podman"
+fi
+
+if [[ "${CONTAINER_ENGINE}" == "podman" ]]; then
+    COMPOSE_ENGINE="podman-compose"
```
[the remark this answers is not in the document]
andrewazores: Actually, it doesn't work with `-p`. Haven't checked on what the problem is yet.

compose-cryostat.yaml
```diff
@@ -0,0 +1,213 @@
...
+      retries: 3
+      start_period: 10s
+      timeout: 5s
+
+  cryostat-duplicate:
+    image: quay.io/cryostat/cryostat:latest
+    container_name: cryostat-duplicate
+    labels: 
+      - io.cryostat.discovery=true
+      - io.cryostat.jmxUrl=service:jmx:rmi:///jndi/rmi://cryostat-duplicate:8282/jmxrmi
```
andrewazores, line 89: Same here re: the three labels

compose-vertx-jmx.yaml
```diff
@@ -0,0 +1,15 @@
...
+  vertx-jmx:
+    image: quay.io/andrewazores/vertx-fib-demo:0.12.3
+    container_name: vertx-jmx
+    deploy:
+      mode: global
+    labels:
+      - io.cryostat.discovery=true
+      - io.cryostat.jmxPort=9093
+      - io.cryostat.jmxHost=vertx-jmx
+      - io.cryostat.jmxUrl=service:jmx:rmi:///jndi/rmi://vertx-jmx:9093/jmxrmi
```
andrewazores, line 12: Same here for the three labels

compose-cryostat-reports.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - CRYOSTAT_REPORT_GENERATOR=http://cryostat-reports:${REPORTS_HTTP_PORT-10001}
+  cryostat-reports:
+    image: ${REPORTS_IMAGE-quay.io/cryostat/cryostat-reports:latest}
+    container_name: cryostat-reports
+    restart: on-failure
+    labels: 
+      - io.cryostat.discovery=true
+      - io.cryostat.jmxPort=${REPORTS_RJMX_PORT-10000}
+      - io.cryostat.jmxHost=cryostat-reports
+      - io.cryostat.jmxUrl=service:jmx:rmi:///jndi/rmi://cryostat-reports:${REPORTS_RJMX_PORT-10000}/jmxrmi
```
andrewazores, line 14: Three labels

andrewazores (commented):
Looking really good. Could you move things into a `compose` subdirectory perhaps? I think basically all of the new files could go in there.

andrewazores (approved):
Very nice work. This should be extremely useful.

Issue #1490, opened 2023-05-11
[Bug] HTTP 415 when creating Automated Rules using `multipart/form-data`


andrewazores:
## Current Behavior:
POSTing a `multipart/form-data` to `/api/v2/rules` always produces an HTTP 415 error response.

## Expected Behavior:
The rule should be created whether submitted as `multipart/form-data`, `application/x-www-form-urlencoded`, or `application/json`.

Related to https://github.com/cryostatio/cryostat/pull/1072

## Steps To Reproduce:
1. `sh run.sh`
2. Try to POST a multipart form:
```
$ curl -vk -F name="firstRule" -F matchExpression="target.labels[\"app\"]==\"jfr-sb-app\"" -F eventSpecifier="template=Profiling,type=TARGET" https://localhost:8181/api/v2/rules -H 'Accept: application/json' -H 'Content-Type: multipart/form-data'
*   Trying 127.0.0.1:8181...
* Connected to localhost (127.0.0.1) port 8181 (#0)
* ALPN: offers h2
* ALPN: offers http/1.1
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS header, Certificate Status (22):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS header, Certificate Status (22):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN: server did not agree on a protocol. Uses default.
* Server certificate:
*  subject: C=CA; O=Cryostat; CN=cryostat
[...]

PR #1499, opened 2023-05-25
fix(multipart): accept multipart/form-data with boundary

aali309:
# Welcome to Cryostat! 👋

Fixes: #1490

## Description of the change:
Accept multipart forms with boundaries when creating automated rules

## Motivation for the change:
This change is helpful because user may send requests with multiple Content-Type header entries including delimiters like semicolons ':' . The automated rule will not fail i.e the content type will discard anything that comes after multipart/form and the semicolon.

## How to manually test:
2. `$ curl -vk -F name="firstRule" -F matchExpression="target.labels[\"app\"]==\"jfr-sb-app\"" -F eventSpecifier="template=Profiling,type=TARGET" https://localhost:8181/api/v2/rules -H 'Accept: application/json' -H 'Content-Type: multipart/form-data'`


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.compile.nullAnalysis.mode": "automatic"
+}
```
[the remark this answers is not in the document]
aali309: ok

src/main/java/io/cryostat/net/web/http/api/v2/RulesPostHandler.java
```diff
@@ -138,8 +138,15 @@ public boolean isOrdered() {
     @Override
     public IntermediateResponse<String> handle(RequestParameters params) throws ApiException {
         Rule rule;
-        HttpMimeType mime =
-                HttpMimeType.fromString(params.getHeaders().get(HttpHeaders.CONTENT_TYPE));
+        String contentType = params.getHeaders().get(HttpHeaders.CONTENT_TYPE);
+
+        if (contentType == null) {
+            contentType = "";
+        }
```
andrewazores, line 145: I think we can collapse the conditions here:

```java
if (contentType != null && contentType.contains(";"))
```

This way the `null` case is handled by the `HttpMimeType.fromString()` later on, and the non-null case is handled here.

Alternatively, you can guard against the null case like this:

```java
String contentType = StringUtils.defaultString(params.getHeaders().get(HttpHeaders.CONTENT_TYPE));
```

This way the `contentType` will never be `null`.

aali309: ok

.gitignore
```diff
@@ -1,6 +1,7 @@
 .gradle/
 .classpath
 .project
+.vscode
```
andrewazores, line 4: Oh, I meant that this would go in your own personal `$HOME/.gitignore` rather than the project's gitignore. Otherwise, we will have to update the project each time a contributor uses a different editor/IDE.

andrewazores: You might also need to update your `$HOME/.gitconfig` to use a personal global gitignore.

```gitconfig
[user]
    email = aazores@redhat.com
    name = Andrew Azores
...
[core]
    excludesfile = ~/.gitignore
...
```

aali309: git config --global core.excludesFile ~/.gitignore .. i did that but the way you did it looks cleaner

src/test/java/io/cryostat/net/web/http/api/v2/RulesPostHandlerTest.java
```diff
@@ -198,6 +198,21 @@ void unknownMimeShouldThrow() {
             MatcherAssert.assertThat(ex.getStatusCode(), Matchers.equalTo(415));
         }
 
+        @Test
+        void shouldAcceptMultipartWithBoundary() {
+            MultiMap headers = MultiMap.caseInsensitiveMultiMap();
+            headers.set(
+                    HttpHeaders.CONTENT_TYPE, "multipart/form-data; boundary=---SOMEBOUNDARY1234");
```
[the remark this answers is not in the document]
aali309: working on this now

.gitignore
andrewazores: That command probably just edits the `~/.gitconfig` the same way that I did it manually.

aali309: yes, the config file now looks like what you did

src/test/java/io/cryostat/net/web/http/api/v2/RulesPostHandlerTest.java
```diff
@@ -218,6 +218,28 @@ void unsupportedFirstMimeShouldThrow() {
             MatcherAssert.assertThat(ex.getStatusCode(), Matchers.equalTo(415));
         }
 
+        @ParameterizedTest
+        @CsvSource(
```
andrewazores, line 222: I think you can use `@ValueSource(strings={...})` in this case. It should do the same thing, but it's a bit more obvious what the intent is. `CsvSource` is good when you need to parameterize multiple arguments to the test method. Here we just need the single contentType parameter.

src/test/java/io/cryostat/net/web/http/api/v2/RulesPostHandlerTest.java
```diff
@@ -218,6 +218,28 @@ void unsupportedFirstMimeShouldThrow() {
...
+              value = {
+	        "multipart/form-data",
+	        "multipart/form-data; boundary=------somecharacters",
+	        "multipart/form-data; unkown characters",
+	        "text/plain; another-directive"
```
andrewazores, line 227: The `text/plain` case will fail because this handler doesn't actually accept that content type. You could move it out to another test with different assertions, or check if there is already another test that covers the same or similar case.

I think more cases like `multipart/form-data; directive1; directive2`, `multipart/form-data;directive` are worth testing as well, and they should ideally pass the same as the rest of your cases here.

andrewazores (approved):
Looks great!

If you're interested, I can also walk you through how to add an integration test so we can be really sure that this is working as intended when the whole application is deployed and interacting over a real network stack with a real HTTP client.

aali309:
> Looks great!
> 
> If you're interested, I can also walk you through how to add an integration test so we can be really sure that this is working as intended when the whole application is deployed and interacting over a real network stack with a real HTTP client.


> Looks great!
> 
> If you're interested, I can also walk you through how to add an integration test so we can be really sure that this is working as intended when the whole application is deployed and interacting over a real network stack with a real HTTP client.

yes that will be great.

andrewazores:
In this case there is already an existing integration test exercising most of the same code, it just missed this particular case.

https://github.com/cryostatio/cryostat/blob/main/src/test/java/itest/RulesPostFormIT.java

Generally, just creating a new file in that directory with the name pattern `SomethingIT.java` is sufficient. The class should probably `extend StandardSelfTest` for convenience, but there are a couple of other options in the `bases` subdirectory.

When the unit tests run it is only testing the direct code under test, just the methods being called in the compiled bytecode. In the integration test harness an actual Cryostat server instance is brought up within a container, and there are tools for adding additional target applications for testing functionality. Rather than calling methods directly like in a unit test, you interact with the server using an HTTP client in a separate process, so it's a lot like the manual testing you have been doing with `curl` where the full stack is actually live and running.

andrewazores:
To build the server and run the integration tests you can do `mvn clean verify`. Once that has been built however, you don't need to rebuild the whole application every time you want to run the tests. `bash repeated-integration-tests.bash` will be useful. This will recompile just the integration tests, then spin up the already-built server container and run the tests against that server again.

Any requests you make to the server in the integration tests should be carefully cleaned up afterward, so that those changes do not affect tests that run later. So for example since your test will be attempting to create a new Automated Rule definition, you should also have an `@AfterAll` cleanup method that deletes that rule, so that when the next integration test class executes it does not find an additional unexpected rule or recordings that may have been created by it.

Once the integration test run finishes (whether tests pass or fail), the database, storage directories, etc. will be dropped, and the next integration test run will use fresh copies of everything. So at least in between test runs things are guaranteed to start from a clean slate, it is only in between individual tests or test classes that cleanup has to be done manually.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,183 @@
...
+
+    static final String TEST_RULE_NAME = "Test_Rule";
+
+    static {
+        NULL_RESULT.put("result", null);
+    }
+
+    @BeforeAll
+    static void setup() throws Exception {
+        testRule = MultiMap.caseInsensitiveMultiMap();
```
andrewazores, line 77: What is this `testRule` used for?

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,183 @@
...
+        testRule.add("name", TEST_RULE_NAME);
+        testRule.add("matchExpression", "target.alias == 'es.andrewazor.demo.Main'");
+        testRule.add("description", "AutoRulesIT automated rule");
+        testRule.add("eventSpecifier", "template=Continuous,type=TARGET");
+    }
+
+    @AfterEach
+    void cleanup() throws Exception {
+        CompletableFuture<JsonObject> deleteResponse = new CompletableFuture<>();
+        webClient.delete(String.format("/api/v2/rules/%s", TEST_RULE_NAME))
```
andrewazores, line 87: Here we do cleanup by deleting a rule named `TEST_RULE_NAME`...

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Test
+    @Order(5)
+    void shouldAcceptMultipartWithBoundarySingleDirective() throws Exception {
+        testMultipartFormData("multipart/form-data; directive");
+    }
+
+    void testMultipartFormData(String contentType) throws ExecutionException, InterruptedException, java.util.concurrent.TimeoutException {
+        MultiMap form = MultiMap.caseInsensitiveMultiMap();
+        form.set("name", "multipart");
```
andrewazores, line 155: Here we will be creating a rule named `multipart`.

andrewazores (changes requested):

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,183 @@
...
+        CompletableFuture<JsonObject> response = new CompletableFuture<>();
+        webClient
+                .post("/api/v2/rules")
+                .putHeader(HttpHeaders.CONTENT_TYPE.toString(), contentType)
+                .sendForm(form, ar -> {
+                    assertRequestStatus(ar, response);
+                });
+    
+                try {
+                    JsonObject result = response.get(5, TimeUnit.SECONDS);
```
andrewazores, line 168: The `response` future is always going to time out here I think, I don't see anywhere that it gets `.complete()`d with a value.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,183 @@
...
+        form.set("name", "multipart");
+        form.set("matchExpression", "false");
+        form.set("eventSpecifier", "template=Continuous");
+    
+        CompletableFuture<JsonObject> response = new CompletableFuture<>();
+        webClient
+                .post("/api/v2/rules")
+                .putHeader(HttpHeaders.CONTENT_TYPE.toString(), contentType)
+                .sendForm(form, ar -> {
+                    assertRequestStatus(ar, response);
```
andrewazores, line 164: The return value from this method should be checked. There are other examples of this method being called in other tests which you can reference.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,172 @@
...
+    static final String TEST_RULE_NAME = "Test_Rule";
+
+    static final Map<String, String> NULL_RESULT = new HashMap<>();
+
+    static {
+        NULL_RESULT.put("result", null);
+    }
+
+    // @BeforeAll
+    // static void setup() throws Exception {}
```
andrewazores, line 71: Please remove this commented block.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,172 @@
...
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.handler.HttpException;
+import itest.bases.StandardSelfTest;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class ShouldAcceptMultipartWithBoundaryIT extends StandardSelfTest {
```
andrewazores, line 61: I think a more conventional name for this test class would be something like `RulesPostMultipartIT` - the `ShouldAccept` would be the convention for an individual `@Test` method, not the whole class.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,172 @@
...
+                                Map.of("type", HttpMimeType.JSON.mime(), "status", "OK"),
+                                "data",
+                                NULL_RESULT));
+
+        try {
+            JsonObject deleteResult = deleteResponse.get(5, TimeUnit.SECONDS);
+            MatcherAssert.assertThat(deleteResult, Matchers.equalTo(expectedDeleteResponse));
+        } catch (TimeoutException e) {
+            System.out.println("Deletion timed out. Reason: " + e.getMessage());
+            throw new RuntimeException("Deletion did not complete within the expected time.", e);
```
andrewazores, line 100: Does the `TimeoutException` need to be wrapped in a `RuntimeException` here? The method signature says `throws Exception` so I think the `TimeoutException` can just be allowed to throw on its own, unless the additional handling here to output log messages is that much more useful than the `TimeoutException` on its own. I would be inclined to just let the `TimeoutException` be - I think the name is fairly self-explanatory, and if it were to occur we would see the stack trace from where it originated as well.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,172 @@
...
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof HttpException) {
+                HttpException httpException = (HttpException) e.getCause();
+                MatcherAssert.assertThat(httpException.getStatusCode(), Matchers.equalTo(400));
+                MatcherAssert.assertThat(
```
andrewazores, line 105: This is in the `@AfterAll` cleanup, not a test case, so I don't think there should be test assertions happening in here. If the case is expected to fail then that behaviour should be captured in the relevant `@Test`, not in the cleanup method.

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,172 @@
...
+                        httpException.getMessage(), Matchers.equalTo("Bad Request"));
+            } else {
+                System.out.println(
+                        "Deletion execution failed. Reason: " + e.getCause().getMessage());
+                throw e;
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
andrewazores, line 113: Why does the current thread need to be interrupted if we are catching an exception due to it just being interrupted by another thread?

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
```diff
@@ -0,0 +1,172 @@
...
+                        ar -> {
+                            if (assertRequestStatus(ar, response)) {
+                                response.complete(ar.result().bodyAsJsonObject());
+                            } else {
+                                response.completeExceptionally(
+                                        new RuntimeException("Request failed"));
+                            }
+                        });
+        try {
+            JsonObject result = response.get(5, TimeUnit.SECONDS);
```
andrewazores, line 149: I think there should be some assertions made about the response here, no?

andrewazores (changes requested):

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
aali309: yes

src/test/java/itest/ShouldAcceptMultipartWithBoundaryIT.java
aali309: yes

src/test/java/itest/RulesPostMultipartIT.java
```diff
@@ -0,0 +1,140 @@
...
+                            }
+                        });
+        try {
+            JsonObject result = response.get(5, TimeUnit.SECONDS);
+            System.out.println("Received response: " + result.toString());
+
+            MatcherAssert.assertThat(result.getJsonObject("meta"), Matchers.notNullValue());
+            MatcherAssert.assertThat(
+                    result.getJsonObject("meta").getString("status"), Matchers.equalTo("Created"));
+            MatcherAssert.assertThat(result.getJsonObject("data"), Matchers.notNullValue());
```
andrewazores, line 134: Might it be worth checking the inner contents of this `data` field to ensure that the response matches the form submitted by the request?

src/test/java/itest/RulesPostMultipartIT.java
```diff
@@ -0,0 +1,140 @@
...
+                            if (assertRequestStatus(ar, response)) {
+                                response.complete(ar.result().bodyAsJsonObject());
+                            } else {
+                                response.completeExceptionally(
+                                        new RuntimeException("Request failed"));
+                            }
+                        });
+        try {
+            JsonObject result = response.get(5, TimeUnit.SECONDS);
+            System.out.println("Received response: " + result.toString());
```
andrewazores, line 129: This println should probably be removed before merge.

src/test/java/itest/RulesPostMultipartIT.java
```diff
@@ -0,0 +1,140 @@
...
+
+            MatcherAssert.assertThat(result.getJsonObject("meta"), Matchers.notNullValue());
+            MatcherAssert.assertThat(
+                    result.getJsonObject("meta").getString("status"), Matchers.equalTo("Created"));
+            MatcherAssert.assertThat(result.getJsonObject("data"), Matchers.notNullValue());
+        } catch (Exception e) {
+            System.err.println("Request failed: " + e.getMessage());
```
andrewazores, line 136: Same question here about catching/logging/rethrowing as I had before about the `TimeoutException`. Why not just let the exception be thrown?

andrewazores (changes requested):

src/test/java/itest/RulesPostMultipartIT.java
aali309: you mean I should get rid of the try?

andrewazores: Yep, pretty much.

src/test/java/itest/RulesPostMultipartIT.java
aali309: for this what Understood is i should just check the existence of the meta field and remove the data?

andrewazores: No, I mean continue checking the data. But check more than just `notNullValue`. Check the contents of `data`, too.

aali309: did i get this  right?
 MatcherAssert.assertThat(
                result.getJsonObject("data").getString("name"), Matchers.equalTo(TEST_RULE_NAME));
            MatcherAssert.assertThat(
                result.getJsonObject("data").getString("matchExpression"), Matchers.equalTo("false"));
            MatcherAssert.assertThat(
                result.getJsonObject("data").getString("eventSpecifier"), Matchers.equalTo("template=Continuous"));

aali309: I mean this is what you mean?

andrewazores: Sure, that looks good.

aali309: i can add one that checks the contentType as well                                                                                                                             result.getJsonObject("data").getString("contentType"), Matchers.equalTo(contentType));

andrewazores (approved):
Nice work! Congrats on your first PR approval :-)

aali309:
Thank you so much @andrewazores for the guidance and help!

Issue #1513, opened 2023-06-01
[Request] Podman Discovery should be able to use container/pod Hostname automatically


andrewazores:
Currently, Podman API discovery requires containers to be labelled like `io.cryostat.connectUrl=service:rmi:jmx://...`. The whole JMX URL, including the hostname and the port number, must be specified in this label. However, the hostname may be automatically generated, in which case the URL cannot be specified in advance. When Cryostat queries the Podman API for the list of running containers it will or can also retrieve information about the hostname of the container. If there is no label specifying the hostname to use for the Cryostat Discovery JMX connection, then the hostname should be inferred from this Podman API information.

PR #1511, opened 2023-06-01
fix(podman): more reliable and informative platform check

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] Signed the last commit: `git commit --amend --signoff`
_______________________________________________

Related to #1510
Related to #1503

## Description of the change:
This does some cleanup and refactoring around the Podman platform strategy/client.

## Motivation for the change:
After this change, Podman strategy selection is more reliable since it checks more required factors before assuming that the `libpod` API is really reachable. If the check cannot be performed due to unavailability of Unix sockets (via netty native transport) then an informative stack trace will be printed.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... sh smoketest.sh...*
2. `https :8181/api/v2.1/discovery` and ensure that the Podman Realm is still populated as normal.
3. Follow the directions in the `README` to put a `jboss-client.jar` into `clientlib/` and restart the Cryostat server. There is already a `wildfly` container in the `smoketest.sh` setup.
4. Check the logs for an exception about a conflict between the `netty` JAR and the `jboss-client.jar`.

PR #1503, opened 2023-05-29
feat(discovery): internal discovery client for Docker API

mwangggg:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] Signed the last commit: `git commit --amend --signoff`
_______________________________________________

Fixes: #1439

PR #1514, opened 2023-06-01
feat(podman): use labels to specify JMX port and optionally host

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1513
Depends on #1511
Related to #1503
Related to #1394

## Description of the change:
This updates the `PodmanPlatformClient` to discover target JMX URLs as follows:
1. Query for containers with the `io.cryostat.discovery` label, then:
2. Check for the `io.cryostat.jmxUrl` label. If it is present, use that. Else continue:
3. Check for the `io.cryostat.jmxPort` label. If it is present, use this to construct the JMX URL.
4. Check for the `io.cryostat.jmxHost` label. If it is not present, default to retrieving the container's hostname from the Podman API. Use this with the `jmxPort` above to construct the JMX URL.

Since there are now two main routes (`jmxUrl`, which replaces the previous `connectUrl`, or `jmxHost`+`jmxPort`), and the Podman HTTP API does not support querying for containers that have any of a list of labels applied, there is a new required label: `io.cryostat.discovery`. This label may have any value. This is the label that Cryostat now queries for against the Podman HTTP API to discover containers.

## Motivation for the change:
As described in #1513, this allows for Podman Discovery to be more flexible and support cases where the container hostname is not known statically when the target application container is started.

## How to manually test:
2. Check that everything still appears in the various discovery API endpoints and Topology view.
3. Change some applications' `--label`s in `smoketest.sh` or `run.sh` according to the description above and ensure they are still picked up after a restart.

### DIFF
--- a/src/main/java/io/cryostat/platform/internal/PodmanPlatformClient.java
+++ b/src/main/java/io/cryostat/platform/internal/PodmanPlatformClient.java
@@ -39,7 +39,6 @@ package io.cryostat.platform.internal;
 
 import java.net.MalformedURLException;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -50,12 +49,18 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 
 import javax.management.remote.JMXServiceURL;
 
 import io.cryostat.core.log.Logger;
+import io.cryostat.core.net.JFRConnectionToolkit;
 import io.cryostat.core.net.discovery.JvmDiscoveryClient.EventKind;
 import io.cryostat.platform.AbstractPlatformClient;
 import io.cryostat.platform.ServiceRef;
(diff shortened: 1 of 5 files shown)


src/main/java/io/cryostat/platform/internal/PodmanPlatformClient.java
```diff
@@ -152,99 +167,148 @@ private void queryContainers() {
...
+                            .forEach(
+                                    spec ->
+                                            notifyAsyncTargetDiscovery(
+                                                    EventKind.FOUND, convert(spec)));
                 });
     }
 
-    private void doPodmanRequest(Consumer<List<ContainerSpec>> successHandler) {
+    private void doPodmanListRequest(Consumer<List<ContainerSpec>> successHandler) {
         URI requestPath = URI.create("http://d/v3.0.0/libpod/containers/json");
```
tthvo, line 188: I am seeing a new `4.0.0` version for this url: https://docs.podman.io/en/latest/_static/api.html?version=v4.5

Not sure what changes are made but response looks pretty much the same to me. I can also access this with podman `3.4.7`.

tthvo (commented):
Looks good! I just have a question about libpod uri above.

src/main/java/io/cryostat/platform/internal/PodmanPlatformClient.java
andrewazores: I've seen that too. I'm not sure what the policy is for podman versions to support older API versions, but I assume that it will remain backward-compatible for some time. So if `3.0.0` does the job and provides the information we need then I think it's better to rely on that version, since more podman versions on more systems will support that than the newer 4.0.0+ API version.

andrewazores: Since https://github.com/cryostatio/cryostat/pull/1511 , the platform detection strategy could also be used to check what API version is available at startup, and then pass the minimum supported version to the platform client after it has determined it and checked that it's within a supported range. Something to look into in the future perhaps.

tthvo: Sounds good! Thanks for explaining!

tthvo (approved):

andrewazores:
Oh, weird. Not sure why my commits aren't signed. Let me try again.

Issue #1433, opened 2023-03-28
[Task] Add GraphQL filter ability for node name from set


andrewazores:
https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/resources/types.graphqls#L19

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java#L69

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java#L108 / https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java#L115

https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java#L97 / https://github.com/cryostatio/cryostat/blob/9bcda3f42ad05349a38d3767f33f11042aa767d2/src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java#L104

I think this functionality can be retrofitted into the existing system pretty cleanly actually:

```diff
diff --git a/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java b/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java
index ee783a7c..a845453d 100644
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/FilterInput.java
@@ -67,6 +67,7 @@ class FilterInput {
     enum Key {
         ID("id"),
         NAME("name"),
+        NAMES("names"),
         LABELS("labels"),
         ANNOTATIONS("annotations"),
         SOURCE_TARGET("sourceTarget"),
diff --git a/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java b/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
[...]

PR #1518, opened 2023-06-05
feat(GraphQL): Add GraphQL filter ability for node name from set

aali309:
# Welcome to Cryostat! 👋

Fixes: #1433

## Description of the change:
*This change adds allows filtering of multiple names instead of just one name*

## Motivation for the change:
*This change is helpful because one can now filter multiple matching names with just one request instead of filtering one name at time for each request*

## How to manually test:
2. * mvn clean verify *
3. * mvn clean verify -Dit=tests=GraphQLIT.java*

### DIFF
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
@@ -100,6 +100,14 @@ class ActiveRecordingsFetcher extends AbstractPermissionedDataFetcher<Active> {
                             .filter(r -> Objects.equals(r.getName(), recordingName))
                             .collect(Collectors.toList());
         }
+        if (filter.contains(FilterInput.Key.NAMES)) {
+            List<String> recordingNames = filter.get(FilterInput.Key.NAMES);
+            recordings =
+                    recordings.stream()
+                            .filter(r -> recordingNames.contains(r.getName()))
+                            .collect(Collectors.toList());
+        }
+
         if (filter.contains(FilterInput.Key.LABELS)) {
             List<String> labels = filter.get(FilterInput.Key.LABELS);
             for (String label : labels) {
--- a/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
+++ b/src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
@@ -110,6 +110,13 @@ class TargetNodesFetcher extends AbstractPermissionedDataFetcher<List<TargetNode
                             .filter(n -> Objects.equals(n.getName(), nodeName))
                             .collect(Collectors.toList());
         }
+        if (filter.contains(FilterInput.Key.NAMES)) {
+            List<String> names = filter.get(FilterInput.Key.NAMES);
+            result =
+                    result.stream()
+                            .filter(n -> names.contains(n.getName()))
+                            .collect(Collectors.toList());
+        }
         if (filter.contains(FilterInput.Key.LABELS)) {
             List<String> labels = filter.get(FilterInput.Key.LABELS);
             for (String label : labels) {
--- a/src/main/resources/types.graphqls
+++ b/src/main/resources/types.graphqls
@@ -10,6 +10,7 @@ scalar Float
 input EnvironmentNodeFilterInput {
     id: Int
     name: String
+    names: [String]
     nodeType: String
     labels: [String]
 }
@@ -17,6 +18,7 @@ input EnvironmentNodeFilterInput {
 input TargetNodesFilterInput {
     id: Int
     name: String
+    names: [String]
     labels: [String]
     annotations: [String]
 }
@@ -24,12 +26,14 @@ input TargetNodesFilterInput {
 input DescendantTargetsFilterInput {
     id: Int
     name: String
+    names: [String]
     labels: [String]
     annotations: [String]
 }
 
 input ActiveRecordingFilterInput {
     name: String
+    names: [String]
     state: String
     continuous: Boolean
     toDisk: Boolean
@@ -42,6 +46,7 @@ input ActiveRecordingFilterInput {
 
 input ArchivedRecordingFilterInput {
     name: String
+    names: [String]
     labels: [String]
     sourceTarget: String
     sizeBytesGreaterThanEqual: Long
--- a/src/test/java/itest/GraphQLIT.java
+++ b/src/test/java/itest/GraphQLIT.java
@@ -37,7 +37,12 @@
  */
 package itest;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.is;
+
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -53,8 +58,13 @@ import java.util.stream.Collectors;
 
 import io.cryostat.MainModule;
 import io.cryostat.core.log.Logger;
+import io.cryostat.net.web.http.HttpMimeType;
 
 import com.google.gson.Gson;
+import io.vertx.core.MultiMap;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import itest.bases.ExternalTargetsTest;
 import itest.util.ITestCleanupFailedException;
@@ -62,6 +72,7 @@ import itest.util.Podman;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.MethodOrderer.OrderAnnotation;
 import org.junit.jupiter.api.Order;
@@ -77,6 +88,8 @@ class GraphQLIT extends ExternalTargetsTest {
 
     static final int NUM_EXT_CONTAINERS = 8;
     static final List<String> CONTAINERS = new ArrayList<>();
+
+    static final String TEST_RECORDING_NAME = "archivedRecording";
 
     @BeforeAll
     static void setup() throws Exception {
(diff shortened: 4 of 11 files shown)


andrewazores:
If the goal of closing and reopening the PR is to get the tests to re-run, please just ping one of @cryostatio/reviewers and we can manually re-trigger the tests. Closing the PR will cause the automation to move it to the Done column on the project board, it might cause other bot automation to re-label it, more CI checks are re-done than need to be, etc.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "java.compile.nullAnalysis.mode": "automatic",
+    "java.configuration.updateBuildConfiguration": "interactive"
```
andrewazores, line 3: Please remove this file

pom.xml
```diff
@@ -661,7 +667,9 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
+
               <argument>until podman ps --filter health=healthy --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.grafana.containerName}; do sleep 5; done</argument>
+
```
andrewazores, line 672: Could you undo the changes to this file please?

src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
```diff
@@ -110,6 +110,13 @@ public List<TargetNode> getAuthenticated(DataFetchingEnvironment environment) th
                             .filter(n -> Objects.equals(n.getName(), nodeName))
                             .collect(Collectors.toList());
         }
+        if (filter.contains(FilterInput.Key.NAMES)) {
+            List<String> names = filter.get(FilterInput.Key.NAMES);
+            result =
+                    result.stream()
+                            .filter(n -> names.contains(n.getName()))
```
andrewazores, line 117: There is another possibility here that I was considering.

Rather than having the `names` filter be a list of exact string matches that the given node's name must be a member of, this could reuse the [LabelSelectorMatcher](https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/net/web/http/api/v2/graph/labels/LabelSelectorMatcher.java). I will give a bit more thought to which approach I would prefer we go for here. Regardless, I suggest you take a look at that class and its unit tests and familiarize yourself with it.

web-client
andrewazores, line 1: Please undo changes to this submodule.

andrewazores (changes requested):

.vscode/settings.json
aali309: done

src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
andrewazores: Maybe a nice flexible way to do this would be:

1. Accept the `name` filter for a single exact match. This is the same thing as `names: ['one-match']` and already done.
2. Accept the `names` filter for a list of exact matches. This is what you have done here.
3. Accept a `nameMatch` filter that accepts the `LabelSelectorMatcher` syntax.

web-client
andrewazores: I don't know if that will help.

I suggest you do `git log web-client` to see what commits have been made to the submodule. You should only see log entries like this:

```
commit feeae5042b194f4ea6f2ceea794731e58a9034e6
Author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
Date:   Thu Jun 8 19:50:58 2023 +0000

    build(web-client): update submodule to db0d4cf

commit 11bfbf6728814f5b6a7e8af55010c86131af9c50
Author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
Date:   Tue Jun 6 19:36:04 2023 +0000

    build(web-client): update submodule to 4889b2b

commit 7e3f5fcf53638cbb8b82e8cf71ac4a52c6b6cced
Author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
Date:   Tue Jun 6 19:22:41 2023 +0000

    build(web-client): update submodule to 0dd044c
```

Since you update the submodule in this PR, there will be an additional commit that you authored in the log. If that commit only updated the submodule then you may simply `git revert abcd1234` the specific commit. You can check what changes were made by that commit with `git show abcd1234`. If the commit made other changes to other files as well, you can try `git revert --no-commit abcd1234` to undo the patch, then use `git add` and `git reset` to stage/unstage changes so that your changes are all still staged except for the `web-client` update. Then you can re-commit the changes and force-push, and do `git submodule update` to reset the `web-client` submodule.

src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
aali309: for now do i need to do any changes?

.vscode/settings.json
aali309: i run --> ```git rm --cached ~/Desktop/workStation/cryostat/.vscode/settings.json ``` the way you showed me the other day and then added this on my .gitconfig
``` [user]
        email = atali@redhat.com
        name = Atif Ali
        signingkey = 44ECBBDDF9A888E0
[commit]
        gpgsign = true
[core]
        excludesFile = ~/.gitignore
```
and i have that on my .gitignore 
```
 .vscode
*.swp
``` 
but still it added the file.. did i do any thing wrong on this set up?

andrewazores: It's probably case sensitive. I have `~/.gitconfig` contents:

```
...
[core]
    autocrlf = input
    excludesfile = ~/.gitignore
...
```

andrewazores (changes requested):

web-client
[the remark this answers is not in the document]
aali309: i must have done the revert in a wrong way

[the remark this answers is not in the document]
aali309: please check again... I reverted a different commit. I hope it fixes it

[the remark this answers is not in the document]
andrewazores: It looks like the web-client and pom.xml are still modified :(

src/main/resources/types.graphqls
```diff
@@ -17,6 +17,7 @@ input EnvironmentNodeFilterInput {
 input TargetNodesFilterInput {
     id: Int
     name: String
+    names: [String]
```
andrewazores, line 20: Change looks good! Could you also add this `names: [String]` filter to the other filter inputs? `EnvironmentNodeFilterInput`, `DescendantTargetsFilterInput`, `ActiveRecordingFilterInput`, and `ArchivedRecordingFilterInput`. I think the implementation will be basically just a copy-paste job across a few other files, and the tests should all be very similar as well.

aali309: will sure do

andrewazores:
Please don't edit the pull request labels to manipulate it into triggering CI runs, either... if you need the tests run, you can either run them on your development machine, or ask a reviewer to trigger a re-run.

src/main/java/io/cryostat/net/web/http/api/v2/graph/TargetNodesFetcher.java
andrewazores: https://github.com/cryostatio/cryostat/issues/1552

src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
```diff
@@ -93,11 +93,20 @@ public Active getAuthenticated(DataFetchingEnvironment environment) throws Excep
...
 
         FilterInput filter = FilterInput.from(environment);
+        List<String> names = new ArrayList<>();
         if (filter.contains(FilterInput.Key.NAME)) {
             String recordingName = filter.get(FilterInput.Key.NAME);
+            names.add(recordingName);
+        }
+        if (filter.contains(FilterInput.Key.NAMES)) {
+            List<String> recordingNames = filter.get(FilterInput.Key.NAMES);
+            names.addAll(recordingNames);
```
andrewazores, line 103: This implementation is going to behave differently from the others:

```
var names = [];
if (filters.contains(NAME_FILTER))
  names.add(filters.get(NAME_FILTER))
if (filters.containers(NAMES_FILTER))
  names.addAll(filters.get(NAMES_FILTER))
recordings = recordings.filter(r -> names.contains(r.getName())) // if r.getName() matches the NAME or any of the NAMES
```

vs

```
if (filters.contains(NAME_FILTER))
  recordings = recordings.filter(r -> r.getName().equals(filters.get(NAME_FILTER))) // if r.getName() matches the one NAME filter
if (filters.contains(NAMES_FILTER))
  recordings = recordings.filter(r -> filters.get(NAMES_FILTER)).contains(r.getName()) // if r.getName() matches any of the NAMES
```

If the client specifies both the `name: ""` and `names: []` filter inputs, we basically have that the first implementation does a set union and the second does a set intersection.

aali309: Should we have both .matches and .
contains implemented?

andrewazores: Those specific method calls are an internal detail that clients don't need to worry about. What matters is that, regardless of which type of resource they are querying for, the behaviour they get when specifying a filter with both `NAME` and `NAMES` should be consistent. It should only be done one way or the other.

Generally in cases like this, I like to follow the "principle of least surprise." A user might come across this API and not realize that there are both `NAME` and `NAMES` filters, and accidentally use one, or the other, or both. If they do that, then that behaviour should be consistent with how any other pair of filters combines.

If a user queries for active recordings and specifies a name filter as well as a label filter, what result do they get? Is it all recordings that have that name or those labels, or is it all recordings that have both that name and those labels?

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +602,392 @@ void testNodesHaveIds() throws Exception {
...
+        target2.name = "service:jmx:rmi:///jndi/rmi://cryostat-itests:9093/jmxrmi";
+        target2.nodeType = "JVM";
+
+        assertThat(targetNodes, hasItem(target1));
+        assertThat(targetNodes, hasItem(target2));
+    }
+
+    @Test
+    @Order(10)
+    public void testQueryForFilteredActiveRecordingsByNames() throws Exception {
```
andrewazores, line 648: Shouldn't this test be calling the `/api/v2.2/graphql` endpoint at some point? I don't think it is currently exercising the active recording "names" filter.

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +602,392 @@ void testNodesHaveIds() throws Exception {
...
+
+        MatcherAssert.assertThat(
+                "list should have size 0 after deleting recordings",
+                listResp.size(),
+                Matchers.equalTo(0));
+    }
+
+    @Test
+    @Order(11)
+    public void shouldReturnArchivedRecordingsFilteredByNames() throws Exception {
```
andrewazores, line 789: Same here, this test exercises a recording creation/archive/deletion workflow, but it doesn't actually use GraphQL to do so.

aali309: I  saw on the  HTTP_API.md  GET/api/vi/recordings will return all archived recordings then I can just match the names?

aali309: i got it though... Also saw an example

andrewazores: You can, yes, but the point of this class is to exercise the GraphQL endpoint, and this test in particular should exercise the new code you have added. If you use `GET /api/v1/recordings` you will get data back, but you will not be using the GraphQL `names` query that you added, so what is being tested exactly?

aali309: ok. got it

src/test/java/itest/GraphQLIT.java
aali309: in my thinking then we do not need to create any active recordings? just retrieve any that are present then if any that are present, based on the data I will filter by names? The created recordings with the code that I have cannot be retrieved with ``` /api/v2.2/graphql ```. its always returning 0 for active recordings available

aali309: never mind i think I saw something useful by creating a snapshot with ``` POST /api/v2/targets/:targetId/snapshot ``` this should work

andrewazores: If you do not create any active recordings, there will be none there when you query and therefore none to filter...

> The created recordings with the code that I have cannot be retrieved with /api/v2.2/graphql. its always returning 0 for active recordings available

Then something is wrong, because these things should work together. Either there is a bug or your test query setup isn't working how you intended.

> never mind i think I saw something useful by creating a snapshot with POST /api/v2/targets/:targetId/snapshot this should work

I don't think so - creating a snapshot when there are no active recordings does not do anything.

aali309: I might need help tomorrow if by any chance you have time. I think the problem is my graphql query. I read the .md files and ```"/api/v2.2/graphql"``` should be able to retrieve the created recordings

andrewazores: Sure, just ping me or send a calendar invite.

aali309: I will ping you... and when you are free we can meet 
mean while this is what i have 
``` CompletableFuture<TargetNodesQueryResponse> resp2 = new CompletableFuture<>();
 String query =
     "query { targetNodes {"
             + "name"
             + "nodeType"
             + "labels"
             + "target {"
             + "alias"
             + "serviceUri"
             + "}"
             + "recordings {"
             + "active {"
             + "data {"
             + "name"
             + "state"
             + "metadata {"
             + "labels"
             + "}"
             + "}"
             + "aggregate {"
             + "count"
             + "}"
             + "}"
             + "}"
             + "}"
             + "}";
 webClient.post("/api/v2.2/graphql")
     .sendJson(new JsonObject().put("query", query), ar -> {
         if (assertRequestStatus(ar, resp2)) {
             resp2.complete(
                     gson.fromJson(
                             ar.result().bodyAsString(),
                             TargetNodesQueryResponse.class));
         }
     });
 
     TargetNodesQueryResponse graphqlResp = resp2.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);``` trying to get all active recordings... i created two named Recording1 and Recording2

File: src/test/java/itest/GraphQLIT.java [hunk 8]
```diff
@@ -835,6 +1235,7 @@ public boolean equals(Object obj) {
     }
 
     static class TargetNodes {
+        public List<EnvironmentNode> environmentNodes;
```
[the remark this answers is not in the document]
aali309: with the changes no, So i removed

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +602,475 @@ void testNodesHaveIds() throws Exception {
...
+                .get(String.format("/api/v1/targets/%s/recordings", SELF_REFERENCE_TARGET_ID))
+                .send(
+                        ar -> {
+                            if (assertRequestStatus(ar, listRespFuture1)) {
+                                listRespFuture1.complete(ar.result().bodyAsJsonArray());
+                            }
+                        });
+        JsonArray listResp = listRespFuture1.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        Assertions.assertTrue(listResp.isEmpty());
+        CountDownLatch latch = new CountDownLatch(2);
```
andrewazores, line 661: I don't think this latch is necessary. Just waiting on the two `CompletableFuture`s in order will do the same thing as what this latch appears to do.

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +602,475 @@ void testNodesHaveIds() throws Exception {
...
+
+        MatcherAssert.assertThat(
+                "list should have size 2 after recording creation",
+                listResp.size(),
+                Matchers.equalTo(2));
+
+        // GraphQL Query to filter Active recordings by names
+        CompletableFuture<TargetNodesQueryResponse> resp2 = new CompletableFuture<>();
+        String query =
+                "query { targetNodes {"
```
andrewazores, line 718: Why not do `targetNodes(filter: { name: "some-node-name" })` here? This way the query looks only for active recordings on the specified target node. I think you could then remove the previous `/api/v1/targets/%s/recordings` query that checks the same data that this query does, and you would not really need the `for (TargetNode targetNode : graphqlResp.data.targetNodes)` loop lower down either.

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +602,475 @@ void testNodesHaveIds() throws Exception {
...
+                        + "recordings {"
+                        + "active(filter: { names: [\"Recording1\", \"Recording2\"] }) {"
```
andrewazores, line 720: If two recordings are created, and this filters for both of their expected values, then we don't know if this filter actually does anything - it might just pass through all results. Should this test perhaps query for only one of them? Or query with a filter that matches neither and assert that the response contains no entries? etc.

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +602,475 @@ void testNodesHaveIds() throws Exception {
...
+        CompletableFuture<TargetNodesQueryResponse> resp2 = new CompletableFuture<>();
+        // List<String> filterNames = Arrays.asList("Recording1", "Recording2");
+        String query =
+                "query { targetNodes {"
+                        + "recordings {"
+                        + "archived(filter: { names: [\"io-cryostat-Cryostat_"
+                        + TEST_RECORDING_NAME
+                        + "_\\\\d{8}T\\\\d{6}Z(\\\\.\\\\d+)?\\\\.jfr\",\"io-cryostat-Cryostat_"
+                        + TEST_RECORDING_NAME
+                        + "_\\\\d{8}T\\\\d{6}Z(\\\\.\\\\d+)?\\\\.jfr\"] }) {"
```
andrewazores, line 887: How does this work? I thought the implementation was looking for exact string matches for the recording names, but this test is passing a regular expression pattern as the recording name and expecting the backend to perform a regex match during the query. Where is this regular expression getting compiled and matched against... ?

src/test/java/itest/GraphQLIT.java
aali309: added one extra that will not be filtered

src/test/java/itest/GraphQLIT.java
aali309: I think that will be filtering node name instead of recording name? Also when I do that, it returns nothing with the logic of creating active recording that I used.

andrewazores: That's right, it will filter the nodes, which you can do first and then on the result filter the recording name. The point would be that this way you only get the query response from that particular node - the one on which you started your test recordings - rather than a query response of all of the recordings on every node.

src/test/java/itest/GraphQLIT.java
```diff
@@ -828,63 +801,51 @@ public void shouldReturnArchivedRecordingsFilteredByNames() throws Exception {
...
+
         String query =
                 "query { targetNodes {"
                         + "recordings {"
-                        + "archived(filter: { names: [\"io-cryostat-Cryostat_"
-                        + TEST_RECORDING_NAME
-                        + "_\\\\d{8}T\\\\d{6}Z(\\\\.\\\\d+)?\\\\.jfr\",\"io-cryostat-Cryostat_"
-                        + TEST_RECORDING_NAME
-                        + "_\\\\d{8}T\\\\d{6}Z(\\\\.\\\\d+)?\\\\.jfr\"] }) {"
+                        + "archived(filter: { names: [\"archivedRecording\",\"someOtherName\"] }) {"
```
andrewazores, line 848: I don't think the archived file names will be like this, they would look like `io-cryostat-Cryostat_some-recording-name_12345678.jfr`, right?

I think the way to test this would be to use a different query, like `GET /api/v1/recordings`, to get the list of all of the archived recordings within the test. Then, within the test, you can apply some logic to filter out the ones that match your expectations, ie the `some-recording-name` matches the name you gave the recordings. Then you use that information to build the GraphQL query filter to pass in here and ensure that the GraphQL response also matches your expectation.

aali309: yes the name expected should be the way you described

aali309: I used ```GET /api/v1/recordings``` and then extracted the name of the archived recording to a variable1 then I used ```.post("/api/v2.2/graphql")``` and extraxted the archived recording to a variable2 then checked if they match.

src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
```diff
@@ -97,9 +97,17 @@ public Active getAuthenticated(DataFetchingEnvironment environment) throws Excep
             String recordingName = filter.get(FilterInput.Key.NAME);
             recordings =
                     recordings.stream()
-                            .filter(r -> Objects.equals(r.getName(), recordingName))
+                            .filter(r -> r.getName().equals(recordingName))
```
andrewazores, line 100: Any reason to change this `Objects.equals()` call?

aali309: just following the consistency of all other classes so they behave the same

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +601,426 @@ void testNodesHaveIds() throws Exception {
...
+        List<String> filterNames = Arrays.asList("Recording1", "Recording2");
+
+        for (TargetNode targetNode : graphqlResp.data.targetNodes) {
+            List<ActiveRecording> recordings = targetNode.recordings.active.data;
+            for (ActiveRecording recording : recordings) {
+                if (filterNames.contains(recording.name)) {
+                    filteredRecordings.add(recording);
+                }
+            }
+        }
```
andrewazores, line 726: What is this for-loop needed for? It seems like it is providing a filtered version of the query response that only contains the expected data, and then the test continues to assert that this filtered version matches the expectations - but it has to, because this loop is trimming the response down to the expectation first. I think the assertions below can just be done directly on the query response, can't they?

src/test/java/itest/GraphQLIT.java
```diff
@@ -588,6 +601,426 @@ void testNodesHaveIds() throws Exception {
...
+                            } else {
+
+                                archiveRecordingFuture.completeExceptionally(
+                                        new RuntimeException("Archive request failed"));
+                            }
+                        });
+
+        archiveRecordingFuture.get(REQUEST_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+
+        Thread.sleep(2_000L);
```
andrewazores, line 826: I think you probably want to put a sleep in between creating the active recording and archiving that recording, so that there is time for the recording to gather data before it is copied to archives. Putting the sleep here between archiving and querying for what archives exist is probably not very useful.

src/test/java/itest/GraphQLIT.java
andrewazores: I'm not sure that this is resolved - the GraphQL query string here still asks the server for `"archivedRecording" and `"someOtherName"`, but those still shouldn't be the actual filenames of archived recordings. I thought that the solution would look like `archived(filter: { names: [" + retrievedArchivedRecordingsName + "] } )`

src/main/java/io/cryostat/net/web/http/api/v2/graph/ActiveRecordingsFetcher.java
andrewazores: Do the others not use `Objects.equals()`? How does `Objects.equals(r.getName(), recordingName))` behave differently from `r.getName().equals(recordingName)`?

src/test/java/itest/GraphQLIT.java
aali309: I erased a big chunk of code that was retrieving to make sure the recording is archived... forgot to remove the sleep

andrewazores: I think the sleep makes sense to keep, but it should be in between creating the new active recording and archiving it.

src/test/java/itest/GraphQLIT.java
aali309: this works too instead of hard cording the name

src/test/java/itest/GraphQLIT.java
aali309: I got rid of the for loop and assert directly

src/test/java/itest/GraphQLIT.java
aali309: yes

andrewazores (approved):

andrewazores:
Looks good! Please rebase and I will merge this.

andrewazores (approved):

Issue #1533, opened 2023-06-09
[Bug] integration tests sometimes time out waiting for containers to start


andrewazores:
![image](https://github.com/cryostatio/cryostat/assets/3787464/9e5c68ff-ea8d-49f8-a3b3-c07e8c2fa243)

This is also pretty commonly seen in CI runs.

PR #1534, opened 2023-06-09
test(itest): improve container startup health detection

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1533

## Description of the change:
This adds Podman health probes to the containers started for integration testing, and changes the `wait-for-X` executions to query podman's view of container health as a startup check.

## Motivation for the change:
Rather than running a `timeout` command that forks subprocess shells executing `curl` to try to check the container availability externally, this ends up using a process run within the container to check for the service availability, which is reported to podman. From outside the containers we ask podman what the status is. This is less prone to getting into a bad state with `curl` trying to open a connection to the services in the containers when they are not yet ready to accept connections and causing `curl` to time out and therefore fail the startup check and the integration test run. This should make itest startup much more reliable.

## How to manually test:
1. `bash repeated-integration-tests.bash N SomeTestsIT`, where `N` is some large number and `SomeTestsIT` are a subset of tests run simply to ensure basic startup succeeded. `bash repeated-integration-tests.bash 10 HealthIT` for example.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -627,7 +627,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.containerName} | grep -E -i '^health'; do sleep 5; done</argument>
+              <argument>until podman ps --filter health=healthy --format '{{.Names}}' | grep ${cryostat.itest.containerName}; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
@@ -644,7 +644,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.jfr-datasource.containerName} | grep -E -i '^health'; do sleep 5; done</argument>
+              <argument>until podman ps --filter health=healthy --format '{{.Names}}' | grep ${cryostat.itest.jfr-datasource.containerName}; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
@@ -661,7 +661,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.grafana.containerName} | grep -E -i '^health'; do sleep 5; done</argument>
+              <argument>until podman ps --filter health=healthy --format '{{.Names}}' | grep ${cryostat.itest.grafana.containerName}; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -627,7 +627,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.containerName} | grep healthy; do sleep 5; done</argument>
+              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.containerName} | grep -i health; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
@@ -644,7 +644,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.jfr-datasource.containerName} | grep healthy; do sleep 5; done</argument>
+              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.jfr-datasource.containerName} | grep -i health; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
@@ -661,7 +661,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.grafana.containerName} | grep healthy; do sleep 5; done</argument>
+              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.grafana.containerName} | grep -i health; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -627,7 +627,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until podman inspect -f '{{.State.Health.Status}}' ${cryostat.itest.containerName} | grep healthy; do sleep 5; done</argument>
+              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.containerName} | grep healthy; do sleep 5; done</argument>
             </arguments>
             <skip>${skipITs}</skip>
           </configuration>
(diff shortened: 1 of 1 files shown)


aali309 (approved):

pom.xml
```diff
@@ -518,7 +522,9 @@
               <argument>run</argument>
               <argument>--pull=${cryostat.itest.pullImages}</argument>
               <argument>--pod=${cryostat.itest.podName}</argument>
-              <argument>--name=grafana-itest</argument>
+              <argument>--name=${cryostat.itest.grafana.containerName}</argument>
+              <argument>--health-cmd</argument>
+              <argument>wget -q --spider http://localhost:${cryostat.itest.grafana.port}</argument>
```
tthvo, line 527: Wondering why we used `wget` here?

andrewazores: This container doesn't have `curl` in it, so this was the closest available equivalent.

`man wget`:
```
       --spider
           When invoked with this option, Wget will behave as a Web spider, which means that it will not download
           the pages, just check that they are there.  For example, you can use Wget to check your bookmarks:

                   wget --spider --force-html -i bookmarks.html

           This feature needs much more work for Wget to get close to the functionality of real web spiders.
```

tthvo: How about https://grafana.com/docs/grafana/latest/developers/http_api/other/#returns-health-information-about-grafana?

```bash
http :3000/api/health
HTTP/1.1 200 OK
Cache-Control: no-store
Content-Length: 70
Content-Type: application/json; charset=UTF-8
Date: Fri, 09 Jun 2023 21:41:15 GMT
X-Content-Type-Options: nosniff
X-Xss-Protection: 1; mode=block

{
    "commit": "4add91f03d",
    "database": "ok",
    "version": "9.4.7"
}

```

pom.xml
```diff
@@ -634,7 +644,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until curl -sSk -o /dev/null http://${cryostat.itest.webHost}:${cryostat.itest.jfr-datasource.port}; do sleep 2; done</argument>
+              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.jfr-datasource.containerName} | grep healthy; do sleep 5; done</argument>
```
tthvo, line 647: Can we use this? 

```
[ -n "$(podman ps -a --filter health=healthy,name=abc --format={{.Names}})" ]
```

pom.xml
```diff
@@ -542,6 +548,10 @@
               <argument>run</argument>
               <argument>--pod=${cryostat.itest.podName}</argument>
               <argument>--name=${cryostat.itest.containerName}</argument>
+              <argument>--health-cmd</argument>
+              <argument>curl --fail http://localhost:8181/health/liveness</argument>
+              <argument>--requires</argument>
```
tthvo, line 553: In k8s/ocp deployment, we do not (maybe cannot) specify such requirement that grafana and datasource must be up first right? Maybe itest should set it up this way too?

pom.xml
andrewazores: I suppose, but I find that harder to read and it should do effectively the same thing, no?

pom.xml
andrewazores: In k8s, the three of these containers are placed within one Pod. They should all have health checks on them, which k8s will use to determine the overall health of the Pod. If any of them are not healthy (ex. not yet running), then the Pod will not be healthy, and k8s will refuse to send traffic to it. If we don't wait for the containers to be healthy and just send traffic, like tests requests in, then the requests could fail because we're just testing things too early. I don't think that's necessarily a productive way to run integration tests...

pom.xml
tthvo: https://developers.redhat.com/blog/2019/04/18/monitoring-container-vitality-and-availability-with-podman#interacting_with_the_results_of_healthchecks

> When using a filter for health, you may use a value of healthy, starting, or unhealthy.

I just think that there are some fixed values we can specify in filters. As for grep, I am not sure if `podman` will ever decide to have a different display string. For example, `Healthy` for `health` or maybe localization.

andrewazores: I suppose that's possible, but those strings are also defined in the libpod API as literals, and there are a lot of scripts out there built around these things, so I doubt podman will break compatibility by renaming the states. If it gets localized and that breaks it then we can just force the localization in the script, too. If capitalization is a concern (which it would be in your solution too?) then I can just do `grep -i` so it's case-insensitive.

tthvo: Yehh `grep -i` is the quickest way to go!

pom.xml
tthvo: Ahh yes make sense then! So the tests are running within the cryostat container?

andrewazores: No, the integration tests run as an external process directly on the host machine, and interact with the integration test container "from the outside" over the network. This setup of waiting for all of the containers to be ready emulates how k8s will wait until the Pod is ready before sending any traffic through.

tthvo: Ahh thanks that clears up lots of stuffs!

Then do we still need `<argument>--requires</argument>` when launching `cryostat` containers? Those 3 cryostat,datasource, and grafana should be launched independently right?

```
--requires strings                         Add one or more requirement containers that must be started before this container will star
```

andrewazores: I suppose it can be removed. This just guards against accidentally launching the server container without the two other services that it requires. It doesn't do anything in this case since (as far as I can tell) all of the scripts and setup do properly launch all three containers and wait for them to be ready. This at least means that the tests will fail early if for example `repeated-integration-tests.bash` is modified and no longer starts one of the required services before starting the server. In that scenario, this `--requires` flag should fail the test launch early, rather than letting it try to proceed and generating garbage results because the tests are written assuming that all services are present and ready.

tthvo: Ah yes right! Just thinking that we can remove that requirement check to be as close to k8s :))

andrewazores: I don't think being close to how k8s lays out containers and dependencies is necessarily the goal, though. In the end k8s and podman are two different container platforms, so there are differences in how we lay out the deployments and get them into the desired state before testing the expected behaviours. It is an "integration" test after all, meant to check the integration between the various components, and so I think it's fair to have a precondition on the tests that the components have actually been deployed completely.

tthvo: Sounds good! Thanks for explaining. Looks good to me!

tthvo (dismissed):

andrewazores:
Seems like after one of the recent commits this actually breaks the CI runner entirely, it always seems to time out... unfortunately, not much in the way of logging.

andrewazores:
I wonder if maybe the CI runner's `grep` doesn't accept `-E` or something.

tthvo:
Ohh yeh just realized this is 8th attempt....What is version of  `grep` on CI?

tthvo:
Actually, I am experiencing the same issue locally too. 

Been stuck at `[INFO] --- exec-maven-plugin:3.1.0:exec (wait-for-cryostat) @ cryostat ---` for a while.

tthvo:
Maybe the regular expresion here is not accepted?

```bash
┌─[thvo][thvo][±][detached:andrew/itest-startup-health U:1 ✗][~/workspace/cryostat]
└─▪ podman ps --filter=health=healthy --format="{{.Names}} {{.Status}}" | grep cryostat-itest | grep -i -E 'health'
cryostat-itest Up 2 minutes (healthy)
┌─[thvo][thvo][±][detached:andrew/itest-startup-health U:1 ✗][~/workspace/cryostat]
└─▪ podman ps --filter=health=healthy --format="{{.Names}} {{.Status}}" | grep cryostat-itest | grep -i -E '^health'
┌─[thvo][thvo][±][detached:andrew/itest-startup-health U:1 ✗][~/workspace/cryostat]
└─▪ echo $?
1

```

pom.xml
```diff
@@ -651,7 +661,7 @@
               <argument>5m</argument>
               <argument>sh</argument>
               <argument>-c</argument>
-              <argument>until curl -sSk -o /dev/null http://${cryostat.itest.webHost}:${cryostat.itest.grafana.port}; do sleep 2; done</argument>
+              <argument>until podman ps --format '{{.Names}} {{.Status}}' | grep ${cryostat.itest.grafana.containerName} | grep -E -i '^health'; do sleep 5; done</argument>
```
tthvo, line 664: How about this?

```bash
$ podman ps --filter=health=healthy --format="{{.Names}}" | grep cryostat-itest
cryostat-itest
```

andrewazores: That's probably the perfect middle-ground between our two solutions, even better than either. Nice!

tthvo (approved):
Looks good to me!

Issue #1546, opened 2023-06-14
[Bug] Built-in discovery plugin failures should log exceptions but not block startup


andrewazores:
While testing an ARM64 image (running on qemu), I see a stack trace like this:

```
SEVERE: FAILED to deploy io.cryostat.discovery.BuiltInDiscovery Verticle
java.net.SocketException: Protocol not available
	at java.base/sun.nio.ch.Net.getInterface6(Native Method)
	at java.base/sun.nio.ch.DatagramChannelImpl.getOption(DatagramChannelImpl.java:427)
	at java.base/sun.nio.ch.DatagramSocketAdaptor.getOption(DatagramSocketAdaptor.java:426)
	at java.base/sun.nio.ch.DatagramSocketAdaptor.outgoingNetworkInterface(DatagramSocketAdaptor.java:657)
	at java.base/sun.nio.ch.DatagramSocketAdaptor.defaultNetworkInterface(DatagramSocketAdaptor.java:672)
	at java.base/sun.nio.ch.DatagramSocketAdaptor.joinGroup(DatagramSocketAdaptor.java:522)
	at java.base/sun.nio.ch.DatagramSocketAdaptor.joinGroup(DatagramSocketAdaptor.java:481)
	at java.base/java.net.MulticastSocket.joinGroup(MulticastSocket.java:325)
	at org.openjdk.jmc.jdp.client.JDPClient.start(JDPClient.java:87)
	at io.cryostat.core.net.discovery.JvmDiscoveryClient.start(JvmDiscoveryClient.java:105)
	at io.cryostat.platform.internal.DefaultPlatformClient.start(DefaultPlatformClient.java:83)
	at io.cryostat.discovery.BuiltInDiscovery.lambda$start$5(BuiltInDiscovery.java:134)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.DistinctOps$1$2.accept(DistinctOps.java:174)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1707)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
[...]

PR #1549, opened 2023-06-14
fix(startup): be tolerant of built-in discovery mechanism failures

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1546

## Description of the change:
This fixes a class cast exception that can occur when startup fails, and changes behaviour from failing startup if any built-in discovery mechanism fails to start to simply logging the cause.

## Motivation for the change:
Since all built-in discovery mechanisms are now tested by default and then automatically started, it is surprising/annoying when one of these failing (such as JDP on aarch64 at the moment) causes the whole server to not start.

## How to manually test:
1. `dnf install qemu-user-static`
2. `mvn clean package -Dbuild.arch=arm64 ; podman image prune -f`
3. `CRYOSTAT_IMAGE=quay.io/cryostat/cryostat:2.4.0-snapshot-linux-arm64 sh run.sh # or other image tag as produced above`

### DIFF
--- a/src/main/java/io/cryostat/Cryostat.java
+++ b/src/main/java/io/cryostat/Cryostat.java
@@ -135,7 +135,7 @@ class Cryostat extends AbstractVerticle {
             if (!(cause instanceof Exception)) {
                 cause = new RuntimeException(cause);
             }
-            logger.error((RuntimeException) cause);
+            logger.error((Exception) cause);
         }
         logger.info("{} shutting down...", instanceName());
         client.vertx().close().onComplete(n -> logger.info("Shutdown complete"));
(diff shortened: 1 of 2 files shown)


ebaron (commented):
Will this make failures in production harder to detect? Like if the Kubernetes API discovery fails, that would mean Cryostat is unusable on OpenShift for example.

ebaron:
To get around the Arm64 case, could we use the `CRYOSTAT_ENABLE_JDP_BROADCAST` environment variable?

andrewazores:
> To get around the Arm64 case, could we use the `CRYOSTAT_ENABLE_JDP_BROADCAST` environment variable?

That environment variable controls whether the Cryostat JVM turns on its own JDP broadcasting in `entrypoint.sh`, but the failure occurring here is when the Cryostat application tries to initialize a JDP discovery client that listens for those broadcasts. We could add an additional environment variable to control whether the `DefaultPlatformStrategy` (the JDP one, it's a bit of a misnomer now) should be enabled - right now it just has `isAvailable() { return true; }`.

> Will this make failures in production harder to detect? Like if the Kubernetes API discovery fails, that would mean Cryostat is unusable on OpenShift for example.

Not necessarily - it would still be usable with custom target definitions, or with external discovery plugins ex. the Cryostat Agent.

ebaron (approved):

andrewazores:
Bah, commit signing...

Issue #1507, opened 2023-05-30
[Bug] `CRYOSTAT_DISABLE_SSL` env var behaves unexpectedly


andrewazores:
> I'm unable to enable SSL for the Cryostat started by the compose file, not sure why. CRYOSTAT_DISABLE_SSL is false, but I can still access the web-client at http:localhost:8181 and unable to at https://...

Looks like this is just a bug. In the `entrypoint` script, this env var is checked for the value `"true"`:

https://github.com/cryostatio/cryostat/blob/f9a7f00f5576a2a81a22ff06a47f0ce20523a5f5/src/main/extras/app/entrypoint.bash#L144

But in the application code that sets up the server's SSL, it just looks for whether the env var is a not-blank string:

https://github.com/cryostatio/cryostat/blob/f9a7f00f5576a2a81a22ff06a47f0ce20523a5f5/src/main/java/io/cryostat/net/SslConfiguration.java#L66

I think the `entrypoint` has it right, the variable should be set to `"true"` to actually turn off the behaviour.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/1480#issuecomment-1565401569_

PR #1570, opened 2023-06-29
fix(ssl): fixed env var behaves unexpectedly 

aali309:
# Welcome to Cryostat! 👋

Fixes: #1507

## Description of the change:
This change allows enabling of SSL when CRYOSTAT_DISABLE_SSL is set to "false"

## Motivation for the change:
This change is helpful because users may want to sue a Secure Sockect Layer(SSL) where as he can just pass the environment variable to false i.e < CRYOSTAT_DISABLE_SSL=false>

## How to manually test:
1. Run `CRYOSTAT_IMAGE=quay.io... sh smoketest.sh...`
2. Run `CRYOSTAT_DISABLE_SSL=false sh smoketest.sh and test it by < http --verify=no https://localhost:8181/health`

### DIFF
--- a/src/main/java/io/cryostat/net/SslConfiguration.java
+++ b/src/main/java/io/cryostat/net/SslConfiguration.java
@@ -65,7 +65,7 @@ public class SslConfiguration {
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
             String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
-            if ("true".equalsIgnoreCase(disableSslValue.toLowerCase())) {
+            if ("true".equals(disableSslValue) || "TRUE".equals(disableSslValue)) {
                 strategy = new NoSslStrategy();
                 logger.info("Selected NoSSL strategy");
                 return;
--- a/src/main/java/io/cryostat/net/SslConfiguration.java
+++ b/src/main/java/io/cryostat/net/SslConfiguration.java
@@ -65,7 +65,7 @@ public class SslConfiguration {
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
             String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
-            if ("true".equals(disableSslValue) || "TRUE".equals(disableSslValue)) {
+            if ("true".equals(disableSslValue) || "TRUE".equals(disableSslValue )) {
                 strategy = new NoSslStrategy();
                 logger.info("Selected NoSSL strategy");
                 return;
--- a/src/main/java/io/cryostat/net/SslConfiguration.java
+++ b/src/main/java/io/cryostat/net/SslConfiguration.java
@@ -65,7 +65,6 @@ public class SslConfiguration {
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
             String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
-            logger.info("++-" + disableSslValue);
             boolean disableSsl = Boolean.parseBoolean(disableSslValue); // Parse the value as a boolean
             if (disableSsl) {
                 strategy = new NoSslStrategy();
--- a/src/main/java/io/cryostat/net/SslConfiguration.java
+++ b/src/main/java/io/cryostat/net/SslConfiguration.java
@@ -65,7 +65,7 @@ public class SslConfiguration {
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
             String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
-            if ("true".equals(disableSslValue) || "TRUE".equals(disableSslValue)) {
+            if ("true".equals(disableSslValue)) {
                 strategy = new NoSslStrategy();
                 logger.info("Selected NoSSL strategy");
                 return;
--- a/src/main/java/io/cryostat/net/SslConfiguration.java
+++ b/src/main/java/io/cryostat/net/SslConfiguration.java
@@ -65,7 +65,7 @@ public class SslConfiguration {
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
             String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
-            if ("true".equalsIgnoreCase(disableSslValue)) {
+            if ("true".equalsIgnoreCase(disableSslValue) || "TRUE".equalsIgnoreCase(disableSslValue)) {
                 strategy = new NoSslStrategy();
                 logger.info("Selected NoSSL strategy");
                 return;
--- a/src/main/java/io/cryostat/net/SslConfiguration.java
+++ b/src/main/java/io/cryostat/net/SslConfiguration.java
@@ -65,8 +65,7 @@ public class SslConfiguration {
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
             String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
-            boolean disableSsl = Boolean.parseBoolean(disableSslValue); // Parse the value as a boolean
-            if (disableSsl) {
+            if ("true".equalsIgnoreCase(disableSslValue)) {
                 strategy = new NoSslStrategy();
                 logger.info("Selected NoSSL strategy");
                 return;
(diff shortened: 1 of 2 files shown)


src/main/java/io/cryostat/net/SslConfiguration.java
```diff
@@ -111,7 +115,7 @@ public class SslConfiguration {
         }
 
         strategy = new NoSslStrategy();
-        logger.info("Selected NoSSL strategy");
+        logger.info("++-Selecteeeeeeeeeed NoSSL strategy");
```
andrewazores, line 118: Probably don't want to include this change

src/main/java/io/cryostat/net/SslConfiguration.java
```diff
@@ -64,9 +64,13 @@ public class SslConfiguration {
         this.logger = logger;
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
-            strategy = new NoSslStrategy();
-            logger.info("Selected NoSSL strategy");
-            return;
+            String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
+            boolean disableSsl = Boolean.parseBoolean(disableSslValue); // Parse the value as a boolean
```
andrewazores, line 68: Part of this bug is that `env.hasEnv(Variables.DISABLE_SSL)` behaves differently than what [`entrypoint.bash`](https://github.com/cryostatio/cryostat/blob/c675a2e113a3d8fc25136ec52919c7a126e31290/src/main/extras/app/entrypoint.bash#L144) does, given the same inputs (environment variable values).

How does `Boolean.parseBoolean()` compare to the `bash` `var = "true"` check? They should behave identically given the same inputs, or else we will still have buggy behaviour due to the mismatch.

src/main/java/io/cryostat/net/SslConfiguration.java
aali309: This was to be able to keep track with ctrl f

src/main/java/io/cryostat/net/SslConfiguration.java
andrewazores: I think there is still a discrepancy between `var = "true"` and `var.equalsIgnoreCase("true")`.

aali309: are you considering in cases where there are white spaces after "true"? example ``` var = "true  "```?

andrewazores: No, I'm considering `CRYOSTAT_DISABLE_SSL="TRUE"`, for example.

aali309: 1. so it should work for all cases.. with ```CRYOSTAT_DISABLE_SSL="TRUE"```
2. ```CRYOSTAT_DISABLE_SSL="TRUE   "```
3. ```CRYOSTAT_DISABLE_SSL="tRue" ```
4. ```CRYOSTAT_DISABLE_SSL="true  "```?

aali309: if needed ignore/remove the whitespace I know i can use trim()

andrewazores: What I am saying is that the Java code and the `entrypoint.bash` need to treat all of these cases the same. Whether that means they do or do not accept (ie. disable SSL) each of those cases is another matter, but it needs to be consistently handled.

src/main/java/io/cryostat/net/SslConfiguration.java
```diff
@@ -64,9 +64,12 @@ public class SslConfiguration {
         this.logger = logger;
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
-            strategy = new NoSslStrategy();
-            logger.info("Selected NoSSL strategy");
-            return;
+            String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
+            if ("true".equalsIgnoreCase(disableSslValue) || "TRUE".equalsIgnoreCase(disableSslValue)) {
```
andrewazores, line 68: What does `equalsIgnoreCase` do? Do you need to check for both `true` and `TRUE` when you use it?

aali309: I dont need, I know ignoreCase takes cares of all those cases.. another way i can maintain the consistency is that I can may be convert the var toLowerCase()

aali309: something like ```    if ("true".equalsIgnoreCase(disableSslValue.toLowerCase())) {
```

Comment (aali309) [reply src/main/java/io/cryostat/net/SslConfiguration.java:68]: so this will change any entry TRUE, true, TRue, to true and then makes the behavior consistent?

File: src/main/java/io/cryostat/net/SslConfiguration.java [hunk 4]
```diff
@@ -64,9 +64,12 @@ public class SslConfiguration {
         this.logger = logger;
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
-            strategy = new NoSslStrategy();
-            logger.info("Selected NoSSL strategy");
-            return;
+            String disableSslValue = env.getEnv(Variables.DISABLE_SSL);
+            if ("true".equals(disableSslValue) || "TRUE".equals(disableSslValue )) {
```
andrewazores, line 68: This still does not agree with `entrypoint.bash`.

```bash
$ function test() {
> if [ $TEST = "true" ]; then
> echo accepted
> else
> echo "not accepted"
> fi
> }

$ TEST=true test
accepted

$ TEST=TRUE test
not accepted
```

aali309: I tested with both ```CRYOSTAT_DISSABLE_SSL=true``` and ```CRYOSTAT_DISSABLE_SSL=TRUE``` and it worked as expected.

andrewazores: Yes, the Java code will accept both of these options. The `entrypoint.bash` will not. This same variable controls both webserver SSL (Java code and the HTTPie test you are performing) and the SSL configuration for JMX and the Java built-in HTTP capabilities (entrypoint.bash, not captured with the HTTPie test you are performing). I will not accept this PR unless both of these places are updated to agree on what values they accept, since that is the true root cause of the original bug.

andrewazores: Please, just make the Java code accept `true` and only `true` for disabling SSL, the same as what `entrypoint.bash` does.

src/main/java/io/cryostat/net/SslConfiguration.java
```diff
@@ -64,9 +64,12 @@ public class SslConfiguration {
         this.logger = logger;
 
         if (env.hasEnv(Variables.DISABLE_SSL)) {
```
andrewazores, line 66: How about just changing `env.hasEnv(Variables.DISABLE_SSL))` to `env.getEnv(Variables.DISABLE_SSL, "").equals("true")`?

aali309: This solution is better by getting the value of the env variable and then see if its value is equal to "true"

andrewazores: I think that's pretty similar to what my suggestion is doing, but if you'd prefer to use `hasEnv` and then `getEnv`, could you combine them into a single `if` condition?

aali309: this is what I have 
 ```String disableSslValue = env.getEnv(Variables.DISABLE_SSL,"");
            if (disableSslValue.equals("true")) {
                strategy = new NoSslStrategy();
                logger.info("Selected NoSSL strategy");
                return;
            }
         ``` 
so far, this works with the idea that you had. This looks good in my opinion than using both ```hasEnv``` and ```getEnv```

aali309: yes... and I think this makes more sense

aali309: my idea of ```hasEnv``` is kinda long cut where as your idea of ```getEnv``` will do both the check and retrieving the value

aali309:
```if ("true".equals(env.getEnv(Variables.DISABLE_SSL))) {
    strategy = new NoSslStrategy();
    logger.info("Selected NoSSL strategy");
    return;
}```
This looks cleaner?

andrewazores:
This looks like the latest push includes some improperly cherry-picked commits from a rebase, or something.

The main change itself looks fine now, but please correct the commit history to include only that change.

Issue #1584, opened 2023-07-07
[Task] Implement PR command /build_test


maxcao13:
## Describe the feature
As per the upstream practices presentation, we should implement a PR command that will build and run the tests everytime the command is invoked, and not on every successful integration-passing commit as the current repo currently does.

Issue should be copied to -web.

PR #1599, opened 2023-07-27
feat(ci): build_test PR command

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #1584

### DIFF
--- a/.github/workflows/ci-build-image.yml
+++ b/.github/workflows/ci-build-image.yml
@@ -18,10 +18,6 @@ on:
       image-version:
         description: the Cryostat application version that will be built
         value: ${{ jobs.get-pom-properties.outputs.image-version }}
-    secrets:
-      GH_PKGS_READ_TOKEN:
-        required: true
-        description: read-only token for pulling artifacts from GitHub Packages
 
 jobs:
   get-pom-properties:
@@ -73,7 +69,7 @@ jobs:
     - run: git submodule update --remote
     - run: mvn -B -U -Dbuild.arch=${{ inputs.build-arch }} clean package
       env:
-        GITHUB_TOKEN_REF: ${{ secrets.GH_PKGS_READ_TOKEN }}
+        GITHUB_TOKEN_REF: ${{ secrets.GITHUB_TOKEN }}
     - name: Save cryostat image
       run: podman save -o cryostat-${{ inputs.build-arch }}.tar --format oci-archive quay.io/cryostat/cryostat
     - uses: actions/upload-artifact@v3
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -1,85 +1,170 @@
 name: CI build and push (PR)
 
 concurrency:
-  group: pr-${{ github.event.number }}
+  group: ci-${{ github.run_id }}
   cancel-in-progress: true
 
 on:
-  pull_request_target:
+  issue_comment:
     types:
-      - opened
-      - reopened
-      - synchronize
-      - labeled
-      - unlabeled
-    branches:
-      - main
-      - v[0-9]+
-      - v[0-9]+.[0-9]+
-      - cryostat-v[0-9]+.[0-9]+
+      - created
 
 jobs:
+  check-before-build:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
+    steps:
+    - name: Fail if needs-triage label applied 
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      run: exit 1
+    - name: Show warning if permission is denied
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'OWNER')
+        && !contains(github.event.issue.labels.*.name, 'safe-to-test')
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
+          to resolve the issue.
+    - name: Fail if command permission is not allowed
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'OWNER')
+        && !contains(github.event.issue.labels.*.name, 'safe-to-test')
+      run: exit 1
+    - name: Show warning if safe-to-test applied and permission denied
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.issue.user.name == github.event.comment.user.name)
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
+          to resolve the issue.
+    - name: Fail if safe-to-test applied and command permission is not allowed
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.issue.user.name == github.event.comment.user.name)
+      run: exit 1
+    - name: React to comment
+      uses: actions/github-script@v4
+      with:
+        script: |
+            const {owner, repo} = context.issue
+            github.reactions.createForIssueComment({
+              owner,
+              repo,
+              comment_id: context.payload.comment.id,
+              content: "+1",
+            });  
+
+  checkout-branch: 
+    runs-on: ubuntu-latest
+    needs: [check-before-build]
+    permissions:
+      contents: read
+      issues: read
+      pull-requests: read
+    outputs: 
+      PR_head_ref: ${{ fromJSON(steps.comment-branch.outputs.result).ref }}
+      PR_head_sha: ${{ fromJSON(steps.comment-branch.outputs.result).sha }}
+      PR_repo: ${{ fromJSON(steps.comment-branch.outputs.result).repo.full_name }}
+    steps:
+    - uses: actions/github-script@v4
+      id: comment-branch
+      with:
+        script: |
+          const result = await github.pulls.get ({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            pull_number: context.issue.number
+          })
+          return result.data.head
+
   code-analysis:
+    needs: [checkout-branch]
     uses: ./.github/workflows/ci-code-analysis.yml
     with:
-      checkout-repo: ${{ github.event.pull_request.head.repo.full_name }}
-      checkout-ref: ${{ github.event.pull_request.head.ref }}
-    secrets: inherit
-    if: github.repository_owner == 'cryostatio' && contains(github.event.pull_request.labels.*.name, 'safe-to-test')
-
+      checkout-repo: ${{ needs.checkout-branch.outputs.PR_repo }}
+      checkout-ref: ${{ needs.checkout-branch.outputs.PR_head_ref }}
+      
   build-and-test:
+    needs: [code-analysis, checkout-branch]
     strategy:
       matrix:
         arch: [amd64, arm64]
     uses: ./.github/workflows/ci-build-image.yml
     with:
       build-arch: ${{ matrix.arch }}
-      checkout-repo: ${{ github.event.pull_request.head.repo.full_name }}
-      checkout-ref: ${{ github.event.pull_request.head.ref }}
+      checkout-repo: ${{ needs.checkout-branch.outputs.PR_repo }}
+      checkout-ref: ${{ needs.checkout-branch.outputs.PR_head_ref  }}
       skip-itests: ${{ matrix.arch != 'amd64' }}
-    secrets: inherit
-    if: github.repository_owner == 'cryostatio' && contains(github.event.pull_request.labels.*.name, 'safe-to-test')
 
   push-to-ghcr:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         arch: [amd64, arm64]
-    needs: [code-analysis, build-and-test]
-    if: always() && github.repository_owner == 'cryostatio'
+    outputs: 
+      amd64_image: ${{ steps.amd64_image.outputs.image }}
+      arm64_image: ${{ steps.arm64_image.outputs.image }}
+    needs: [build-and-test, checkout-branch]
+    env:
+      head_sha: ${{ needs.checkout-branch.outputs.PR_head_sha}}
     steps:
-    - name: Fail if needs-triage label applied
-      if: ${{ contains(github.event.pull_request.labels.*.name, 'needs-triage') }}
-      run: exit 1
-    - name: Fail if safe-to-test label NOT applied
-      if: ${{ !contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}
-      run: exit 1
     - uses: actions/download-artifact@v3
       with:
         name: cryostat-${{ matrix.arch }}
-      if: github.repository_owner == 'cryostatio'
     - name: Load cryostat image
       run: podman load -i cryostat-${{ matrix.arch }}.tar
-      if: github.repository_owner == 'cryostatio'
     - name: Tag cryostat image
-      run: podman tag cryostat ghcr.io/${{ github.repository_owner }}/cryostat:pr-${{ github.event.number }}-${{ github.event.pull_request.head.sha }}-linux-${{ matrix.arch }}
-      if: github.repository_owner == 'cryostatio'
+      run: podman tag cryostat ghcr.io/${{ github.repository_owner }}/cryostat:pr-${{ github.event.number }}-${{ env.head_sha }}-linux-${{ matrix.arch }}
     - name: Push PR test image to ghcr.io
       id: push-to-ghcr
       uses: redhat-actions/push-to-registry@v2
       with:
         image: cryostat
-        tags: pr-${{ github.event.number }}-${{ github.event.pull_request.head.sha }}-linux-${{ matrix.arch }}
+        tags: pr-${{ github.event.number }}-${{ env.head_sha }}-linux-${{ matrix.arch }}
         registry: ghcr.io/${{ github.repository_owner }}
-        username: ${{ github.event.pull_request.user.login }}
-        password: ${{ secrets.GHCR_PR_TOKEN }}
-      if: github.repository_owner == 'cryostatio'
-    - name: Comment test image link
-      uses: thollander/actions-comment-pull-request@v1
+        username: ${{ github.event.comment.user.login }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Store images as output amd64
+      if: ${{ matrix.arch == 'amd64' }}
+      id: amd64_image
+      run: echo "image=${{ steps.push-to-ghcr.outputs.registry-path }}" >> "$GITHUB_OUTPUT"
+    - name: Store images as output
+      if: ${{ matrix.arch == 'arm64' }}
+      id: arm64_image
+      run: echo "image=${{ steps.push-to-ghcr.outputs.registry-path }}" >> "$GITHUB_OUTPUT"
+
+  comment-image:
+    runs-on: ubuntu-latest
+    needs: [push-to-ghcr]
+    env:
+      amd64_image: ${{ needs.push-to-ghcr.outputs.amd64_image }}
+      arm64_image: ${{ needs.push-to-ghcr.outputs.arm64_image }}
+    steps:
+    - name: Create markdown table
+      id: md-table
+      uses: petems/csv-to-md-table-action@v3.0.0
+      with: 
+        csvinput: |
+          ARCH, IMAGE
+          AMD64, ${{ env.amd64_image }}
+          ARM64, ${{ env.arm64_image }}
+    - uses: thollander/actions-comment-pull-request@v2
       with:
         message: |-
-          Test image available:
+          ${{ steps.md-table.outputs.markdown-table }}
+
+          To run smoketest:
           ```
-          $ CRYOSTAT_IMAGE=${{ steps.push-to-ghcr.outputs.registry-path }} sh smoketest.sh
+          # amd64          
+          CRYOSTAT_IMAGE=${{ amd64_image }} sh smoketest.sh
+
+          # or arm64
+          CRYOSTAT_IMAGE=${{ arm64_image }} sh smoketest.sh
           ```
-      if: github.repository_owner == 'cryostatio'
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -19,35 +19,17 @@ jobs:
       run: exit 1
     - name: Show warning if permission is denied
       if: |
-        !(github.event.comment.author_association == 'MEMBER' ||
-        github.event.comment.author_association == 'OWNER')
-        && !contains(github.event.issue.labels.*.name, 'safe-to-test')
+        !(github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+        && (!contains(github.event.issue.labels.*.name, 'safe-to-test') || github.event.issue.user.name != github.event.comment.user.name)
       uses: thollander/actions-comment-pull-request@v2
       with:
         message: |-
           You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
           to resolve the issue.
-    - name: Fail if command permission is not allowed
+    - name: Fail if command permission is denied
       if: |
-        !(github.event.comment.author_association == 'MEMBER' ||
-        github.event.comment.author_association == 'OWNER')
-        && !contains(github.event.issue.labels.*.name, 'safe-to-test')
-      run: exit 1
-    - name: Show warning if safe-to-test applied and permission denied
-      if: |
-        !(github.event.comment.author_association == 'MEMBER' ||
-        github.event.comment.author_association == 'OWNER' ||
-        github.event.issue.user.name == github.event.comment.user.name)
-      uses: thollander/actions-comment-pull-request@v2
-      with:
-        message: |-
-          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
-          to resolve the issue.
-    - name: Fail if safe-to-test applied and command permission is not allowed
-      if: |
-        !(github.event.comment.author_association == 'MEMBER' ||
-        github.event.comment.author_association == 'OWNER' ||
-        github.event.issue.user.name == github.event.comment.user.name)
+        !(github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+        && (!contains(github.event.issue.labels.*.name, 'safe-to-test') || github.event.issue.user.name != github.event.comment.user.name)
       run: exit 1
     - name: React to comment
       uses: actions/github-script@v4
@@ -113,7 +95,7 @@ jobs:
       arm64_image: ${{ steps.arm64_image.outputs.image }}
     needs: [build-and-test, checkout-branch]
     env:
-      head_sha: ${{ needs.checkout-branch.outputs.PR_head_sha}}
+      head_sha: ${{ needs.checkout-branch.outputs.PR_head_sha }}
     steps:
     - uses: actions/download-artifact@v3
       with:
@@ -153,8 +135,8 @@ jobs:
       with: 
         csvinput: |
           ARCH, IMAGE
-          AMD64, ${{ env.amd64_image }}
-          ARM64, ${{ env.arm64_image }}
+          amd64, ${{ env.amd64_image }}
+          arm64, ${{ env.arm64_image }}
     - uses: thollander/actions-comment-pull-request@v2
       with:
         message: |-
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -145,8 +145,8 @@ jobs:
           To run smoketest:
           ```
           # amd64          
-          CRYOSTAT_IMAGE=${{ amd64_image }} sh smoketest.sh
+          CRYOSTAT_IMAGE=${{ env.amd64_image }} sh smoketest.sh
 
           # or arm64
-          CRYOSTAT_IMAGE=${{ arm64_image }} sh smoketest.sh
+          CRYOSTAT_IMAGE=${{ env.arm64_image }} sh smoketest.sh
           ```
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -24,7 +24,7 @@ jobs:
       uses: thollander/actions-comment-pull-request@v2
       with:
         message: |-
-          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
+          You do not have permission to run the /build_test command. Please ask @cryostatio/reviewers
           to resolve the issue.
     - name: Fail if command permission is denied
       if: |
--- a/.github/workflows/ci-code-analysis.yml
+++ b/.github/workflows/ci-code-analysis.yml
@@ -7,10 +7,6 @@ on:
       checkout-ref:
         required: false
         type: string
-    secrets:
-      GH_PKGS_READ_TOKEN:
-        required: true
-        description: read-only token for pulling artifacts from GitHub Packages
 
 jobs:
   spotless:
@@ -47,7 +43,7 @@ jobs:
     - name: Run spotbugs
       run: mvn -B -U -Dheadless=true compile spotbugs:check
       env:
-        GITHUB_TOKEN_REF: ${{ secrets.GH_PKGS_READ_TOKEN }}
+        GITHUB_TOKEN_REF: ${{ secrets.GITHUB_TOKEN }}
 
   shellcheck:
     runs-on: ubuntu-latest


.github/workflows/pr-command.yml
```diff
@@ -1,31 +1,36 @@
-name: CI build and push (PR)
```
tthvo, line 1: I think we should still keep it as `pr-ci.yml` and the name as `CI build and push (PR)`.

tthvo: Maybe something like: `CI build and push (PR) on demand` is also fine.

.github/workflows/pr-command.yml
```diff
@@ -1,31 +1,36 @@
...
-    branches:
-      - main
-      - v[0-9]+
-      - v[0-9]+.[0-9]+
-      - cryostat-v[0-9]+.[0-9]+
+  issue_comment:
+    types: [created]
 
 jobs:
+  create-command:
```
tthvo, line 12: Wondering what purpose of this job? 

I think the ` xt0rted/slash-command-action@v1` action should be used as the first step in other jobs that run on on PR command. Basically, an additional check if the command is supported (i.e other check is with the `if` clause).

.github/workflows/pr-command.yml
```diff
@@ -74,6 +79,9 @@ jobs:
         username: ${{ github.event.pull_request.user.login }}
         password: ${{ secrets.GHCR_PR_TOKEN }}
       if: github.repository_owner == 'cryostatio'
+    - name: check-status
```
maxcao13, line 82: What is this workflow for? Seems like it links to the relevant pr command workflow that triggered the comment, would the comment be commented at all if the workflow wasn't successful?

.github/workflows/pr-command.yml
tthvo: I think we can just not use this. Instead, use `if` clause only + an action to add a reaction to the PR comment. That should do the exact same thing.

.github/workflows/pr-command.yml
maxcao13: Oh, is it to signal the success of the /build_test to test the mergibility?

tthvo: Seems to be printed out the PR discussions. We probably do not need to show this. The action status is shown at the bottom itself.

.github/workflows/pr-command.yml
```diff
@@ -82,4 +90,5 @@ jobs:
           ```
           $ CRYOSTAT_IMAGE=${{ steps.push-to-ghcr.outputs.registry-path }} sh smoketest.sh
           ```
-      if: github.repository_owner == 'cryostatio'
```
tthvo, line 85: Can u do the same thing for other steps in this job? The job's `if` clause already checks this.

.github/workflows/pr-command.yml
```diff
@@ -82,4 +90,5 @@ jobs:
...
+          ${{ github.workflow   }}: ${{steps.workflow-status.outputs.status}}
```
tthvo, line 93: As said above with Max, I am not sure we need to print these out. The action table at the bottom of the PR will show all these.

tthvo: However, one idea is that you can group both the amd64 and arm64 image tag into a single PR comment instead of 2 separate ones.

tthvo (commented):
Foundation looking good!

~However, we will need check if the command belongs to cryostat reviewers or admin. This would likely require the GITHUB_TOKEN to be granted organization permission.~

.github/workflows/pr-command.yml
mwangggg: how do I add a reaction to a PR comment?

tthvo:
This action might be useful: https://github.com/tspascoal/get-user-teams-membership

.github/workflows/pr-ci.yml
```diff
@@ -1,32 +1,44 @@
...
+            repo,
+            comment_id: context.payload.comment.id,
+            content: "+1",
+          });  
+    - uses: tspascoal/get-user-teams-membership@v2
+      id: checkMember
+      with:
+        username: ${{ github.actor }}
+        team: 'cryostat/reviewers'
+        GITHUB_TOKEN: ${{ secrets.PAT }}
```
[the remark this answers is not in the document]
tthvo: I think we can elevate the permissions of this token for action to allow querying organization?

.github/workflows/pr-ci.yml
```diff
@@ -1,32 +1,44 @@
 name: CI build and push (PR)
 
 concurrency:
-  group: pr-${{ github.event.number }}
+  group: ci-${{ github.run_id }}
   cancel-in-progress: true
 
 on:
```
tthvo, line 7: We also want to retain the current behaviour of `safe-to-test`.
```yaml
  pull_request_target:
    types:
      - opened
      - reopened
      - synchronize
      - labeled
      - unlabeled
    branches:
      - main
      - v[0-9]+
      - v[0-9]+.[0-9]+
      - cryostat-v[0-9]+.[0-9]+
```

.github/workflows/pr-ci.yml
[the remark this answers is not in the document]
tthvo: Also, its worth exploring if that can extend to the use of pushing images to `ghcr.io`. That way, we can use `GITHUB_TOKEN` and eliminate the need for `target_pull_target`.

.github/workflows/pr-ci.yml
tthvo: Also, there might be issues accessing `${{ secrets.GHCR_PR_TOKEN }}` on `issue_comment` event. See https://github.com/cryostatio/cryostat/pull/1599/files#r1282452369

.github/workflows/pr-ci.yml
```diff
@@ -1,32 +1,44 @@
...
+            comment_id: context.payload.comment.id,
+            content: "+1",
+          });  
+    - uses: tspascoal/get-user-teams-membership@v2
+      id: checkMember
+      with:
+        username: ${{ github.actor }}
+        team: 'cryostat/reviewers'
+        GITHUB_TOKEN: ${{ secrets.PAT }}
+    - name: Fail if not a Cryostat reviewer
```
tthvo, line 33: Can we grab this checking step, along with the `safe-to-test` and `needs-triage` label check into a separate job? Then, I think other build and push jobs can just depends on it.

.github/workflows/pr-ci.yml
```diff
@@ -73,13 +82,25 @@ jobs:
...
+      id: arm64_image
+      outputs:
+        amd64_image: ${{ steps.push-to-ghcr.outputs.registry-path }}
+      if: matrix.arch == 'arm64'
     - name: Comment test image link
       uses: thollander/actions-comment-pull-request@v1
       with:
         message: |-
-          Test image available:
+          arm64 test image available:
```
tthvo, line 99: Can we use markdown table?

.github/workflows/pr-ci.yml
```diff
@@ -73,13 +82,25 @@ jobs:
...
+          ```
+          $ CRYOSTAT_IMAGE=${{ steps.amd64_image.outputs.amd64_image }} sh smoketest.sh
           ```
-          $ CRYOSTAT_IMAGE=${{ steps.push-to-ghcr.outputs.registry-path }} sh smoketest.sh
+          amd64 test image available:
           ```
-      if: github.repository_owner == 'cryostatio'
+          $ CRYOSTAT_IMAGE=${{ steps.arm64_image.outputs.arm64_image }} sh smoketest.sh
+          ```
```
maxcao13, line 106: Files need new lines: see here for why: https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b

You can enable this automatically when you save by enabling this setting for VSCode:
image

.github/workflows/pr-ci.yml
tthvo: | ARCH  | IMAGE                                                                                      |
|-------|--------------------------------------------------------------------------------------------|
| AMD64 | `ghcr.io/cryostatio/cryostat:pr-1599-cadf449151f9e1b5e15bdd3555c91fc97924864b-linux-amd64` |
| ARM64 | `ghcr.io/cryostatio/cryostat:pr-1599-25b3afaf6b2b63be0835808068a335dee918ce0b-linux-arm64` |

.github/workflows/pr-ci.yml
```diff
@@ -73,13 +82,25 @@ jobs:
...
-          Test image available:
+          arm64 test image available:
+          ```
+          $ CRYOSTAT_IMAGE=${{ steps.amd64_image.outputs.amd64_image }} sh smoketest.sh
           ```
-          $ CRYOSTAT_IMAGE=${{ steps.push-to-ghcr.outputs.registry-path }} sh smoketest.sh
+          amd64 test image available:
           ```
-      if: github.repository_owner == 'cryostatio'
+          $ CRYOSTAT_IMAGE=${{ steps.arm64_image.outputs.arm64_image }} sh smoketest.sh
```
maxcao13, line 105: I personally prefer having the command with no `$` shell prompt, because then we can just copy the command without removing the $ later.

.github/workflows/pr-ci.yml
```diff
@@ -73,13 +82,25 @@ jobs:
         registry: ghcr.io/${{ github.repository_owner }}
         username: ${{ github.event.pull_request.user.login }}
         password: ${{ secrets.GHCR_PR_TOKEN }}
-      if: github.repository_owner == 'cryostatio'
+    - name: store image amd64
```
tthvo, line 85: For this, seems like it will print twice the comment that contains the 2 images.

How about moving the `Comment test image link` step into a job that runs last? Then, build jobs can define outputs that this last "print" job can run.

.github/workflows/pr-ci.yml
tthvo: How is https://github.com/cryostatio/cryostat/pull/1599#discussion_r1282457303?

maxcao13: Makes sense!

.github/workflows/pr-ci.yml
mwangggg: I thought we only wanted to build and run tests when there's a comment with the /build_test command and not when there's other PR events?

mwangggg: I thought we only wanted to build and run tests when there's a comment with the /build_test command and not when there's other PR events?

.github/workflows/pr-ci.yml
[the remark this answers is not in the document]
mwangggg: the token needs read:org scope so I think users need to grant it.... I'm not certain though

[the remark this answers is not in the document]
mwangggg: @tthvo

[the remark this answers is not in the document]
tthvo: Sorry to clarify, i mean we could, instead of using PAT, use the `GITHUB_TOKEN` with elevated permissions.

Check this out: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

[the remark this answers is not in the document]
tthvo: Unfortunately, there is no option for org query:

```yaml
permissions:
  actions: read|write|none
  checks: read|write|none
  contents: read|write|none
  deployments: read|write|none
  id-token: read|write|none
  issues: read|write|none
  discussions: read|write|none
  packages: read|write|none
  pages: read|write|none
  pull-requests: read|write|none
  repository-projects: read|write|none
  security-events: read|write|none
  statuses: read|write|none
```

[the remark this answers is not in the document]
tthvo: I mention this because we will run into issues with accessing repo secrets if the event is `issue_comment`.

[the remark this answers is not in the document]
tthvo: Well there could be another way:

- `safe-to-test` label: Enable the PR author to run the `/build_test` command.
- `auto-test`: A new label to do behave as what the `safe-to-test` label does now.

This way, we don't have to ensure the commenter is the cryostat reviewer. The author can just do it themselves if granted/trusted -> save us the need for PAT.

.github/workflows/pr-ci.yml
```diff
@@ -73,13 +103,32 @@ jobs:
...
         username: ${{ github.event.pull_request.user.login }}
         password: ${{ secrets.GHCR_PR_TOKEN }}
-      if: github.repository_owner == 'cryostatio'
-    - name: Comment test image link
-      uses: thollander/actions-comment-pull-request@v1
+    - name: store image amd64
+      id: amd64_image
+      run: echo "image=${{ steps.push-to-ghcr.outputs.registry-path }}" >> "$GITHUB_OUTPUT"
+      if: matrix.arch == 'amd64'
+    - name: store image arm64
```
tthvo, line 110: We don't need to specify separate steps (with `if`) for each arrch. This should work?

```yaml
      id: amd64_image
      run: echo "${{ matrix.arch }}_image=${{ steps.push-to-ghcr.outputs.registry-path }}" >> "$GITHUB_OUTPUT"
```

Also, u need to map job output to step output.

https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

tthvo:
Comment for visibility after further discussions:

**Current direction:**

> Well there could be another way:
>    safe-to-test label: Enable the PR author/or just anyone to run the /build_test command.
>    auto-test: A new label to do behave as what the safe-to-test label does now.

The auto-test label can be done in separate PR (possibly separate workflow) to avoid confusion.

**Potential issue:** 

> Also, there might be issues accessing ${{ secrets.GHCR_PR_TOKEN }} on issue_comment event. 

After some research, I believe the `issue_comment` might run in the context of the upstream and is allowed to request write access in `GITHUB_TOKEN`. 

Reference: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#accessing-secrets

So, the original issue probably won't be the case. However, we should be cautious in protecting our token by adding checks as above (i.e. current direction) and also requests only necessary scope.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+      id: arm64_image
+      run: echo "image=${{ steps.push-to-ghcr.outputs.registry-path }}" >> "$GITHUB_OUTPUT"
+
+  comment-image:
+    runs-on: ubuntu-latest
+    needs: [push-to-ghcr]
+    if: ithub.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
+    steps:
+    - name: Create markdown table
+      uses: petems/csv-to-md-table-action@master
```
tthvo, line 140: Can we use a specific version?

`petems/csv-to-md-table-action@v3.0.0`

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+    - uses: thollander/actions-comment-pull-request@v1
       with:
         message: |-
-          Test image available:
+          ${{ steps.md-table.outputs.markdown-table }}
+
+          To run smoketest:
           ```
-          $ CRYOSTAT_IMAGE=${{ steps.push-to-ghcr.outputs.registry-path }} sh smoketest.sh
+          CRYOSTAT_IMAGE=${PULL_REQUEST_IMAGE} sh smoketest.sh
```
tthvo, line 157: How about? This way, it is more convenient to copy and avoid confusion about the `PULL_REREQUEST_IMAGE` variable.

```
CRYOSTAT_IMAGE=${{ env.amd64_image }} sh smoketest.sh
```

.github/workflows/pr-ci.yml
tthvo: Also, u can remove the `id` for this step since we don't need anything from it.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
     with:
       build-arch: ${{ matrix.arch }}
-      checkout-repo: ${{ github.event.pull_request.head.repo.full_name }}
-      checkout-ref: ${{ github.event.pull_request.head.ref }}
+      checkout-repo: ${{ needs.checkout-branch.outputs.PR_repo }}
+      checkout-ref: ${{ needs.checkout-branch.outputs.PR_head_ref  }}
       skip-itests: ${{ matrix.arch != 'amd64' }}
-    secrets: inherit
-    if: github.repository_owner == 'cryostatio' && contains(github.event.pull_request.labels.*.name, 'safe-to-test')
+    if: github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test') && steps.checkMember.outputs.isTeamMember == 'true'
```
tthvo, line 96: Remove `steps.checkMember.outputs.isTeamMember == 'true'`?

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
     runs-on: ubuntu-latest
     strategy:
       matrix:
         arch: [amd64, arm64]
-    needs: [code-analysis, build-and-test]
-    if: always() && github.repository_owner == 'cryostatio'
+    outputs: 
+      amd64_image: ${{ steps.amd64_image.outputs.image }}
+      arm64_image: ${{ steps.arm64_image.outputs.image }}
+    needs: [code-analysis, build-and-test, checkout-branch]
```
tthvo, line 106: Just  `[build-and-test,checkout-branch]` should be fine since we want to make sure the image is built.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+            pull_number: context.issue.number
+          }
+          core.info(`Getting PR #${request.pull_number} from ${request.owner}/${request.repo}`)
+          try {
+            const result = await github.pulls.get(request)
+            return result.data
+          } catch (err) {
+            core.setFailed(`Request failed with error ${err}`)
+          } 
+    - uses: actions/checkout@v3
```
tthvo, line 71: There is no need to run checkout here since no steps to be run after that.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+        script: |
+          const request = {
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            pull_number: context.issue.number
+          }
+          core.info(`Getting PR #${request.pull_number} from ${request.owner}/${request.repo}`)
+          try {
+            const result = await github.pulls.get(request)
+            return result.data
```
tthvo, line 67: Can we extract the field into a simplified object and return it? This way, the parsing of `fromJSON` should be lighter (i.e. the full original result object is quite large).

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
       with:
         name: cryostat-${{ matrix.arch }}
-      if: github.repository_owner == 'cryostatio'
     - name: Load cryostat image
       run: podman load -i cryostat-${{ matrix.arch }}.tar
-      if: github.repository_owner == 'cryostatio'
     - name: Tag cryostat image
-      run: podman tag cryostat ghcr.io/${{ github.repository_owner }}/cryostat:pr-${{ github.event.number }}-${{ github.event.pull_request.head.sha }}-linux-${{ matrix.arch }}
-      if: github.repository_owner == 'cryostatio'
+      run: podman tag cryostat ghcr.io/${{ needs.checkout-branch.outputs.PR_repo_owner }}/cryostat:pr-${{ github.event.number }}-${{ needs.checkout-branch.outputs.PR_head_sha }}-linux-${{ matrix.arch }}
```
tthvo, line 115: I would suggest extracting necessary field in `needs.checkout-branch.outputs.` into the `env` of the job. This way, we can just use either:

- `${{ env.VAR }}`
- `${VAR}`

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
-      - unlabeled
-    branches:
-      - main
-      - v[0-9]+
-      - v[0-9]+.[0-9]+
-      - cryostat-v[0-9]+.[0-9]+
+      - created
 
 jobs:
+  check-before-build:
```
tthvo, line 13: Excluding the `needs-triage` case and unexpected issue, looks like this job will also complete successfully. This will cause the other jobs (i.e. build, test, code-analysis) to run.

tthvo: How about:

1. Create a step "Command permission allowed" that checks if the author is MEMBER or ADMIN  or `safe-to-test` label is applied.
2. Create a next step. If step 1 result is false, print the message.
3. Create another step. If step 1 result is false, exit 1.
4. For step "React to comment", only react "+1" if step 1 returns true. Otherwise, ignore or use "eyes emoji".

tthvo: Once this is done, all other jobs can remove

```
github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
```

since they declare needs on `check-before-build` job which already does the check.  Other jobs will be skipped if the `check-before-build`  fails or skips.

tthvo (commented):
Very nice! Some more comments. Please let me know if u need clarifications :))

.github/workflows/pr-ci.yml
mwangggg: should I include a line for the arm64 image too?

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+          This PR still has the needs-triage label applied. Please remove it, then try run again.
+    - name: Fail if needs-triage label applied 
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      run: exit 1
+    - name: Check command permission allowed
+      if: |
+        github.event.comment.author_association != 'MEMBER' &&
+        github.event.comment.author_association != 'OWNER' &&
+        !contains(github.event.issue.labels.*.name, 'safe-to-test')
+      uses: thollander/actions-comment-pull-request@v1
```
maxcao13, line 31: This can probably be updated to v2 https://github.com/thollander/actions-comment-pull-request

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+      contents: read
+      issues: read
+      pull-requests: read
+    outputs: 
+      PR_head_ref: ${{ fromJSON(steps.comment-branch.outputs.result).ref }}
+      PR_head_sha: ${{ fromJSON(steps.comment-branch.outputs.result).sha }}
+      PR_repo: ${{ fromJSON(steps.comment-branch.outputs.result).repo.full_name }}
+      PR_repo_owner: ${{ fromJSON(steps.comment-branch.outputs.result).repo.owner.login }}
+    steps:
+    - uses: actions/github-script@v3
```
maxcao13, line 67: And this should be upgraded to v4 like the actions/github-script above

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+      uses: petems/csv-to-md-table-action@v3.0.0
+      env:
+        amd64_image: ${{ needs.push-to-ghcr.outputs.amd64_image }}
+        arm64_image: ${{ needs.push-to-ghcr.outputs.arm64_image }}
+      with: 
+        csvinput: |
+          ARCH, IMAGE
+          AMD64, ${{ env.amd64_image }}
+          ARM64, ${{ env.arm64_image }}
+    - uses: thollander/actions-comment-pull-request@v1
```
maxcao13, line 149: Same here

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
-        registry: ghcr.io/${{ github.repository_owner }}
-        username: ${{ github.event.pull_request.user.login }}
-        password: ${{ secrets.GHCR_PR_TOKEN }}
-      if: github.repository_owner == 'cryostatio'
-    - name: Comment test image link
-      uses: thollander/actions-comment-pull-request@v1
+        tags: pr-${{ github.event.number }}-${{ env.head_sha }}-linux-${{ matrix.arch }}
+        registry: ghcr.io/${{ env.repo_owner }}
+        username: ${{ github.event.comment.user.login }}
+        password: ${{ secrets.GITHUB_TOKEN}}
```
maxcao13, line 125: put a space in between `TOKEN` and `}}`

.github/workflows/pr-ci.yml
tthvo: Good question. I think we just need one example.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+  check-before-build:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
+    steps:
+    - name: Comment if needs-triage label applied
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          This PR still has the needs-triage label applied. Please remove it, then try run again.
```
tthvo, line 22: I think outside contributors cannot add/remove labels. Just remove this message.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+      run: exit 1
+    - name: Check command permission allowed
+      if: |
+        github.event.comment.author_association != 'MEMBER' &&
+        github.event.comment.author_association != 'OWNER' &&
+        !contains(github.event.issue.labels.*.name, 'safe-to-test')
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          You do not have permission to run the /build_test command. Please ask a member of the cryostatio
```
tthvo, line 34: Same as above. Labels cannot be editted by outside contributors. Maybe:

```
Permission denied for `/build_test` command. Please ask cryostat @reviewers to resolve the issue.
```

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
       with:
         name: cryostat-${{ matrix.arch }}
-      if: github.repository_owner == 'cryostatio'
     - name: Load cryostat image
       run: podman load -i cryostat-${{ matrix.arch }}.tar
-      if: github.repository_owner == 'cryostatio'
     - name: Tag cryostat image
-      run: podman tag cryostat ghcr.io/${{ github.repository_owner }}/cryostat:pr-${{ github.event.number }}-${{ github.event.pull_request.head.sha }}-linux-${{ matrix.arch }}
-      if: github.repository_owner == 'cryostatio'
+      run: podman tag cryostat ghcr.io/${{ env.repo_owner }}/cryostat:pr-${{ github.event.number }}-${{ env.head_sha }}-linux-${{ matrix.arch }}
```
tthvo, line 116: We actually want to push the image to `ghcr.io/cryostatio` registry.

So, we should not use `${{ needs.checkout-branch.outputs.PR_repo_owner }}`. It must be the upstream owner (i.e. `${{ github.repository_owner }}`).

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
         image: cryostat
-        tags: pr-${{ github.event.number }}-${{ github.event.pull_request.head.sha }}-linux-${{ matrix.arch }}
-        registry: ghcr.io/${{ github.repository_owner }}
-        username: ${{ github.event.pull_request.user.login }}
-        password: ${{ secrets.GHCR_PR_TOKEN }}
-      if: github.repository_owner == 'cryostatio'
-    - name: Comment test image link
-      uses: thollander/actions-comment-pull-request@v1
+        tags: pr-${{ github.event.number }}-${{ env.head_sha }}-linux-${{ matrix.arch }}
+        registry: ghcr.io/${{ env.repo_owner }}
```
tthvo, line 123: Same as above. Must be upstream repo name.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,158 @@
...
+    - uses: thollander/actions-comment-pull-request@v2
       with:
         message: |-
-          Test image available:
+          ${{ steps.md-table.outputs.markdown-table }}
+
+          To run smoketest:
           ```
-          $ CRYOSTAT_IMAGE=${{ steps.push-to-ghcr.outputs.registry-path }} sh smoketest.sh
+          CRYOSTAT_IMAGE=${{ env.amd64_image }} sh smoketest.sh
```
tthvo, line 156: How about?

```bash
# amd64
CRYOSTAT_IMAGE=${{ env.amd64_image }} sh smoketest.sh

# or arm64
CRYOSTAT_IMAGE=${{ env.arm64_image }} sh smoketest.sh
```

tthvo (commented):
Looks good! Some comments and rebase please?

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,153 @@
...
 
 jobs:
+  check-before-build:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
+    steps:
+    - name: Fail if needs-triage label applied 
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      run: exit 1
+    - name: Check command permission allowed
```
tthvo, line 20: This should be 'Show warnings if permission is denied/not allowed`. Just a quick fix for name haha ;))

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,153 @@
...
+      if: |
+        github.event.comment.author_association != 'MEMBER' &&
```
tthvo, line 22: This might be silly and depends on preferences. Do u think its better to write:

```bash
!(github.event.comment.author_association == 'MEMBER' ||
        github.event.comment.author_association == 'OWNER')
&& !contains(github.event.issue.labels.*.name, 'safe-to-test')
```

This translate to:

`If the user is neither member or owner and PR does not have the `safe-to-test` label.`

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,153 @@
...
+        github.event.comment.author_association != 'OWNER' &&
+        !contains(github.event.issue.labels.*.name, 'safe-to-test')
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
+          to resolve the issue.
+    - name: Fail if command permission is not allowed
+      if: |
+        github.event.comment.author_association != 'MEMBER' &&
```
tthvo, line 32: We should also have a check so that only member, owner, and the author themselves can run the command in their own PR.

This prevents anyone from randomly run the command.

tthvo (commented):
Nicee! Almost there. After the next patch, can u show us a sample PR on ur fork to test these out?

.github/workflows/pr-ci.yml
tthvo: Also, should it be `github.event.issue.comment.author_association` tho? 

https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,169 @@
...
+        !(github.event.comment.author_association == 'MEMBER' ||
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.issue.user.name == github.event.comment.user.name)
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers 
+          to resolve the issue.
+     - name: Fail if safe-to-test applied and command permission is not allowed
+      if: |
```
tthvo, line 47: I think u can just use this if for step `Fail if command permission is not allowed` and remove this.

tthvo: Btw, pls rename it to `Fail if command permission is denied` just for consistency.

mwangggg:
https://github.com/mwangggg/cryostat/pull/19

.github/workflows/pr-ci.yml
tthvo: Something like should work?

```yaml
!(github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
&& 
(!contains(github.event.issue.labels.*.name, 'safe-to-test') || github.event.issue.user.name != github.event.comment.user.name))
```

This should translate to:

> Permission denied if the user is nether OWNER or MEMBER and either `safe-to-test` label is not applied or the user is not the author when the label is applied.

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,152 @@
...
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
+    steps:
+    - name: Fail if needs-triage label applied 
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      run: exit 1
+    - name: Show warning if permission is denied
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+        && (!contains(github.event.issue.labels.*.name, 'safe-to-test') || github.event.issue.user.name == github.event.comment.user.name)
```
tthvo, line 23: Should be `github.event.issue.user.name != github.event.comment.user.name` here?

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,152 @@
...
       matrix:
         arch: [amd64, arm64]
-    needs: [code-analysis, build-and-test]
-    if: always() && github.repository_owner == 'cryostatio'
+    outputs: 
+      amd64_image: ${{ steps.amd64_image.outputs.image }}
+      arm64_image: ${{ steps.arm64_image.outputs.image }}
+    needs: [build-and-test, checkout-branch]
+    env:
+      head_sha: ${{ needs.checkout-branch.outputs.PR_head_sha}}
```
tthvo, line 98: Need a space at the closing `}}`.

web-client
tthvo, line 1: Ah also remove this pls :))

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,152 @@
...
+    needs: [push-to-ghcr]
+    env:
+      amd64_image: ${{ needs.push-to-ghcr.outputs.amd64_image }}
+      arm64_image: ${{ needs.push-to-ghcr.outputs.arm64_image }}
+    steps:
+    - name: Create markdown table
+      id: md-table
+      uses: petems/csv-to-md-table-action@v3.0.0
+      with: 
+        csvinput: |
```
tthvo, line 136: Lets use lower `amd64` and `arm64` here.

tthvo:
@mwangggg Rebase?

tthvo (commented):
Looks good! Do you have a sample run with latest changes?

I don't have permissions in https://github.com/mwangggg/cryostat/pull/19

.github/workflows/pr-ci.yml
```diff
@@ -1,85 +1,152 @@
...
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      run: exit 1
+    - name: Show warning if permission is denied
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+        && (!contains(github.event.issue.labels.*.name, 'safe-to-test') || github.event.issue.user.name != github.event.comment.user.name)
+      uses: thollander/actions-comment-pull-request@v2
+      with:
+        message: |-
+          You do not have permission to run the /build_test command. Please ask Cryostat @reviewers
```
maxcao13, line 27: Looks good to me too, one more thing: this should be changed to @cryostatio/reviewers

tthvo: Seems like we have been tagging some random person lately :D

andrewazores:
Changes look like they make sense, but reading Actions YAML is always a pain. Sample run, please?

.github/workflows/ci-build-image.yml
```diff
@@ -18,10 +18,6 @@ on:
       image-version:
         description: the Cryostat application version that will be built
         value: ${{ jobs.get-pom-properties.outputs.image-version }}
-    secrets:
-      GH_PKGS_READ_TOKEN:
-        required: true
-        description: read-only token for pulling artifacts from GitHub Packages
```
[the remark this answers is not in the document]
tthvo: I am reading from this:

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry

Looks like `GITHUB_TOKEN` should work as long as they have `packages:read` scope on it, which it should since its a `issue_comment` event.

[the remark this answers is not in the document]
tthvo: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

andrewazores:
Any other review comments?

maxcao13 (approved):
Looks good to me, great work on this!

andrewazores (approved):

PR #1614, opened 2023-08-17
test(smoketest): set jfr-datasource for JDP and Podman discovery

andrewazores:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/jfr-datasource/pull/204
Depends on https://github.com/cryostatio/jfr-datasource/pull/204

## Description of the change:
After the above PR is merged, `jfr-datasource` images will be published in JVM mode and supporting JMX connections. This adds configuration matching how `cryostat-reports` is deployed in `smoketest.sh` since both containers are built similarly.

## Motivation for the change:
This allows `jfr-datasource` to be a JMX target in the smoketest environment.

## How to manually test:
2. Select the target with JMX port `11223`, which is the `jfr-datasource`. Test that general Cryostat functionality works on this target.

### DIFF
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -254,7 +254,7 @@ runJfrDatasource() {
     fi
     # limits set to match operator defaults:
     # https://github.com/cryostatio/cryostat-operator/blob/2d386930dc96f0dcaf937987ec35874006c53b61/internal/controllers/common/resource_definitions/resource_definitions.go#L66
-    local RJMX_PORT=11223
+    local RJMX_PORT=10111
     podman run \
         --name jfr-datasource \
         --pull "${PULL_IMAGES}" \
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -261,6 +261,11 @@ runJfrDatasource() {
         --pod cryostat-pod \
         --cpus 0.1 \
         --memory 512m \
+        --label io.cryostat.discovery="true" \
+        --label io.cryostat.jmxHost="localhost" \
+        --label io.cryostat.jmxPort="${RJMX_PORT}" \
+        --restart on-failure \
+        --env JAVA_OPTS="-Dcom.sun.management.jmxremote.autodiscovery=true -Dcom.sun.management.jmxremote.port=${RJMX_PORT} -Dcom.sun.management.jmxremote.rmi.port=${RJMX_PORT} -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false" \
         --rm -d "${DATASOURCE_IMAGE}"
 }
 


smoketest.sh
```diff
@@ -256,6 +256,11 @@ runJfrDatasource() {
         --name jfr-datasource \
         --pull "${PULL_IMAGES}" \
         --pod cryostat-pod \
+        --label io.cryostat.discovery="true" \
+        --label io.cryostat.jmxHost="localhost" \
+        --label io.cryostat.jmxPort="${RJMX_PORT}" \
```
tthvo, line 261: Need to define `RJMX_PORT`?

tthvo (commented):
I am experiencing some thread blocks while running smoketest.

https://gist.github.com/tthvo/19ef7e7f9a2c12c974f6710cebb9fdde

andrewazores:
> I am experiencing some thread blocks while running smoketest. Does this have something to do with container resource constraints?
> 
> https://gist.github.com/tthvo/19ef7e7f9a2c12c974f6710cebb9fdde

These particular thread blocks you're seeing are happening on startup, so perhaps there is something that can be done to move those startup tasks onto worker threads instead of the main Vertx event loop, but I'm not sure.

In any case, this probably highlights that the limits we set are too low and cause performance issues.

> Looks like setting `--memory` is actually configuring the memory/cpu limit, not a request.
> 
> ```shell
> -m, --memory <number>[<unit>]                  Memory limit (format: <number>[<unit>], where unit = b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes))
>       --memory-reservation <number>[<unit>]      Memory soft limit (format: <number>[<unit>], where unit = b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes))
> --cpus float                               Number of CPUs. The default is 0.000 which means no limit
> ```
> 
> I think the table here suggested the request resources only, not limits. https://github.com/cryostatio/cryostat-operator/blob/main/docs/config.md#resource-requirement

It's true, but if I understand the requests/limits mechanisms correctly, the requests are used for scheduling containers/Pods onto nodes to ensure that at least that amount of resource is available. So if these are the default requests we set then it should result in a Cryostat deployment that runs reasonably well with only those resources, since there is no guarantee there are actually any more resources than that. Of course, not every resource is actually equivalent across different hardware - perhaps the memory usage is different between amd64 and amd64 for example, and certainly CPU time means different things when running on different kinds of CPU.

tthvo:
> Basically, the point here is to make smoketest.sh simulate the worst-case scenario for an Operator deployment, so we can make sure we aren't only seeing that things run well when the Cryostat deployment is allowed to use a whole machine's worth of resources, but that it can still run well with some reasonable resource constraints that allow it to run with reasonable and justifiable cost in real deployments.

Oh right!! Make sense! I have been wondering that

> In any case, this probably highlights that the limits we set are too low and cause performance issues.

True, I changed it to `--cpu 0.3` and the issue is solved.

andrewazores:
> > Basically, the point here is to make smoketest.sh simulate the worst-case scenario for an Operator deployment, so we can make sure we aren't only seeing that things run well when the Cryostat deployment is allowed to use a whole machine's worth of resources, but that it can still run well with some reasonable resource constraints that allow it to run with reasonable and justifiable cost in real deployments.
> 
> Oh right!! Make sense! I have been wondering that
> 
> > In any case, this probably highlights that the limits we set are too low and cause performance issues.
> 
> True, I changed it to `--cpu 0.3` and the issue is solved.

File a bug please - the container limits on `smoketest.sh`/`run.sh` and the corresponding default requests set by the Operator (and -helm?) should be investigated and probably increased then.

tthvo:
https://github.com/cryostatio/cryostat/issues/1629

tthvo (approved):
Tho, these changes look good! Worked well while testing with commit before resource limits.

Issue #1613, opened 2023-08-14
[Story] Implement HTTP JFR snapshot creation


andrewazores:
Depends on https://github.com/cryostatio/cryostat-agent/issues/185

PR #1566, opened 2023-06-28
chore(svc): extract EventOptionsBuilder to -core and use new CryostatFlightRecorderService

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] Signed the last commit: `git commit --amend --signoff`
_______________________________________________

Related to https://github.com/cryostatio/cryostat-agent/issues/162
Depends on https://github.com/cryostatio/cryostat-core/pull/245

## Description of the change:
Refactor to extract some logic out to `-core` and use an additional abstraction on top of the JMC `IFlightRecorderService`. The new `CryostastFlightRecorderService` subinterface provides higher-level operation abstractions to support the `AgentService`/`AgentConnection`/`AgentClient` HTTP-based communications, so that existing code needs minimal changes to perform operations like dynamically starting recordings by providing arguments like event template name+type, or later the whole event template XML document (for custom event templates), rather than using the JMC `IConstrainedMap` model.

## How to manually test:
1. *Run CRYOSTAT_IMAGE=quay.io... sh smoketest.sh...*
2. Everything should work as usual

PR #1627, opened 2023-08-23
feat(agent): implement HTTP JFR snapshot creation

aali309:
# Welcome to Cryostat! 👋

Fixes: #1613 

## Description of the change:
This change allows users to start snapshot via HTTP

## Motivation for the change:
This continues the effort to bring Cryostat application feature parity between JMX and HTTP connection methods.

## How to manually test:
1. Same setup as in [https://github.com/cryostatio/cryostat/pull/1566](https://github.com/cryostatio/cryostat/pull/1566)
2. Test against the smoketest.sh http://localhost:9988/ agent target. Go to the Cryostat Web UI, Recordings, Create, and check that a recording can be started and a snapshot can be started.

### DIFF
--- a/src/main/java/io/cryostat/net/AgentJFRService.java
+++ b/src/main/java/io/cryostat/net/AgentJFRService.java
@@ -24,7 +24,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
 import org.openjdk.jmc.common.unit.IConstrainedMap;
@@ -251,23 +250,14 @@ class AgentJFRService implements CryostatFlightRecorderService {
 
     @Override
     public void updateRecordingOptions(
-        IRecordingDescriptor recordingDescriptor, IConstrainedMap<String> newSettings)
-        throws FlightRecorderException {
+            IRecordingDescriptor recordingDescriptor, IConstrainedMap<String> newSettings)
+            throws FlightRecorderException {
         try {
             long recordingId = recordingDescriptor.getId();
-                
-            Set<String> keys = newSettings.keySet();
-            boolean nameKeyExists = keys.contains("name");
-        
-        if (nameKeyExists) {
-            // Get the new name from newSettings
-            String newName = (String) newSettings.get("name");
-                // Update the recording name using the AgentClient's method
-                client.updateRecordingOptions(recordingId, newName)
+            client.updateRecordingOptions(recordingId, newSettings)
                     .toCompletionStage()
                     .toCompletableFuture()
                     .get();
-            }
         } catch (ExecutionException | InterruptedException e) {
             throw new FlightRecorderException("Failed to update recording options", e);
         }
--- a/src/main/java/io/cryostat/net/AgentJFRService.java
+++ b/src/main/java/io/cryostat/net/AgentJFRService.java
@@ -24,6 +24,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
 import org.openjdk.jmc.common.unit.IConstrainedMap;
@@ -250,14 +251,23 @@ class AgentJFRService implements CryostatFlightRecorderService {
 
     @Override
     public void updateRecordingOptions(
-            IRecordingDescriptor recordingDescriptor, IConstrainedMap<String> newSettings)
-            throws FlightRecorderException {
+        IRecordingDescriptor recordingDescriptor, IConstrainedMap<String> newSettings)
+        throws FlightRecorderException {
         try {
             long recordingId = recordingDescriptor.getId();
-            client.updateRecordingOptions(recordingId, newSettings)
+                
+            Set<String> keys = newSettings.keySet();
+            boolean nameKeyExists = keys.contains("name");
+        
+        if (nameKeyExists) {
+            // Get the new name from newSettings
+            String newName = (String) newSettings.get("name");
+                // Update the recording name using the AgentClient's method
+                client.updateRecordingOptions(recordingId, newName)
                     .toCompletionStage()
                     .toCompletableFuture()
                     .get();
+            }
         } catch (ExecutionException | InterruptedException e) {
             throw new FlightRecorderException("Failed to update recording options", e);
         }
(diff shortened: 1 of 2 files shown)


aali309:
On this, what differentiates the endpoint between a recording and a snapshot recording? or the logic will be taken care of on the agent side depending on the template created? i.e custom or if a recordings exist to successfully create a snapshot? on the method `handleStartRecordingOrSnapshot`

andrewazores:
> On this, what differentiates the endpoint between a recording and a snapshot recording? or the logic will be taken care of on the agent side depending on the template created? i.e custom or if a recordings exist to successfully create a snapshot? on the method `handleStartRecordingOrSnapshot` in `cryostat-agent`[src/main/java/io/cryostat/agent/remote/RecordingsContext.java](src/main/java/io/cryostat/agent/remote/RecordingsContext.java)

This needs to be coordinated with Ming's PR on the Agent (in this case acting as the server): https://github.com/cryostatio/cryostat-agent/pull/186

aali309:
> > On this, what differentiates the endpoint between a recording and a snapshot recording? or the logic will be taken care of on the agent side depending on the template created? i.e custom or if a recordings exist to successfully create a snapshot? on the method `handleStartRecordingOrSnapshot` in `cryostat-agent`[src/main/java/io/cryostat/agent/remote/RecordingsContext.java](src/main/java/io/cryostat/agent/remote/RecordingsContext.java)
> 
> This needs to be coordinated with Ming's PR on the Agent (in this case acting as the server): [cryostatio/cryostat-agent#186](https://github.com/cryostatio/cryostat-agent/pull/186)

I did.. Just confirming coz the endpoint is the same and the logic is taken care of in the `handleStartRecordingOrSnapshot` which @mwangggg  made

andrewazores:
The endpoint is the same, but the Agent-side change does expect some particular request format in the payload to indicate that the request is for a snapshot:

https://github.com/cryostatio/cryostat-agent/pull/186/files#diff-fb08efc7a13d2b872d8abd6e8d6f983bc6e1960cb08979998949d371aa66803dR373

andrewazores (commented):
Looking good. I think the only remaining thing to do is a one-liner change in `AgentJFRService` to call this method at the appropriate place (`getSnapshotRecording()` I think).

src/main/java/io/cryostat/net/AgentJFRService.java
```diff
@@ -174,7 +174,11 @@ public List<String> getServerTemplates() throws FlightRecorderException {
 
     @Override
     public IRecordingDescriptor getSnapshotRecording() throws FlightRecorderException {
-        throw new UnimplementedException();
+        try {
+            return client.startSnapshot().toCompletionStage().toCompletableFuture().get();
+        } catch (ExecutionException | InterruptedException e) {
+            throw new FlightRecorderException("Failed to creat snapshot recording");
```
andrewazores, line 180: creat -> create

aali309: ooh my GOD!! so sorry for this

andrewazores:
@aali309 @mwangggg please test this out between the two of you and both sides of the changes. I think I am seeing a bug when I try to use this, but perhaps I just built the pieces wrong.

mwangggg:
I'm getting a 500 Internal Server Error too

andrewazores:
IIRC what I saw in the web-client was a `500`, but looking at the logs the cause of that was that Cryostat received a `400` from the Agent on the request.

mwangggg:
I think I know what the problem is- the IsValid() logic is incorrect on the agent side

andrewazores:
That's what I thought, too.

mwangggg:
hmm there's another error (or the same one)

andrewazores:
- https://github.com/cryostatio/cryostat/blob/4afbb8f4264cd0cda123854cf101bfb2f524e80c/src/main/java/io/cryostat/recordings/RecordingTargetHelper.java#L325
- https://github.com/openjdk/jmc/blob/master/application/org.openjdk.jmc.flightrecorder.configuration/src/main/java/org/openjdk/jmc/flightrecorder/configuration/recording/RecordingOptionsBuilder.java

Basically, after creating the remote snapshot, the server also wants to be able to rename the recording. Since we have more control over this process when doing it over HTTP it might make sense to simply provide that name on the original request, rather than perform the request to create the snapshot and immediately try to rename it, but that would require some deeper refactoring and changing of logic since the JMX side does not support that flow. For simplicity I think we can just proceed with maintaining that same logic even over HTTP: create the snapshot first, then follow-up with a request to rename it.

The `updateRecordingOptions` method allows for the properties handled by the `RecordingOptionsBuilder` to be updated - these are generally the same properties that can be set when first creating a recording, so the name and duration in addition to the "Advanced Options" we expose in the UI (toDisk, maxAge, maxSize). I don't think we want/need to support the other properties.

I would suggest these might be done by the Cryostat server sending a `PATCH /recordings/:id` to the Agent, and the request body being JSON like `{ "propertyName": updatedValue, "propertyName2": updatedValue2 }`, so more concretely in this case it would be `{ "name": "snapshot-rename" }`. The Agent can accept this request, parse the body, and use the JSON structure to determine what changes it should apply to the recording with the given ID.

aali309:
### Need help debugging this 
When I run test if  `snapShot` is being created... It appears that it is created but I get the error `500`
`Request failed (500 Internal Server Error)
org.openjdk.jmc.rjmx.services.jfr.FlightRecorderException: Failed to create snapshot recording caused by RuntimeException: Unknown failure`
 but checking the recordings, a snapshot is created 
image
but viewing the report, this is what I see `{"meta":{"type":"text/plain","status":"Internal Server Error"},"data":{"reason":"java.lang.NullPointerException: Cannot invoke \"io.vertx.core.buffer.Buffer.getBytes()\" because \"b\" is null"}}` 
if I am right the error is pointing at `handleGetRecording' method in the `recordingContext.java` class in cryostat-agent

andrewazores:
If you check the Cryostat server logs, you should be able to find the stack trace of that NullPointerException and this will show you the exact line of code where it is occurring, so you will be able to tell which `Buffer b` is null.

andrewazores:
image

image

@aali309 @mwangggg I haven't looked into which side is causing this, but it looks to me like the rename attempt is failing because there are quotation marks surrounding the new name.

So from the evidence in these screenshots it seems:
1. the Cryostat server asks the Agent to create a new snapshot
2. the Agent creates the snapshot with its default name (just "Snapshot")
3. the Cryostat server asks the Agent to rename that recording to "snapshot-n" where n is a numeric ID. The rename occurs, but either the server actually asked for "\"snapshot-n\"" or the agent interpreted it that way
4. the Cryostat server then checks with the Agent whether there is a recording name "snapshot-n" (or "\"snapshot-n\""), and this fails, either because the server asked for the wrong thing or the Agent interpreted it the wrong way

andrewazores (approved):

Issue #1634, opened 2023-08-24
[Task] `/build_test` command enhancements


andrewazores:
### Describe the feature

Since https://github.com/cryostatio/cryostat/pull/1599 , reviewers can comment `/build_test` on PRs to trigger a CI run for integration tests and preview images. For example,

- https://github.com/cryostatio/cryostat/pull/1631#issuecomment-1692396952
- https://github.com/cryostatio/cryostat/actions/runs/5968826929

It would be nice if:
- the bot could leave a comment of the Actions run when it starts
- the bot could leave a comment noting whether the tests passed or failed
- the workflow could be parallelized (again) - static analysis (shellcheck/spotbugs/spotless) can be done separately, with the unit testing and OCI packaging done separately from that with no dependency on static analysis, then integration testing and registry pushing can be done separately with only a dependency on the OCI packaging.
- there could be a separate comment for retrying integration tests using the most recently-built images, ie `/build_test` first, then `/retest` to simply rerun the tests without a rebuild, if the reviewer suspects that the tests are simply flaky
- is it possible for the `/build_test` to add a Status Check to the PR? Since moving these checks off from the `pull_request_target` and into a comment trigger, the checks do not seem to be linked anymore.

![image](https://github.com/cryostatio/cryostat/assets/3787464/aa8fe6e0-56e8-4404-9c41-4230e7de93c1)

![image](https://github.com/cryostatio/cryostat/assets/3787464/2df4053c-20fc-454e-9316-9ace82dada4e)

andrewazores:
@mwangggg you're probably a bit sick of this work by now, but perhaps later on you can revisit it?

andrewazores:
I think it would also be nice for the automatic CI run on every PR update (every push to the PR branch) to still trigger the build and unit tests, just skip the image build/push and integration testing until manually triggered.

andrewazores:
Another potentially useful thing: https://github.com/marketplace/actions/slash-command-dispatch

PR #1635, opened 2023-08-25
ci(permissions): set explicit write permission requests with restrictive GITHUB_TOKEN

tthvo:
# Welcome to Cryostat! 👋

Related to #1599 

## Description of the change:

Explicitly specify `package:write` (for push) permission. This would also mean other permissions that are listed for this job are denied.

We should implement strict permission requests in #1634 after taking in consideration of third-party actions. Also, some scope might enclose other scopes.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,46 +1,23 @@
 categories:
-    - title: 'Features'
-      label: 'feat'
-    - title: 'Fixes'
-      label: 'fix'
-    - title: 'Documentation'
-      label: 'docs'
-    - title: 'Maintenance'
-      labels:
-          - 'chore'
-          - 'ci'
-          - 'cleanup'
-          - 'perf'
-          - 'refactor'
-          - 'style'
-          - 'test'
-
-autolabeler:
-    - label: 'feat'
-      title: '/^feat:/'
-    - label: 'fix'
-      title: '/^fix:/'
-    - label: 'docs'
-      title: '/^docs:/'
-
-    - label: 'chore'
-      title: '/^chore:/'
-    - label: 'ci'
-      title: '/^ci:/'
-    - label: 'cleanup'
-      title: '/^cleanup:/'
-    - label: 'perf'
-      title: '/^perf:/'
-    - label: 'refactor'
-      title: '/^refactor:/'
-    - label: 'style'
-      title: '/^style:/'
-    - label: 'test'
-      title: '/^test:/'
+  - title: 'Features'
+    label: 'feat'
+  - title: 'Fixes'
+    label: 'fix'
+  - title: 'Documentation'
+    label: 'docs'
+  - title: 'Maintenance'
+    labels:
+      - 'chore'
+      - 'ci'
+      - 'cleanup'
+      - 'perf'
+      - 'refactor'
+      - 'style'
+      - 'test'
 
 template: |
-    ## Contributors
-    $CONTRIBUTORS
-    ## What's Changed
-    $CHANGES
+  ## Contributors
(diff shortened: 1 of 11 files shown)


tthvo:
I think forks are using permissive mode on the `GITHUB_TOKEN` so we was not able to catch this. This is what I am seeing my default on my fork.

image

tthvo:
Opps I spoke too soon :(( From this [run](https://github.com/cryostatio/cryostat/actions/runs/5968826929/job/16194311076), seems like we are also using permissive mode and the token already has `packages:write` permission.

image

I wonder if it has something to do with organization settings?

tthvo:
If the above is true, should we also change the default scope of `GITHUB_TOKEN` to ready-only? If not, I think we can close the PR.

andrewazores:
Okay, I updated the package settings so that the Actions runners within this source repo should also have access to the corresponding container registry repo. The token should be read-only unless it's being used specifically by an action/workflow step that actually requires write access to something (like the final push step).

tthvo:
Sounds good! I will do some more research into actions and figure out the permissions. Marked as draft.

tthvo:
Sample runs: https://github.com/tthvo/cryostat/pull/35
Actions: https://github.com/tthvo/cryostat/actions/workflows/release-drafter.yml

No changes to push event except release drafter. All pull request workflows seem to work fine.

tthvo:
I also remove autolabel option for `release-drafter` since it does not seem to work (perhaps because of the custom `.github/labeler.yaml`.

Some workflows:
- `.github/workflows/pr-labeled.yml`
- `.github/workflows/semantic-pr.yml`
- `.github/workflows/linked-issue.yml`) 

need to run `pull_request_target` in order to comment on the PR.

tthvo:
After this, we can select Read Only option as seen here: https://user-images.githubusercontent.com/68053619/263415796-2bd41f11-cc3d-418c-9f11-2fe37bff890b.png

tthvo:
Thats odd. Not sure why there is 2 mergify comments.

tthvo:
Looks like we can use https://docs.mergify.com/actions/dismiss_reviews/ to mark mergify review as stale automatically.

tthvo:
Image cleanup is failing due to missing permissions (tho, already added `packages:write`). Investigating...

tthvo:
It failed silently while trying to find all images that match the expression (i.e. 404). Tried setting all permissions to `write` here but the action cannot find the images.

https://github.com/tthvo/cryostat/actions/runs/6032732709/job/16368343556

 I am seeing https://github.com/bots-house/ghcr-delete-image-action/issues/5 issue with the source code at https://github.com/r26D/ghcr-delete-image-action/blob/2dcff984378112992d20afbfa628308196434d4f/utils.js#L110C17-L110C17

Looks like user's owned packages are not supported. So, I can't verify the use of `GITHUB_TOKEN` on my fork. Lets try it first on main and if it failed, we can fall back to PAT for this action @andrewazores ?

tthvo:
This run uses a PAT that has `packages:delete` but stills failed (i.e. passed but 404 ignored) on my fork: https://github.com/tthvo/cryostat-operator/actions/runs/6033226424/job/16369823264

andrewazores:
We can try merging this to `main` and see how it behaves. Worst case scenario is we just revert it.

andrewazores:
Needs a rebase first.

tthvo:
Sounds good!!

tthvo:
Looks good here! https://github.com/cryostatio/cryostat/actions/runs/6041751746/job/16395384777

Issue #1633, opened 2023-08-24
[Bug] Agent discovery desync


andrewazores:
### Current Behavior

Occasionally, the Cryostat server and Agent instances become desynchronized regarding the Agent's registration state as a discovery plugin. This results in both the server and Agent failing to communicate with each other and sending each other `4xx` responses.

### Expected Behavior

The server and Agent should stay in a good synchronization state when properly configured to point to communicate with each other. If there is ever an interruption, ex. if one or the other briefly goes offline, then the self-healing behaviour protocol should restore communications the next time one of the components tries to contact the other.

### Steps To Reproduce

TBD

andrewazores:
```
sample-app_1          | Aug 25, 2023 2:54:30 PM io.cryostat.agent.Agent lambda$main$1
sample-app_1          | INFO: Registration state: REFRESHING
sample-app_1          | Aug 25, 2023 2:54:30 PM io.cryostat.agent.CryostatClient register
sample-app_1          | INFO: POST http://cryostat:8181/api/v2.2/discovery HTTP/1.1
cryostat_1            | 2023-08-25 14:54:30,889 INFO  [io.cry.dis.DiscoveryPlugin$PluginCallback] (executor-thread-1) Using stored credentials id:3 referenced in ping callback userinfo
sample-app_1          | Aug 25, 2023 2:54:30 PM io.cryostat.agent.WebServer$RequestLoggingFilter doFilter
sample-app_1          | INFO: GET /
sample-app_1          | Aug 25, 2023 2:54:30 PM io.cryostat.agent.WebServer$CompressionFilter doFilter
sample-app_1          | INFO: Using no encoding
sample-app_1          | Aug 25, 2023 2:54:30 PM io.cryostat.agent.WebServer$RequestLoggingFilter doFilter
sample-app_1          | INFO: GET / : 401 0ms
cryostat_1            | 2023-08-25 14:54:30,897 ERROR [io.cry.dis.DiscoveryPlugin$Listener] (executor-thread-1) Discovery Plugin ping failed: org.jboss.resteasy.reactive.ClientWebApplicationException: Received: 'Unauthorized, status code 401' when invoking: Rest Client method: 'io.cryostat.discovery.DiscoveryPlugin$PluginCallback#ping'
cryostat_1            | 	at org.jboss.resteasy.reactive.client.impl.RestClientRequestContext.unwrapException(RestClientRequestContext.java:185)
[...]

PR #1636, opened 2023-08-29
test(smoketest): tinker with resource limits

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #1629
See also https://github.com/cryostatio/cryostat-reports/pull/136 https://github.com/cryostatio/jfr-datasource/pull/213

## How to manually test:
1. Ask me for a sample 10MB JFR file, or collect your own
2. Run `smoketest.sh` with test image
3. Upload 10MB JFR file into Archives
4. Check that `View Report` and `View in Grafana` both work
5. Click around UI and ensure other functionalities generally work as expected
6. Bonus: repeat this with a larger sample JFR file. The next larger file I have is 50MB and that's too large for the resource limits applied here.

PR #1646, opened 2023-08-31
fix(rules): rule triggering on late-connecting targets

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits using a GPG signature](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification)

**To recreate commits with GPG signature** `git fetch upstream && git rebase --force --gpg-sign upstream/main`
_______________________________________________

Fixes: #1494

## Description of the change:
This changeset:
1. handles `MODIFIED` JVM discovery events in a couple of places. Primarily to avoid an `UnsupportedOperationException` that was wrongly thrown and could cause rule activations to fail, and then secondarily to use these events to trigger a proactive re-check of non-connectable targets in case the discovery update made the target connectable.
2. Enhances the non-connectable target recheck logic. Previously these were always checked on a fixed 15 second timer. Now, the timer fires every 2 seconds, but each non-connectable target will only be rechecked according to some backoff timing logic, and after 1 minute of retry failures, that target will not be rechecked again.

## Motivation for the change:
This should fix the bug where rules may not activate against targets that appear but are not immediately connectable, and should also shorten the time it takes for Cryostat to detect that a slowly-connectable target becomes connectable.

## How to manually test:
Deploy test image in `crc` or other OpenShift/k8s. Deploy a sample application. Quite often, the k8s server notifies Cryostat of the new Endpoints object before the target JVM is actually ready to accept JMX connections, so the initial JVM ID retrieval attempt fails and so does the initial attempt to trigger automated rules. After this change, this workflow should again work as expected and Cryostat should notice that the target becomes connectable, and then triggers automated rules against it.

PR #1650, opened 2023-09-01
fix(discovery): plugin registration bugfixes

andrewazores:
# Welcome to Cryostat! 👋

Fixes: https://github.com/cryostatio/cryostat/issues/1633
See also https://github.com/cryostatio/cryostat-agent/pull/193
Based on #1636

## Description of the change:
Improves error handling and cleanup when plugin registrations fail and are associated with stored credentials.

## Motivation for the change:
This along with https://github.com/cryostatio/cryostat-agent/pull/193 increases the resiliency of the server/agent registration system so that temporary networking failures or registration conflicts or other bugs are less likely to leave the server and agent both in a state where neither recognizes the other and yet neither is able to clean up and reset the registration status.

## How to manually test:
2. Check that agent instances properly register & publish, and are visible in the Topology view
3. `podman kill` some agent instances to prevent clean shutdown
4. `podman run` (reference `smoketest.sh` for exact invocation) to restart some of the killed agent instances to spin them back up
5. stop the smoketest, then restart it without clearing databases, and ensure that the agent instances are able to re-register. This may take a couple of minutes for the server to recognize that the old state is stale, clear it, and allow agents to register again.

### DIFF
--- a/smoketest.sh
+++ b/smoketest.sh
@@ -125,7 +125,7 @@ runDemoApps() {
         --label io.cryostat.discovery="true" \
         --label io.cryostat.jmxHost="localhost" \
         --label io.cryostat.jmxPort="9089" \
-        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.0
+        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.1
 
     podman run \
         --name vertx-fib-demo-1 \
@@ -142,9 +142,9 @@ runDemoApps() {
         --env CRYOSTAT_AGENT_AUTHORIZATION="Basic $(echo user:pass | base64)" \
         --pod cryostat-pod \
         --label io.cryostat.discovery="true" \
-        --label io.cryostat.jmxHost="localhost" \
+        --label io.cryostat.jmxHost="vertx-fib-demo-1" \
         --label io.cryostat.jmxPort="9093" \
-        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.0
+        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.1
 
     podman run \
         --name vertx-fib-demo-2 \
@@ -162,10 +162,10 @@ runDemoApps() {
         --env CRYOSTAT_AGENT_AUTHORIZATION="Basic $(echo user:pass | base64)" \
         --pod cryostat-pod \
         --label io.cryostat.discovery="true" \
-        --label io.cryostat.jmxHost="localhost" \
+        --label io.cryostat.jmxHost="vertx-fib-demo-2" \
         --label io.cryostat.jmxPort="9094" \
         --label io.cryostat.jmxUrl="service:jmx:rmi:///jndi/rmi://localhost:9094/jmxrmi" \
-        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.0
+        --rm -d quay.io/andrewazores/vertx-fib-demo:0.13.1
 
     podman run \
         --name vertx-fib-demo-3 \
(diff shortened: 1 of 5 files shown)


src/main/java/io/cryostat/net/NetworkModule.java
```diff
@@ -152,7 +152,7 @@ static Vertx provideVertx() {
         return Vertx.vertx(
                 new VertxOptions()
                         .setPreferNativeTransport(true)
-                        .setEventLoopPoolSize(defaults.getEventLoopPoolSize() + 4));
+                        .setEventLoopPoolSize(defaults.getEventLoopPoolSize() + 6));
```
tthvo, line 155: What is the choice of the additional 4 threads previously and 6 here?

andrewazores: It's pretty hacky.

https://github.com/cryostatio/cryostat/pull/1658

I noticed that the Podman discovery was broken again with this PR even after rebasing on top of that one. I haven't chased exactly why or what is causing the additional blocking work at startup, but there is some thread blockage going on that prevents the platform discovery strategy check from correctly finishing, and increasing the size of the worker pool allows the check to succeed.

smoketest.sh
```diff
@@ -125,7 +125,7 @@ runDemoApps() {
         --label io.cryostat.discovery="true" \
         --label io.cryostat.jmxHost="localhost" \
```
tthvo, line 126: Update here too.

tthvo (commented):
Comments for testing Mergify config :))

smoketest.sh
andrewazores: Not sure what you mean...

src/main/java/io/cryostat/net/NetworkModule.java
tthvo: hmm yehhh make sense! Probably, we should document it somewhere.

Tho, in environment where resources are more relaxed, I suppose these settings might just be a little "overkill"?

smoketest.sh
tthvo: Oh sorry, lets use `vertx-fib-demo-0`?

tthvo: Btw, Mergify seems to work as expected :))

src/main/java/io/cryostat/net/NetworkModule.java
andrewazores: Vertx's default setting is to have `cpuCores * 2` threads in the pool, so definitely adding a bunch of extra threads here like this is going to increase the size of the pool substantially for most deployments. The proper fix would be to avoid doing blocking work on startup, or at least move that blocking work off of a Vertx event loop thread, but that's a relatively deep change and I haven't had enough time to look into it yet. If I have time before the next 2.4 release I will, but it seems riskier to do that than to simply bump up the minimum pool size a bit for now.

smoketest.sh
tthvo: Opps, https://github.com/cryostatio/cryostat/pull/1646#issuecomment-1710598137

andrewazores: It's close to working as expected but slightly different. I think what it's doing is good enough for now, since if reviewers have their GitHub notifications turned on, then they should get notified once they are tagged for review, as well as when the author comments with `/request_review` later, even if the label doesn't stick.

tthvo: SOunds good!

src/main/java/io/cryostat/net/NetworkModule.java
tthvo: Sounds good! I think we can file a separate issue to track. Perhaps, this could be fixed in 3.0.

andrewazores: It will already be fundamentally different in 3.0 since the whole startup mechanism is completely different. I think the 3.0 prototype already completely avoids this kind of problem...

aali309:
> Comments for testing Mergify config :))

Looks like this worked as expected.

src/main/java/io/cryostat/net/NetworkModule.java
tthvo: Oh yes right that just slipped me mind. Thanks :))

mwangggg (approved):

Issue #1649, opened 2023-09-01
[Bug] Podman discovery not working


mwangggg:
### Current Behavior

When I run `sh smoketest.sh`, the topology view in Cryostat doesn't show the podman realm. 
![Screenshot from 2023-09-01 14-38-43](https://github.com/cryostatio/cryostat/assets/90855268/c470dfbf-f398-42c3-86bd-f32a77a501c8)

The logs show:

``` bash
Sep 01, 2023 6:35:26 PM io.cryostat.core.log.Logger info
INFO: HTTPS service running on https://localhost:8181
Sep 01, 2023 6:35:26 PM io.cryostat.core.log.Logger info
INFO: Deployed io.cryostat.net.HttpServer Verticle [0139d32f-0fa8-4605-a66b-c07f70dc582c]
Sep 01, 2023 6:35:28 PM io.cryostat.core.log.Logger info
INFO: Testing DockerPlatformStrategy Availability via /var/run/docker.sock
Sep 01, 2023 6:35:28 PM io.cryostat.core.log.Logger info
INFO: DockerPlatformStrategy available? false
Sep 01, 2023 6:35:28 PM io.cryostat.core.log.Logger info
INFO: Testing PodmanPlatformStrategy Availability via /run/user/0/podman/podman.sock
Sep 01, 2023 6:35:29 PM org.slf4j.impl.JDK14LoggerAdapter fillCallerData
WARNING: Thread Thread[vert.x-eventloop-thread-0,5,main] has been blocked for 2502 ms, time limit is 2000 ms
Sep 01, 2023 6:35:30 PM io.cryostat.core.log.Logger error
SEVERE: Exception thrown
java.util.concurrent.TimeoutException
	at java.base/java.util.concurrent.CompletableFuture.timedGet(CompletableFuture.java:1960)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2095)
	at io.cryostat.platform.internal.PodmanPlatformStrategy.testPodmanApi(PodmanPlatformStrategy.java:122)
	at io.cryostat.platform.internal.PodmanPlatformStrategy.isAvailable(PodmanPlatformStrategy.java:86)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
	at java.base/java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1707)
[...]

Issue #1629, opened 2023-08-23
[Bug] Resource limits are too low for smoketest


tthvo:
### Current Behavior

              > I am experiencing some thread blocks while running smoketest. Does this have something to do with container resource constraints?
> 
> https://gist.github.com/tthvo/19ef7e7f9a2c12c974f6710cebb9fdde

These particular thread blocks you're seeing are happening on startup, so perhaps there is something that can be done to move those startup tasks onto worker threads instead of the main Vertx event loop, but I'm not sure.

In any case, this probably highlights that the limits we set are too low and cause performance issues.

> Looks like setting `--memory` is actually configuring the memory/cpu limit, not a request.
> 
> ```shell
> -m, --memory <number>[<unit>]                  Memory limit (format: <number>[<unit>], where unit = b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes))
>       --memory-reservation <number>[<unit>]      Memory soft limit (format: <number>[<unit>], where unit = b (bytes), k (kibibytes), m (mebibytes), or g (gibibytes))
> --cpus float                               Number of CPUs. The default is 0.000 which means no limit
> ```
> 
> I think the table here suggested the request resources only, not limits. https://github.com/cryostatio/cryostat-operator/blob/main/docs/config.md#resource-requirement

It's true, but if I understand the requests/limits mechanisms correctly, the requests are used for scheduling containers/Pods onto nodes to ensure that at least that amount of resource is available. So if these are the default requests we set then it should result in a Cryostat deployment that runs reasonably well with only those resources, since there is no guarantee there are actually any more resources than that. Of course, not every resource is actually equivalent across different hardware - perhaps the memory usage is different between amd64 and amd64 for example, and certainly CPU time means different things when running on different kinds of CPU.

Basically, the point here is to make `smoketest.sh` simulate the worst-case scenario for an Operator deployment, so we can make sure we aren't only seeing that things run well when the Cryostat deployment is allowed to use a whole machine's worth of resources, but that it can still run well with some reasonable resource constraints that allow it to run with reasonable and justifiable cost in real deployments.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/1614#issuecomment-1690016286_
            

### Expected Behavior

The limit should allow the smoketest to run without issue, at least.

> The container limits on smoketest.sh/run.sh and the corresponding default requests set by the Operator (and -helm?) should be investigated and probably increased then.

### Steps To Reproduce

[...]

PR #1658, opened 2023-09-06
fix(startup): platform detection progresses with limited parallelism

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1649
Related to #1629 #1636

## Description of the change:
This:
- updates calls to `ForkJoinPool.commonPool()` to use specific Executor instances where sensible. This may increase some overhead by allowing more threads to be created than there are CPUs to run them, but also helps to isolate various components from each other so that they do not share worker pools and threading bugs in one component do not starve every other component.
- forces the minimum Vert.x event loop pool size to be the same as the default, plus 4 additional at minimum. This worker pool owns the threads that perform the application startup and initialization of all the various Verticles, including the DiscoveryStorage verticle which triggers the platform detection checks. Ideally all verticles should have non-blocking startup procedures but that is not always possible for us, and this same thread pool is also the one that owns threads that handle the Vert.x web-client async responses. If this pool is too small - for example if there are not many CPUs assigned to the container - then the pool can starve itself because startup workers are trying to initialize Verticles, and those Verticles require more workers to perform async HTTP requests to check startup conditions.

## How to manually test:
1. `run.sh` with `--cpus 0.5` or other smaller values
2. Check Topology view or read logs or otherwise validate that all expected platform discoveries succeeded
3. `smoketest.sh`, click around the UI and ensure everything still works as expected

### DIFF
--- a/src/main/java/io/cryostat/discovery/DiscoveryModule.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryModule.java
@@ -17,6 +17,7 @@ package io.cryostat.discovery;
 
 import java.time.Duration;
 import java.util.Set;
+import java.util.concurrent.Executors;
 
 import javax.inject.Named;
 import javax.inject.Singleton;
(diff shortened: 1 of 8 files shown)


andrewazores:
https://github.com/cryostatio/cryostat/actions/runs/6102200022

mwangggg (approved):
Screenshot from 2023-09-07 12-03-49
LGTM

Issue #1670, opened 2023-09-15
[Bug] "Restart if recording already exists" checkbox doesn't work


andrewazores:
### Current Behavior

Creating a recording with name Foo, then trying to create another recording with name Foo, does fails with an HTTP 400.

### Expected Behavior

The request should fail if the "restart" checkbox is not checked, but should pass and result in the recording being restarted if the box is checked.

### Steps To Reproduce

1. Run sh smoketest.sh
2. Go to web UI
3. Go to Recordings
4. Select a target, etc. `vertx-fib-demo-1`
5. Create a recording manually, ex. named `test`, continuous duration, Continuous template
6. Try to repeat step 5 with and without the "restart" checkbox checked

### Anything else?

![image](https://github.com/cryostatio/cryostat/assets/3787464/ad0a7153-3b7f-4b86-8290-1ebdde2e7491)

tthvo:
Looks like we are ignoring the `restart` parameter here and the replace policy will return as NEVER?

https://github.com/cryostatio/cryostat/blob/0197f7b20ca9a90c0831e4936b95c04108378afc/src/main/java/io/cryostat/net/web/http/api/v1/TargetRecordingsPostHandler.java#L156

tthvo:
This is call from web client: https://github.com/cryostatio/cryostat-web/blob/ee99ad168f707e4342444589a466f9abb0ca6599/src/app/Shared/Services/Api.service.tsx#L308

andrewazores:
Right -

https://github.com/cryostatio/cryostat/pull/1587

https://github.com/cryostatio/cryostat/pull/1587/files#diff-52b7be8bc2188ea065259592ef8710555d3d73dabd58cd2ce0b3a57ab4c6f999L155

I recall we had some discussion around this change before but I can't remember exactly what we landed on. Since this is changing in a minor release version the API "shouldn't" change in incompatible ways, but I'm okay with changing a query parameter like this. Or we could have the endpoint handler handle both forms and prefer the newer one if both are supplied.

andrewazores:
ie.

```
ReplacementPolicy replace = ReplacementPolicy.NEVER;
if (attrs.contains("restart")) {
  replace = Boolean.parseBoolean(attrs.get("restart")) ? ReplacementPolicy.ALWAYS : ReplacementPolicy.NEVER;
}
if (attrs.contains("replace")) {
  replace = ReplacementPolicy.fromString(attrs.get("replace"));
}
}
```

[...]

PR #1672, opened 2023-09-18
fix(recording): backwards compatibility for restart parameter

tthvo:
# Welcome to Cryostat! 👋

Fixes: #1670 

## Description of the change:

Added back the `restart` parameter for backwards compatibility. Future major releases can remove it if needed but should highlight this.

## Motivation for the change:

See #1670 

> I recall we had some discussion around this change before but I can't remember exactly what we landed on. Since this is changing in a minor release version the API "shouldn't" change in incompatible ways, but I'm okay with changing a query parameter like this. Or we could have the endpoint handler handle both forms and prefer the newer one if both are supplied.

## How to manually test:

The old legacy graphQL query with `restart` parameter should work fine:

```graphql
query {
        descendantTargets {
            doStartRecording(recording: {
                name: "foo",
                template: "Profiling",
                templateType: "TARGET",
                duration: 30,
                restart: true
                }) {
                    name
                    state
            }
        }
    }
}
```

Also, operations on web client also succeeds.

### DIFF
--- a/docs/HTTP_API.md
+++ b/docs/HTTP_API.md
@@ -861,6 +861,10 @@
 
     **The request may include the following fields:**
 
+    `restart`: Whether to restart the recording if one already exists with the same name. **DEPRECATED**: See `replace` below.
+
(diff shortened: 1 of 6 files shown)


tthvo:
https://github.com/cryostatio/cryostat/actions/runs/6219769657

andrewazores (approved):
Thanks @tthvo !

PR #1683, opened 2023-09-22
feat(ci): build_test command enhancements

aali309:
# Welcome to Cryostat! 👋

Fixes: #1634 

## Description of the change:
- Leave a comment of the Actions run when it starts
- Leave a comment noting whether the tests passed or failed
- Retry integration test by typing `/retest` to simply rerun the tests without a rebuild

## Motivation for the change:
provide enhancements to the existing command `/build_test`

## How to manually test:
2. *...*


.github/workflows/pr-ci.yml
```diff
@@ -85,14 +118,38 @@ jobs:
...
       skip-itests: ${{ matrix.arch != 'amd64' }}
+  
+  check-test-results:
+    runs-on: ubuntu-latest
+    needs: [build-and-test]
+    steps:
+      - name: Check Test Results
+        id: check-results
+        run: |
+          if [ ... ]; then
```
andrewazores, line 129: Does this if-condition still need to be filled in?

.github/workflows/pr-ci.yml
```diff
@@ -73,9 +73,42 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Capture Date and Time
+        id: date-time
+        run: echo "::set-output name=date-time::$(date +'%Y-%m-%d %H:%M:%S')"
+      - name: Start Comment
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            Workflow started at ${{steps.date-time.outputs.date-time}}.
```
andrewazores, line 87: Is it feasible to add a link to the workflow run here as well? I'm not sure how the link can be retrieved but perhaps there is some Actions API for that or a reusable Action that can help.

mwangggg:
do you have a demo we can take a look at?

.github/workflows/pr-ci.yml
```diff
@@ -73,9 +73,42 @@ jobs:
...
+        with:
+         build-arch: ${{ matrix.arch }}
+         checkout-repo: ${{ needs.checkout-branch.outputs.PR_repo }}
+         checkout-ref: ${{ needs.checkout-branch.outputs.PR_head_ref  }}
+         skip-itests: ${{ matrix.arch != 'amd64' }}
+      - name: Set BUILD_TEST_EXECUTED flag
+        id: set-build-test-executed
+        run: echo "BUILD_TEST_EXECUTED=true" >> $GITHUB_ENV
+
+  retest-integration:
```
mwangggg, line 109: I'm pretty sure you don't intend on rebuilding the image here? Might need to add an input to ci-build-image.yml to specify that you only want to run integration tests.

.github/workflows/pr-ci.yml
```diff
@@ -73,9 +73,42 @@ jobs:
     with:
       checkout-repo: ${{ needs.checkout-branch.outputs.PR_repo }}
       checkout-ref: ${{ needs.checkout-branch.outputs.PR_head_ref }}
+
+  start-comment:
+    runs-on: ubuntu-latest
+    steps:
```
mwangggg, line 79: should probably have an `if: [check-before-build]`

.github/workflows/pr-ci.yml
```diff
@@ -73,9 +73,42 @@ jobs:
...
+        id: date-time
+        run: echo "::set-output name=date-time::$(date +'%Y-%m-%d %H:%M:%S')"
+      - name: Start Comment
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            Workflow started at ${{steps.date-time.outputs.date-time}}.
       
   build-and-test:
     needs: [code-analysis, checkout-branch]
```
mwangggg, line 90: we want this running in parallel so we should remove the dependency on the `code-analysis` job

.github/workflows/pr-ci.yml
aali309: Yes, I kind of have difficulties in figuring out this part.. The option that makes sense is to read test results in a file and grep if there is a test that failed. Any help with this part will be highly appreciated.. on may be another logic to log the test pass or failed???

andrewazores: The exit status of the `mvn` command will suffice. If `mvn test` exits with the normal 0 status then all tests passed, if it exits with `1` then at least one unit test failed. Likewise for `mvn verify` and integration tests.

The `repeated-integration-tests.bash` uses this exit status to count passes/failures, for example, and this script also uses the number of failures as its exit status, so if the exit status is 0 then all tests passed, otherwise there was a failure.

aali309: YES. Thank you.

.github/workflows/pr-ci.yml
andrewazores: Or perhaps it would be better to create a separate reusable workflow that pulls an already-built image tag from ghcr.io and runs the integration test suite against it?

aali309:
> do you have a demo we can take a look at?

Couldn't run a demo yesterday because of the missing logic on `if[...]`. should do it today, I have no pending issues

mwangggg:
https://github.com/cryostatio/cryostat-core/pull/266#discussion_r1341717352 - following this discussion, could you remove the permissions in the `CI-build-image.yml` file too?

aali309:
https://github.com/aali309/cryostat/actions/runs/6410504392/job/17403950446.. This is a demo run.. will need help to tackle this where `/retest` does not build images but just runs the integrated tests only.

.github/workflows/pr-ci.yml
```diff
@@ -73,9 +73,47 @@ jobs:
...
+         checkout-ref: ${{ needs.checkout-branch.outputs.PR_head_ref  }}
+         skip-itests: ${{ matrix.arch != 'amd64' }}
+         run-tests-only: true
+      - name: Set BUILD_TEST_EXECUTED flag
+        id: set-build-test-executed
+        run: echo "BUILD_TEST_EXECUTED=true" >> $GITHUB_ENV
+
+  retest-integration:
+    needs: [code-analysis, checkout-branch, build-and-test]
+    if: startsWith(github.event.comment.body, '/retest') && env.BUILD_TEST_EXECUTED == 'true'
```
andrewazores, line 116: I don't know if this `env.BUILD_TEST_EXECUTED` is going to persist between completely different runs - when the workflow is re-triggered by a new `/retest` comment, I would expect that the `$GITHUB_ENV` from the last workflow run is not still here for the new run to read. Maybe instead, this needs to check that it is able to pull a `ghcr.io` test image with the expected image tag?

aali309:
`https://github.com/aali309/cryostat/actions/runs/6474602388/job/17579876114` sample run with the update

.github/workflows/ci-build-image.yml
```diff
@@ -129,3 +141,40 @@ jobs:
     - uses: skjolber/maven-cache-github-action@v1
       with:
         step: save
+
+  retest-integration-tests:
+    runs-on: ubuntu-latest
+    if: ${{ !inputs.skip-itests && inputs.run-tests-only }}
```
andrewazores, line 147: Maybe this should also have `&& inputs.pr-number != ''` or something like that? Otherwise it seems like this will always fail on pulling the image later on.

andrewazores: Same for the `sha-value`.

This would really just be to safeguard against doing something silly in the future and trying to reuse this workflow without passing the inputs that it actually needs, since the inputs here are scoped to the whole file rather than this individual job.

I suppose that's also an argument to split this up into a separate reusable workflow so that the inputs can be properly defined as required for this particular job...

aali309:
Sample run:

/retest `https://github.com/aali309/cryostat/actions/runs/6512927455`

image

image

/build-test `https://github.com/aali309/cryostat/actions/runs/6512926521`

image

image

.github/workflows/ci-build-image.yml
aali309: will try the `&& inputs.pr-number != ''` and same for `sha-value` for now and I will look into separating the building image and integrated testing into separate workflows next week or if i have time on the weekend

aali309:
sample run after extra conditions `https://github.com/aali309/cryostat/actions/runs/6513631944/job/17693607675`

tthvo:
I am afraid it might complicate things more but is it possible to explore writing commit status to PR status section at the bottom?

Reference: https://github.com/orgs/community/discussions/25141

.github/workflows/ci-build-image.yml
```diff
@@ -129,3 +141,40 @@ jobs:
...
+          ref: ${{ inputs.checkout-ref }}
+          submodules: true
+          fetch-depth: 0
+    - uses: actions/setup-java@v3
+      with:
+        java-version: '17'
+        distribution: 'temurin'
+    - name: Pull cryostat image
+      run: podman pull ghcr.io/${{ github.repository_owner }}/cryostat:pr-${{ inputs.pr-number }}-${{ inputs.sha-value }}-linux-${{ inputs.build-arch }}
+    - name: Run integration tests
```
mwangggg, line 171: the integration tests are failing right now because it can't access the maven files, such as the -core jar. You'll need to set-up maven with maven-settings-action

.github/workflows/pr-ci.yml
```diff
@@ -86,13 +87,85 @@ jobs:
...
+    with:
+      build-arch: ${{ matrix.arch }}
+      run-tests-only: true
+      pr-number: ${{ github.event.issue.number }}
+      sha-value: ${{ needs.checkout-branch.outputs.PR_head_sha }}
+
+  start-comment:
+    runs-on: ubuntu-latest
+    needs: [check-before-build]
+    if: ${{ needs.check-before-build.result == 'success' }}
```
mwangggg, line 107: don't need this if statement- the `needs: [check-before-build]` ensures that this job only runs if the `check-before-build` job succeeded

mwangggg: same with [L126](https://github.com/aali309/cryostat/blob/e474dae3a41e9da2ee214aaa1af3388ffc6b5a6c/.github/workflows/pr-ci.yml#L126) and [L145](https://github.com/aali309/cryostat/blob/e474dae3a41e9da2ee214aaa1af3388ffc6b5a6c/.github/workflows/pr-ci.yml#L145C1-L145C1)

.github/workflows/pr-ci.yml
```diff
@@ -86,13 +87,85 @@ jobs:
...
+          });
+
+  build-and-test-pass:
+    runs-on: ubuntu-latest
+    needs: [build-and-test]
+    if: ${{ needs.build-and-test.result == 'success' }}
+    steps:
+    - name: Leave Actions Run Comment
+      uses: actions/github-script@v6
+      if: always()
```
mwangggg, line 130: don't need this- there is no previous step that could fail

mwangggg: same with [L149](https://github.com/aali309/cryostat/blob/e474dae3a41e9da2ee214aaa1af3388ffc6b5a6c/.github/workflows/pr-ci.yml#L149)

.github/workflows/pr-ci.yml
aali309: > don't need this if statement- the `needs: [check-before-build]` ensures that this job only runs if the `check-before-build` job succeeded

Yes that is the intention, it should run only when the jobs mentioned succeed

aali309: yes

mwangggg:
You should also consider separating the `retest-integration` job into a separate workflow. That way it doesn't need to be called through the `ci-build-image.yml` file and when you call `/retest_integration`, you don't rerun all the earlier jobs in `pr-ci.yml` like `code-analysis`

aali309:
Looks like both `/build_test` and `\retest` work now.
Latest workflow run: [https://github.com/aali309/cryostat/pull/97](https://github.com/aali309/cryostat/pull/97)
  link to  `/build_test`  : [https://github.com/aali309/cryostat/actions/runs/6631674187](https://github.com/aali309/cryostat/actions/runs/6631674187)
link to `/retest`: [https://github.com/aali309/cryostat/actions/runs/6631898132](https://github.com/aali309/cryostat/actions/runs/6631898132)

image

image


**Error** not related to `/build_test` or `/retest`

image

image

mwangggg:
can you rebase all your branches so they are up to date? I wonder why the integration tests are failing...

.github/workflows/integrated-test.yml
```diff
@@ -0,0 +1,79 @@
...
+        sudo apt-get install -y libxml-xpath-perl
+    - name: Install qemu
+      if: ${{ inputs.build-arch != 'amd64' }}
+      continue-on-error: true
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-user-static
+    - uses: actions/checkout@v4
+      with:
+          repository: ${{ github.repository_owner }}/cryostat
```
mwangggg, line 48: change this to `inputs.checkout-repo`

.github/workflows/pr-ci.yml
```diff
@@ -150,9 +169,66 @@ jobs:
...
           CRYOSTAT_IMAGE=${{ env.arm64_image }} sh smoketest.sh
           ```
+
+  integration-test:
+    needs: [checkout-branch, push-to-ghcr]
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+    uses: ./.github/workflows/integrated-test.yml
+    with:
```
mwangggg, line 185: add `skip-itests` or only run with `build-arch: amd64` since tests run with `arm64` image will surely fail

mwangggg: same at https://github.com/aali309/cryostat/blob/cd43ca4aa0389b949287d0de21793d437da49a10/.github/workflows/pr-ci.yml#L197

.github/workflows/pr-ci.yml
```diff
@@ -150,9 +169,66 @@ jobs:
...
+      build-arch: ${{ matrix.arch }}
+      pr-number: ${{ github.event.issue.number }}
+      sha-value: ${{ needs.checkout-branch.outputs.PR_head_sha }}
+      
+  integration-test-pass:
```
mwangggg, line 202: this can be moved into the `integration-test.yml` file and added on as the last step of the `integration-tests` job? Then you wont need separate jobs here for `integration-test-pass` and `retest-integration-pass`

.github/workflows/pr-ci.yml
mwangggg: also `checkout-repo` and `checkout-ref`?

.github/workflows/ci-build-image.yml
```diff
@@ -10,7 +12,7 @@ on:
       build-arch:
         required: true
         type: string
-      skip-itests:
+      run-tests-only:
```
mwangggg, line 15: we can get rid of this :)

.github/workflows/pr-ci.yml
aali309: > this can be moved into the `integration-test.yml` file and added on as the last step of the `integration-tests` job? Then you wont need separate jobs here for `integration-test-pass` and `retest-integration-pass`

The idea was to produce different outputs depending on the test

.github/workflows/ci-build-image.yml
aali309: Im blind 😂

mwangggg:
[Line59](https://github.com/aali309/cryostat/blob/952d91124c10b1953e33ae3c159c98ffb0e729ef/.github/workflows/ci-build-image.yml#L59) in ci-build-image.yml can also be deleted... I missed it earlier with the other `always()` oops

.github/workflows/ci-build-image.yml
aali309: I will celebrate if this PR is merged. I see all related code in my sleep too😂

mwangggg:
[Line61](https://github.com/aali309/cryostat/blob/cd43ca4aa0389b949287d0de21793d437da49a10/.github/workflows/ci-build-image.yml#L61) in `ci-build-image.yml` can also be deleted

.github/workflows/ci-build-image.yml
```diff
@@ -64,8 +71,8 @@ jobs:
     - name: maven-settings
       uses: s4u/maven-settings-action@v2
       with:
-        servers: '[{"id": "github", "username": "dummy", "password": "${env.GITHUB_TOKEN_REF}"}]'
-        githubServer: false
+        servers: '[{"id": "github", "username": "dummy", "password": "${env.GITHUB_TOKEN}"}]'
```
mwangggg, line 74: Like how you've done in `integrated-test.yml`, you can get rid of this line

aali309:
Tests look. I will clean the code and push the current changes including the ones mentioned ^

image

.github/workflows/ci-build-image.yml
mwangggg: that's how I felt with /build_test haha

.github/workflows/pr-ci.yml
aali309: Last time i tried it dint work, il try it again now

aali309:
Looks good now.  PR LINK: [https://github.com/aali309/cryostat/pull/104](https://github.com/aali309/cryostat/pull/104)
`build_test` Work flow LINK: [https://github.com/aali309/cryostat/actions/runs/6642460460](https://github.com/aali309/cryostat/actions/runs/6642460460)
`retest` Work flow Link: [https://github.com/aali309/cryostat/actions/runs/6642747025](https://github.com/aali309/cryostat/actions/runs/6642747025)

.github/workflows/pr-ci.yml
aali309: Yah, does not work.. I remember had a conversation with @Andrew the other day about this

andrewazores:
Those test runs look absolutely awesome. Nice work.

.github/workflows/integrated-test.yml
```diff
@@ -0,0 +1,78 @@
+name: Retest Integrated Tests
```
andrewazores, line 1: Nitpick: it should be "integration" tests rather than "integrated" tests

.github/workflows/ci-build-image.yml
```diff
@@ -64,8 +63,13 @@ jobs:
     - name: maven-settings
       uses: s4u/maven-settings-action@v2
       with:
-        servers: '[{"id": "github", "username": "dummy", "password": "${env.GITHUB_TOKEN_REF}"}]'
-        githubServer: false
+        githubServer: true
+    - name: ghcr login
+      uses: redhat-actions/podman-login@v1
```
andrewazores, line 68: What causes this to be required now?

.github/workflows/integrated-test.yml
```diff
@@ -0,0 +1,78 @@
...
+  integration-tests:
+    runs-on: ubuntu-latest
+    if: ${{ !inputs.skip-itests }}
+    steps:
+    - name: Install xpath
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libxml-xpath-perl
+    - name: Install qemu
+      if: ${{ inputs.build-arch != 'amd64' }}
```
andrewazores, line 40: Maybe this can be refactored? It ends up looking like this in pseudocode terms:

```
ci(false)
ci(skip) {
  run_tests(skip)
}
run_tests(skip) {
  if (skip) return
  ... # do tests
}
```

But perhaps it could just be:

```
ci(false)
ci(skip) {
  if (!skip) run_tests()
}
run_tests() {
  ... # do tests
}
```

ie. whoever is calling this workflow can just decide not to call it, rather than passing in this parameter that makes the whole thing do nothing?

.github/workflows/pr-ci.yml
```diff
@@ -150,9 +169,68 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [integration-test]
+    steps:
+    - name: Leave Actions Run Comment
+      uses: actions/github-script@v6
+      with:
+        script: |
+          const runURL = `https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${{ github.run_id }}`;
+          const commentBody = `both build and itest completed successfully ✅. \n[View Actions Run](${runURL}).`;
```
andrewazores, line 213: Is there any way to leave a similar comment in the case that the (re)test run fails?

aali309: I really want this to be part of this PR. I had tried it earlier and still trying and testing it. I tried several ways, with the status itself i.e `success` and status code `$? != 0` still not successful. Looks like the `needs` updates the current tests. The results were different with many times the test being skipped. I'll change the PR status until this is done to be reviewed again.

.github/workflows/ci-build-image.yml
aali309: I thought by setting `githubServer: false`,   the Mvn settings will not include GitHub Packages as a repository, which is needed and that Maven won't  fetch artifacts from GitHub Packages.

andrewazores: I'm not sure I understand... where does podman come into that equation?

aali309:
Added the test failure message as well & intentionally failed the test to see sample run
`/retest` failure LINK: [https://github.com/aali309/cryostat/actions/runs/6659054066](https://github.com/aali309/cryostat/actions/runs/6659054066)

image

image

andrewazores:
Very nice, that looks like it's working really well. I still had those two comments before - one small question and one small refactor/cleanup suggestion. Anyone else have any review feedback? I think this is really close to being done and ready for merge.

.github/workflows/ci-build-image.yml
aali309: if I have it as `false` on the `integrated-test ` file I get this error: I am checking it again ATM. 
Screenshot from 2023-10-26 10-03-30

aali309: Also, 

> I'm wondering what happened to this workflow that requires the new step to have podman log in to ghcr.io.

The new step to have podman log in to ghcr.io is to be able to pull the container image. Did I address the question right or may be I din't understand?

.github/workflows/integrated-test.yml
aali309: The updated code uses the default values of the inputs

.github/workflows/ci-build-image.yml
andrewazores: Right, I'm only asking about the podman part, not about the `githubServer` maven settings part. It makes sense that a step called `build-image` needs to be able to connect to GitHub Packages maven repo in order to pull the -core dependency and build the image.

But I'm not sure I understand why a step called `build-image` requires pulling the same image, and I don't see where that new requirement for this step came from - it doesn't look like the step itself really changed, so I suppose it comes from some other reusable workflow step that this one calls into. If that's the case, should the login step not be moved closer to where it is actually required?

Issue #1698, opened 2023-10-06
[Bug] Startup failure due to discovery ping failure


andrewazores:
### Current Behavior

Server startup can fail due to DiscoveryStorage ping failure:

```
Oct 06, 2023 8:46:42 PM io.cryostat.core.log.Logger info
INFO: Stale discovery service http://storedcredentials:4@localhost:9977/ removed
Oct 06, 2023 8:46:42 PM io.cryostat.core.log.Logger error
SEVERE: FAILED to deploy io.cryostat.discovery.DiscoveryStorage Verticle
java.util.concurrent.CompletionException: io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: localhost/127.0.0.1:9944
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:874)
	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:841)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
	at io.cryostat.discovery.DiscoveryStorage.lambda$ping$12(DiscoveryStorage.java:332)
	at io.vertx.core.impl.future.FutureImpl$3.onFailure(FutureImpl.java:153)
	at io.vertx.core.impl.future.FutureBase.emitFailure(FutureBase.java:75)
	at io.vertx.core.impl.future.FutureImpl.tryFail(FutureImpl.java:230)
	at io.vertx.core.impl.future.PromiseImpl.tryFail(PromiseImpl.java:23)
	at io.vertx.core.impl.future.PromiseImpl.onFailure(PromiseImpl.java:54)
	at io.vertx.core.impl.future.PromiseImpl.handle(PromiseImpl.java:43)
	at io.vertx.core.impl.future.PromiseImpl.handle(PromiseImpl.java:23)
	at io.vertx.ext.web.client.impl.HttpContext.handleFailure(HttpContext.java:393)
	at io.vertx.ext.web.client.impl.HttpContext.execute(HttpContext.java:387)
[...]

PR #1700, opened 2023-10-10
fix(discovery): do not fail startup if plugin ping fails

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1698

## Description of the change:
Modifies the `DiscoveryStorage` plugin ping logic to be tolerant of request failures or network exceptions by remapping the ping task future to one that logs exceptions but does not propagate the `Future` failure to the aggregate `Future` representing the entire task.

## Motivation for the change:
When this plugin ping prune task is done periodically failures are already ignored. However, when this task is done initially at startup to check plugin registrations already present in the database from the last run, ping failures would fail the whole task, cascading into failing the server startup. This should not happen - the server is not dependent upon the discovery plugins and can handle their absence fine, so it should not be startup-blocking if an exception occurs when probing them.

## How to manually test:
1. Tough to nail down because it requires a narrow condition where the discovery plugin is present but a network error occurs during the registration ping request, ex. the network stack says that the plugin hostname can be resolved but the connection is refused because the plugin JVM is not yet ready to accept connections.

### DIFF
--- a/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
+++ b/src/main/java/io/cryostat/discovery/DiscoveryStorage.java
@@ -276,19 +276,25 @@ public class DiscoveryStorage extends AbstractPlatformClientVerticle {
         }
     }
 
-    private CompletableFuture<?> pingPrune() {
-        List<CompletableFuture<Boolean>> futures =
+    private CompletableFuture<Void> pingPrune() {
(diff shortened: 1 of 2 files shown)


andrewazores:
https://github.com/cryostatio/cryostat/actions/runs/6470118720

aali309 (approved):
Looks good to me

Issue #1783, opened 2023-11-20
[Meta] CI workflows must request write permissions


andrewazores:
### Current Behavior

https://github.com/cryostatio/cryostat-operator/pull/684#discussion_r1399577188

andrewazores:
Looks like cryostat3 CI needs an update: https://github.com/cryostatio/cryostat3/actions/runs/6934496324/job/18862719115

PR #1784, opened 2023-11-20
chore(ci): update ci

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #1783

### DIFF
--- a/.github/workflows/ci-build-image.yml
+++ b/.github/workflows/ci-build-image.yml
@@ -17,10 +17,6 @@ on:
         description: the Cryostat application version that will be built
         value: ${{ jobs.get-pom-properties.outputs.image-version }}
 
-permissions:
-  contents: read
-  packages: read
-
 jobs:
   get-pom-properties:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-code-analysis.yml
+++ b/.github/workflows/ci-code-analysis.yml
@@ -8,10 +8,6 @@ on:
         required: false
         type: string
 
-permissions:
-  contents: read
-  packages: read
-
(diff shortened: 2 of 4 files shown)


mwangggg:
one thing- I added the `permissions: pull-requests: write` to each job separately, but wondering if it would be better to just include the permission for the whole workflow?

mwangggg:
Also, all the other repos use status checks instead of comments. For consistency, should I update Cryostat to use status checks instead of, or on top of, the comments that are currently being printed?

andrewazores:
> but wondering if it would be better to just include the permission for the whole workflow?

I would rather keep it granular and applied only to each specific job/step that actually needs it. Otherwise, we may as well just leave the default setting to grant it all the time if we're going to broadly request it at the workflow level everywhere.

andrewazores:
> Also, all the other repos use status checks instead of comments. For consistency, should I update Cryostat to use status checks instead of, or on top of, the comments that are currently being printed? Or, if we think comments are better, I can update the other way around.

Status checks are good because they allow PR merge to be gated, ie the "Merging is blocked" status can be made dependent on the outcome of these workflows, which is super nice.

On the other hand, comments are nice because it lets us add various pieces of detailed information. It might be possible to add that to the statuses as well but then you have to click through somewhere else to see the details, and the history of those statuses probably won't be as clear over time as the comment history is.

So IMO both have their place and I'd like to have both where it makes sense. If a repo has a simple CI setup with no `/build_test` etc. then the statuses alone are fine, but once it gets a bit more advanced, comments are good to add.

andrewazores (approved):

Issue #1861, opened 2024-01-23
[Bug] Cryostat should not always assume Kubernetes ports numbered 9091 are JMX


andrewazores:
### Current Behavior

Since forever, Cryostat's Kubernetes API discovery mechanism has assumed that if it sees a port named `jfr-jmx` or numbered 9091 that this is a JMX port. More recent Cryostat versions try to proactively connect to discovered target applications to determine their JVM IDs.

In the case that 9091 is exposed for a different service and is **NOT** a JMX port, this results in Cryostat trying to open a JMX connection to a non-JMX port as soon as it sees such a target application. This connection attempt will fail, but it takes time for that to happen. If enough of these attempts happen simultaneously (ie there is a large number of non-JMX 9091 applications discovered), it can lead to thread starvation, which degrades Cryostat performance - possibly to the point that the container platform kills and restarts Cryostat.

The `jfr-jmx`/`9091` pattern has been in Cryostat for so long that it is best not to change it now, at least not as the default setting. However, it is about time that we make this configurable. These values can be kept as defaults but they should be overrideable by environment variables, and these variables should also be configurable by Operator CRD fields.

PR #1862, opened 2024-01-23
fix(k8s): Endpoints discovery port names/numbers are configurable

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1861

## Description of the change:
Adds environment variables for customizing the behaviour of the builtin k8s Endpoints discovery mechanism. Rather than always looking for ports named `jfr-jmx` or numbered `9091`, this allows the deploying user to specify a list of acceptable names and a list of acceptable numbers that should be matched, both of them comma-separated. The names list can be set to the empty list by using the value `-`, and the number list can be set to the empty list by using the value `0`. If these variables are not defined or have empty values then they default to the singleton lists of `jfr-jmx` and `9091` respectively, maintaining the previous behaviour.

## Motivation for the change:
For users deploying Cryostat into scenarios where a port named `jfr-jmx` or numbered `9091` already exists for another purpose, this allows Cryostat to be dropped in with only minor configuration so that the Kubernetes Endpoints discovery mechanism can still be used, but does not mistakenly see non-JMX service ports as connectable targets when they really are not.

## How to manually test:
1. I used helm, ex.: `helm install cryostat --set core.image.repository=quay.io/andrewazores/cryostat --set core.image.tag=2.5.0-k8s-discovery-3 --set pvc.enabled=true --set core.route.enabled=true ./charts/cryostat`
2. Then I used the Operator's `make sample_app` to get a `quarkus-test` sample application
3. Then I used `oc edit deployment cryostat` to set/edit `CRYOSTAT_DISCOVERY_K8S_PORT_NAMES` and `CRYOSTAT_DISCOVERY_K8S_PORT_NUMBERS`, and `oc edit service quarkus-test` to change the port name/number on the sample app.

### DIFF
--- a/src/main/java/io/cryostat/platform/internal/KubeApiPlatformStrategy.java
+++ b/src/main/java/io/cryostat/platform/internal/KubeApiPlatformStrategy.java
@@ -39,7 +39,7 @@ import org.apache.commons.lang3.StringUtils;
 class KubeApiPlatformStrategy implements PlatformDetectionStrategy<KubeApiPlatformClient> {
 
     public static final String NO_PORT_NAME = "-";
-    public static final String NO_PORT_NUMBER = "0";
+    public static final Integer NO_PORT_NUMBER = 0;
 
     protected final Lazy<? extends AuthManager> authMgr;
     protected final Environment env;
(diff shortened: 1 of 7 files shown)


src/main/java/io/cryostat/platform/internal/KubeApiPlatformStrategy.java
```diff
@@ -39,21 +38,18 @@
 
 class KubeApiPlatformStrategy implements PlatformDetectionStrategy<KubeApiPlatformClient> {
 
+    public static final String NO_PORT_NAME = "-";
+    public static final Integer NO_PORT_NUMBER = 0;
```
tthvo, line 42: Just curious if we should also use `-` for the empty port number list? I guess we can filter it out first before parsing int.

andrewazores: Sure, that's an option too. I don't really mind either way. @ebaron any preference?

andrewazores: https://github.com/cryostatio/cryostat-operator/pull/715#issuecomment-1907560164

aali309 (approved):

ebaron (approved):

Issue #1864, opened 2024-01-24
[Bug] `BUILTIN_DISCOVERY_DISABLED` is ignored if `PLATFORM` is set


andrewazores:
### Current Behavior

https://github.com/cryostatio/cryostat/blob/ece2e2862510b4c0c3f6c75bdd86fbd202c27b73/src/main/java/io/cryostat/platform/PlatformModule.java#L89

If the `CRYOSTAT_PLATFORM` environment variable is set, then it is impossible to disable builtin discovery.

The `CRYOSTAT_PLATFORM` should still be respected for auto-detecting platforms and associated AuthManagers, but if `BUILTIN_DISCOVERY_DISABLED=true` then this should always override the platform detection clients.

PR #1865, opened 2024-01-24
fix(discovery): BUILTIN_DISCOVERY_DISABLED can be specified even when PLATFORM is set

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1864

## Description of the change:
Removes `BUILTIN_DISCOVERY_DISABLED` environment variable from the platform detection selection logic. This is an unfortunate and very old design decision I made that the platform integration is coupled to both the default auth manager as well as platform-specific target discovery mechanism. This made sense when originally implemented because it never made sense for platform-specific discovery mechanisms to overlap, but with the advent of custom targets and later the pluggable discovery API, and then with support for both Podman and Docker API, this is clearly no longer a good assumption and no longer a good abstraction.

This results in a bug where the variable used to control turning off all of these built-in discovery mechanisms is evaluated in the same place as where we determine which platform detection strategies to enable, which therefore also implicates the AuthManagers. We want to be able to do automatic platform auth (OpenShift is the only one that exists now) integration while also allowing users to turn off built-in discovery in case it causes problems or they intend not to use it.

The fix applied in this patch is to allow platform detection as a whole to either work by the usual automatic detection mechanisms, or to allow the user to use the `CRYOSTAT_PLATFORM` environment variable to specify the list of platforms to enable explicitly. The `BUILTIN_DISCOVERY_DISABLED` variable is not checked until later, after the platform strategies are selected. Once all the strategies are selected the DiscoveryStorage, which integrates all of the mechanisms into one data repository, turns each one on only if its new `isEnabled()` method returns true. For Custom Targets and Discovery Storage itself (the pluggable API) this is always true, but for the other built-in discovery mechanisms (Kubernetes API, Podman, Docker) the check depends on `BUILTIN_DISCOVERY_DISABLED` not being set to `true`.

## How to manually test:
1. `./mvnw clean package ; podman image prune -f ; podman tag quay.io/cryostat/cryostat:latest quay.io/$MY_QUAY_USERNAME/cryostat:pr-1865-1`
2. `podman push quay.io/$MY_QUAY_USERNAME/cryostat:pr-1865-1`
3. `cd cryostat-operator`
4. `oc new-project pr1865`
5. `DEPLOY_NAMESPACE=$(oc project -q) CORE_IMG=quay.io/$MY_QUAY_USERNAME/cryostat:pr-1865-1 make deploy`
6. `make create_cryostat_cr ; make sample_app`
7. Open web-client, log in with OpenShift, accept permissions, etc. Verify that Cryostat itself and the sample app appear under the KubernetesApi heading in the target dropdown and in the KubernetesApi Realm in Topology.
8. `oc patch -n $(oc project -q) cryostat/cryostat-sample -p '{"spec":{"targetDiscoveryOptions":{"builtInDiscoveryDisabled":true}}}'` to flip the switch and turn off builtin discovery
9. Wait for rollout replacement to complete, then repeat Step 7. There should be no more KubernetesApi target dropdown header or Realm.

### DIFF
--- a/src/main/java/io/cryostat/discovery/BuiltInDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/BuiltInDiscovery.java
@@ -109,9 +109,11 @@ public class BuiltInDiscovery extends AbstractVerticle implements Consumer<Targe
                                     .onSuccess(
                                             subtree -> storage.update(id, subtree.getChildren()));
                             try {
-                                platform.start();
-                                platform.load(promise);
-                                enabledClients.add(platform);
+                                if (platform.isEnabled()) {
+                                    platform.start();
+                                    platform.load(promise);
+                                    enabledClients.add(platform);
+                                }
                             } catch (Exception e) {
                                 logger.warn(e);
                             }
--- a/src/main/java/io/cryostat/discovery/BuiltInDiscovery.java
+++ b/src/main/java/io/cryostat/discovery/BuiltInDiscovery.java
@@ -74,6 +74,12 @@ public class BuiltInDiscovery extends AbstractVerticle implements Consumer<Targe
                 .distinct()
                 .forEach(
                         platform -> {
+                            if (!platform.isEnabled()) {
+                                logger.info(
+                                        "Skipping built-in discovery with {} : not enabled",
+                                        platform.getClass().getSimpleName());
+                                return;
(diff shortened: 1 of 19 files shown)


mwangggg (approved):
looks good to me

